PR #28, opened 2025-05-16
new files needed to support onboarding

clubanderson:


SECURITY.md
```diff
@@ -0,0 +1,35 @@
+## Security Announcements
+
+Join the [llm-d-security-announce](https://groups.google.com/u/1/g/llm-d-security-announce) group for emails about security and major API announcements.
```
smarterclayton, line 3: The public group and the private vulnerability reporting email address below are the same - the private one seems wrong.

PR #83, opened 2025-05-21
Expand details of components in the llm-d proposal

smarterclayton:
Bring details from some of the northstar documents up and
into the proposal to answer questions raised by community.

Call out performance benchmarking as a project goal.

Clarify multi-host model support preferences via LWS and Ray
as it was asked in multiple forums.

### DIFF
--- a/docs/proposals/llm-d.md
+++ b/docs/proposals/llm-d.md
@@ -27,11 +27,14 @@ llm-d is successful if it:
 * Offers an extensible and flexible inference scheduler to balance traffic
 * Supports multiple emerging LLM workloads (agents, multimodal, RAG/search) with clear reference architectures
 * Composes with existing Kubernetes infrastructure choices
+* Is not opinionated about model server deployment orchestration and model server lifecycle
+* Is reliably and consistently tested for performance in our development and testing and in end-user production 
 
 ### Non-Goals
 
 * Prioritize non-Transformer model architectures (initially)
 * Fork upstream repositories or carry unmerged upstream changes
+* Control the exact configuration of the end-user vLLM deployments
 
 ## Proposal
 
@@ -81,6 +84,46 @@ Our current Northstar designs lay out the initial scope (join llm-d-contributors
 * [Prefix Cache Hierarchy](https://docs.google.com/document/d/1inTneLEZTv3rDEBB9KLOB9K6oMq8c3jkogARJqdt_58/edit)
 * [Variant Autoscaling over Hardware, Workload, and Traffic](https://docs.google.com/document/d/1inTneLEZTv3rDEBB9KLOB9K6oMq8c3jkogARJqdt_58/edit)
 
+### Components
+
+llm-d streamlines deployment and integration of the following components:
+
+1. An **inference scheduler** directing traffic to model servers
+    1. Integrates with the Kubernetes [inference gateway API project](https://gateway-api-inference-extension.sigs.k8s.io/) to have Kubernetes control plane API
+    2. Performing model routing and rollout, flow control, kv- and prefix- cache aware load balancing
+    3. Balances traffic to the optimal model server based on the request, workload type, and current load
+2. **vLLM model servers** deployed onto Kubernetes
+    1. In single host or multi-host (using [LeaderWorkerSets](https://lws.sigs.k8s.io/) and [Ray](https://docs.vllm.ai/en/latest/serving/distributed_serving.html#running-vllm-on-multiple-nodes) as best practice) configurations
+    2. With native support for disaggregated serving and optional curated plugins for advanced features
+    3. Using project recommended defaults or highly customized user settings
+    4. May be deployed in multiple deployment **variants** (hardware, software, topology) that offer different performance tradeoffs
+    5. Can be rapidly repurposed as load shifts and started/restarted to reduce wasted capacity
+    6. Can dynamically load new LoRA adapters or even new models with low configuration / coordination
+3. vLLM default prefix caching and zero or more **prefix cache integrations**
+    1. At least one remote prefix cache option
+    2. At least one prefix cache option using the local SSD drives for each replica efficiently
+4. A **variant autoscaler** working with the inference scheduler and Kubernetes horizontal pod autoscaling
+    1. Can reassign prefill and decode roles between model server instances dynamically
+    2. Is aware of multiple deployment variants and their performance and can optimize across them
+    3. Can perform more advanced bucketization of traffic on latency or throughput objectives
+
+### APIs
+
+llm-d intends to drive the following APIs in our upstreams:
+
+1. vLLM public workload APIs to support inference scheduler-driven disaggregated serving
+2. vLLM management APIs to support rapid reconfiguration of vLLM where appropriate
+2. vLLM internal APIs to support point-to-point disaggregated serving, remote prefix cache, and testing
+3. Kubernetes Gateway APIs to support static or dynamic tuning of disaggregation
+4. Kubernetes Gateway APIs to enable latency or throughput optimization and autoscaling
+5. vLLM/[LMCache](https://lmcache.ai) remote prefix cache APIs that are interoperable across implementations
+
+llm-d may incubate new Kubernetes APIs via custom resources or new vLLM APIs, but our primary path is to land APIs upstream.
+
+### Dependencies
+
+llm-d intends to use [NIXL](https://github.com/ai-dynamo/nixl) to optimize GPU originating and terminating transfers. A follow-on proposal will identify gaps across accelerators and in host to host scenarios and recommend a solution.
+
 ## Alternatives
 
 ### Use NVIDIA Dynamo
@@ -94,3 +137,7 @@ AIBrix provides a strong research-focused and fast iterating integrated serving 
 ### Use Production Stack
 
 production-stack is the easiest way to deploy vLLM on Kubernetes. llm-d intends to work closely with the production-stack team to find common components and patterns to integrate, especially around prefix cache configuration.
+
+### Use kServe
+
+kServe assists platform teams in running large numbers of traditional and generative models on Kubernetes densely. Consider kServe if you have lots of LLM deployments smaller than several hosts or if you have many teams that need distinct deployments of models. llm-d focuses on operationalizing large models in very large deployments, as well as having multiple teams using a single shared deployment efficiently.


docs/proposals/llm-d.md
```diff
@@ -81,6 +84,39 @@ Our current Northstar designs lay out the initial scope (join llm-d-contributors
...
+    1. Can reassign prefill and decode roles between model server instances dynamically
+    2. Is aware of multiple deployment variants and their performance and can optimize across them
+    3. Can perform more advanced bucketization of traffic on latency or throughput objectives
+
+### APIs
+
+llm-d intends to drive the following APIs in our upstreams:
+
+1. vLLM public APIs to support inference scheduler-driven disaggregated serving
+2. vLLM internal APIs to support point-to-point disaggregated serving, remote prefix cache, and testing
```
ahg-g, line 116: Also APIs to dynamically load models or LoRA adapters?

docs/proposals/llm-d.md
```diff
@@ -81,6 +84,43 @@ Our current Northstar designs lay out the initial scope (join llm-d-contributors
...
+    3. Balances traffic to the optimal model server based on the request, workload type, and current load
+2. **vLLM model servers** deployed onto Kubernetes
+    1. In single host or multi-host (using [LeaderWorkerSets](https://lws.sigs.k8s.io/) and Ray as best practice) configurations
+    2. With native support for disaggregated serving and optional curated plugins for advanced features
+    3. Using project recommended defaults or highly customized user settings
+    4. May be deployed in multiple deployment **variants** (hardware, software, topology) that offer different performance tradeoffs
+3. vLLM default prefix caching and zero or more **prefix cache integrations**
+    1. At least one remote prefix cache option
+    2. At least one prefix cache option using the local SSD drives for each replica efficiently
+4. A **variant autoscaler** working with the inference scheduler and Kubernetes horizontal pod autoscaling
```
lionelvillard, line 105: should this be generalized to cover all use cases related autoscaling. As I understand it, the variant autoscaler concern is scaling Ps and Ds for a given *model*. We also want to include autoscaling across models, minimizing downtime due to frequent scaling up and down, optimizing costs and meet SLAs.

docs/proposals/llm-d.md
```diff
@@ -94,3 +134,7 @@ AIBrix provides a strong research-focused and fast iterating integrated serving
 ### Use Production Stack
 
 production-stack is the easiest way to deploy vLLM on Kubernetes. llm-d intends to work closely with the production-stack team to find common components and patterns to integrate, especially around prefix cache configuration.
+
+### Use kServe
```
sriumcp, line 141: This is missing a reference to ModelService.

***

### Use ModelService

[ModelService](https://github.com/llm-d/llm-d-model-service) is purpose built for llm-d users and is intended to simplify model deployments on llm-d. A ModelService custom resource encapsulates the desired state of workloads and routing associated with a given model, and is designed with the goals of modularity, interoperability with the eco-system components (including Gateway API Inference Extension APIs), support for PD disaggregation, and model sourcing from various providers.

sriumcp (changes requested):

docs/proposals/llm-d.md
sriumcp: LoRA Controller also happens to be one of the roadmap items in ModelService.

docs/proposals/llm-d.md
eilamt: This is really a great design and necessary to support the requirement of modularity and extensibility. The global autoscaler should generate a suggested new desired state. This should be a requirement for the autoscaling implementation.

docs/proposals/llm-d.md
```diff
@@ -27,11 +27,14 @@ llm-d is successful if it:
 * Offers an extensible and flexible inference scheduler to balance traffic
 * Supports multiple emerging LLM workloads (agents, multimodal, RAG/search) with clear reference architectures
 * Composes with existing Kubernetes infrastructure choices
+* Is not opinionated about how model server deployment orchestration
```
sriumcp: Voices heard re: ModelService in the community call on May 21 include the following.

* the path forward for llm-d user facing vs platform separation (@tumido)
* useful; lets continue to incubate (@robertgshaw2-redhat , @chcost )
* llm-d user base we're targeting is advanced and know their k8s; won't need modelservice (@smarterclayton)

Could we please reword this sentence to reflect the diversity above? For example:

***

* llm-d is open to any model deployment tool or pattern; that said, consider [ModelService](https://github.com/llm-d/llm-d-model-service), and if it doesn't fit your needs, consider filing an issue in the repo to describe why not.

sriumcp (changes requested):

docs/proposals/llm-d.md
smarterclayton: The scope for variant autoscaler is heterogenous workload and traffic scaling as per the north star to rebalance capacity availability vs capacity demand (due to current traffic mix).

I do not know what autoscaling across models means in this context - can you elaborate.

docs/proposals/llm-d.md
smarterclayton: Model server is a proposed component within this project, not an ecosystem alternative that we are framing against.

docs/proposals/llm-d.md
smarterclayton: We need a proposal to cover model service, its role, and scope before I think we are ready to recommend it here.  A point of initial scoping prior to launch was that we will consider model service and the use case after launch.

docs/proposals/llm-d.md
```diff
@@ -81,6 +84,46 @@ Our current Northstar designs lay out the initial scope (join llm-d-contributors
...
+### Components
+
+llm-d streamlines deployment and integration of the following components:
+
+1. An **inference scheduler** directing traffic to model servers
+    1. Integrates with the Kubernetes [inference gateway API project](https://gateway-api-inference-extension.sigs.k8s.io/) to have Kubernetes control plane API
+    2. Performing model routing and rollout, flow control, kv- and prefix- cache aware load balancing
+    3. Balances traffic to the optimal model server based on the request, workload type, and current load
+2. **vLLM model servers** deployed onto Kubernetes
+    1. In single host or multi-host (using [LeaderWorkerSets](https://lws.sigs.k8s.io/) and Ray as best practice) configurations
```
terrytangyuan: "Ray as best practice" - is there any supporting reference to learn more about this statement?

docs/proposals/llm-d.md
lionelvillard: thanks for clarifying the scope, that's basically what I meant by "across model".

My second item, how quickly vllm instances are reconfigured/created is covered by item 2.5.

docs/proposals/llm-d.md
```diff
@@ -94,3 +137,7 @@ AIBrix provides a strong research-focused and fast iterating integrated serving
 ### Use Production Stack
 
 production-stack is the easiest way to deploy vLLM on Kubernetes. llm-d intends to work closely with the production-stack team to find common components and patterns to integrate, especially around prefix cache configuration.
+
+### Use kServe
+
+kServe assists platform teams in running large numbers of traditional and generative models on Kubernetes densely. Consider kServe if your model deployments are small and you have many teams that need distinct deployments. llm-d focuses on running very large models in very large deployments, as well as having multiple teams using a single shared deployment efficiently.
```
terrytangyuan: Note that KServe supports Envoy AI Gateway (which is integrated with GIE): https://kserve.github.io/website/master/admin/ai-gateway_integration/. Although KServe does not support disaggregated deployments yet.

docs/proposals/llm-d.md
sriumcp: Thanks @smarterclayton. Yes, will plan on circulating this proposal shortly.

docs/proposals/llm-d.md
smarterclayton: https://docs.vllm.ai/en/latest/serving/distributed_serving.html#running-vllm-on-multiple-nodes

docs/proposals/llm-d.md
smarterclayton: Is this a suggested edit?  I think from a structural point of view the difference in focus is serving very large deployments vs not.

docs/proposals/llm-d.md
terrytangyuan: Thanks. It might be a good idea to provide a link to this in the doc. cc @njhill @tlrmchlsmth @robertgshaw2-redhat Do we have any plans to relax the dependency on Ray in vLLM?

docs/proposals/llm-d.md
terrytangyuan: I think that still needs to be discussed further since we originally had plans on supporting GenAI workloads well. The community is working on a new CRD in KServe to support additional use cases cc @danielezonca

asm582: @terrytangyuan can you share any docs on the new CRD that is being discussed?

terrytangyuan: https://github.com/kserve/kserve/issues/4433 so far the public discussion has been around multi-node but there's also an interest in supporting disaggregated serving. cc @jooho

yuzisun: From what I see ModelService essentially is very similar to KServe’s InferenceService, just that we are not supporting PD deployment yet but we are actively working on the distributed inference API as @terrytangyuan stated above. Can we work together instead of adding more confusions to the community?

yuzisun: Just to be clear KServe’s mission is to support scalable model deployment not small deployment. We internally already run a large fleet of vLLM deployments on KServe with many LLM models, we are looking to improve inference efficiently with PD.

smarterclayton: As discussed above in my comment to @sriumcp, llm-d is not opinionated about model deployment, and the model-service component must go through a proposal process for incubation. However, it is within scope for this project to reduce complexity of deployment and lifecycle for LLMs, and I can believe there are a set of people who would like an LLM CRD who do not wish to install kserve to do so.

I would suggest @sriumcp reviews his proposal for incubation with you (@terrytangyuan and @yuzisun) to identify what the differences in user base and core principles (for instance, model service might be unidimensional). That would then be a good input in the proposal to clarify framing that justifies incubation under the principles of this project and looks for opportunities to collaborate with kServe.

Re: AI Gateway:

Integration with the inference gateway project for the inference-scheduler is fundamental to this project for the dynamic distribution of requests to P and D servers as load and traffic changes.  Dynamic disaggregation requires an active process on the frontend (the scheduler) and the backend (autoscaler) to optimize.  If there is an opportunity for kServe to work with llm-d to leverage the capabilities we will develop in that vein, I think we'd be happy to discuss.

smarterclayton: Removed the use of the word "small" from the description of kserve.

chcost (approved):
lgtm

PR #103, opened 2025-06-04
Adding Variant Autoscaler proposal

eilamt:
This is the community driven Variant Autoscaler proposal.


docs/proposals/autoscaler.md
```diff
@@ -0,0 +1,154 @@
...
+
+In this section and below, wrap lines to a reasonable length to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+-->
+
+The variant autoscaler dynamically calculates the optimal mix of accelerator topologies (including, scaling up and down as necessary) for each available model server configuration that optimally serves the current multidimensional traffic load.
+
+1. We strive to have a loosely coupled design of a variant autoscaler relative to the request scheduler while having both working together towards a common goal, which is maximizing latency SLO attainment with the least cost of hardware. 
+2. The variant autoscaler manages the mix of topology assignments (e.g. accelerator type, single-node vs. multi-node, intra-node and inter-node RMDA and DCN connection type, different TP, EP, DP and max batch size) and dynamic traffic patterns.
```
aarnphm, line 30: for "max batch size", iiuc we are referring to the max number of requests that can be used in a mixed batch settings? (`max_num_seqs` setting)

docs/proposals/autoscaler.md
```diff
@@ -0,0 +1,154 @@
...
+3. The request scheduler manages individual requests and makes point-in-time scheduling decisions for each request according to the bucketization guidance from the variant autoscaler.
```
aarnphm, line 31: I'm not sure if I understand `point-in-time` here? is it just-in-time decision?

docs/proposals/autoscaler.md
ev-shindin: In general a separate discussion on max_batch_size or `max_num_seqs` required. I think the best option to put this parameter as big as possible and have an external control loop that managing actual number of concurrent requests that accepted by the particular instance, based on latency requirements. This allow to manage latency dynamically (`max_num_seqs` is configuration parameter that set on deployment (variant) level so to change it you should have a new deployment and then perform scale-up for new variant and scale-down ob previous variant). We are preparing dynamic concurrency management proposal and hope will make it availilble during next week.

docs/proposals/diagrams/autoscaler.png
ev-shindin, line 1: I would change this picture, but let us go over design document first.

docs/proposals/autoscaler.md
```diff
@@ -0,0 +1,154 @@
...
+List the specific goals of the proposal. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+1. Method to adjust the accelerator and configuration set to serve requests in a performant way with the least amount of resources and cost. 
+2. Show the advantages of the approach relative to the baseline, and additionally, the benefits of a heterogeneous set of resources and topology.  
+3. The method should work independently of KVCache configuration (it is orthogonal).
+4. The method should work for different and dynamically changing traffic patterns. 
+5. Architecture should be extensible, allowing for experimentation with new auto-scalers or the use of different auto-scalers for different models. 
+6. The method should minimize changes and avoid oscillations and identify the sequence of steps to get to the desired state without any disruptions.
```
pierDipi, line 70: Have you considered adding a "recommendation mode" (or dry-run mode) as a formal goal?

Allowing operators to see the proposed changes without applying them would be invaluable for building trust, validating the logic or tuning configurations, and previewing cost savings before enabling active scaling.

docs/proposals/autoscaler.md
eilamt: it really just means decision per request based on current state. The point is that the request scheduler does not decide to add or remove capacity for a particular model, it just decides based on the current set where to route a particular request.

docs/proposals/diagrams/autoscaler.png
eilamt: picture is very high level, we will have detailed pictures in the design document

docs/proposals/autoscaler.md
eilamt: No. max batch size controls how many *independent* requests/prompts can be processed in parallel. Each one is a separate conversation/prompt. If i understand correctly, Max_num_seq controls how many sequences can be processed *within a single batch* - more relevant for parallel deciding. In theory, if it makes sense this is another dimension that the autoscaler can control.

docs/proposals/autoscaler.md
eilamt: yes I think that is an excellent idea, we need to think how it can be incorporated into the design.

docs/proposals/autoscaler.md
chcost, line 30: RMDA -> RDMA

aarnphm: got it, I think `max_num_seqs` is important in a context of serving LLMs given that we will need to determine when the schedulers are being overloaded.

eilamt: yes agreed we can certainly incorporate it

smarterclayton:
This proposal is LGTM, i will wait for sign off from other project maintainers.

docs/proposals/autoscaler.md
atantawi: Maybe having a core decision maker which just produces decisions and a separate Actuator which acts on those decisions. If the Actuator is not running no changes will happen. A simple Actuator could only be a dashboard showing the decisions as recommendations, along with the potential savings.

chcost:
This proposal also looks good to me.

chcost (approved):
LGTM.

PR #111, opened 2025-06-16
Update autoscaler proposal

smarterclayton:
Clarify that the autoscaler must compose with existing infrastructure such as montoring and HPA.

Fix whitespace and title.


eilamt (changes requested):
Can you change to "exposing metrics readable by HPA or any other actuation mechanisms" ? I think this captures the intention better.  Actuation is the architectural component in Variant AutoScaler that is responsible to actually affect the change. It can be HPA or other mechanisms. The point is that we do not prescribe that it will be HPA.

liu-cong:
Closing this to clean up the backlog. This  is likely obsolete. Feel free to reopen a new one.

PR #112, opened 2025-06-16
Update the section regarding KServe in llm-d.md

terrytangyuan:
This is basically the same from https://github.com/llm-d/llm-d/blob/dev/docs/proposals/modelservice.md#alternative-extending-kserve but with the link to KServe proposal for details.


smarterclayton:
This is pretty verbose.  I'd prefer to integrate the proposal link and update the framing in #112.

PR #113, opened 2025-06-16
Describe the core design choices for llm-d

smarterclayton:
Based on recent discussions.

### DIFF
--- a/docs/proposals/llm-d.md
+++ b/docs/proposals/llm-d.md
@@ -57,6 +57,16 @@ The project will measure success against:
 * Efficiency of serving (perf/$ at target latency)
 * Reduction of operational toil, especially with increasing workload density
 
+The key design choices in `llm-d` are:
+
+1. Assume multiple workloads are consuming a shared model server pool and standardize those workload APIs via Kubernetes
+2. Leverage scheduler-directed RPC for disaggregated serving to allow latency and throughput to be traded off dynamically
+3. Strongly separate in-memory and scheduler-directed prefix-caching per replica from durable disaggregated prefix caching to prevent replicas from becoming stateful and to support key sets larger than replica working memory
+4. Use NIXL to abstract high performance point to point KV-cache transfer without a metadata sidechannel
+5. Adapt to existing user deployment patterns and infrastructure vs creating a single monolithic solution
+6. Integrate APIs and core capabilities into upstreams to enforce modularity
+7. Support multiple accelerator hardware architectures with a single set of abstractions, including multiple GPU vendors and systolic architectures like Google TPUs
+
 ### User Stories (Optional)
 
 #### Story 1
@@ -77,7 +87,7 @@ The [unified architecture diagram](https://docs.google.com/drawings/d/1PNGNsicSF
 
 ![Architecture diagram](../assets/images/llm-d-arch-initial-large.svg)
 
-Our current Northstar designs lay out the initial scope (join llm-d-contributors@googlegroups.com to comment). They will be converted into project proposals:
+Our current north star designs lay out the initial scope (join llm-d-contributors@googlegroups.com to comment). They will be converted into project proposals:
 
 * [vLLM-Optimized Inference Scheduler](https://docs.google.com/document/d/1kE1LY8OVjiOgKVD9-9Po96HODbTIbgHp4qgvw06BCOc/edit)
 * [Disaggregated Serving with vLLM](https://docs.google.com/document/d/1FNN5snmipaTxEA1FGEeSH7Z_kEqskouKD1XYhVyTHr8/edit)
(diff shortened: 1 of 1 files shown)


docs/proposals/llm-d.md
```diff
@@ -130,14 +138,28 @@ llm-d intends to use [NIXL](https://github.com/ai-dynamo/nixl) to optimize GPU o
...
 
+Unlike production-stack, llm-d:
+* Is focused on the needs of large scale production serving and expects users to have significant opinions about the rest of the infrastructure.
+
 ### Use KServe
 
 KServe assists platform teams in running large numbers of traditional and generative models on Kubernetes densely. Consider KServe if you have lots of LLM deployments smaller than several hosts or if you have many teams that need distinct deployments of models. llm-d focuses on operationalizing large models in very large deployments, as well as having multiple teams using a single shared deployment efficiently.
+
+Unlike KServe, llm-d:
+* Is focused on the needs of large scale production serving and expects users to desire to consume a limited set of dependencies required for that workload.
```
robertgshaw2-redhat: I would drop the first part and just say:

- "Expects expects users to desire to consume a limited set of dependencies required for their workloads"

terrytangyuan: Instead of adding this sentence which might not be accurate (KServe's raw deployment has minimal dependencies as well), could you merge https://github.com/llm-d/llm-d/pull/112?

docs/proposals/llm-d.md
```diff
@@ -57,6 +57,14 @@ The project will measure success against:
...
 * Reduction of operational toil, especially with increasing workload density
 
+The key design choices in `llm-d` are:
+
+1. Assume multiple workloads are consuming a shared model server pool and standardize those workload APIs via Kubernetes
+2. Leverage scheduler-directed RPC for disaggregated serving to allow latency and throughput to be traded off dynamically
+3. Strongly separate in-memory and scheduler-directed prefix-caching per replica from durable disaggregated prefix caching to prevent replicas from becoming stateful and to support key sets larger than replica working memory
+4. Use NIXL to abstract high performance point to point KV-cache transfer without a metadata sidechannel
+5. Adapt to existing user deployment patterns and infrastructure vs creating a single monolithic solution
+
```
robertgshaw2-redhat, line 69: I think that we should also highlight the cross-platform support here:
- nvidia gpu
- amd gpu
- google tpu

docs/proposals/llm-d.md
```diff
@@ -57,6 +57,14 @@ The project will measure success against:
 * Efficiency of serving (perf/$ at target latency)
 * Reduction of operational toil, especially with increasing workload density
 
+The key design choices in `llm-d` are:
```
vMaroon, line 60: I think one important point missing here is that `llm-d` builds directly on open-source projects and actively integrates with their upstreams. While other projects are moving in that direction, `llm-d` treats this as a core design principle. This makes it much more flexible, especially when customers already have preferences for certain parts of the stack or manners of deployment. To be precise, technically we are not 100% there but are significantly close and it is a design principle. This is definitely strengthened by `llm-d` being a joint effort between leads of the relevant project.

This goes hand in hand with modularity and true pluggability of components. A good example is the ongoing work on the prefix-cache northstar, where the team ensures users only adopt the pieces they need, without being forced into unnecessary dependencies/features.

These points perhaps fit with the 5th.

smarterclayton: Tried to condense this into its own option 6

docs/proposals/llm-d.md
smarterclayton: I think #112 is too verbose.  I don't agree that KServe is seen as minimal dependencies to the users we are discussing, and I do not believe it is modular in the same way we are defining our own modularity.

I've integrated the proposal, please see here.

terrytangyuan: > I think https://github.com/llm-d/llm-d/pull/112 is too verbose

Could you be more specific? #112 has less words so I may misunderstood what your meant by "verbose". I also don't think being verbose is necessarily bad. My intention was to highlight the collaboration.

> I don't agree that KServe is seen as minimal dependencies to the users we are discussing

Could you please elaborate? In case you missed it, here's the installation guide: https://kserve.github.io/website/latest/admin/kubernetes_deployment/

I don't think I fully agree what was written originally here. My comment https://github.com/llm-d/llm-d/pull/83#discussion_r2102653701 (as well as others from KServe community) were unaddressed before merging.

johnugeorge: +1 @smarterclayton Can you explain more about the KServe dependencies that you referred to? It doesn't need any external packages

terrytangyuan:
Please mark my comment https://github.com/llm-d/llm-d/pull/113#discussion_r2150358953 back to unresolved. Thank you.

docs/proposals/llm-d.md
smarterclayton: I am looking through the comment in #83 you mentioned and I'm not sure what is unaddressed.  It would help to highlight that here.

Regarding minimal dependencies - kserve, even the raw deployment, states it requires:

* cert-manager: Users may not wish to leverage cert-manager
* The kserve gateway: Users may prefer to not include the kserve gateway
* kserve controller, which depends on:
  * 9 CRDs
  * 8 webhooks
  
Those are dependencies llm-d and a reasonable portion of the target users would not need.  I agree RawDeployment has fewer dependencies than the knative mode, but in the specific example we are highlighting here the only additional components llm-d requires in the user's environment are the gateway extension components for the inference scheduler.

In terms of verbosity trying to keep all the sections roughly the same length to peer projects, and clearly defining why a user of llm-d should consider those instead.

yuzisun: @smarterclayton KServe’s goal is to support both predictive inference and generative inference. For generative inference we are working on a separate helm chart which does not include all the crds and webhook from predictive inference once kserve/llm-d integration is landed https://github.com/kserve/kserve/pull/4522, so it will be only installing the components needed for llm-d and kserve LLMInferenceService API. 

- kserve does not strictly require cert manager, there is a way to generate self signed cert. However for production I believe most of users are using cert manager. 
- kserve gateway is really an optional pre-configured gateway resource. For llm installation it will be the gateway that supports inference extension API.

Let me know if you have any more questions.

ahg-g: In kServe's 'minimal dependency' mode, it's unclear to me what advantage it offers over using basic deployment APIs. If I'm seeking value from kServe beyond model deployment, wouldn't I likely need to add more dependencies?

yuzisun: KServe API offers seamless cloud model storage integration, model cache, LLM metrics based autoscaling(via KEDA), KV Cache manager. All I am saying is that we only install the component needed for generative inference if that addresses the question Clayton has that llm user may not need all the dependencies from kserve universe. You can read more from this blog post https://kserve.github.io/website/latest/blog/articles/2025-05-27-KServe-0.15-release/

ahg-g: Right, but all those good features bring with them extra dependencies, which is what the point above tries to make, and I agree with. If the user desires to consume a limited set of dependencies, then a reasonable recommendation is to use basic APIs for deploying the model server and IGW.

yuzisun: coming back on this. Deploying LLMs using basic Kubernetes APIs remains complex for most end-users. KServe API is intentionally designed to support wide spectrum of users, offering a low barrier to entry while providing a high ceiling for advanced use cases. KServe's focus is on workload orchestration while the llm-d project's focus on inference efficiency is highly valuable. I think the two projects can work well together with the KServe/llm-d integration https://github.com/orgs/kserve/projects/23.

smarterclayton: Right, I believe they can.  But the llm-d focus is not on the wide spectrum, so I want to capture that.  Let me think about a revision.

smarterclayton: Updated to capture:

* llm-d does not orchestrate inference workloads directly (uses indirect scaling)
* Focused on single workload and MaaS enablement rather than offering a MaaS

docs/proposals/llm-d.md
```diff
@@ -130,14 +140,31 @@ llm-d intends to use [NIXL](https://github.com/ai-dynamo/nixl) to optimize GPU o
...
+
 ### Use KServe
 
-KServe assists platform teams in running large numbers of traditional and generative models on Kubernetes densely. Consider KServe if you have lots of LLM deployments smaller than several hosts or if you have many teams that need distinct deployments of models. llm-d focuses on operationalizing large models in very large deployments, as well as having multiple teams using a single shared deployment efficiently.
+KServe assists platform teams in running large numbers of traditional and generative models on Kubernetes densely. Consider KServe if you have lots of LLM deployments smaller than several hosts or if you have many teams that need distinct deployments of models. llm-d focuses on operationalizing large models in very large deployments, as well as having multiple teams using a single shared deployment efficiently. llm-d intends to offer its capabilities into the KServe ecosystem as a [supporting dependency for serving](https://docs.google.com/document/d/11ZQJ2VhTc42S9K4yau2dMs3Q3f4jqWJL_7Sq14C3hzY/edit?usp=sharing).
```
yuzisun, line 165: KServe is designed to help platform teams efficiently manage and deploy a large number of traditional and generative AI models on Kubernetes. It excels at workload orchestration, while llm-d focuses on operationalizing large models in very large deployments, as well as having multiple teams using a single shared deployment efficiently. llm-d intends to offer its capabilities into KServe ecosystem as a [supporting dependency for serving](https://docs.google.com/document/d/11ZQJ2VhTc42S9K4yau2dMs3Q3f4jqWJL_7Sq14C3hzY/edit?usp=sharing) to provide a more comprehensive solution for running a single shared deployment efficiently, enhancing the overall user experience by combining KServe's robust deployment and management capabilities with llm-d's specialized inference optimizations.

yuzisun:
```suggestion
KServe is designed to help platform teams efficiently manage and deploy a large number of traditional and generative AI models on Kubernetes. It excels at workload orchestration, while llm-d focuses on operationalizing large models in very large deployments, as well as having multiple teams using a single shared deployment efficiently. llm-d intends to offer its capabilities into KServe ecosystem as a [supporting dependency for serving](https://docs.google.com/document/d/11ZQJ2VhTc42S9K4yau2dMs3Q3f4jqWJL_7Sq14C3hzY/edit?usp=sharing) to provide a more comprehensive solution for running a single shared deployment efficiently, enhancing the overall user experience by combining KServe's robust deployment and management capabilities with llm-d's specialized inference optimizations.
```

smarterclayton: Thanks, I will try to integrate the core elements but in fewer words, because it's getting long.  Will have something up shortly.

smarterclayton: Integrated (without the "vs", to keep this section consistent with others).  Please let me know if I captured the core message.

yuzisun: Looks good! Thanks @smarterclayton!

PR #118, opened 2025-06-20
[AMD] Dockerfile for building on AMD hardware

rasmith:
This PR adds support for running on AMD hardware.  This has currently been tested on MI325 and MI300 using single-node and multi-node outside of a k8s cluster.  We are currently working on k8s support, but this work is ongoing.   A few notes:

1) Until ROCm 7 is released, we will need to build `libamdhip64.s`o with a cherry-pick in order for this to work,
2) We need a special build of NIXL for the time being, until there is a more permanent solution.

The goal of adding Dockerfile now is to add support for now that developers can use and experiment with and to also avoid huge PRs in the future and iterate on functional pieces until full support is achieved.

To test, there is a useful `justfile` here:

https://github.com/[tlrmchlsmth/pd_examples](https://github.com/tlrmchlsmth/pd_examples/)/

Set `vllm-directory` to be `/app/vllm` and be sure to set the host/port properly if using a multi-node setup.


rasmith:
> Once you are good with this PR, I would like to merge #189 into this, which will enable builds in CI for it. If you are agreable to this let me know when its ready and I can rebase ... We will also need to configure private access to `nixl-hip` repo for CI if you want to keep that private.

OK, sounds good, I will update when the nixl-hip situation gets sorted out.

Gregory-Pereira:
@rasmith, next steps here are you to rebase against main since we moved our default branch. After this I will open a PR to merge my CI stuff directly into your feature PR, and then the final blocker is the NIXL-HIP stuff which we discussed which we will wait for you to provide a solution on that front. Sound like a plan?

Gregory-Pereira:
Implemented already: https://github.com/llm-d/llm-d/pull/642

PR #119, opened 2025-06-21
distributed tracing proposal

sallyom:
Distributed tracing proposal outlines llm-d components current status wrt opentelemetry instrumentation and the benefits of enabling tracing with each component.


docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,289 @@
...
+#### Story 2
+
+As a cost-conscious organization using llm-d, I want to track token usage and costs per application and request type so that
+I can optimize my prompt engineering and model selection to reduce operational expenses.
+
+#### Story 3
+
+As an llm-d developer validating new optimizations, I want to measure the impact of KV-cache aware routing and P/D disaggregation on request 
+latency so that I can quantify the benefits of these advanced features.
+
```
frzifus, line 60:
```suggestion

#### Story 4

As an llm-d developer/administrator, I have noticed a significant change in performance since the last upgrade. I want to compare the execution, caching and decision-making in routing between the two releases.
```

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,289 @@
...
+  * **Component Architecture**: Provides the Endpoint Picker Protocol (EPP) that integrates with the 
+Inference Gateway. It operates as a gRPC service that receives routing requests from the gateway and makes 
+intelligent endpoint selection decisions.
+
+  * **Instrumentation Focus**: This component is responsible for making smart load-balancing and routing decisions,
+applying filtering and scoring algorithms based on awareness of P/D, KV-cache, SLA, and load.
+
+  * **Requests to Instrument**: Trace the lifecycle of routing requests as they:
+    - Enter the EPP gRPC service from the Inference Gateway
+    - Flow through the scheduler's internal logic
```
frzifus: Not sure if this might become overwhelming.  Maybe it makes sense to focus on the ingress/egress part first and evolve over time. wdyt?

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,289 @@
...
+* **Granular Control**: While top-level enablement is unified, individual components retain fine-grained controls.
+
+* **Scalability**: Single collector endpoint configuration supports scalable trace ingestion and forwarding to various observability backends.
+
+### Configuration
+
+```yaml
+tracing:
+  enabled: false
+  otelCollectorEndpoint: "http://otel-collector:4317"
```
frzifus: Are we always using gRPC? If thats not the case, we may want to provide a few more options. E.g. also for TLS.
```suggestion
  otelEndpoint: "otel-collector:4317"
  otelProtocol: "grpc"
  tls:
    enabled: false
    ...
```
or control the proto via scheme: `http`, `https`, `grpc` and `grpcs`. 
```
  otelEndpoint: "grpc://otel-collector:4317"
```

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,289 @@
...
+
+The instrumentation strategy focuses on the critical path of LLM inference requests through the llm-d stack,
+covering key components responsible for routing, caching, and serving.
+
+**Current Implementation Status Summary:**
+- **Inference Gateway (gateway-api-inference-extension)**: Has OpenTelemetry dependencies, needs activation
+- **llm-d-inference-scheduler (EPP)**: Has OpenTelemetry dependencies, needs activation
+- **llm-d-kv-cache-manager**: No OpenTelemetry instrumentation
+- **Routing Proxy (llm-d-routing-sidecar)**: No OpenTelemetry instrumentation  
+- **vLLM v1**: No tracing support, was explicitly removed in v1 (v0.x had tracing support)
```
frzifus: I think we also can not simply copy it.
- https://github.com/vllm-project/vllm/issues/17528

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,289 @@
...
+- `llm_d.cache.hit_ratio`: KV cache hit ratio
+- `llm_d.routing.decision_time`: Routing algorithm duration
+
+## Alternatives
+
+### Manual Instrumentation Per Component
+
+Platform operators could manually instrument each llm-d component independently, configuring OpenTelemetry for each service separately.
+While this provides maximum flexibility, it significantly increases operational complexity and error surface, particularly for correlating
+traces across the distributed serving pipeline.
```
frzifus: In case a single component is not instrumented, it would cut the relation between the traces, right? Due to the lag of context propagation. Even when disabling traces for a single component, it should still propagate the trace header.

docs/proposals/distributed-tracing.md
sallyom: agreed - let's capture this in the proposal.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Configuration
+
+```yaml
+tracing:
+  enabled: false
+  otelCollectorEndpoint: "http://otel-collector:4317"
+  apiToken: ""
+  samplingRate: 0.1
+  detailedTracing: false
```
frzifus: What does it mean?

docs/proposals/distributed-tracing.md
frzifus: How is it solved?

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+  # Always propagate trace context, even when component tracing is disabled
+  alwaysPropagateContext: true
```
frzifus:
```suggestion
```
I assume that would be the case if traceing is enabled, but no endpoint to export the data is set?

docs/proposals/distributed-tracing.md
sallyom: I updated to prioritize simple e2e visibility over comprehensive tracing with each component with this proposal.

docs/proposals/distributed-tracing.md
sallyom: left over from a previous change, please remove it or i'll remove it with next push, ty!

docs/proposals/distributed-tracing.md
sallyom: added a section about Trace Context Propagation

docs/proposals/distributed-tracing.md
frzifus:
```suggestion
```

sallyom:
I've instrumented `llm-d-kv-cache-manager` and `llm-d-inference-scheduler` and have a few traces being generated, creating a PoC to work from
See from this trace:
  - Root Span: epp.kvcache_scorer.score (26.42μs) - The EPP scoring operation
  - Child Span: kv-cache-manager.GetPodScores (9.53μs) - The KV cache manager call
  
Screenshot 2025-06-30 at 4 41 45 PM

sallyom:
I've instrumented `llm-d-kv-cache-manager`, `llm-d-inference-scheduler`, `gateway-api-inference-extension`, and `llm-d-routing-sidecar` repos for a PoC (all sallyom/tracing branches). I've built 
* **quay.io/sallyom/llm-d-inference-scheduler:tracing** 
    * local checkouts of the llm-d-kv-cache-manager & gateway-api-inference-extension tracing branches
* **quay.io/sallyom/llm-d-routing-sidecar:tracing**

I updated `llm-d-deployer` with a tracing option & deployed with my images, then:

Screenshot 2025-07-01 at 10 36 56 PM

This shows:
Screenshot 2025-07-01 at 10 39 04 PM

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+applying filtering and scoring algorithms based on awareness of P/D, KV-cache, SLA, and load.
+
+  * **Requests to Instrument**:
+    - Request entry into the EPP gRPC service from the Inference Gateway
+    - Final endpoint selection response back to the Inference Gateway
+    - Request duration and basic success/failure status
+ 
+  * **Key Spans and Attributes**:
+    - EPP gRPC request receipt and processing
+    - Final endpoint selection decision
```
pierDipi: Does final here mean decode pod selection?

I think another key aspect is the prefill pod selection, should we also add the prefill pod selection decision as attribute?

https://github.com/llm-d/llm-d-inference-scheduler/blob/881aaa05dc2b4319f405506cf56cb523065ec940/pkg/plugins/pre-request/pd_prerequest.go#L58-L68

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+prefill worker. This component is deployed when P/D disaggregation is enabled.
+
+  * **Requests to Instrument**:
+    - Request entry into the routing proxy
+    - Request forwarding to prefill worker
+    - Request duration and basic success/failure status
+
+  * **Key Spans and Attributes**:
+    - Routing proxy request receipt and processing
+    - Prefill worker selection and forwarding
```
pierDipi: prefill worker selection is done in the scheduler 

https://github.com/llm-d/llm-d-inference-scheduler/blob/881aaa05dc2b4319f405506cf56cb523065ec940/pkg/plugins/pre-request/pd_prerequest.go#L58-L68

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+* **Enhanced Performance Diagnostics**: Provide detailed, request-level visibility into llm-d pipeline bottlenecks, enabling optimization of TTFT,
+ITL, and overall throughput across distributed serving components.
+
+* **Cost Efficiency and Attribution**: Enable per-request token usage tracking and cost attribution across applications and workloads, crucial for
+managing high LLM computational costs.
+
+* **Quality and Accuracy Validation**: Enable inspection of LLM inputs/outputs and validation of response quality, particularly in complex RAG
+pipelines where context enrichment occurs across multiple services.
+
+* **Simplified Debugging**: Provide end-to-end request tracing across llm-d components, to reduce mean time to resolution (MTTR) for performance
```
pavolloffay, line 29: also enhanced root cause analysis

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+degradation and error scenarios.
+
+* **Optimization Validation**: Provide concrete, per-request data to validate the effectiveness of llm-d's advanced optimizations like KV-cache aware
+routing and disaggregated serving.
+
+### Non-Goals
+
+* **Comprehensive Internal Instrumentation**: This proposal focuses on end-to-end visibility through ingress/egress tracing, not exhaustive instrumentation of every internal operation, function call, or database query within components.
```
pavolloffay: I would alter this sentence.

Tracing usually brings the highest value when it is appled to API/RPC calls and database queries.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+
+* **Metrics Collection**: This proposal focuses on distributed tracing, not metrics collection, though OpenTelemetry collectors can export to both.
```
pavolloffay: A single opentelemetry instrumentation can emit metrics data OOTB from instrumented processes e.g. for HTTP servers. 

The collector can extract metrics from collected traces in some scenarios. Essentially gives users important RED metrics without direct metrics instrumentation.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+
+* **Log Aggregation**: While OpenTelemetry supports logs, this proposal addresses distributed tracing only.
+
+* **Real-time Alerting**: Tracing data analysis and alerting are out of scope, though trace data can feed into alerting systems.
```
pavolloffay: usually metrics are exported from traces which enables alerting.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+The tracing implementation will instrument key llm-d components: the llm-d-inference-scheduler, llm-d-kv-cache-manager,
+routing proxy, vLLM instances, and inference gateway. Comprehensive instrumentation enables validation of llm-d's
+optimizations while maintaining operational simplicity through centralized configuration.
+
+### User Stories
+
+#### Story 1
+
+As a platform operator running llm-d in production, I want to quickly identify which component in my distributed serving
+pipeline is causing high latency so that I can optimize resource allocation and meet my SLAs.
```
pavolloffay: is causing high latency and properly identify root-cause of the problem

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+* **Granular Control**: While top-level enablement is unified, individual components retain fine-grained controls.
+
+* **Scalability**: Single collector endpoint configuration supports scalable trace ingestion and forwarding to various observability backends.
+
+### Configuration
+
+```yaml
+tracing:
+  enabled: false
+  otelCollectorEndpoint: "http://otel-collector:4317"
```
pavolloffay: I would be more specific (protocol wise) and create a section for the exporter which can be extended for auth. 

e.g.

```yaml
exporter:
  endpoint: 
  tls:
   ....
  headers:
 ...
```

It's also better option to start with the OTLP http which is more commonly used and works better across ingresses.

docs/proposals/distributed-tracing.md
pavolloffay: > it significantly increases operational complexity and error surface, particularly for correlating 

the correlation might not be even possible if the context propag is broken

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Manual Instrumentation Per Component
+
+Platform operators could manually instrument each llm-d component independently, configuring OpenTelemetry for each service separately.
+While this provides maximum flexibility, it significantly increases operational complexity and error surface, particularly for correlating
+traces across the distributed serving pipeline.
+
+### Third-party APM Solutions
+
+Commercial APM solutions could provide automatic instrumentation. However, these solutions may lack the GenAI-specific semantic
```
pavolloffay: Most vendors already base their agents on otel instrumentation anyways

sallyom:
Draft PRs for collaboration:
* https://github.com/llm-d/llm-d-inference-scheduler/pull/231
    * https://github.com/llm-d/llm-d-kv-cache-manager/pull/48
    * https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1125
* https://github.com/llm-d/llm-d-routing-sidecar/pull/39 

vLLM v1 tracing PR:
* https://github.com/vllm-project/vllm/pull/20372

docs/proposals/distributed-tracing.md
sallyom: yes, final == decode, I clarified.
added span in the PoC tracing branch at pkg/plugins/pre-request/pd_prerequest.go - see https://github.com/llm-d/llm-d-inference-scheduler/pull/231

docs/proposals/distributed-tracing.md
sallyom: updated, ty

docs/proposals/distributed-tracing.md
JeffLuoo: I have created a FR for it in vLLM: https://github.com/vllm-project/vllm/issues/17794

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,449 @@
...
+      - Request parsing and validation span
+      - Model instance selection span  
+      - Request/response transformation spans
+    - **Advanced Attributes**:
+      - `gen_ai.usage.input_tokens`: Input token count (requires parsing)
+      - `gen_ai.usage.output_tokens`: Output token count (from response)
+      - `llm_d.gateway.routing_algorithm`: Algorithm used for selection
+      - `llm_d.gateway.load_balancing_decision`: Load balancing choice
+      - `llm_d.gateway.request_size`: Request payload size
+      - `llm_d.gateway.response_size`: Response payload size
```
JeffLuoo: We already have the input and output tokens count in attributes. I wonder what would be the use case of llm_d.gateway request and response size? If this is for tracking the payload size in bytes, can we propose to add it as common attributes in https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/ like `gen_ai.usage.request_size` and `gen_ai.usage.response_size`?

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+#### **Component: `llm-d-inference-scheduler (Endpoint Picker Protocol)`**
+
+  * **Component Architecture**: Provides the Endpoint Picker Protocol (EPP) that integrates with the 
+Inference Gateway. It operates as a gRPC service that receives routing requests from the gateway and makes 
+intelligent endpoint selection decisions.
+
+  * **Instrumentation Focus**: This component is responsible for making smart load-balancing and routing decisions,
+applying filtering and scoring algorithms based on awareness of P/D, KV-cache, SLA, and load.
+
+  * **Auto-instrumentation Implementation**:
```
JeffLuoo: Can you clarify on the meaning of `Auto-instrumentation` here? It may refer to using the Go zero-code instrumentation https://opentelemetry.io/docs/zero-code/go/.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+    - **Tracing Infrastructure**: Auto-instrumentation using `otel.Tracer("llm-d-inference-scheduler")` without explicit initialization
+    - **Initial Spans**: 
+      - `epp.pd_prerequest`: P/D disaggregation pre-request plugin operation
+    - **Basic Attributes**: 
+      - `llm_d.pd.disaggregation_enabled`: Whether P/D disaggregation is active
+      - `llm_d.pd.prefill_pod_address`: Selected prefill pod address (when applicable)
+      - `operation.outcome`: success/error
```
JeffLuoo: Can we just use the status of the span?

docs/proposals/distributed-tracing.md
damemi: yes this is the intent

JeffLuoo: I see. Thanks for the info. I saw the refreshed version of the proposal the phase 1 will be using auto-instrumentation to generate basic tracing and it's the future enhancements to have detailed span. That makes sense to me!

damemi: @JeffLuoo no problem, to clarify auto-instrumentation can do detailed spans as well. It's possible to "manually instrument" (that is, add the span code, attributes, etc) without actually instantiating a tracerprovider (so the code on its own is a no-op with 0 overhead). The auto-instrumentation can then hook into this no-op code to provide detailed spans with all of the overhead offloaded to ebpf (so it is still a no-op in the code itself)

JeffLuoo: @damemi For ebpf here, do you refer to the epf of go zero-code sdk or other ebpf stacks like beyla?

damemi: @JeffLuoo I am referring specifically to the go zero-code sdk, but for reference that otel zero-code sdk is what Beyla/OBI uses to provide Go ebpf instrumentation. I'm proposing we use the zero-code sdk directly rather than take a dependency on a full tool like that

.version.json
```diff
@@ -1,6 +1,6 @@
 {
-  "dev-version": "v0.2.1",
+  "dev-version": "v0.2.2",
   "dev-registry": "ghcr.io/llm-d/llm-d-dev",
-  "prod-version": "v0.2.0",
+  "prod-version": "v0.2.1",
```
smarterclayton: I assume this is accidental?

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+## Motivation
+
+LLM inference workloads present unique observability challenges due to their expensive, non-uniform, and often slow request patterns. In distributed
+systems like llm-d, understanding request flow across components like the inference scheduler, KV cache manager, and vLLM instances becomes critical
+for operationalizing inference at scale.
+
+Current monitoring approaches lack the granular, request-level visibility needed to optimize Time to First Token (TTFT), Inter-Token Latency (ITL),
+and cost efficiency in complex serving topologies involving disaggregated serving, KV-cache aware routing, and multi-model deployments.
+
+### Goals
```
smarterclayton, line 18: Are there security implications to this work?  Is there any delta in our use case vs normal tracing?

I.e. the query / response for many models is very sensitive.  What care does llm-d need to take to ensure that data is treated as such, even during tracing?  What implications on this design come from that sensitivity?

A few sentences in the proposal plus some goals and non-goals (and justifications why) are important when we are discussing `production serving`.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+covering key components responsible for routing, caching, and serving.
+
+**Implementation Approach:**
+Initial implementation will focus on ingress/egress instrumentation to establish end-to-end visibility with minimal complexity.
+Implementation prioritizes request entry and exit points from each component rather than internal operation tracing.
+
+**Auto-instrumentation Implementation:**
+The llm-d stack will implement distributed tracing using an auto-instrumentation approach with external observability agents:
+- **llm-d-inference-scheduler (EPP)**: Auto-instrumentation support with P/D disaggregation pre-request plugin instrumentation
+- **llm-d-kv-cache-manager**: Auto-instrumentation with utility functions and Redis instrumentation via redisotel
```
smarterclayton: Is the redis work really high priority?  I thought we were moving to a world where we didn't need that metadata store in favor of direct scheduler integration?

From a completeness perspective its fine, but I would have thought we were more aggressively moving to remove the need for the additional component in a distributed system?

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+- **Routing Proxy (llm-d-routing-sidecar)**: Auto-instrumentation support with minimal tracing package
+- **vLLM v1**: No tracing support (was removed in v1, though v0.x had tracing support)
```
smarterclayton: This will be a significant gap, given that our goals include:

1. Removing the routing sidecar
2. The attempt to further disaggregate vLLM (the work @NickLucche is doing) will involve calls through vLLM

At a first pass, I don't think these two are quite correct, I'd probably make vLLM in scope with Nicolo's work.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+- `gen_ai.request.model`: Model identifier
+- `gen_ai.usage.input_tokens`: Input token count (where available)
+- `gen_ai.usage.output_tokens`: Output token count (where available)  
+- `operation.outcome`: success/error/timeout
+- Request duration (automatic via span timing)
+
+**llm-d Specific Attributes**:
+- `llm_d.kv_cache.hit_ratio`: KV cache hit ratio
+- `llm_d.kv_cache.pod_count`: Number of pods considered
+- `llm_d.pd.disaggregation_enabled`: P/D disaggregation status
```
smarterclayton: This is almost certainly going to change later.  That's probably ok, but it would be better to scope this to a specific versioned component like `llmd_d.pd_proxy` and stay focused on how it's implemented today.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+- `llm_d.proxy.connector`: Connector type (nixlv2, lmcache, nixl)
```
smarterclayton: Suggest changing this to `llm_d.pd_proxy` so we can version and remove it later.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+  * **Instrumentation Focus**: This component manages a global view of KV cache states and localities, for optimizing LLM inference by reusing
+computed key/value attention vectors. It interacts with storage to index KV block availability.
+
+  * **Auto-instrumentation Implementation**:
+    - **Tracing Infrastructure**: Auto-instrumentation using `otel.Tracer("llm-d-kv-cache-manager")` without explicit initialization
+    - **Initial Spans**: `GetPodScores` operation (entry → response)
+    - **Basic Attributes**:
+      - `gen_ai.request.model`: Model identifier
+      - `llm_d.kv_cache.hit_ratio`: Cache hit ratio for the request
+      - `llm_d.kv_cache.pod_count`: Number of pods considered
```
smarterclayton: Make this more specific to the component that is sending it, kv-cache-manager is *not* "generic" to all kv-caching.

Suggest `llm_d.kv_cache_manager` instead

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+  * **Instrumentation Focus**: This component is responsible for making smart load-balancing and routing decisions,
+applying filtering and scoring algorithms based on awareness of P/D, KV-cache, SLA, and load.
+
+  * **Auto-instrumentation Implementation**:
+    - **Tracing Infrastructure**: Auto-instrumentation using `otel.Tracer("llm-d-inference-scheduler")` without explicit initialization
+    - **Initial Spans**: 
+      - `epp.pd_prerequest`: P/D disaggregation pre-request plugin operation
+    - **Basic Attributes**: 
+      - `llm_d.pd.disaggregation_enabled`: Whether P/D disaggregation is active
+      - `llm_d.pd.prefill_pod_address`: Selected prefill pod address (when applicable)
```
smarterclayton: As commented below, make this `llm_d.pd_proxy` to make it specific.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+
+**Implementation Approach:**
+Initial implementation will focus on ingress/egress instrumentation to establish end-to-end visibility with minimal complexity.
+Implementation prioritizes request entry and exit points from each component rather than internal operation tracing.
+
+**Auto-instrumentation Implementation:**
+The llm-d stack will implement distributed tracing using an auto-instrumentation approach with external observability agents:
+- **llm-d-inference-scheduler (EPP)**: Auto-instrumentation support with P/D disaggregation pre-request plugin instrumentation
+- **llm-d-kv-cache-manager**: Auto-instrumentation with utility functions and Redis instrumentation via redisotel
+- **Routing Proxy (llm-d-routing-sidecar)**: Auto-instrumentation support with minimal tracing package
```
smarterclayton: This component has three names and all of them are vague and bad. :)

It is not a routing sidecar.  It is not a routing proxy.  It is not a proxy.

Suggest for the sake of future clarity we call this the `pd_proxy` or `pd_sidecar` going forward, and update the text both here (for spans) and in the repo.  The name should be descriptive enough for it not to be confused with any other component, but it's ok to keep it relatively terse since we think it will go away in time.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+      - `gen_ai.usage.input_tokens`: Input token count
+      - `gen_ai.usage.output_tokens`: Output token count
+      - `operation.outcome`: success/error/timeout
+    - **Context Propagation**: Extract incoming trace context from inference requests using global tracer
+    - **Benefit**: Establishes vLLM visibility in end-to-end traces, essential token usage for cost attribution
+
+  * **Future Enhancement Opportunities**:
+    vLLM contributors could add detailed spans for request preprocessing, model execution, and response generation. Enhanced attributes could include TTFT/ITL measurements, cache utilization metrics, and batch processing details.
+
+#### **Component: `Inference Gateway (gateway-api-inference-extension)`**
```
smarterclayton: EDIT: Technically this is inference scheduler from llm-d perspective.  I don't think the gateway should write spans that say "inference scheduler" though, but I'm also unclear whether it should use spans that reference it as inference gateway vs the more component specific spans (`epp`).

Suggest here we clarify that we are talking about the inference-scheduler component in llm-d, which is an endpoint picker in inference gateway, and use `epp` or `igw_epp`  as the span identifier (i.e. more specific to what the component actually is).

docs/proposals/distributed-tracing.md
smarterclayton: This would be a good sentence to add in the proposal, since it strengthens the argument.

docs/proposals/distributed-tracing.md
smarterclayton: If these are stronger arguments than what is in the proposal, please integrate them into the proposal.

docs/proposals/distributed-tracing.md
Frapschen: OTel have span status, if you want add extral messages you can add them to span event

docs/proposals/distributed-tracing.md
Frapschen: I doubt that auto-instrumentation (eBPF-generated) spans will be able to join with manually created spans in the future. Perhaps we should create a proof of concept (POC) to verify this.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+  * **Auto-instrumentation Implementation**:
+    - **Tracing Infrastructure**: Auto-instrumentation using `otel.Tracer("llm-d-kv-cache-manager")` without explicit initialization
+    - **Initial Spans**: `GetPodScores` operation (entry → response)
+    - **Basic Attributes**:
+      - `gen_ai.request.model`: Model identifier
+      - `llm_d.kv_cache.hit_ratio`: Cache hit ratio for the request
+      - `llm_d.kv_cache.pod_count`: Number of pods considered
+      - `operation.outcome`: success/error/timeout
+    - **Context Propagation**: Maintains trace context across cache operations using global tracer
+    - **Redis Instrumentation**: Automatic tracing of Redis operations via `redisotel`
```
Frapschen: current [opentelemetry-go-instrumentation](https://github.com/open-telemetry/opentelemetry-go-instrumentation)  can't automatic instrument redis. Only packages listed in https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/internal/pkg/inject/offset_results.json can do it.

we can find the instruemt sdk at https://opentelemetry.io/ecosystem/registry/?language=go&s=redis

docs/proposals/distributed-tracing.md
Frapschen: Should we also add a attribute to record which pod has selected?

Frapschen: OTel attribtue support list value, we can save all the considered pods into a attribtue which is more clarity than just pod count.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+The implementation follows OpenTelemetry semantic conventions for GenAI operations:
+
+**Core Attributes** (implemented in auto-instrumentation):
+- `gen_ai.request.model`: Model identifier
+- `gen_ai.usage.input_tokens`: Input token count (where available)
+- `gen_ai.usage.output_tokens`: Output token count (where available)  
+- `operation.outcome`: success/error/timeout
+- Request duration (automatic via span timing)
+
+**llm-d Specific Attributes**:
```
Frapschen: Since LLM-D has multiple components, I suggest giving each component a consistent prefix:

 component | component prefix |
|--------|--------|
| llm-d-kv-cache-manager | kv_cache | 
| llm-d-inference-scheduler | TODO | 
| Routing Proxy | TODO |

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,310 @@
...
+#### Story 3
+
+As an llm-d developer validating new optimizations, I want to measure the impact of KV-cache aware routing and P/D disaggregation on request 
+latency so that I can quantify the benefits of these advanced features.
+
+#### Story 4
+
+As an llm-d developer/administrator, I have noticed a significant change in performance since the last upgrade. I want to compare the execution, caching and decision-making in routing between the two releases.
+
+## Design Details
```
Frapschen, line 104: We should also consider the sampling strategy for LLM-d.

As a subsystem of the LLM backend service, the upstream may already be instrumented with spans. Requests sent to LLM-d will carry parent span information along with the sampling decision. Should LLM-d adopt a "parent-based" sampling strategy, or an "always-sample" approach?

Kubernetes follows a practice of linking upstream span information as a "span link" and independently setting its own sampling rate.

docs/proposals/distributed-tracing.md
damemi: @Frapschen they do join with manually created spans :) The OTel SDK is instrumented just like any other library, and the context is propagated across manual and auto spans

docs/proposals/distributed-tracing.md
sallyom: added to this section.

docs/proposals/distributed-tracing.md
sallyom: added

docs/proposals/distributed-tracing.md
sallyom: updated (added you as co-author to latest commit, too)

docs/proposals/distributed-tracing.md
sallyom: added, and with you as co-author to this commit,ty

docs/proposals/distributed-tracing.md
sallyom: added

docs/proposals/distributed-tracing.md
sallyom: added

docs/proposals/distributed-tracing.md
sallyom: I've updated to remove this favoring the operator Instrumentation instead

docs/proposals/distributed-tracing.md
sallyom: added

docs/proposals/distributed-tracing.md
sallyom: removed, updated

.version.json
sallyom: updated - this file is removed

docs/proposals/distributed-tracing.md
sallyom: added a security section 
added you as co-author, too

docs/proposals/distributed-tracing.md
sallyom: I've removed the Redis mentions.

docs/proposals/distributed-tracing.md
sallyom: updated

docs/proposals/distributed-tracing.md
sallyom: updated

docs/proposals/distributed-tracing.md
sallyom: updated

docs/proposals/distributed-tracing.md
sallyom: updated

docs/proposals/distributed-tracing.md
sallyom: i've removed redis instrumentation from proposal, ty

docs/proposals/distributed-tracing.md
sallyom: added some details on "Sampling Strategy"

sallyom:
I've added several co-authors since I would not have known many of the specific details included in this proposal without them:
/cc @smarterclayton @damemi @pavolloffay @JeffLuoo @Frapschen @pierDipi @frzifus 

    Co-Authored-By: Claude <noreply@anthropic.com>
    Co-Authored-By: ploffay <ploffay@redhat.com>
    Co-Authored-By: smarterclayton <smarterclayton@gmail.com>
    Co-Authored-By: Frapschen <minquan.chen@daocloud.io>
    Co-Authored-By: JeffLuoo <jeffluoo@google.com>
    Co-Authored-By: damemi <mike@odigos.io>
    Co-Authored-By: frzifus <ben@klimlive.de>
    Co-Authored-By: PierDipi <pdipilat@redhat.com>

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,424 @@
...
+
+The implementation establishes end-to-end tracing across llm-d components using auto-instrumentation patterns. The approach progresses from proven vLLM v1 tracing support to proposed implementations in llm-d components.
+
+**Current Implementation Status:**
+- **vLLM v1**: Tracing support with `from vllm.tracing import init_tracer`
+- **llm-d Components**: Working branch implementations demonstrate feasibility of auto-instrumentation approach
+
+**Proposed Auto-instrumentation Pattern:**
+Based on working branch prototypes, each component implements:
+- Auto-instrumentation using `otel.GetTracerProvider().Tracer()` without explicit initialization
```
Frapschen: I'm a little confused, for my knowledge, `otel.GetTracerProvider().Tracer()` is more like the manual instrumentation.

As the manifest illustrated

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/6e8e0c1849cea07253c822b2d8c5ad1758544343/docs/tutorial/emojivoto-instrumented.yaml#L41-L50

The auto-instrumentation approach only runs a sidecar with the target go binary:

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,424 @@
...
+- Context propagation is lightweight and stateless
+- Header extraction/injection operations are constant-time
+- No spans are created or exported without an active TraceProvider
+
+### Semantic Conventions and Attributes
+
+The implementation follows OpenTelemetry semantic conventions for GenAI operations:
+
+**Core Attributes** (implemented in auto-instrumentation):
+- `gen_ai.request.model`: Model identifier
```
Frapschen: Have we tested that we can add custom attributes with auto-instrument? I can't find any document or demo to illustrate this feature.

docs/proposals/distributed-tracing.md
damemi: @Frapschen go auto instrumentation provides the Auto SDK which allows you to define custom spans that are processed by ebpf and linked to ebpf spans. The agent does run as a sidecar, but it essentially instruments the `tracer.Start()` and `tracer.End()` functions with a special tracerprovider. We are adding docs on this here https://github.com/open-telemetry/opentelemetry.io/pull/7842

docs/proposals/distributed-tracing.md
damemi: Yes, see comment above https://github.com/llm-d/llm-d/pull/119#discussion_r2379773567

docs/proposals/distributed-tracing.md
sallyom: updated

docs/proposals/distributed-tracing.md
JeffLuoo: I still see the attribute here..

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,424 @@
...
+    - **Context Propagation**: Maintains trace context across cache operations using global tracer
+    - **Benefit**: Establishes KV cache manager visibility in end-to-end traces with zero configuration
+
+  * **Future Enhancement Opportunities**:
+    Component owners can add detailed spans for cache lookup operations and scoring algorithms. Enhanced attributes could include cache hit metrics and lookup timing data.
+
+#### **`P/D Proxy (llm-d-routing-sidecar)` - Transitional**
+
+  * **Instrumentation Focus**: This component currently acts as a reverse proxy for P/D (Prefill/Decode) disaggregation. However, this component is
+    planned for removal as part of the architectural evolution toward direct vLLM disaggregation.
```
JeffLuoo: Can you link a bug or a public for the discussion of removing it from the architecture? Otherwise, we don't know when the component will be removed.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,424 @@
...
+    - **Spans**: vLLM inference request processing with configurable tracing backend
+    - **Attributes**: Model execution metadata and performance characteristics
+    - **Context Propagation**: Native trace context handling through vLLM processing pipeline
+    - **Integration**: Works directly with OpenTelemetry without requiring external auto-instrumentation agents
+    - **Security Compliance**: vLLM's tracing implementation fully aligns with the security goals outlined in this proposal, capturing only performance metrics, token counts, and request parameters while avoiding any prompt or completion content
+
+  * **Enhanced Integration Opportunities**:
+    Future vLLM disaggregation work will enable direct tracing through vLLM components, eliminating dependency on P/D proxy patterns and providing native P/D tracing visibility.
+
+#### **`Inference Gateway (gateway-api-inference-extension)`**
```
JeffLuoo: Shall we place `llm-d-inference-scheduler` a sub-component of Inference Gateway (or merge in an other way)? I am a little confused about the role of both llm-d-inference-scheduler and gateway-api-inference-extension.

docs/proposals/distributed-tracing.md
```diff
@@ -0,0 +1,424 @@
...
+
+LLM inference workloads process highly sensitive data including proprietary prompts, personal information, confidential business data, and intellectual property.
+LLM queries and responses frequently contain:
+
+- Confidential communications
+- Personal identifiable information (PII) and regulated data
+- Proprietary code, algorithms, and technical specifications
+- Sensitive healthcare, financial, or legal information
+
+This sensitive data requires specialized handling in observability systems to prevent inadvertent exposure through trace data.
```
JeffLuoo: Can you provide some examples of which specialized handling we can use in the pipeline? Does OTel have any built-in plugins from the collector that can sanitize the data?

docs/proposals/distributed-tracing.md
sallyom: GAIE and vLLM now include otel trace initialization, this proposal is moving to manual instrumentation over zero-code/auto.

PR #120, opened 2025-06-23
Model caching proposal

lionelvillard:
This is the model caching proposal owned by the autoscaling sig

@eilamt @chcost

### DIFF
--- a/docs/proposals/modelcaching.md
+++ b/docs/proposals/modelcaching.md
@@ -1,0 +1,102 @@
+# Model Caching Proposal
+
+## Summary
+
+This proposal introduces a new llm-d component controlling the location of models to make vLLM startup time faster and more reliable. It also includes an implementation design focusing on caching models to local storage, with plans to expand functionalities based on community feedback.
+
+# Motivation
+
+vLLM startup time can greatly be reduced when models are cached on local storage. All supported vLLM model loaders, safetensor (builtin), [fastsafetensors](https://docs.vllm.ai/en/latest/models/extensions/fastsafetensor.html?h=fastsafete), [Run.ai model streamer](https://docs.vllm.ai/en/latest/models/extensions/runai_model_streamer.html), [Coreweave’s tensoriser](https://docs.vllm.ai/en/latest/models/extensions/tensorizer.html) support loading models from local storage. Each has consistently shown faster startup compared to loading models from HuggingFace (see [\#1](https://github.com/run-ai/runai-model-streamer/blob/master/docs/src/benchmarks.md), [\#2](https://github.com/vllm-project/vllm/pull/10647), [\#3](https://v1.docs.coreweave.com/coreweave-machine-learning-and-ai/inference/tensorizer#benchmarks)).
+
+In addition, loading models from local storage is considered more reliable than loading from the network (eg. S3, Huggingface, etc...), as it reduces dependency on external services, mitigates issues related to connectivity and latency, and makes system operations more predictable.
+
+# Goals
+
+* Be agnostic about which model loader is being used.
+* Provide a low-level API to cache models on local storage.
+
+# Non-Goals
+
+* Provisioning Persistent Volumes, which is better handled by existing external static provisioners, such as [Local Persistent Volume Static Provisioner](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner), [Openshift Local Storage Operator](https://catalog.redhat.com/software/container-stacks/detail/66993e0f4f284e980ee072d3) and [Open EBS](https://openebs.io).
+* Supporting other storage types such as S3, HTTP, etc…
+* Supporting model revision (GIE does not support it yet).
+* Providing an high-level CRD-based API.
+* Supporting sharing models across namespaces.
+
+# Proposal
+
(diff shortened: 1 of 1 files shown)


docs/proposals/modelcaching.md
```diff
@@ -0,0 +1,102 @@
...
+    requests:
+      storage: 100Gi
+  storageClassName: local-sc
+```
+
+## Downloading Models to Local Storage
+
+This proposal introduces a new controller for applying the desired LPVCs state as specified by the annotations defined above. This controller watches LPVCs in the cluster and performs the following actions depending on its operating mode:
+
+* **agent mode**. An agent is created for each local Persistent Volume Claim (PVC). This agent is deployed on the node matching the PVC. It monitors and updates the PVC’s annotations and ensures that the required models are cached in the PV.
```
jeremyeder, line 70: Here you say the agent ensures models are cached, and then you mention volume populator below. Is that what the agents are intended to use?

docs/proposals/modelcaching.md
```diff
@@ -0,0 +1,102 @@
...
+* Support for other storage types (S3, HTTP, etc..).
+* Reformat model files to be compatible with specific model loaders (eg. Tensorizer) and sharding techniques (TP, PP).
+* Caching pytorch compilation artifacts.
+* Better UX.
+
+# Alternatives
+
+## Reusing [KServe Local Model Cache](https://kserve.github.io/website/master/modelserving/storage/modelcache/localmodel/)
+
+KServe local model cache provides an opinionated way to cache models on local storage and has been designed to be an optional KServe extension, as opposed to an independent component. This proposal borrows many ideas from KServe local model cache and adapts them to fit nicely with llm-d.
```
terrytangyuan, line 90: We've recently started discussing this on CNCF Slack: [thread](https://cloud-native.slack.com/archives/C090YR4B6BF/p1751395558138719). It would be great if we can reuse what KServe already has.

cc @yuzisun

docs/proposals/modelcaching.md
MikeSpreitzer, line 70: An agent is created _by what_? When?

Really for every PVC in the namespace?

docs/proposals/modelcaching.md
```diff
@@ -0,0 +1,102 @@
...
+# Goals
+
+* Be agnostic about which model loader is being used.
+* Provide a low-level API to cache models on local storage.
+
+# Non-Goals
+
+* Provisioning Persistent Volumes, which is better handled by existing external static provisioners, such as [Local Persistent Volume Static Provisioner](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner), [Openshift Local Storage Operator](https://catalog.redhat.com/software/container-stacks/detail/66993e0f4f284e980ee072d3) and [Open EBS](https://openebs.io).
+* Supporting other storage types such as S3, HTTP, etc…
+* Supporting model revision (GIE does not support it yet).
```
MikeSpreitzer, line 22: By the time the bits hit the GPU, a particular version will have been selected. Let's give users control over that.

MikeSpreitzer: Specifying a branch name is already known to be a vulnerability. Might as well start right out with users specifying commit ID.

MikeSpreitzer: BTW, I have seen commits added to the default branch of a model at HF after I started using it.

docs/proposals/modelcaching.md
```diff
@@ -0,0 +1,102 @@
...
+    llm-d.io/desired-cached-models: "ibm-granite/granite-3.3-8b-instruct,deepseek-ai/DeepSeek-R1-0528"
+    llm-d.io/cached-models: "ibm-granite/granite-3.3-8b-instruct"
+spec:
+  accessModes:
+  - ReadWriteOnce
+  volumeMode: Filesystem
+  resources:
+    requests:
+      storage: 100Gi
+  storageClassName: local-sc
```
MikeSpreitzer, line 63: Isn't this normally called "local-storage"?

docs/proposals/modelcaching.md
```diff
@@ -0,0 +1,102 @@
...
+
+* Be agnostic about which model loader is being used.
+* Provide a low-level API to cache models on local storage.
+
+# Non-Goals
+
+* Provisioning Persistent Volumes, which is better handled by existing external static provisioners, such as [Local Persistent Volume Static Provisioner](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner), [Openshift Local Storage Operator](https://catalog.redhat.com/software/container-stacks/detail/66993e0f4f284e980ee072d3) and [Open EBS](https://openebs.io).
+* Supporting other storage types such as S3, HTTP, etc…
+* Supporting model revision (GIE does not support it yet).
+* Providing an high-level CRD-based API.
```
MikeSpreitzer, line 23: It seems to me that defined structured data would be preferable to annotations on PVCs.

And it would provide a smoother path forward into the anticipated future that includes storage that is NOT through a PVC.

docs/proposals/modelcaching.md
ritzdevp: Recently we gave a talk on model caching design in KServe https://www.youtube.com/watch?v=uUjUWaMn5K4

Issue #122, opened 2025-06-26
[Bug]: Published v0.0.8 image leads to failure on vLLM startup


achandrasekar:
### What happened?

Using the published llm-d image in GCP leads to this issue - https://github.com/vllm-project/vllm/issues/18859 where vLLM is not able to find the GPU devices. This is due to a recent change in the NVIDIA base image. Fix is to set the right CUDA binary location in LD_LIBRARY_PATH. 

### Version

0.0.1

diegocastanibm:
This issue refers to v0.0.8 and we are v0.6.0

PR #123, opened 2025-06-26
Add nvidia driver path for backwards compatibility

achandrasekar:
Fixes https://github.com/llm-d/llm-d/issues/122


achandrasekar:
@clubanderson PTAL. In its current form, llm-d deployment on GCP fails by default with the new vLLM / nvidia images. This will address that.

PR #125, opened 2025-07-07
Update README.md

carlory:
llm-d-model-service is now deprecated and no longer under active maintenance. see https://github.com/llm-d/llm-d-model-service/commit/403e98362eee3db82f4c7e8f0c630605e47f040c


carlory:
/hold
because llm-d-deployer is using the unmaintained project.

PR #140, opened 2025-07-20
feat: Add OpenAI API server dependencies to Docker image

yankay:
The current llm-d image does not support modelscope.  

This PR adds essential dependencies for enhanced OpenAI API server functionality to the existing Dockerfile; reference implementation at https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile#L503 .


yankay:
HI @robertgshaw2-redhat @tlrmchlsmth
Would you please help to review.

PR #148, opened 2025-07-25
installing vllm rc as wheel

Gregory-Pereira:
cc @robertgshaw2-redhat. I think this will break future builds but we can get our RC and release image this way


Gregory-Pereira (commented):
/lgtm

robertgshaw2-redhat (approved):

PR #159, opened 2025-08-10
vllm wheel nightly-builds

Gregory-Pereira:
This action should build golden commits of vllm upstream and CI and publish them to the public registry: https://gitlab.com/llm-d/vllm/-/packages

### DIFF
--- a/.github/workflows/playbooks/build-vllm-nvidia.yml
+++ b/.github/workflows/playbooks/build-vllm-nvidia.yml
@@ -1,0 +1,225 @@
+- name: Build vLLM from source for Nvidia GPUs
+  hosts: test_environments
+  remote_user: ubuntu
+  become: true
+  gather_facts: false
+
+  tasks:
+
+    - name: Wait until the instance is ready
+      ansible.builtin.wait_for_connection:
+        delay: 15
+        timeout: 180
+
+    - name: Install Python, GCC 10, AWS CLI, and dev Deps
+      shell: |
+        export DEBIAN_FRONTEND=noninteractive
+        sudo ln -fs /usr/share/zoneinfo/Etc/UTC /etc/localtime
+        sudo apt update
+        sudo apt install -y tzdata && \
+        sudo dpkg-reconfigure -f noninteractive tzdata && \
+        sudo apt install -y python{{ PYTHON_VERSION }} python{{ PYTHON_VERSION }}-dev python{{ PYTHON_VERSION }}-venv
+        # Upgrade to GCC 10 to avoid https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92519
+        sudo apt-get install -y gcc-10 g++-10
+        sudo apt install -y git cmake ninja-build
+        # install dev tools in case something goes wrong
+        sudo apt install -y vim tree curl unzip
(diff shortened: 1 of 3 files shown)


.github/workflows/playbooks/build-vllm-nvidia.yml
```diff
@@ -0,0 +1,225 @@
...
+        fi
+        export NVCC_THREADS=8
+        export CUDA_HOME=/usr/local/cuda
+        export PATH=$CUDA_HOME/bin:$PATH
+        export LD_LIBRARY_PATH=$CUDA_HOME/lib64:$LD_LIBRARY_PATH
+        
+        export RUSTC_WRAPPER="sccache"
+        export CC="gcc"
+        export CXX="g++"
+        export SCCACHE_BUCKET="vllm-nightly-sccache"
```
smarterclayton, line 161: Who owns this bucket?

.github/workflows/playbooks/build-vllm-nvidia.yml
```diff
@@ -0,0 +1,225 @@
...
+    - name: Install Python dependencies
+      shell: |
+        /home/ubuntu/vllm/venv/bin/pip install --upgrade pip setuptools setuptools_scm wheel
+        /home/ubuntu/vllm/venv/bin/pip install -r /home/ubuntu/vllm/requirements/build.txt --extra-index-url {{ PYTORCH_CUDA_INDEX_BASE_URL }}/{{ CUDA_SHORT }}
+        /home/ubuntu/vllm/venv/bin/pip install -r /home/ubuntu/vllm/requirements/cuda.txt --extra-index-url {{ PYTORCH_CUDA_INDEX_BASE_URL }}/{{ CUDA_SHORT }}
+
+        if [[ "$(uname -m)" == "arm64" || "$(uname -m)" == "aarch64" ]]; then
+          /home/ubuntu/vllm/venv/bin/pip install \
+            --index-url {{ PYTORCH_CUDA_NIGHTLY_INDEX_BASE_URL }}/{{ CUDA_SHORT }} \
+            "torch==2.8.0.dev20250318+cu128" "torchvision==0.22.0.dev20250319"
```
smarterclayton, line 135: What's the outcome of this?  Just curious why this is so old?

smarterclayton (commented):
A couple of questions, not a wheel build expert.

.github/workflows/playbooks/build-vllm-nvidia.yml
Gregory-Pereira: This is a static S3 bucket in my organizations account. Its access still falls under the main account credentials, I just use separate auth for managing access to only the s3 bucket to reduce changes something goes wrong during the upload step. I can easily re-create this somewhere else if we want to migrate to another account. It just holds the build Cache + built wheels

.github/workflows/playbooks/build-vllm-nvidia.yml
Gregory-Pereira: This was pulled from this section of the upstream dockerfile: https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile#L143-L155

smarterclayton (approved):

PR #161, opened 2025-08-14
Add XPU support

yuanwu2017:
The vLLM image of XPU is not released. This patch is for building the XPU image for llm-d project.  The image will be used in llm-d-modelservices and llm-d-infra examples. For example, [PR90](https://github.com/llm-d-incubation/llm-d-modelservice/pull/90) for adding the XPU single node deployment in modelservice


yao-matrix:
@yuanwu2017 , could you add a description here of Intel xpu? so reviewers can know what you are doing.

Gregory-Pereira:
Also interested if there's a way I can get access to XPU for Ci or just testing in general for the project.

yao-matrix:
> Also interested if there's a way I can get access to XPU for Ci or just testing in general for the project. Happy to rent them from a cloud provider if that's the path forward.

@Gregory-Pereira , sorry for late response, we were check the machine availability internal. Yes, we can provide one 4-card XPU system(which is hosted in Intel cloud) for llm-d ci access. Is it enough for CI or you think 2 or more machines are needed? Thx very much.

Gregory-Pereira:
I believe there are two considerations. First would be building the vllm-wheel for XPU (see how we do this for [nvidia GPUs](https://github.com/llm-d/llm-d/blob/dev/.github/workflows/build-vllm-wheel-on-dispatch.yaml)), we would extend that to support XPU option. If I understand that process correctly, a single XPU system would be fine there, and the the llm-d image would not need any XPU, just install that wheel build. 

The second concern is which quick-start / well lit paths you want supported for regression testing. The default basic example is [inference-scheduling](https://github.com/llm-d-incubation/llm-d-infra/tree/main/quickstart/examples/inference-scheduling) which doesn't do Prefill Decode desegregation, but only uses 2 GPUs and so would also need 2 XPU systems. [precise-prefix-cache-aware](https://github.com/llm-d-incubation/llm-d-infra/tree/main/quickstart/examples/precise-prefix-cache-aware) does PD but also uses 2 total so that could be supported, but PD needs 8 and wide-ep needs 24. 

TLDR: If we don't care about E2Es for regression testing, 1 system is fine for the builds. If you want to support the inference-scheduling basic example or precise example, you will need 2.

yao-matrix:
Thx very much for clear explanation, will go back and return to you.

yao-matrix:
> 2.

@Gregory-Pereira , sorry, i'm back for more clarifications. my summarizations is:
    | Usage | Minimum Number of GPU cards needed |
   | -------- | ------- |
   | vllm build | 1    |
   | inference-scheduling|2|
   |precise-prefix-cache-aware|2|
   |PD|8|
   |wide-ep|24|

Gregory-Pereira:
Can you Rebase @yuanwu2017

Makefile
```diff
@@ -2,28 +2,46 @@ SHELL := /usr/bin/env bash
...
-PROD_VERSION ?= 0.0.0
-IMAGE_TAG_BASE ?= ghcr.io/llm-d/$(PROJECT_NAME)
-IMG = $(IMAGE_TAG_BASE):$(DEV_VERSION)
-NAMESPACE ?= hc4ai-operator
+DOCKERFILE_DIR = docker
+ifeq ($(DEVICE), xpu)
+	DOCKERFILE ?= Dockerfile.xpu
+else
+	DOCKERFILE ?= Dockerfile.cuda
+endif Maybe we break out version per image because they share no common bits --> independent releas cycles
```
Gregory-Pereira:
```suggestion
endif # Maybe we break out version per image because they share no common bits --> independent releas cycles
```
Should this be a comment? If I left it as a non-comment In my PR that would definetly be my fault

Gregory-Pereira (dismissed):
One small question but overall /lgtm

Gregory-Pereira:
Ready to merge this whenever, we just need signed commit signatures.

yuanwu2017:
This may be a CI test environment configuration problem and has nothing to do with patch
Error: Password required

Gregory-Pereira:
> This may be a CI test environment configuration problem and has nothing to do with patch
> Error: Password required

Yes, you got it. This works for testing when pushing to an upstream branch, which is why I opened and then closed #190 to build and test these changes. I will be working on fixing the build infrastructure sometime after release - I don't want to brick image builds right before we need them.

Gregory-Pereira:
Based on using merge commits and not really getting the value of signing commits we have dropped the requirement. This PR can be merged as soon as it gets a rebase

yao-matrix:
@yuanwu2017 , pls help rebase, thx

yuanwu2017:
> Can we use the image from dockerhub: [`intel/vllm:0.10.0-xpu`](https://hub.docker.com/layers/intel/vllm/0.10.0-xpu/images/sha256-76c5d41a3766a0ee7844c91d1e98ba704b0bc1277daf145cc0b5494bf9ab0eea) instead of building new one here?
[intel/vllm:0.10.0-xpu](https://hub.docker.com/layers/intel/vllm/0.10.0-xpu/images/sha256-76c5d41a3766a0ee7844c91d1e98ba704b0bc1277daf145cc0b5494bf9ab0eea) does not include the nixl support for XPU. It will be included in 0.10.1.  I will keep tracking the intel released image. Actually we are one team. If a suitable image released and does not need this patch, I will delete this dockerfile and modify the makefile.

PR #160, opened 2025-08-11
Image builds to gha

Gregory-Pereira:
cc @smarterclayton @robertgshaw2-redhat

Depends on: #159 .

Changes:
- allows users to specify if they want to consume our nightly builds via: `VLLM_PREBUILT=1`
- It also converts our tekton build files to a simpler github action
- Bumps LMCache Commit sha to: `0db8ae4746a207a72483d716b5f82545a2ead24b`, ([see commit](https://github.com/neuralmagic/LMCache/commit/0db8ae4746a207a72483d716b5f82545a2ead24b)

PR #162, opened 2025-08-15
Image builds to gha

Gregory-Pereira:
Testing #160 on an upstream branch so I have access to secrets...

Changes:

- Removing all unused action code from 3 months ago. Anything in use has not been touched. Separate PR out to fix markdown link checker in #163.
- Removing all legacy tekton pipeline builds
- Implementing pipeline for Building the `llm-d-dev` image to any change to source on PR and on Merged PR (push to main, we have branch protections so only triggerable through PR merges anyway)
- Implementing pipeline for building the `llm-d` on any tag creation or release.
- Fixing NVSHMEM distribution links and associated build methodology
    - Updating NVSHMEM to 3.3.20 while were at it and keeping patches of supported versions going forward 
- Removing `.version.json` for versioning. `dev` images get tags for their build sha + the PR. Prod images will be versioned by the tag that triggers the workflow.

Future changes:
- get rid of this `dev` and `main` branches. We need 1 master branch, and with the versioning changes we can drop 1 of them


Gregory-Pereira:
Artifacts are being properly created and tagged: https://github.com/llm-d/llm-d/pkgs/container/llm-d-dev/488459895?tag=pr-162. I would consider this ready for merge

Dockerfile.ubi
```diff
@@ -47,30 +51,38 @@ RUN dnf install -y dnf-plugins-core && \
...
         rdma-core-devel \
+        libibverbs \
+        libibverbs-devel \
         numactl-libs \
         subunit \
         pciutils \
         pciutils-libs \
         ninja-build \
+        xz \
+        rsync \
```
wseaton, line 62: What is pulling in `xz` and `rsync` out of curiousity?

wseaton (approved):
LGTM to get going on these builds

Dockerfile.ubi
Gregory-Pereira: ` xz` was required with the new link for source, where we are dowloading a `.tar.xz` file for libnvshmem. I pulled in `rsync` because compared to the last version of nvshmem we had, this one does not have a global `CMakelists.txt`, or global build target. Instead they have four directories in source, `bootstrap-plugins`, `examples`, `perftest`, and `transport-plugins`, each (except for examples) with its own individual `CMakelists.txt` build target.  In [the first step](https://github.com/llm-d/llm-d/pull/162/files#diff-3fc20524636ade6cd833fcabb28a94921781961cee8d7fce631922cc10704c76R124-R136) were placing the full source into our nvshmem dir (this uses `rsync`), and then in the second, we are applying our coreweave patch over JUST the transport-plugins path, building and installing it. If we didn't want to use `rsync` im sure we could find another way to do it

PR #168, opened 2025-08-21
Editable install but removed source fix

Gregory-Pereira:
Dupe of #167 to push to registry

### DIFF
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -362,8 +362,8 @@ RUN --mount=type=cache,target=/var/cache/git \
     source /opt/vllm/bin/activate && \
     uv pip install cuda-python 'huggingface_hub[hf_xet]' && \
     uv pip install /tmp/wheels/*.whl && \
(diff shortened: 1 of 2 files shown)


Dockerfile.ubi
```diff
@@ -362,8 +362,8 @@ RUN --mount=type=cache,target=/var/cache/git \
     source /opt/vllm/bin/activate && \
     uv pip install cuda-python 'huggingface_hub[hf_xet]' && \
     uv pip install /tmp/wheels/*.whl && \
-    git clone https://github.com/vllm-project/vllm.git /tmp/vllm-${VLLM_COMMIT_SHA} && \
-    cd /tmp/vllm-${VLLM_COMMIT_SHA} && \
+    git clone https://github.com/vllm-project/vllm.git /vllm-source && \
+    cd /vllm-source && \
```
wseaton: Can we put this in `/opt/` or `/etc/` to avoid permissions issues in the container at /root?

wseaton (approved):
LGTM, getting the build back green

PR #171, opened 2025-08-22
Greg fixes

Gregory-Pereira:
cc @robertgshaw2-redhat @wseaton


Dockerfile.ubi
```diff
@@ -350,17 +336,20 @@ COPY --from=builder /wheels/*.whl /tmp/wheels/
...
 # Public LLM-D vllm wheels index
 ARG VLLM_WHEEL_URL="https://gitlab.com/api/v4/projects/72482892/packages/pypi/simple"
 
+
 # Install PyTorch and cuda-python
 # Install all compiled wheels (DeepEP, DeepGEMM, pplx-kernels, LMCache, nixl)
 RUN --mount=type=cache,target=/var/cache/git \
     source /opt/vllm/bin/activate && \
     uv pip install cuda-python 'huggingface_hub[hf_xet]' && \
+    uv pip install torch==2.7.1 && \
```
robertgshaw2-redhat, line 352: why is this needed?

Gregory-Pereira: Trying to pin torch before it gets resolved naturally by other packages like vllm.

PR #176, opened 2025-08-25
bump cuda version to 12.9.1

Gregory-Pereira:
cc @smarterclayton lets see if this builds out of the box with this


Dockerfile.ubi
```diff
@@ -279,22 +301,24 @@ RUN --mount=type=cache,target=/root/.cache/uv \
...
-RUN cat > /tmp/constraints.txt << 'EOF'
+RUN cat > /tmp/constraints.txt <<'EOF'
 torch==2.7.1
 EOF
 
 ENV LANG=C.UTF-8 \
     LC_ALL=C.UTF-8 \
     PYTHON_VERSION=${PYTHON_VERSION:-3.12} \
-    UV_TORCH_BACKEND=${UV_TORCH_BACKEND:-cu${CUDA_MAJOR}${CUDA_MINOR}} \
+    # UV_TORCH_BACKEND=${UV_TORCH_BACKEND:-cu${CUDA_MAJOR}${CUDA_MINOR}} \
```
wseaton, line 330: is this safe from an ABI perspective? Should we attempt to build with torch 2.8 (lowest version that is built on 12.9.1) even though that means we can't use precompiled vllm wheels?

Gregory-Pereira: As far as I understand t’s safe ABI-wise to stay on torch 2.7.1 with CUDA 12.9. PyTorch wheels should vendor the CUDA libs they need, and CUDA 12.x has minor-version “enhanced compatibility,” so a 12.8-built wheel runs fine on a 12.9 driver/runtime.

I think if we wanted 12.9-toolkit features/bugfixes or you’re compiling CUDA extensions / building torch from source that would require 12.9, but otherwise I think were ok pinning to 2.7.1

PR #178, opened 2025-08-26
support l40

Gregory-Pereira:
cc @ckhordiasma


Gregory-Pereira:
Turns out PPLX kernels only work with `TORCH_CUDA_ARCH_LIST=9.0a+PTX` so were gonna have to keep our build support at 9.0a,10.0,12.0+PTX then dont use perplexity kernel backend in CI on the nvidia L40s

Gregory-Pereira:
Depends on: #176, I plan to reset the TORCH_CUDA_ARCH_LIST just for PPLX and set it back for the parent image but this depends on breaking up wheels builds into 1 phase each

PR #180, opened 2025-08-26
try adding sccache for 1 build layer

Gregory-Pereira:
cc @wseaton


Dockerfile.ubi
```diff
@@ -12,14 +12,60 @@ ARG CUDA_MINOR=8
 ARG CUDA_PATCH=1
 ARG PYTHON_VERSION
 
+ARG USE_SCCACHE=true
+
+COPY sccache.config.toml /tmp
+
+RUN --mount=type=secret,id=aws_access_key_id \
```
wseaton, line 19: We should be be able to clean this up via herdoc docker syntax :)

wseaton (dismissed):
LGTM, let's get it in and start testing

wseaton (approved):

PR #190, opened 2025-09-06
[DO_NOT_MERGE] XPU ci builds

Gregory-Pereira:
cc @yuanwu2017

This should implement CI for #161 . After you are comfortable with it we should merge this into it before the dev merge.


Gregory-Pereira:
This adopts the same pattern as #189 for extending builds

build-xpu.sh
```diff
@@ -0,0 +1,106 @@
+#!/bin/bash
```
yuanwu2017, line 1: This script is not needed anymore, just use makefile and dockerfile to compile

Gregory-Pereira:
closing in favor of #161

PR #192, opened 2025-09-08
enable builds from PR

Gregory-Pereira:
This PR should break builds into 2 categories, regular builds which just test if image builds work, and then builds with the `ok-to-push` label which will push to the registry as we currently have it. Both will require approval to run based on their environments.


Gregory-Pereira:
I want to wait on this PR until after release, so as not to break our build infrastructure leading up to it

PR #193, opened 2025-09-08
Enable builds from forks

Gregory-Pereira:
dupe of #192 for testing


Gregory-Pereira:
This did not test as I anticipated, closing

PR #194, opened 2025-09-08
[DRAFT] Move llm-d-infra content to llm-d

smarterclayton:
This PR moves the content previous located at https://github.com/llm-d-incubation/llm-d-infra/tree/main/quickstart to llm-d
docs/ and /guides.

Sourced from https://github.com/llm-d-incubation/llm-d-infra/pull/207

TODO:
* [ ] Update links
* [ ] Update blog posts and release docs
* [ ] Review structure


guides/prereq/client-setup/README.md
```diff
@@ -0,0 +1,47 @@
...
+## Required Tools
+
+| Binary      | Minimum Required Version | Download / Installation Instructions                                                            |
+| ----------- | ------------------------ | ----------------------------------------------------------------------------------------------- |
+| `yq`        | v4+                      | [yq (mikefarah) – installation](https://github.com/mikefarah/yq?tab=readme-ov-file#install)     |
+| `git`       | v2.30.0+                 | [git – installation guide](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)       |
+| `helm`      | v3.12.0+                 | [Helm – quick-start install](https://helm.sh/docs/intro/install/)                               |
+| `helmfile`  | v1.1.0+                  | [Helmfile - installation](https://github.com/helmfile/helmfile?tab=readme-ov-file#installation) |
+| `kubectl`   | v1.28.0+                 | [kubectl – install & setup](https://kubernetes.io/docs/tasks/tools/install-kubectl/)            |
+| ----------- | ------------------------ | ----------------------------------------------------------------------------------------------- |
```
Gregory-Pereira: Lets remove this line, its not showing up as intended in preview

guides/prereq/client-setup/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Optional Tools
+
+| Binary             | Recommended Version      | Download / Installation Instructions                                                             |
+| ------------------ | ------------------------ | ------------------------------------------------------------------------------------------------ |
+| `stern`            | 1.30+                    | [stern - installation](https://github.com/stern/stern?tab=readme-ov-file#installation)           |
+| `helm diff` plugin | v3.10.0+                 | [helm diff installation docs](https://github.com/databus23/helm-diff?tab=readme-ov-file#install) |
+| ------------------ | ------------------------ | ------------------------------------------------------------------------------------------------ |
```
Gregory-Pereira: Same as above

guides/prereq/client-setup/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## HuggingFace Token
+
+Most guides download their model from Huggingface directly in the `llm-d` image. There are exceptions to this like the [`simulated-accelerators` guide](../examples/simulated-accelerators/) that uses no model, or the [`wide-ep-lws` guide](../examples/wide-ep-lws/) which uses a model loaded from storage directly on the nodes for faster development cycle iterations.
```
Gregory-Pereira: Both of these links dont work, relative paths are incorrect here:

1. `../examples/wide-ep-lws/` --> `../../wide-ep-lws`
2. `../examples/simulated-accelerators/` --> `../../wide-ep-lws`

guides/prereq/gateway-provider/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+This document will guide you through choosing a Gateway provider that can support the llm-d `inference-scheduler` component.
+
+This prerequisite generally requires cluster administration rights.
+
+## Before you begin
+
+Prior to deploying a Gateway control plane, you must install the custom resource definitions (CRDs) that add the Kubernetes API objects:
+
+- [Gateway API v1.3.0 CRDs](https://github.com/kubernetes-sigs/gateway-api/config/crd?ref=v1.3.0)
```
Gregory-Pereira: Not sure this should be a link like this, because this is the kustomize URL but returns empty page in github UI

guides/prereq/gateway-provider/README.md
```diff
@@ -0,0 +1,67 @@
...
+  - for more information see their [docs](https://gateway-api.sigs.k8s.io/guides/)
+- [Gateway API Inference Extension CRDs v0.5.1](https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd?ref=v0.5.1)
```
Gregory-Pereira: Same as above

guides/prereq/gateway-provider/common-configurations/README.md
Gregory-Pereira, line 1: This whole directory should move back to the guides section, because this contains values and customizations to the guide examples. It becomes confusing because these configurations are not at all related to the control plane, which is what this parent `gateway-provider` directory represents. We should also keep the `common/gateway-configurations` structure, because with the new specialized hardware backends (TPU, XPU, etc.) were going to create another directory for all those guide value configurations specific to that backend hardeware type

guides/README.md
```diff
@@ -0,0 +1,30 @@
+# Deploying distributed inference on Kubernetes with llm-d
+
+Our guides provide tested and benchmarked recipes and Helm charts to start serving large language models (LLMs) quickly with best practices common to production deployments. A familiarity with basic deployment and operation of Kubernetes is assumed.
+
+If you want to learn by doing, follow the [first deployment quickstart](./QUICKSTART.md).
+
+These guides are intended to be used as a starting point for your own configuration and deployment of model servers. Our Helm charts provide basic reusable building blocks for vLLM deployments and inference scheduler configuration within these guides but will not support the full range of all possible configurations. Both guides and charts depend on features provided and supported in the [vLLM](https://github.com/vllm-project/vllm) and [inference gateway](https://github.com/kubernetes-sigs/gateway-api-inference-extensions) open source projects.
```
Gregory-Pereira:
```suggestion
These guides are intended to be used as a starting point for your own configuration and deployment of model servers. Our Helm charts provide basic reusable building blocks for vLLM deployments and inference scheduler configuration within these guides but will not support the full range of all possible configurations. Both guides and charts depend on features provided and supported in the [vLLM](https://github.com/vllm-project/vllm) and [inference gateway](https://github.com/kubernetes-sigs/gateway-api-inference-extension) open source projects.
```
Link is incorrect

guides/README.md
```diff
@@ -0,0 +1,30 @@
...
+* Deploying a self-hosted LLM behind a single workload across tens or hundreds of nodes
+* Running a production model-as-a-service platform that supports many users and workloads sharing one or more LLM deployments
+
+## Our Well-Lit Path Guides
+
+A well-lit path is a documented, tested, and benchmarked solution of choice to reduce risk and maintenance cost. These are the central best practices common to production deployments of large language model serving.
+
+We currently offer three tested and benchmarked paths to help you deploy large models:
+
+1. [Intelligent Inference Scheduling](./inference-scheduling) - Deploy [vLLM](https://docs.vllm.ai) behind the [Inference Gateway (IGW)](https://github.com/kubernetes-sigs/gateway-api-inference-extension) to decrease latency and increase throughput via [precise prefix-cache aware routing](https://github.com/llm-d-incubation/llm-d-infra/tree/main/guides/precise-prefix-cache-aware) and [customizable scheduling policies](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md).
```
Gregory-Pereira: All these links are wrong, they reference `llm-d-incubation/llm-d-infra` these should be relative paths to the new guides

guides/inference-scheduling/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Gateway options
+
+To see specify your gateway choice you can use the `-e <gateway option>` flag, ex:
+
+```bash
+helmfile apply -e kgateway -n ${NAMESPACE}
+```
+
+To see what gateway options are supported refer to our [gateway control plane docs](../../gateway-control-plane-providers/README.md#supported-providers). Gateway configurations per provider are tracked in the [gateway-configurations directory](../common/gateway-configurations/).
```
Gregory-Pereira: This link is incorrect:
```suggestion
To see what gateway options are supported refer to our [gateway control plane docs](../prereq/gateway-provider/common-configurations/README.md#supported-providers). Gateway configurations per provider are tracked in the [gateway-configurations directory](../prereq/gateway-provider/).
```

guides/pd-disaggregation/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Gateway options
+
+To see specify your gateway choice you can use the `-e <gateway option>` flag, ex:
+
+```bash
+helmfile apply -e kgateway -n ${NAMESPACE}
+```
+
+To see what gateway options are supported refer to our [gateway control plane docs](../../gateway-control-plane-providers/README.md#supported-providers). Gateway configurations per provider are tracked in the [gateway-configurations directory](../common/gateway-configurations/).
```
Gregory-Pereira: Same as above, this link does not exist:
```suggestion
To see what gateway options are supported refer to our [gateway control plane docs](../prereq/gateway-provider/common-configurations/README.md#supported-providers). Gateway configurations per provider are tracked in the [gateway-configurations directory](../prereq/gateway-provider/).
```

guides/pd-disaggregation/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+You can also customize your gateway, for more information on how to do that see our [gateway customization docs](../../docs/customizing-your-gateway.md).
+
+#### GKE specific workarounds
+
+While this example out of the box requires Infiniband RDMA, GKE does not support this. Therefore we patch out these values in [the helmfile](./helmfile.yaml.gotmpl#L73-80).
```
Gregory-Pereira:
```suggestion
While this example out of the box requires Infiniband RDMA, GKE does not support this. Therefore we patch out these values in [the helmfile](./helmfile.yaml.gotmpl#L88-97).
```
Lines do not match up to rdma patches

guides/prereq/gateway-provider/README.md
smarterclayton: Agreed, this is unmodified from your change here: https://github.com/llm-d-incubation/llm-d-infra/commit/91b86d4bed3c40a5e870a0d341e349dc3840c5a9

Do you know where you sourced them from?

Gregory-Pereira: Yeah, this was my mistake in the first place for sure. I would just reference the branch here: `https://github.com/kubernetes-sigs/gateway-api/tree/v1.3.0/config/crd`. The actual scripts still use the `kubectl` complient link so that should be fine

guides/prereq/gateway-provider/README.md
Gregory-Pereira: As per our discussion above:
```suggestion
- [Gateway API Inference Extension CRDs v0.5.1](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/v0.5.1/config/crd)
```

guides/prereq/gateway-provider/common-configurations/README.md
smarterclayton: These are common configurations of the gateway-provider, which is a prerequisite to deploying an InferencePool + deployment + other config specific to a running model server.

I don't think a casual reader would expect to see gateway specific configuration outside of the gateway-provider directory, especially when the config you are describing is shared by multiple guides.  I would expect top level guides to deep link to appropriate prerequisites which are configured before the guide is run, and these look like prerequisites run before the guide.

guides/README.md
smarterclayton: There is only one link referencing llm-d infra, where are the "all"?

guides/pd-disaggregation/README.md
smarterclayton: Deep linking in to the helmfile feels fragile.  We should probably not do that.

guides/prereq/gateway-provider/common-configurations/README.md
Gregory-Pereira: I do want to push back on this, it is not:

>  a prerequisite to deploying an InferencePool + deployment + other config specific to a running model server

These values get calculated into the deployment of the inferencepool + deployment at deploy time based on the gateway provider you have selected (that decision is the only prereq), and determine how the stack is composed. This is why I thought in kustomize structure it is more apt to keep it in a directory common to all guides but separate from the control plane configurations

That being said I do agree with your second point so if you feel strongly we can keep the structure as you have it.

smarterclayton: Perhaps "common configurations" is not the right name, but the README in the common-configurations directory does at least explain the guides directly reference these.

I propose we keep the structure here for clarity and then make a new decision based on:

> because with the new specialized hardware backends (TPU, XPU, etc.) were going to create another directory for all those guide value configurations specific to that backend hardeware type

Is that 0.3 or no?

guides/pd-disaggregation/README.md
smarterclayton: The actual text probably doesn't make sense.  @liu-cong when we want to add RoCE this will be different.

I think it's more clear to say "Different platforms request accelerator network resources with different values.  See infra-providers for details".

smarterclayton: Updated with a specific commit https://github.com/llm-d-incubation/llm-d-infra/pull/207/commits/d432be47f67f69105982f2f7c37daca7fdb4d0c6

@liu-cong / @robertgshaw2-redhat can you review this for accuracy?

.github/workflows/e2e-pd-disaggregation-accelerator-test.yaml
```diff
@@ -0,0 +1,586 @@
...
+            helmfile apply -e ${GATEWAY_TYPE}
+          EOF
+
+      - name: Run installer to deploy llm-d infrastructure
+        run: |
+          ssh -o StrictHostKeyChecking=no -i key.pem ubuntu@$INSTANCE_IP "HF_TOKEN=$HF_TOKEN GATEWAY_TYPE=$GATEWAY_TYPE bash -s" <<'EOF'
+            set -euo pipefail
+            set -x
+            # TODO: can we avoid this here?
+            cd llm-d-infra
```
smarterclayton, line 411: @Gregory-Pereira Not sure how to remove this, should this be a helm chart deployment instead?

Gregory-Pereira: I have a fix for this in: https://github.com/llm-d-incubation/llm-d-infra/pull/214/files#diff-fd859de5d5dd1e7d0028ba445dc36abd78c517b12e2b456140518835303a0075R403-R412. You don't need to worry about it

guides/README.md
```diff
@@ -0,0 +1,30 @@
...
+* Deploying a self-hosted LLM behind a single workload across tens or hundreds of nodes
+* Running a production model-as-a-service platform that supports many users and workloads sharing one or more LLM deployments
+
+## Our Well-Lit Path Guides
+
+A well-lit path is a documented, tested, and benchmarked solution of choice to reduce risk and maintenance cost. These are the central best practices common to production deployments of large language model serving.
+
+We currently offer three tested and benchmarked paths to help you deploy large models:
+
+1. [Intelligent Inference Scheduling](./inference-scheduling) - Deploy [vLLM](https://docs.vllm.ai) behind the [Inference Gateway (IGW)](https://github.com/kubernetes-sigs/gateway-api-inference-extension) to decrease latency and increase throughput via [precise prefix-cache aware routing](./precise-prefix-cache-aware/) and [customizable scheduling policies](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md).
```
liu-cong, line 22: The precise prefix-cache aware scheduling is currently a separate example, the default `inference-scheduling` is still using the approximate prefix cache aware routing.

PR #203, opened 2025-09-10
Refactor wide ep quickstart to use kustomize

liu-cong:
I am using this example to demonstrate the idea of deprecating the modelservice and llm-d-infra charts, and use kustomize to manage variations in different environments. 

Pros:
- Significantly reduces the helm chart maintenance  overhead for the maintainers, as well as cognitive overhead for users.
- Cleanly presents the "out of band" manifests that are not covered from upstream charts that will likely require deep customizations from the users. 
- This makes it easier for my next PR to add GKE specific manifest overlays for the vllm deployment.

Cons:
- Requires the users to clone the repo to apply the manifests (I don't think this is too bad)
- Currently the Istio destination rule has a “host” which contains the dns name of the epp service, which contains a hard-coded namespace. This can be addressed by adding destination rule to the upstream helm chart.

Changes:
- Use inferencepool helm chart to install the inference pool and epp instead of modelservice chart
- Use kustomize to install gateway instead of using llm-d-infra chart
- Use kustomize to install vllm instead of modelservice chart
- Use default service account for vllm instead of creating the `ms-wide-ep-llm-d-modelservice` 
- Removed helm related labels

TESTED: NOTE I haven't tested this. My plan is to generate the manifests and diff the before/after. 

If we are cool with this, I will test this change, and follow up with 1. GKE specific changes for wide ep and 2. migrate other guides to use this pattern.

@Gregory-Pereira initiated https://github.com/llm-d-incubation/llm-d-infra/pull/215, which focuses on the `inference-scheduling` example. This PR should have no/little conflict with that.

cc @smarterclayton @Gregory-Pereira


smarterclayton:
Not for 0.3, right?

PR #206, opened 2025-09-11
Refactor wide ep quickstart to use kustomize

liu-cong:
I am using this example to demonstrate the idea of deprecating the modelservice and llm-d-infra charts, and use kustomize to manage variations in different environments.  More details can be found in [doc](https://docs.google.com/document/d/1HdPPcs1HamKHFt6eqzVJDpp9W8_U12GitCG85EJTW_U/edit?resourcekey=0-5QY-K4EfLmQKrCWUflIeZg&tab=t.0)

Pros:
- Significantly reduces the helm chart maintenance  overhead for the maintainers, as well as cognitive overhead for users.
- Cleanly presents the "out of band" manifests that are not covered from upstream charts that will likely require deep customizations from the users. 
- This makes it easier for my next PR to add GKE specific manifest overlays for the vllm deployment.

Cons:
- Requires the users to clone the repo to apply the manifests (I don't think this is too bad)
- Currently the Istio destination rule has a “host” which contains the dns name of the epp service, which contains a hard-coded namespace. This can be addressed by adding destination rule to the upstream helm chart.

Changes:
- Use inferencepool helm chart to install the inference pool and epp instead of modelservice chart
- Use kustomize to install gateway instead of using llm-d-infra chart
- Use kustomize to install vllm instead of modelservice chart
- Use default service account for vllm instead of creating the `ms-wide-ep-llm-d-modelservice` 
- Removed helm related labels

TESTED: 
### Diff with existing state
I compared the yaml generated by kustomize with the ones generated by helm. 
- The inferencepool related resources diff: https://www.diffchecker.com/5QDkfgtn/
- GKE gateway diff: https://www.diffchecker.com/2SA537IH/
- Istio gateway diff: https://www.diffchecker.com/vs6Wj2k8/
- kgateway diff: https://www.diffchecker.com/dFbvCdET/
### Manual test
* NOTE I didn't verify the vllm yamls work. I just made sure the there is no diff between the existing state. Testing of vllm on GKE is done in the follow up PR #215.
* I tested all 3 gateways paths (GKE, Istio, kgateway) with sim vllm server.

If we are cool with this, I will test this change, and follow up with 1. GKE specific changes for wide ep and 2. migrate other guides to use this pattern.

@Gregory-Pereira initiated https://github.com/llm-d-incubation/llm-d-infra/pull/215, which focuses on the `inference-scheduling` example. This PR should have no/little conflict with that.

cc @smarterclayton @Gregory-Pereira


guides/wide-ep-lws/manifests/gateway/istio/destinationrule.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: networking.istio.io/v1beta1
+kind: DestinationRule
+metadata:
+  name: infra-wide-ep-inference-gateway
+spec:
+  host: ms-wide-ep-llm-d-modelservice-epp.llm-d-wide-ep.svc.cluster.local
```
Gregory-Pereira: If you are creating this through GIE instead of MS now this needs to change to the GIE service name

guides/wide-ep-lws/manifests/gateway/istio/configmap.yaml
```diff
@@ -0,0 +1,32 @@
...
+            resources: 
+              limits:
+                cpu: "4"
+                memory: 4Gi
+              requests:
+                cpu: "1"
+                memory: 1Gi
+  service: |
+    spec:
+      type: "LoadBalancer"
```
Gregory-Pereira: This file should have comments supporting what other types of customization patterns we see that users want, ex specifying non default ports under the service from the charts: https://github.com/llm-d-incubation/llm-d-infra/blob/main/charts/llm-d-infra/templates/gateway-infrastructure/configmap.yaml#L43-L45. Yes this is not something we want to do long term, but I think it will help people adapt to the new pattern

guides/wide-ep-lws/manifests/gateway/istio/configmap.yaml
```diff
@@ -0,0 +1,32 @@
...
+              - proxy
+              - router
+              - --domain
+              - $(POD_NAMESPACE).svc.cluster.local
+              - --proxyLogLevel
+              - error
+              - --proxyComponentLogLevel
+              - misc:error
+              - --log_output_level
+              - default:error
```
Gregory-Pereira: All of this is only required when adjusting the log level of the container, you can comment this out as per my comment below

guides/wide-ep-lws/manifests/gateway/istio/configmap.yaml
```diff
@@ -0,0 +1,32 @@
...
+            resources: 
+              limits:
+                cpu: "4"
+                memory: 4Gi
+              requests:
+                cpu: "1"
+                memory: 1Gi
```
Gregory-Pereira: Note: These are the resource usage defaults for benchmark cases, the default values for regular are smaller. I guess in this pattern we lose this concept of bench vs debug setups

guides/wide-ep-lws/manifests/gateway/base/kustomization.yaml
Gregory-Pereira, line 1: We also had an optional [ingress](https://github.com/llm-d-incubation/llm-d-infra/blob/main/charts/llm-d-infra/templates/gateway-infrastructure/ingress.yaml) that users could stick in front of their gateway if they chose service clusterIP, not sure if we want to keep this as an option, or even if so how it gets kept

guides/wide-ep-lws/manifests/modelserver/base/kustomization.yaml
Gregory-Pereira, line 1: Your missing pod monitors, retemplate with metrics on that is the default were using for v0.3.0. Should look something like this: https://github.com/llm-d-incubation/llm-d-infra/pull/215/files#diff-c0c94050623c1ef48e204c5695488a8c0d5b2574358fe34557b3a1b1d48bfe52

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
Gregory-Pereira, line 1: It looks like you dropped the TPU file in this: https://github.com/llm-d-incubation/llm-d-infra/pull/215/files#diff-ca26b7c79f09895bed335ffa833e22be208ae2d1b791955087603307b4a7f07a

guides/wide-ep-lws/manifests/modelserver/openshift/kustomization.yaml
Gregory-Pereira, line 1: Remove, if there are no customizations that we are providing we don't need to have a holding directory for it, just adds noise. If there become custom things we need to do in our OCP Path we will need to redo this. Same with GKE dir

guides/wide-ep-lws/manifests/gateway/kgateway/kustomization.yaml
Gregory-Pereira, line 1: You are missing the gateway parameters: https://github.com/llm-d-incubation/llm-d-infra/blob/main/charts/llm-d-infra/templates/gateway-infrastructure/gatewayparameters.yaml

guides/wide-ep-lws/manifests/gateway/istio/destinationrule.yaml
smarterclayton: IIUC you shouldn't need this to be fully specified (or I would hope not), otherwise this breaks namespace idempotency.

guides/wide-ep-lws/manifests/gateway/istio/configmap.yaml
smarterclayton: I think debug setups are the not normal cases.  Are you debugging on non H100 boxes?  I think a reasonable prereq is at least a full H100 host, which should have oodles of CPU.

guides/wide-ep-lws/manifests/gateway/istio/destinationrule.yaml
Gregory-Pereira: It can work without the namespace but again if you have multiple implementations with the same name (ie doing guide `inference-scheduling` in 2 different namespaces) it breaks if you don't specifiy the NS

liu-cong: We can use the short name and let Istio default to the namespace of the DestionRule. Does that sound good?

From Istio docs 

```
Note for Kubernetes users: When short names are used (e.g. “reviews” instead of “reviews.default.svc.cluster.local”), Istio will interpret the short name based on the namespace of the rule, not the service. A rule in the “default” namespace containing a host “reviews” will be interpreted as “reviews.default.svc.cluster.local”, irrespective of the actual namespace associated with the reviews service. To avoid potential misconfigurations, it is recommended to always use fully qualified domain names over short names.
```

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
liu-cong: we don't have TPU example in the wide-ep path yet.

guides/wide-ep-lws/manifests/gateway/kgateway/kustomization.yaml
liu-cong: Yeah I was wondering about this. When I did `helmfile apply -e kgateway` I didn't see this generated. Perhaps a bug in the existing guide?

guides/wide-ep-lws/manifests/gateway/istio/destinationrule.yaml
liu-cong: Changed the name to match the new EPP service name, and use the short name to allow istio to default to current namespace.

guides/wide-ep-lws/manifests/modelserver/base/kustomization.yaml
liu-cong: Are there any prerequisites required for this to work? I don't see this generated in the current helm approach or maybe I am behind. Do we need to add the monitoring for other guides as well?

guides/wide-ep-lws/manifests/modelserver/openshift/kustomization.yaml
liu-cong: added RDMA customization here, we can add more if needed.

guides/wide-ep-lws/manifests/gateway/istio/configmap.yaml
liu-cong: are you saying the entire `args` section is optional? If so it's nice I can remove them all

guides/wide-ep-lws/manifests/gateway/base/kustomization.yaml
liu-cong: I will bias toward picking an opinionated default here. We are focusing on llm and not on gateway. Advanced users can adapt accordingly. So can we just default to `LoadBalancer` type and remove the ingress?

guides/wide-ep-lws/manifests/gateway/kgateway/kustomization.yaml
Gregory-Pereira: Its possible it was a bug, we don't get too much testing coverage on the `kgateway` side. Somethings i noticed in the past were that the gateway class name is case sensitive, also it relies on gateway Params being turned on: https://github.com/llm-d-incubation/llm-d-infra/blob/main/charts/llm-d-infra/templates/gateway-infrastructure/gatewayparameters.yaml#L1C84-L1C125. I can recreate the manifest for you if were still having issues

liu-cong: thanks, if you have the manifest pls share so I can add it here. If not I will try to generate it myself.

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -0,0 +1,164 @@
...
+              - containerPort: 8000
+            resources: {}
+            restartPolicy: Always
+            securityContext:
+              allowPrivilegeEscalation: false
+              runAsNonRoot: true
+        volumes:
+          - emptyDir:
+              medium: Memory
+              sizeLimit: 1Gi
```
smarterclayton: You need to bump this to 2Gi for NCCL 2.27 to be safe (on all workers)

guides/wide-ep-lws/manifests/modelserver/openshift/kustomization.yaml
```diff
@@ -0,0 +1,34 @@
...
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/affinity
+        value:
+          nodeAffinity:
+            requiredDuringSchedulingIgnoredDuringExecution:
+              nodeSelectorTerms:
+                - matchExpressions:
+                  - key: gpu.nvidia.com/model
+                    operator: In
+                    values:
+                      - H200
```
Gregory-Pereira: This is not an openshift thing, the default example works out of the box on H200s, this shouldn't be in this overlay. Maybe not in the base manifests but it gets confusing to keep non provider flavored customizations in their directory

guides/wide-ep-lws/manifests/modelserver/openshift/kustomization.yaml
```diff
@@ -0,0 +1,34 @@
...
+          hostPath:
+                path: /mnt/local/hf-cache
+                type: DirectoryOrCreate
+      # Add RMDA config.
+      - op: add
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/resources/limits/rdma~1ib
+        value: "1"
+      - op: add
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/resources/requests/rdma~1ib
+        value: "1"
```
Gregory-Pereira: Other way around, you should patch this out of GKE rather than patching it in. The default conceptual model of this example uses RDMA, that is a requirement. We do have workarounds to get around that but its intended for it

guides/wide-ep-lws/manifests/modelserver/openshift/kustomization.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - target:
+      kind: LeaderWorkerSet
+    patch: |-
+      - op: add
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/volumes/-
+        value: 
+          name: hf-cache
+          hostPath:
+                path: /mnt/local/hf-cache
+                type: DirectoryOrCreate
```
Gregory-Pereira: Again not OCP specific, Rob does this for the CKS cluster so its a faster development experience but it actually has no relation to the example other than thats how he choses to load the model in - we will need to think about how we want to maintain these QOL customizations that fall outside scope of the base example

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../base
+# TODO: Add GKE specific RDMA patches here.
```
Gregory-Pereira: Linting: May I suggest the VSCode setting: User --> Text Editor --> Files --> Insert Final Newline
```suggestion
# TODO: Add GKE specific RDMA patches here.

```

guides/wide-ep-lws/manifests/modelserver/base/prefill.yaml
```diff
@@ -0,0 +1,133 @@
...
+              cpu: 32
+              ephemeral-storage: 64Gi
+              memory: 512Gi
+              nvidia.com/gpu: "8"
+          
+          volumeMounts:
+            - mountPath: /dev/shm
+              name: dshm
+            - mountPath: /huggingface-cache
+              name: hf-cache
```
Gregory-Pereira, line 132: This is only required to be set when you are trying to load the model as we are now from the volume hf-cache (see below). Either keep both or removem same with all other references above

guides/wide-ep-lws/manifests/gateway/kgateway/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
...
+  target:
+    kind: Gateway
+- target:
+    kind: HTTPRoute
+  patch: |-
+    - op: add
+      path: /spec/rules/0/timeouts
+      value:
+        backendRequest: 0s
+        request: 0s
```
Gregory-Pereira: This is an optional one, I think the default should be setting max timeout and it should be patched out rather than vice-versa but I can get how my stance on this is just as opionated as yours so I could go either way

guides/wide-ep-lws/manifests/modelserver/openshift/kustomization.yaml
smarterclayton: RDMA is not a kube standard resource spec, unfortunately.  So we should not endorse it here.

smarterclayton: We have to wait for something in DRA to standardize those.

guides/wide-ep-lws/manifests/modelserver/openshift/kustomization.yaml
smarterclayton: That label is not a kube standard label - it will not work on every Kubernetes cluster, only those with gpu operator installed.

guides/wide-ep-lws/manifests/modelserver/base/prefill.yaml
smarterclayton: I think we should always have a mount path to a volume, whether it is an empty dir, host path, or PV is irrelevant.  I mentioned this in my infra PR (https://github.com/llm-d/llm-d/tree/main/guides/prereq/infrastructure#optional-vllm-container-image)

guides/wide-ep-lws/manifests/modelserver/openshift/kustomization.yaml
Gregory-Pereira: That I actually did not know thank you.

guides/wide-ep-lws/manifests/modelserver/base/prefill.yaml
Gregory-Pereira: What I got from reading that was specifically for compile caches, which I considered as fundamentally different from the model and dataset caches. Not sure if this was just a misunderstanding on my side or if we should reference both in those docs

guides/wide-ep-lws/manifests/gateway/base/kustomization.yaml
Gregory-Pereira: Sounds like a plan

guides/wide-ep-lws/manifests/modelserver/openshift/kustomization.yaml
liu-cong: yeah, the reason I did this is that `/mnt/local/hf-cache` is not a valid mount path for the Container Optimized OS (COS) on GKE, so I had to choose a different mount path. Not sure about other clouds but things like this can happen. How about I add a comment on why it's done this way and move forward for now? 

Agree it's not the cleanest and let's follow up later.

guides/wide-ep-lws/manifests/gateway/istio/configmap.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: v1
+kind: ConfigMap
```
liu-cong: TODO: Remove this entirely

guides/wide-ep-lws/manifests/gateway/kgateway/kustomization.yaml
liu-cong: This is optional, remove. We will capture this in a FAQ style doc. Created an issue to track https://github.com/llm-d/llm-d/issues/217

guides/wide-ep-lws/manifests/modelserver/base/kustomization.yaml
liu-cong: I will do this as a followup, as users can choose different monitoring stacks, I think we should have a section in the doc to link to the docs, and provide examples. I just don't have the bandwidth to test it for this PR.

guides/wide-ep-lws/manifests/gateway/istio/destinationrule.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: networking.istio.io/v1beta1
+kind: DestinationRule
+metadata:
+  name: wide-ep-inference-gateway
+spec:
+  host: deepseek-r1-epp
```
Gregory-Pereira:
```suggestion
  host: deepseek-r1-epp.cluster.svc.local
```

guides/wide-ep-lws/manifests/gateway/base/httproute.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: llm-route
```
Gregory-Pereira: It should convey that this route targets the deepseek inferencepool
```suggestion
  name: deepseek-r1
```

guides/wide-ep-lws/manifests/gateway/kgateway/openshift/gatewayParameters.yaml
Gregory-Pereira, line 1: Tested this, turns out I was wrong, it is required to have the gatewayParameters for OCP, otherwise the pod does not spin from the deployment, it also needs the SCC changes, all other cloud providers should be fine without the gateway parameters though

Gregory-Pereira (commented):
Some more progress but ill need to fix kgateway tomorrow see above

liu-cong:
@Gregory-Pereira Tested all 3 gateways and made sure things are wired up correctly. Will test vllm deployment on GKE as a follow up. This is ready to be merged. Thank you!

guides/wide-ep-lws/README.md
```diff
@@ -27,67 +27,100 @@ Use the helmfile to compose and install the stack. The Namespace in which the st
...
+
+### Deploy InferencePool
+
+```bash
+# For GKE
+helm install deepseek-r1 \
+  -n ${NAMESPACE} \
+  -f inferencepool.values.yaml \
+  --set provider.name=gke \
+  oci://us-central1-docker.pkg.dev/k8s-staging-images/gateway-api-inference-extension/charts/inferencepool --version v0 # TODO: Upgrade to v1.0 version.
```
Gregory-Pereira: Are we doing the v1 upgrade in this pr or not? I see above you ref `NOTE the required \`GATEWAY_API_INFERENCE_EXTENSION_CRD_REVISION\` version is \`v1.0.0\`.`

guides/wide-ep-lws/README.md
```diff
@@ -27,67 +27,100 @@ Use the helmfile to compose and install the stack. The Namespace in which the st
...
+
+# For non-GKE
+helm install deepseek-r1 \
+  -n ${NAMESPACE} \
+  -f inferencepool.values.yaml \
+  oci://us-central1-docker.pkg.dev/k8s-staging-images/gateway-api-inference-extension/charts/inferencepool --version v0 # TODO: Upgrade to v1.0 version.
```
Gregory-Pereira: Same as above

guides/wide-ep-lws/inferencepool.values.yaml
Gregory-Pereira, line 1: Your missing the additional flag definitions:
```
  flags:
    - name: "pool-group"
      value: "inference.networking.x-k8s.io"
```
See deployment pieces: https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/v1.0.0/config/charts/inferencepool/templates/epp-deployment.yaml#L34C27-L37

To enforce the old pool type, otherwise it uses the new one, see: https://github.com/llm-d-incubation/llm-d-infra/pull/214/files#diff-75f5438681b5c8291568b4ac2d183e4ecf6e1ae9506468ad23676206d96ca81cR16-R19.

This should be required to make the EPP actual matchup with the inferencepool because its using the old API.

liu-cong: Good catch! Updated.

guides/wide-ep-lws/README.md
liu-cong: Updated, when I started this PR v1.0.0 wasn't available.

smarterclayton:
Did we intentionally merge this to dev instead of main?

PR #208, opened 2025-09-11
[DO_NOT_MERGE] Weka

Gregory-Pereira:
cc @sdimitro @robertgshaw2-redhat @chcost 
Ignore the non-weka pieces for now i just pull in congs changes for a pattern here as a guide and to track upgrade changes

### DIFF
--- a/guides/weka/manifests/modelserver/base/decode.yaml
+++ b/guides/weka/manifests/modelserver/base/decode.yaml
@@ -55,7 +55,7 @@ spec:
             type: File 
       containers:
         - name: vllm
-          image: ghcr.io/llm-d/llm-d-dev:pr-246@sha256:41219315abda2cdf61121c6bc8a747795dfc8872568d21a392d335e7361e350a
+          image: ghcr.io/llm-d/llm-d-dev:pr-246@sha256:41219315abda2cdf61121c6bc8a747795dfc8872568d21a392d335e7361e350a # This image has updated LMCache with upstream but not new vLLM for support
           command: ["vllm", "serve"]
           args:
             - RedHatAI/Llama-3.3-70B-Instruct-FP8-dynamic
@@ -68,7 +68,7 @@ spec:
             - --block-size
             - "128"
             - --kv-transfer-config
-            - '{"kv_connector":"NixlConnector", "kv_role":"kv_both"}'
+            - '{"kv_connector":"MultiConnector","kv_role":"kv_both","kv_connector_extra_config":{"connectors":[{"kv_connector":"NixlConnector","kv_role":"kv_both"},{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}]}}'
             - --disable-log-requests
             - --disable-uvicorn-access-log
             - --max-model-len
@@ -92,6 +92,17 @@ spec:
               value: "INFO"
             - name: NVIDIA_GDRCOPY
               value: "enabled"
+            # LMCache GPU Direct Vars
+            - name: LMCACHE_USE_EXPERIMENTAL
+              value: true
+            - name: LMCACHE_CHUNK_SIZE
+              value: 256
+            - name: LMCACHE_GDS_PATH
+              value: /mnt/weka
+            - name: LMCACHE_CUFILE_BUFFER_SIZE
+              value: "8192"
+            - name: LMCACHE_LOCAL_CPU
+              value: false
           ports:
             - containerPort: 8200
               name: metrics
--- a/guides/weka/manifests/modelserver/base/prefill.yaml
+++ b/guides/weka/manifests/modelserver/base/prefill.yaml
@@ -38,7 +38,7 @@ spec:
             type: File 
       containers:
         - name: vllm
-          image: ghcr.io/llm-d/llm-d-dev:pr-246@sha256:41219315abda2cdf61121c6bc8a747795dfc8872568d21a392d335e7361e350a
+          image: ghcr.io/llm-d/llm-d-dev:pr-246@sha256:41219315abda2cdf61121c6bc8a747795dfc8872568d21a392d335e7361e350a # This image has updated LMCache with upstream but not new vLLM for support
           command: ["vllm", "serve"]
           args:
             - RedHatAI/Llama-3.3-70B-Instruct-FP8-dynamic
@@ -49,7 +49,7 @@ spec:
             - --block-size
             - "128"
             - --kv-transfer-config
-            - '{"kv_connector":"NixlConnector", "kv_role":"kv_both"}'
+            - '{"kv_connector":"MultiConnector","kv_role":"kv_both","kv_connector_extra_config":{"connectors":[{"kv_connector":"NixlConnector","kv_role":"kv_both"},{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}]}}'
             - --disable-log-requests
             - --disable-uvicorn-access-log
             - --max-model-len
@@ -73,6 +73,17 @@ spec:
               value: "INFO"
             - name: NVIDIA_GDRCOPY
               value: "enabled"
+            # LMCache GPU Direct Vars
+            - name: LMCACHE_USE_EXPERIMENTAL
+              value: true
+            - name: LMCACHE_CHUNK_SIZE
+              value: 256
+            - name: LMCACHE_GDS_PATH
+              value: /mnt/weka
+            - name: LMCACHE_CUFILE_BUFFER_SIZE
+              value: "8192"
+            - name: LMCACHE_LOCAL_CPU
+              value: false
           ports:
           - containerPort: 8000
             name: metrics


docs/infra-providers/weka/weka-check-cufile
```diff
@@ -0,0 +1,11 @@
...
+# /opt/probes/check-cufile.sh
+set -eu
+TARGET="${1:-/host/cufile}"   # we'll mount the node path here
+if [ -e "$TARGET" ]; then
+  # Other verification on cufile.json goes here
+  exit 0
+else
+  echo "Missing: $TARGET" >&2
+  exit 1
+fi
```
sdimitro: Ok so if I understand things correctly this will be placed in to the Docker container and it will be ran by a readiness probe in the decode pods, correct?

If so, as a follow up, does this mean that the prefill pods will get that too or are they hosted together? Basically I want to make sure that wherever GDS is needed we have that readiness probe.

As for the actual file name and path. This is not Weka-specific so I'll rename it to `gds-sanity-check` or something. I'll also change the default target to be `/etc/cufile.json` since this is the default by NVIDIA.

Dockerfile.ubi
```diff
@@ -435,6 +435,8 @@ RUN --mount=type=cache,target=/var/cache/git \
     uv pip install "nvidia-nccl-cu12>=2.26.2.post1" && \
     rm -rf /tmp/wheels
 
+COPY docs/infra-providers/weka/weka-check-cufile /usr/local/bin
+
```
sdimitro: BTW I was looking at this Dockerfile - I haven't pulled it yet but there's a change we may want to install a couple NVIDIA-related packages for GDS. Is it ok if we add them here? I guess I can always make the change and we discuss afterwards.

guides/weka/README.md
```diff
@@ -0,0 +1,8 @@
+# Well-lit Path: WEKA GPU Direct
+
+Full doc coming soon, this is just meant to track the changes in this PR for now.
+
+1. Add a readiness prob script to the `llm-d` image to detect if the cufile exists on the node and is properly formatted
+2. InitContainer added to run `amgctl` to create the cufile on the node (only implemented for decode so far, will need to add this for prefill too)
+3. Mount the cufile.json from ~/amg_stable/cufile.json on the host to ~/amg_stable/cufile.json on the container
```
sdimitro, line 7: I wonder if it would make sense to load it under `/etc/cufile.json` in the container. E.g. in the decode.yaml we do:
```
      volumes:
        ...
        - name: host-node-cufile
          hostPath:
            path: /etc/cufile.json
            type: File 
```
Will get back to you on this one.

guides/weka/manifests/gateway/base/gateway.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: Gateway
+metadata:
+  name: infra-pd-inference-gateway
+  labels:
+    app.kubernetes.io/instance: infra-pd
+    app.kubernetes.io/gateway: infra-pd-inference-gateway
+    app.kubernetes.io/component: inference-gateway
+spec:
+  listeners:
+    - port: 80
```
sdimitro, line 11: Apologies as this is probably a silly question. I see later in `httproute.yaml` we open 8000 which I assume is for `vLLM`. What's this port 80 about?

guides/weka/manifests/gateway/istio/destinationrule.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: networking.istio.io/v1beta1
+kind: DestinationRule
+metadata:
+  name: infra-wide-ep-inference-gateway
```
sdimitro: These `istio` files are going a bit over my head, though couldn't help but notice that this file is not `infra-pd-inference-gateway`. Is that on purpose?

guides/weka/manifests/modelserver/base/decode.yaml
```diff
@@ -0,0 +1,145 @@
...
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+        - name: create-cufile-on-node # this init container will create the cufile on the node and runs before vllm
+          image: # OUR DOCKER IMAGE GOES HERE
+          command:
+            - amgctl
+          args:
+            - host
+            - config
+            - cufile
```
sdimitro: Nice! Thanks for pointing that out - I'll fill it out!

guides/weka/manifests/gateway/istio/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - op: add
+      path: /spec/rules/0/timeouts
+      value:
+        backendRequest: 0s
+        request: 0s
+labels:
+- pairs:
+    app.kubernetes.io/instance: infra-wide-ep
+    app.kubernetes.io/gateway: infra-wide-ep-inference-gateway
+    app.kubernetes.io/component: inference-gateway
```
sdimitro, line 23: Similar `wide-ep` vs `pd`.

guides/weka/manifests/modelserver/base/decode.yaml
```diff
@@ -0,0 +1,145 @@
...
+          name: metrics-volume
+        - emptyDir:
+            medium: Memory
+            sizeLimit: 16Gi
+          name: shm
+        - emptyDir: {}
+          name: torch-compile-cache
+        - name: weka-pvc
+          persistentVolumeClaim:
+            claimName: wekafs-amg
```
sdimitro: Nice! I think I try changing that to:
```
    - name: weka-storage
      {{- if eq .Values.volume.type "pvc" }}
      persistentVolumeClaim:
        claimName: {{ .Values.volume.pvc.name }}
      {{- else if eq .Values.volume.type "hostPath" }}
      hostPath:
        path: {{ .Values.volume.hostPath.path | quote }}
      {{- else }}
      {{- fail "volume.type must be either 'pvc' or 'hostPath'" }}
      {{- end }}
```
So it is easy for us to experiment with both the actual PVC and a direct mount for development/testing.

guides/weka/manifests/modelserver/base/decode.yaml
```diff
@@ -0,0 +1,145 @@
...
+            - --port
+            - "8200"
+            - --served-model-name
+            - "RedHatAI/Llama-3.3-70B-Instruct-FP8-dynamic"
+            - --tensor-parallel-size
+            - "4"
+            - --block-size
+            - "128"
+            - --kv-transfer-config
+            - '{"kv_connector":"NixlConnector", "kv_role":"kv_both"}'
```
sdimitro: I'd probably want to change this to LMCache. Two questions for you:
(1) Should I add a conditional or environment variable in the Dockerfile to make the upstream LMCache repo parametrizable? This way people can use the NeuralMesh by default, but others can change it (e.g. either use the upstream or their own forks).

(2) We actually wrote the GDS_MT backend for upstream NIXL and have used it with Weka. Out of curiosity have you been able run GDS workloads through LLM-D through NIXL?

guides/weka/manifests/modelserver/base/decode.yaml
```diff
@@ -0,0 +1,145 @@
...
+                fieldRef:
+                  fieldPath: status.podIP
+            - name: VLLM_LOGGING_LEVEL
+              value: DEBUG
+            # - name: DP_SIZE # @gregory-pereira to verify if we still need this
+            #   value: "1"
+            # - name: TP_SIZE # @gregory-pereira to verify if we still need this
+            #   value: "1"
+            - name: HF_HOME
+              value: /model-cache
```
sdimitro: Is this path an LLM-D convention? We're generally used to save our models in the Weka mount so we can load them quickly too, but I don't think we care for now as long as the inference is fast.

Update: I see it further down written as `HF_PATH` which I don't think means anything to Hugging Face as an env variable.

guides/weka/manifests/modelserver/base/decode.yaml
```diff
@@ -0,0 +1,145 @@
...
+              rdma/ib: 1
+            requests:
+              cpu: "16"
+              memory: 64Gi
+              nvidia.com/gpu: "4"
+              rdma/ib: 1
+          readinessProbe: # KUSTOMIZATION overlay
+            exec:
+              command:
+                - "weka-check-cufile ~/amg_stable/cufile.json" # path on node goes here, this is a binary we need to add to our `llm-d` image
```
sdimitro: Cool! Thanks for the guidance!

docs/infra-providers/weka/weka-check-cufile
Gregory-Pereira: Yes this will get run on any inference worker so both prefill and decode. Feel free to rename as you see fit

guides/weka/manifests/gateway/base/gateway.yaml
Gregory-Pereira: Port 8000 is the port of the inference pool, so the port used after traffic has entered the system and is communicating internally. The port 80 is the port the gateway is listening on as the ingress and egress to the system

Dockerfile.ubi
Gregory-Pereira: My fork is always open if you wana push directly to the PR

guides/weka/manifests/modelserver/base/decode.yaml
Gregory-Pereira: So `kustomize` doesn't support templating like helm does, the solution would be for us to either create different `kustomize` directories and patch in host storage path vs using `pvc`, ive updated the example to do that

guides/weka/manifests/gateway/istio/destinationrule.yaml
Gregory-Pereira: Great catch, updating this

guides/weka/manifests/modelserver/base/decode.yaml
Gregory-Pereira: No, you guys would be the first GDS provider we will have worked with. Yes we can both make LMCache repo parametrizable and swap the default to upstream - this is long overdue anyway.

guides/weka/manifests/modelserver/base/decode.yaml
Gregory-Pereira: We definitely can use a model from the weka mount point, we just need to swap the refs to support that

guides/weka/manifests/modelserver/base/decode.yaml
```diff
@@ -0,0 +1,145 @@
...
+          imagePullPolicy: Always
+          ports:
+            - containerPort: 8000
+          resources: {}
+          restartPolicy: Always
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+        - name: create-cufile-on-node # this init container will create the cufile on the node and runs before vllm
+          image: # OUR DOCKER IMAGE GOES HERE
```
Gregory-Pereira: I took a look at your `amg-utils` image and its pretty big - the `nvidia` base image itself is about 3.5 Gb. I know there are other utilities in there such as `amg-qad`, `bpftrace`, etc. What is actually the minimum that we need here just to poll the system and create the `cufile` on the host node? It would create a better experience to use a minimally sized container to do this creation, especially because this run quickly at startup, and thats it. So ideally we create a new slim version of the container.

sdimitro: Oh yeah I'm 100% creating a new container for this. The image is so big because `amgctl` installs PyTorch+LMCache+vLLM and their dependencies on it. We don't need any of these stuff for this :)

guides/weka/README.md
Gregory-Pereira: Sounds good. I think the pieces to note here are:

- `volumes` section, this `hostPath` refers to the location of the file on the local node
- `volumeMount` section, this `mountPath` refers to where we are going to mount this onto the container. My current mental model is that we will mount from the host node at `~/amg_stable/cufile.json` into the container at `/etc/cufile.json` - that way you are using the weka specific implementation on the nodes, but we can consume the cufile in a pattern consistent with other GDS providers

guides/weka/manifests/modelserver/base/decode.yaml
Gregory-Pereira: More context here: https://llm-d.slack.com/archives/C08TB7ZDV7S/p1758376542824289. We will need to test this.

Gregory-Pereira:
Also tracking: https://github.com/weka/amg-utils/pull/4/files

Dockerfile.ubi
Gregory-Pereira: Are these enough: https://github.com/llm-d/llm-d/pull/208/files#diff-3fc20524636ade6cd833fcabb28a94921781961cee8d7fce631922cc10704c76R120-R130? Add whatever else we need I just thought this should be a good start / cover most of it

guides/weka/manifests/modelserver/base/decode.yaml
Gregory-Pereira: Ive updated to try upstream

guides/weka/manifests/gateway/base/gateway.yaml
sdimitro: ok that makes sense. Thank you!

PR #214, opened 2025-09-12
Add monitoring info

sallyom:
I think this would be a nice addition for monitoring docs. I've been testing with these promQL queries and load-generation script. 

This depends on/includes/is a follow-up for https://github.com/llm-d/llm-d/pull/213:  ServiceMonitors in-place for EPP metrics


sallyom:
will open against main branch, closing this one

PR #215, opened 2025-09-12
GKE wide ep

liu-cong:
This is based on previous PR https://github.com/llm-d/llm-d/pull/206

### DIFF
--- a/guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
+++ b/guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
@@ -39,7 +39,7 @@ spec:
           - name: dshm
             emptyDir:
               medium: Memory
-              sizeLimit: 4Gi # 4Gi for NCCL 2.27 to be safe
+              sizeLimit: 3Gi # 3Gi for NCCL 2.27 to be safe
         containers:
         - name: vllm-worker
           image: ghcr.io/llm-d/llm-d-cuda-dev:pr-291@sha256:cc15e6538d654e83ccddea19c665735549bc9fe915b020a2145b5d713bb637cc
@@ -95,7 +95,6 @@ spec:
                                 "num_redundant_experts":"32",
                                 "log_balancedness":"False"}' \
                 --compilation_config '{"cudagraph_mode": "FULL_DECODE_ONLY"}'
-                --gpu-memory-utilization 0.75
           env:
           # VLLM_MOE_DP_CHUNK_SIZE is a tunable parameter.
           # Higher values increase the memory footprint of the hidden states in the MoE layers,
@@ -147,12 +146,12 @@ spec:
               port: 8200
           resources:
             limits:
-              ephemeral-storage: 3Ti
+              ephemeral-storage: 1Ti
               memory: 512Gi
               nvidia.com/gpu: "8"
             requests:
               cpu: 32
-              ephemeral-storage: 3Ti
+              ephemeral-storage: 1Ti
               memory: 512Gi
               nvidia.com/gpu: "8"
           volumeMounts:
--- a/guides/wide-ep-lws/manifests/modelserver/base/prefill.yaml
+++ b/guides/wide-ep-lws/manifests/modelserver/base/prefill.yaml
@@ -122,12 +122,12 @@ spec:
               port: 8000
           resources:
             limits:
-              ephemeral-storage: 3Ti
+              ephemeral-storage: 1Ti
               memory: 512Gi
               nvidia.com/gpu: "8"
             requests:
               cpu: 32
-              ephemeral-storage: 3Ti
+              ephemeral-storage: 1Ti
               memory: 512Gi
               nvidia.com/gpu: "8"
           volumeMounts:


guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
Gregory-Pereira, line 1: May I suggest just re-creating the decode and prefill deployments entirely here? You are welcome to patch them, but the whole reason we split them out of Modelservice in the first place was UX, and this to me does not seem like a very controllable GKE DevX

liu-cong: don't read this for now, most of this file is for debugging, etc. It should be cleaned up significantly

Gregory-Pereira:
Should I tag this as part of release criteria or is this more of a testing setup?

liu-cong:
> Should I tag this as part of release criteria or is this more of a testing setup?

We need something to track the wide wp on GKE validation. For now we can use this one so we don't miss it.

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -78,6 +79,8 @@ spec:
           env:
           - name: VLLM_FUSED_MOE_CHUNK_SIZE
             value: "1024"
+          - name: VLLM_ATTENTION_BACKEND
+            value: "FLASHMLA"
```
smarterclayton: Switch to CUTLASS_MLA now that we have fixed the hang

guides/wide-ep-lws/manifests/modelserver/base/prefill.yaml
```diff
@@ -59,6 +60,8 @@ spec:
                 --trust-remote-code \
                 --kv_transfer_config '{"kv_connector":"NixlConnector","kv_role":"kv_both"}'
           env:
+          - name: VLLM_ATTENTION_BACKEND
+            value: "FLASHMLA"
```
smarterclayton: CUTLASS_MLA

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,329 @@
...
+        value: us-central1-docker.pkg.dev/conliu-gke-dev/llm-d/vllm:clayton-v0.10.2
+      # Env vars 
+      # TODO these should be added to the base yaml
+      - op: add
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/env/-
+        value:
+          name: VLLM_NIXL_SIDE_CHANNEL_PORT
+          value: "6555"
+      - op: add
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/env/-
```
smarterclayton, line 72: Let's drop these for now, we can test again with them later.

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,329 @@
...
+        value:
+          name: TRITON_CACHE_DIR
+          value: "/root/.nv/ComputeCache/.triton"
+      - op: add
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/env/-
+        value:
+          name: DG_JIT_CACHE_DIR
+          value: "/root/.nv/ComputeCache/.deepgemm"
+      - op: add
```
smarterclayton, line 80: Drop since this should be covered by the base image

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,329 @@
...
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/volumeMounts/-
+        value:
+          name: vllm
+          mountPath: /code
+      # Replace container image
+      - op: replace # TODO: delete once we have the official image
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/image
+        value: us-central1-docker.pkg.dev/conliu-gke-dev/llm-d/vllm:clayton-v0.10.2
+      # Env vars
```
smarterclayton, line 70: Why aren't these already in the base image?

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,329 @@
...
+        value:
+          name: vllm
+          emptyDir: {}
+      - op: add
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/volumeMounts/-
+        value:
+          name: vllm
+          mountPath: /code
+      # Replace container image
+      - op: replace # TODO: delete once we have the official image
```
smarterclayton: Are we blocked on the image still?

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,329 @@
...
+          name: gib
+          hostPath:
+            path: /home/kubernetes/bin/gib
+            type: ""
+      - op: add
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/volumeMounts/-
+        value:
+          mountPath: /usr/local/gib
+          name: gib
+      - op: add
```
smarterclayton: I would drop this for now, let's remove scope and keep simpler.

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,329 @@
...
+              {"interfaceName":"eth2","network":"rdma-0"},
+              {"interfaceName":"eth3","network":"rdma-1"},
+              {"interfaceName":"eth4","network":"rdma-2"},
+              {"interfaceName":"eth5","network":"rdma-3"},
+              {"interfaceName":"eth6","network":"rdma-4"},
+              {"interfaceName":"eth7","network":"rdma-5"},
+              {"interfaceName":"eth8","network":"rdma-6"},
+              {"interfaceName":"eth9","network":"rdma-7"}
+            ]
+      - op: add
```
smarterclayton: Are you removing the RDMA requests?

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,329 @@
...
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/affinity
+        value:
+          nodeAffinity:
+            requiredDuringSchedulingIgnoredDuringExecution:
+              nodeSelectorTerms:
+                - matchExpressions:
+                    - key: cloud.google.com/gke-accelerator
+                      operator: In
+                      values:
+                        - nvidia-h200-141gb
```
smarterclayton: I think this is too specific for our guide, the user is expected to select working hardware and our docs should state that.  `nvidia.com/gpu: 8` is sufficient.

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -50,6 +50,7 @@ spec:
               - SYS_RAWIO
             runAsGroup: 0
             runAsUser: 0
+            privileged: true
```
smarterclayton: This is a GKE thing (because we choose not to set PeerMappingOverride by default).  I think this should be set and documented in the GKE patch

smarterclayton: In the GKE side, add a comment link to https://github.com/llm-d/llm-d/pull/296/files (the RDMA section)

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,329 @@
-
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../base
-# TODO: Add GKE specific RDMA patches here.
+patches:
```
smarterclayton, line 5: Add `NVSHMEM_DISABLED_GDRCOPY=true` as per the updated RDMA guidance.

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,329 @@
...
+  - target:
+      kind: LeaderWorkerSet
+      name: wide-ep-llm-d-prefill
+    patch: |-
+      # Replace command and args
+      - op: replace
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/args
+        value:
+          - |
+            if command -v apt-get >/dev/null 2>&1; then
```
smarterclayton: We shouldn't need this.  What are we waiting for on the image?

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
Gregory-Pereira: This is the non default NIXL port (default is 5557), if you specify as seen here it will overwrite, otherwise you don't need to pass it at all

liu-cong:
@smarterclayton Did a major cleanup and moved all the temporary stuff in the end of the kustomization.yaml file to cleanly separate. Once we get the image working we will remove that part. This current state works with your image.

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -145,12 +148,12 @@ spec:
               port: 8200
           resources:
             limits:
-              ephemeral-storage: 64Gi
+              ephemeral-storage: 3Ti
```
liu-cong: I didn't test what's the minimal required storage, just took Clayton's current config. 64Gi is not enough for me.

guides/wide-ep-lws/manifests/modelserver/gke/kustomization.yaml
```diff
@@ -1,6 +1,222 @@
...
+      # This is recommended on gke https://github.com/llm-d/llm-d/pull/296/files.
+      # TODO: Update with doc link once the PR is merged.
+      - op: add
+        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/env/-
+        value:
+          name: NVSHMEM_DISABLED_GDRCOPY
+          value: "true"
+  
+  ######################
+  # TODO: The following is for debug only and will be removed in the final PR.
```
liu-cong: Everything below this should be removed. No need to review.

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -92,6 +95,7 @@ spec:
                                 "num_redundant_experts":"32",
                                 "log_balancedness":"False"}' \
                 --compilation_config '{"cudagraph_mode": "FULL_DECODE_ONLY"}'
+                --gpu-memory-utilization 0.75
```
Gregory-Pereira: Remove this, it was intentional:

```suggestion
               --compilation_config '{"cudagraph_mode": "FULL_DECODE_ONLY"}'
```

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
Gregory-Pereira: 3Ti is a ton but as long as it starts on CKS (which it does) I really don't care right now. Lets scale this back later

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
liu-cong: Without --gpu-memory-utilization 0.75 I ran into cuda out of memory issue on H200.

Gregory-Pereira: From @tlrmchlsmth:

> [it's] intentional. Working around an OOM that's happening in the prefill case. We need to look into it in vLLM. I think it's related to DBO but unsure.

I was able to run this on CKS H200s so I am a bit confused by the OOM issue you mention. Additionally that line of code shouldn't even be running because you dont have a `\` on the line above

PR #219, opened 2025-09-15
Update QUICKSTART and gateway-prereq with more context

smarterclayton:
Provide more information about the key decisions users should make
and the reasons why those decisions matter.

Update the QUICKSTART (where users may encounter these concepts
first) and then update the gateway prereq as it makes the most
choices.

### DIFF
--- a/guides/prereq/gateway-provider/README.md
+++ b/guides/prereq/gateway-provider/README.md
@@ -1,19 +1,71 @@
 # Gateway Provider Prerequisite
 
-This document will guide you through choosing a Gateway provider that can support the llm-d `inference-scheduler` component.
+This document will guide you through configuring a [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/) provider that can support the llm-d [`inference-scheduler`](https://github.com/llm-d/llm-d-inference-scheduler) component.
+
+The key elements are:
+
+* The `inference-scheduler` is an **endpoint picker (EPP)** that decides which model server a given request should go to
+* The **Inference Gateway `InferencePool` Custom Resource** that provisions and configures an `inference-scheduler` on a Kubernetes cluster
+* The **Gateway Custom Resources** that define the Kubernetes-native Gateway API and how traffic reaches an `InferencePool`
+* A **compatible Gateway implementation (control plane)** that provisions and configures load balancers and endpoint pickers in response to the Gateway API and InferencePool API
+
+After this prerequisite is complete you will be able to create `InferencePool` objects on cluster and route traffic to them.
 
 This prerequisite generally requires cluster administration rights.
 
-## Before you begin
+## Why do you need a Gateway?
 
-Prior to deploying a Gateway control plane, you must install the custom resource definitions (CRDs) that add the Kubernetes API objects:
+The inference scheduler provides an extension to [compatible Gateway providers](https://gateway-api-inference-extension.sigs.k8s.io/implementations/gateways/) that optimizes load balancing of LLM traffic across model server replicas.
+
+The integration with a Gateway allows self-hosted models to be exposed in a [wide variety of network topologies including](https://gateway-api.sigs.k8s.io/concepts/use-cases/):
+
+* Internet-facing services
+* Internal to your cluster
+* Through a service mesh
+
+and take advantage of key Gateway features like:
+
+* Traffic splitting for incremental rollout of new models
+* TLS encryption of queries and responses
+
+By integrating with a Gateway -- instead of developing an llm-d specific proxy layer -- llm-d can leverage the high performance of mature proxies like [Envoy](https://www.envoyproxy.io/) and take advantage of existing operational tools for managing traffic to services.
+
+## Select and install an `inference-scheduler` compatible Gateway implementation
+
+llm-d requires you select a [Gateway implementation that supports the inference-scheduler](https://gateway-api-inference-extension.sigs.k8s.io/implementations/gateways/). Your infrastructure may provide a default compatible implementation, or you may choose to deploy a gateway implementation onto your cluster.
+
+### Use an infrastructure provided Gateway implementation
+
+We recommend using the infrastructure provided Gateway with our guides if available.
+
+#### Google Kubernetes Engine (GKE)
+
+GKE automatically enables an inference-compatible Gateway control plane when you enable the `HttpLoadBalancing` addon.  
+
+The key choice for deployment is whether you want to create a regional internal Application Load Balancer - accessible only workloads within your VPC (class name: `gke-l7-rilb`) - or a regional external Application Load Balancer - accessible to the internet (class name: `gke-l7-regional-external-managed`).
+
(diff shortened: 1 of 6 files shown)


guides/prereq/gateway-provider/README.md
```diff
@@ -1,19 +1,71 @@
 # Gateway Provider Prerequisite
 
-This document will guide you through choosing a Gateway provider that can support the llm-d `inference-scheduler` component.
+This document will guide you through configuring a [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/) provider that can support the llm-d [`inference-scheduler`](https://github.com/llm-d/llm-d-inference-scheduler) component.
+
+The key elements are:
+
+* The `inference-scheduler` is an **endpoint picker (EPP)** that decides which model server a given request should go to
+* The **Inference Gateway `InferencePool` CRD** that provisions and configures an `inference-scheduler`
```
liu-cong:
```suggestion
* The **Inference Gateway `InferencePool` Custom Resource Definition (CRD)** that configures an `inference-scheduler`
```

guides/prereq/gateway-provider/README.md
```diff
@@ -1,19 +1,71 @@
...
+The following steps from the [GKE Inference Gateway deployment documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/deploy-gke-inference-gateway) should be run:
 
-Prior to deploying a Gateway control plane, you must install the custom resource definitions (CRDs) that add the Kubernetes API objects:
+1. [Verify your prerequisites](https://cloud.google.com/kubernetes-engine/docs/how-to/deploy-gke-inference-gateway#before-you-begin)
+2. [Prepare your environment](https://cloud.google.com/kubernetes-engine/docs/how-to/deploy-gke-inference-gateway#prepare-environment)
+3. [Create the Gateway](https://cloud.google.com/kubernetes-engine/docs/how-to/deploy-gke-inference-gateway#create-gateway)
+
+The other steps are optional and are not necessary to continue with your guide.
+
+### Verified Gateway implementations
```
liu-cong: Is this section meant to be "self installed gateway implementations" as opposed to the "infrastructure provided" above? If so can we rename?

guides/prereq/gateway-provider/README.md
```diff
@@ -1,19 +1,71 @@
...
+llm-d verifies our guides with a Helm chart that installs and configures the `kgateway` or `istio` Gateway implementations.
+
+#### Before you begin
+
+Prior to deploying a Gateway control plane, you must install the custom resource definitions (CRDs) configuration that adds the Kubernetes API objects:
 
 - [Gateway API v1.3.0 CRDs](https://github.com/kubernetes-sigs/gateway-api/tree/v1.3.0/config/crd)
   - for more information see their [docs](https://gateway-api.sigs.k8s.io/guides/)
-- [Gateway API Inference Extension CRDs v0.5.1](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/v0.5.1/config/crd)
+- [Gateway API Inference Extension CRDs v1.0.0](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/v1.0.0/config/crd)
```
liu-cong, line 65: If I read correctly, this applies to the "self installed" gateways, and for GKE, the CRDs are managed. That implies we need to switch to the v1 inferencepool for GKE examples, which is fine. I can create a follow up PR to update the guides for GKE. 

Alternatively we can keep the alpha inferencepool, but then we need to explain for GKE users also need to install the alpha inferencepool, which can be annoying (they may see an error message saying the v1 inference CRD is already installed)

guides/prereq/gateway-provider/README.md
smarterclayton: Tried another wording, PTAL

guides/prereq/gateway-provider/README.md
smarterclayton: I was creating three buckets:

1. Installed for you
2. We have guides / docs to accelerate you deploying that we test
3. No guides for my gateway provider

Agree wording is not great.

guides/prereq/gateway-provider/README.md
smarterclayton: No, we should not switch to the v1 inference pool yet (discussion yesterday), because we want the examples to be consistent across all gateways.  We verified that the GA GKE gateway deploys the alpha resources and will continue to do so.

liu-cong: OK, then we will have some inconsistency with GKE user guide, to trade off for the consistency in llm-d.

We need to add the alpha CRD installation guide for GKE (since GKE only installs v1). Not sure if we will see an error though due to the v1 CRD collision. Need to test that out.

liu-cong: GKE user guide separates the inferenceobjective CRD installation: https://cloud.google.com/kubernetes-engine/docs/how-to/deploy-gke-inference-gateway#prepare-environment

So I think here we need to document that for GKE, users should apply the alpha CRDs separately (to avoid the collision with GKE managed v1 inf pool) :

```
kubectl apply -f https://github.com/kubernetes-sigs/gateway-api-inference-extension/raw/v1.0.0/config/crd/bases/iinference.networking.x-k8s.io_inferenceobjectives.yaml

kubectl apply -f https://github.com/kubernetes-sigs/gateway-api-inference-extension/raw/v1.0.0/config/crd/bases/iinference.networking.x-k8s.io_inferenceopools.yaml
```

smarterclayton: My understanding is that the alpha inference pool is already installed, but could be wrong.

@kfswain is this the best split we can come up with for managed vs unmanaged?

kfswain: For managed, the v1a2 InferencePool should not be installed.

We can package all experimental APIs into a single manifest & attach it as part of the release artifacts (so cutting down to a single apply and not relying on a branch). 

I can add this fairly quickly, just running it by the other maintainers first so I'm not acting unilaterally.

kfswain: https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v1.0.0/experimental-manifests.yaml will have v1a2 `InferencePool` & v1a1 `InferenceObjective`

Alternatively you can use: https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/latest/download/experimental-manifests.yaml
But that becomes a moving target.

(FWIW I also added a `v1-manifests.yaml` file if someone wants to only install the stable APIs)

liu-cong: Cool! That makes it much easier to split the install. We can ask to add that to GKE docs and just link it here.

smarterclayton: Can we open an internal issue to track the change?  Merging this now and will update when we have that.

liu-cong: Yes GKE public docs are updated with CRD matrix and installation guide. In llm-d user guides we will be consistent with GKE recommendations (use v1 inferencepool). No further action item for this PR.

Resolving this.

PR #220, opened 2025-09-16
Fix version for the helm-diff plugin in pre-req script

natoscott:
The version of helm is fixed at v3.17.3 however helm-diff uses the
latest version, which results in a compatibility-related error now:

"failed to load plugins: [...] error unmarshaling JSON: while decoding JSON: json: unknown field "platformHooks"

Resolve by fixing helm-diff at a similar, known-good vintage as helm.

Signed-off-by: Nathan Scott <nathans@redhat.com>


Gregory-Pereira (dismissed):
/lgtm

natoscott:
Updated to add a commit signature.

PR #221, opened 2025-09-16
docs: Clarify curl requests and responses in getting-started-inferencing.md

JaredTan95:
Separate curl requests from results to make it easier for users to understand


Gregory-Pereira (approved):
/lgtm

PR #228, opened 2025-09-17
Refactor prefill and decode to share a base

liu-cong:
A bit more organized :) And makes it easier to see the diff between prefill and decode, and make common changes.

This shows the diff: https://www.diffchecker.com/VkkFPsPT/, the only diff are:
* Removed the `model-storage` volume from prefill, it's not used
* Set NVIDIA_GDRCOPY to `enabled` for both prefill and decode. Previously it was set to `disabled` for decode, this was a unintended change from #206


liu-cong:
cc @nicolexin

nicolexin:
lgtm!

Gregory-Pereira:
While this is less code and more efficient, I think this adds back some of the less clear configurability that stripping out model service gave us. What do you think @liu-cong?

liu-cong:
> While this is less code and more efficient, I think this adds back some of the less clear configurability that stripping out model service gave us. What do you think @liu-cong?

We can defer this. My hope is that we can further clean up the args section and the diff will be very minimal, and actually makes sense to do so as a way to highlight different configs such as the different `VLLM_ALL2ALL_BACKEND` for prefill and decode.

Gregory-Pereira:
I think I understand your point more clearly now, and yes we do want to highlight that. I think however we are also catering to an audience that deploys vLLM enough to recognize those changes in the yaml, and we could even include some documentation just `diff`ing the two manifests. For now lets hold this and continue to discuss.

PR #230, opened 2025-09-17
swap to stable release of deepep

wseaton:
Change the install of deepep from main branch to recently released stable tag

### DIFF
--- a/docker/Dockerfile.cuda
+++ b/docker/Dockerfile.cuda
@@ -227,10 +227,7 @@ WORKDIR /workspace
 # Define commit SHAs as build args to avoid layer invalidation
 ARG LMCACHE_REPO=https://github.com/neuralmagic/LMCache.git
 ARG LMCACHE_COMMIT_SHA=0db8ae4746a207a72483d716b5f82545a2ead24b
-# ARG VLLM_COMMIT_SHA="8ce5d3198d00631a76e1aa02a57947b46bc7218c" # CORRECT COMMIT FOR RELEASE
-
-# COMMIT FOR TESTING SWAPABILITY
-ARG VLLM_COMMIT_SHA="ef283548f75122b0e8ce49ce2548fbb49446d7c7"
+ARG VLLM_COMMIT_SHA="8ce5d3198d00631a76e1aa02a57947b46bc7218c"
 
 # Define if lmcache should be built 
 # Clone repositories with cache mounts
@@ -250,9 +247,7 @@ WORKDIR /workspace/vllm
 
 # Install DeepEP and DeepGEMM dependencies
 ARG DEEPEP_REPO="https://github.com/deepseek-ai/DeepEP"
-# Pin to commit at stable v1.2.1 tag
 ARG DEEPEP_VERSION="v1.2.1"
-# ARG DEEPEP_VERSION="da6ca24ef7c2594e058def142adfb8f2e31a7f85" # claytons stable
 ARG DEEPGEMM_REPO="https://github.com/deepseek-ai/DeepGEMM"
 ARG PPLX_KERNELS_REPO="https://github.com/perplexityai/pplx-kernels"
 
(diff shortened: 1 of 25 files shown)


Gregory-Pereira:
I just rebased

Gregory-Pereira:
Next steps to work on: Deep EP commit still not being honored, and its still ignoring pulling the precompiled shared libraries from the vllm wheel index - make sure it exists in the index and debug with pip in a container

docker/Dockerfile.cuda
```diff
@@ -225,15 +225,16 @@ RUN echo "/usr/local/lib" > /etc/ld.so.conf.d/local.conf && \
 WORKDIR /workspace
 
 # Define commit SHAs as build args to avoid layer invalidation
+ARG LMCACHE_REPO=https://github.com/neuralmagic/LMCache.git
 ARG LMCACHE_COMMIT_SHA=0db8ae4746a207a72483d716b5f82545a2ead24b
-ARG VLLM_COMMIT_SHA=e017120ed146cc3069d18428322d67881cb95e67
+ARG VLLM_COMMIT_SHA=8ce5d3198d00631a76e1aa02a57947b46bc7218c
```
tlrmchlsmth: looks good, tip of v0.11.0 branch as I'm writing this comment https://github.com/vllm-project/vllm/commits/releases/v0.11.0/

docker/Dockerfile.cuda
```diff
@@ -245,9 +246,12 @@ RUN --mount=type=cache,target=/var/cache/git \
 WORKDIR /workspace/vllm
 
 # Install DeepEP and DeepGEMM dependencies
-ARG DEEPEP_URL="https://github.com/deepseek-ai/DeepEP"
-ARG DEEPGEMM_URL="https://github.com/deepseek-ai/DeepGEMM"
-ARG PPLX_KERNELS_URL="https://github.com/perplexityai/pplx-kernels"
+ARG DEEPEP_REPO="https://github.com/deepseek-ai/DeepEP"
+# Pin to commit at stable v1.2.1 tag
+ARG DEEPEP_VERSION="9af0e0d0e74f3577af1979c9b9e1ac2cad0104ee"
```
tlrmchlsmth: Can this be a tag instead? I think it would work and be more readable to do:
```suggestion
ARG DEEPEP_VERSION="v1.2.1"
```

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -107,7 +108,7 @@ spec:
           - name: NCCL_IB_HCA
             value: ibp
           - name: VLLM_LOGGING_LEVEL
-            value: INFO
+            value: DEBUG
```
tlrmchlsmth: Should we support production vs debug configs somehow? It's kind of hard to track down all the places to turn the logging in order to run benchmarks

helpers/interactive-pod/manifests/deployment.yaml
```diff
@@ -0,0 +1,42 @@
...
+    type: Recreate # don't run two interactive pods at once during updates
+  selector:
+    matchLabels:
+      app: interactive-pod
+  template:
+    metadata:
+      labels:
+        app: interactive-pod
+    spec:
+      serviceAccountName: intereactive-pod
```
tlrmchlsmth: there's a couple of other spots too
```suggestion
      serviceAccountName: interactive-pod
```

docker/Dockerfile.cuda
```diff
@@ -432,9 +466,7 @@ ENV PATH="${VIRTUAL_ENV}/bin:/usr/local/nvidia/bin:${PATH}" \
     TORCH_NCCL_DUMP_ON_TIMEOUT=0 \
     VLLM_SKIP_P2P_CHECK=1 \
     VLLM_CACHE_ROOT=/tmp/vllm \
-    # NOTE: workaround for hang with XET. This should be removed 
-    # in a future release when we get to the bottom of the issue.
-    HF_HUB_DISABLE_XET=1
```
Gregory-Pereira, line 437: Removing because we can set this in coreweave kustomize overlay

Gregory-Pereira:
cc @wseaton @tlrmchlsmth PTAL, this works for RC wheels and old wheels. If started with `/bin/bash` instead of the generic serve command it should output a warning message to users that it was built with pre-compiled binaries and if they want to do live development in the image they need to account for that. I tested the image against all the examples today.

Gregory-Pereira (approved):
/lgtm

PR #233, opened 2025-09-18
Readme nits

chcost:
Fixing logo and header alignment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
-<p>
+<p style="text-align: center;">
   <picture>
     <source media="(prefers-color-scheme: dark)">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,13 @@
-<p>
+<p align="center">
```
smarterclayton, line 1: Heh, I actually moved this to the left because I thought it looked better.  If centered is what people want, I'm ok with that :)

chcost: Ah, I wasn't sure it was intentional (and didn't look at the commit history). I don't feel super strongly about it, but I'm more on the centered camp :) . If we are all okay with it, I'd go with centered.

smarterclayton (approved):

PR #234, opened 2025-09-18
Deprecate all references to InferenceModel

### LINKED ISSUE #226: Remove all references to InferenceModel
InferenceModel is deprecated, its successor is InferenceObjective but it is an optional API that we don't really need to introduce right now.

ahg-g:
By default creation is set to false in modelservice, so it should be safe to remove from the well lit path charts: https://github.com/llm-d-incubation/llm-d-modelservice/blob/main/charts/llm-d-modelservice/values.yaml#L121; @Gregory-Pereira can you confirm?

Fixes #226


Gregory-Pereira (approved):
/lgtm

PR #238, opened 2025-09-19
Update the image pull policy

JiaweiGithub:


Gregory-Pereira:
Is there any reason we want to set the `imagePullPolicy` to `IfNotPresent` rather than `Always`? From our perspective were frequently testing feature working coming from DBO or other upstream vLLM and kernel fixes, so its helpful for us to always pull for new images when we have use PR images like `ghcr.io/llm-d/llm-d-dev:pr-188`

PR #254, opened 2025-09-24
Update Dockerfile.ubi to CUDA 12.9.1

tlrmchlsmth:
Update to CUDA 12.9.1. I think this is low risk to upgrade to. We know about performance optimizations related to instruction scheduling that are new in 12.9, so expect to see an increase in performance.

Also squash this warning in our startup logs:
```
Warning: please use at least NVCC 12.9 for the best DeepGEMM performance
```


smarterclayton (dismissed):

smarterclayton:
Will test in https://github.com/smarterclayton/vllm-dp-lws/blob/main_20250923_cuda129/Dockerfile as well

tlrmchlsmth:
@Gregory-Pereira yeah, let's get this into the release

Gregory-Pereira (approved):
/lgtm

PR #255, opened 2025-09-24
Remove TRANSFORMERS_CACHE env

tlrmchlsmth:
See the following warning. I don't think we should set it anymore. If we're setting it for backwards-compatibility, we should make sure to set it the same as HF_HOME:

```
wide-ep-llm-d-decode-0-1 vllm-worker /opt/vllm/lib64/python3.12/site-packages/transformers/utils/hub.py:111: FutureWarning: Using `TRANSFORMERS_CACHE` is deprecated and will be removed in v5 of Transformers. Use `HF_HOME` instead.
```


Gregory-Pereira (dismissed):
yup been seeing this a lot too.
/lgtm

tlrmchlsmth:
> Woops please also remove the directory creation: https://github.com/llm-d/llm-d/pull/255/files#diff-3fc20524636ade6cd833fcabb28a94921781961cee8d7fce631922cc10704c76R385

nice catch - removed

Gregory-Pereira (approved):
/lgtm

Image builds fine now.

tlrmchlsmth:
anybody know what's up with these trivy vulnerability scanner failures?

Gregory-Pereira:
> anybody know what's up with these trivy vulnerability scanner failures?

No but don't block on them, they are not a piece of merge criteria. Ill look into it tomorrow

PR #267, opened 2025-09-26
unify darwin/arm64 with other platforms when install helmfile in install-deps.sh

yitingdc:
when install helmfile in install-deps.sh, darwin/arm64 used a hard code name, it can be unified with other os/arch with parameters


Gregory-Pereira (approved):
/lgtm

yitingdc:
Hi @Gregory-Pereira , thanks for reviewing and approving, it seems that ci is stuck for about 12 hours, what should I do now? or could you help to make it working? thanks very much!

PR #278, opened 2025-09-30
Add container with EFA support

andjsmi:
This pull request adds a Dockerfile built with EFA support.

Currently the Dockerfile (`Dockerfile.cuda`) uses the release of UCX bundled in with NIXL which is missing some support for EFA so on AWS, disaggregated serving is unable to use the srd transport.

### DIFF
--- a/docker/Dockerfile.aws
+++ b/docker/Dockerfile.aws
@@ -12,7 +12,7 @@ ARG CUDA_MINOR=8
 ARG CUDA_PATCH=1
 ARG PYTHON_VERSION
 
-ARG USE_SCCACHE=false
+ARG USE_SCCACHE=true
 
 COPY sccache.config.toml /tmp
 
@@ -22,7 +22,7 @@ RUN --mount=type=secret,id=aws_access_key_id \
     if [ "${USE_SCCACHE}" = "true" ]; then \
         set -Eeuo pipefail && \
         # install sccache
-        apt-get install -y openssl-devel && \
+        apt-get update && apt-get install -y libssl-dev && \
         mkdir -p /tmp/sccache && \
         cd /tmp/sccache && \
         curl -sLO https://github.com/mozilla/sccache/releases/download/v0.10.0/sccache-v0.10.0-x86_64-unknown-linux-musl.tar.gz && \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.aws
```diff
@@ -0,0 +1,493 @@
...
+
+# Copy patches before build
+COPY patches/ /tmp/patches/
+
+# Build and install NVSHMEM from source with coreweave patch
+RUN cd /tmp && \
+    wget https://developer.download.nvidia.com/compute/redist/nvshmem/${NVSHMEM_VERSION}/source/nvshmem_src_cuda12-all-all-${NVSHMEM_VERSION}.tar.gz -O nvshmem_src_cuda${CUDA_MAJOR}.tar.gz && \
+    tar -xf nvshmem_src_cuda${CUDA_MAJOR}.tar.gz && \
+    cd nvshmem_src && \
+    git apply /tmp/patches/cks_nvshmem${NVSHMEM_VERSION}.patch && \
```
Gregory-Pereira: I don't think you need the cks_nvshmem patch if your not running this on Coreweave. Also above on line 181

docker/Dockerfile.aws
```diff
@@ -0,0 +1,493 @@
...
+# BUILD STAGE - Install build dependencies and create wheels
+# ============================================================================
+FROM nvcr.io/nvidia/cuda-dl-base:25.01-cuda12.8-devel-ubuntu24.04 AS builder
+
+ARG CUDA_MAJOR=12
+ARG CUDA_MINOR=8
+ARG CUDA_PATCH=1
+ARG PYTHON_VERSION
+
+ARG USE_SCCACHE=false
```
Gregory-Pereira: If you don't want to use our SCCache bucket thats perfectly fine, but in that case I would remove it from builds

docker/Dockerfile.aws
```diff
@@ -0,0 +1,493 @@
...
+        echo "output=json"; \
+        } > /root/.aws/config && \
+        export AWS_SHARED_CREDENTIALS_FILE=/root/.aws/credentials AWS_CONFIG_FILE=/root/.aws/config AWS_DEFAULT_REGION="us-west-2" AWS_PROFILE=default && \
+        export SCCACHE_IDLE_TIMEOUT=0 && \
+        export SCCACHE_LOG=debug && \
+        export RUSTC_WRAPPER="sccache" && \
+        /usr/local/bin/sccache --start-server && \
+        # verify sccache is working with the s3 backend
+        /usr/local/bin/sccache --show-stats; \
+    fi
```
Gregory-Pereira, line 59: Again, if you don't want to use SCCache you don't have to but then remove this.

docker/Dockerfile.aws
andjsmi: That was just for my own testing when running it. Have switched to use sccache

andjsmi:
> Depends on: [andjsmi#1](https://github.com/andjsmi/llm-d/pull/1)
> 
> Tested in: #282 --> produced build: ghcr.io/llm-d/llm-d-aws-dev:pr-282@sha256:a7a6186e44e2de9f35cabdb9b259839acd52e71a6f446c1dac593f8232d798b5
> 
> Testing for images were leaving up to the providers themselves so if that works for you were good to go.

Works for me.

Gregory-Pereira:
Sounds good if it works. Lets merge my CI PR into yours and then we can look to merge yours.

Gregory-Pereira (approved):
/lgtm

PR #290, opened 2025-10-01
Update WideEP well-lit-path with benchmarking configs

tlrmchlsmth:


guides/wide-ep-lws/manifests/gateway/istio/configmap.yaml
```diff
@@ -0,0 +1,35 @@
...
+              - misc:warn
+              - --log_output_level
+              - default:warn
+            resources:
+              limits:
+                cpu: "8"
+                memory: 16Gi
+              requests:
+                cpu: 2
+                memory: 4Gi
```
tlrmchlsmth, line 32: needed to keep isio from getting oomkilled

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: leaderworkerset.x-k8s.io/v1
+  apiVersion: leaderworkerset.x-k8s.io/v1
```
Gregory-Pereira:
```suggestion
apiVersion: leaderworkerset.x-k8s.io/v1
```
linting

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -59,25 +59,54 @@ spec:
               #################
               # RUN vLLM decode worker
               #################
+              set -euo pipefail
```
Gregory-Pereira: Do we want this pipefail? This means if any variable is unbound it will fast fail, is this something we want? Wouldn't we see that in vLLM startup log anyway which seems like it would be more helpful in debugging the issue than just "unbound variable XYZ"

Gregory-Pereira (dismissed):
/lgtm

Gregory-Pereira (approved):
/lgtm

PR #295, opened 2025-10-01
Remove pool-group flag, it is configured by the apiVersion already

liu-cong:
This is a regression introduced by previous PR. The `pool-group` will be configured by the `apiVersion` value which is already set.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 *.swp
+*.history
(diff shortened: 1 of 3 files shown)


Gregory-Pereira:
We should keep verbosity 1 though, this is a production example

Gregory-Pereira (approved):
/lgtm

PR #296, opened 2025-10-01
Provide more information about RDMA nodes for GKE

smarterclayton:


wseaton (dismissed):
Thanks @smarterclayton, this is excellent. Setting the standard for provider docs.

PR #311, opened 2025-10-03
gateway-provider: fix proxyMetadata type in istio MeshConfig

asharkhan3101:
As [per docs](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig-proxy_metadata) for proxyMetadata in MeshConfig is a `map<string, string>` and the current config was failing with 
```
2025-10-03T11:16:58.109506859Z 2025-10-03T11:16:58.109347Z      warn    invalid mesh config, using last known state: failed to convert to proto. json: cannot unmarshal bool into Go value of type string
```
because a boolean value was being passed.

### DIFF
--- a/guides/prereq/gateway-provider/istio.helmfile.yaml
+++ b/guides/prereq/gateway-provider/istio.helmfile.yaml
@@ -19,7 +19,7 @@ releases:
       - meshConfig:
           defaultConfig:
             proxyMetadata:
-              SUPPORT_GATEWAY_API_INFERENCE_EXTENSION: true
+              SUPPORT_GATEWAY_API_INFERENCE_EXTENSION: "true"
         pilot:
           env:
             SUPPORT_GATEWAY_API_INFERENCE_EXTENSION: true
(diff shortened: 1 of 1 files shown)


guides/prereq/gateway-provider/istio.helmfile.yaml
```diff
@@ -19,7 +19,7 @@ releases:
       - meshConfig:
           defaultConfig:
             proxyMetadata:
-              SUPPORT_GATEWAY_API_INFERENCE_EXTENSION: true
+              SUPPORT_GATEWAY_API_INFERENCE_EXTENSION: "true"
         pilot:
           env:
             SUPPORT_GATEWAY_API_INFERENCE_EXTENSION: true
```
Gregory-Pereira: After looking at this I believe the same is true with pilot env values:
```suggestion
            SUPPORT_GATEWAY_API_INFERENCE_EXTENSION: "true"
```

Gregory-Pereira (changes requested):
Great catch and fix! Could you do the same with pilot env? I found its the same.

asharkhan3101:
Done!

Gregory-Pereira (approved):
/lgtm

PR #320, opened 2025-10-06
Create a GKE specific Dockerfile to work around RDMA issue

smarterclayton:
While we debug the issue, avoid blocking the release. Take the
tested version of the Ubuntu image from smarterclayton/vllm-dp-lws
and make it as consistent as possible with the existing UBI image.


Gregory-Pereira (approved):
/lgtm

PR #324, opened 2025-10-07
refactor dockerfiles to a set bash scripts

wseaton:
Inine docker RUN commands are complicated. Let's move the complex build logic into standalone scripts, so that we can run shellcheck and other tools against them, but it also makes looking at diffs and reviews simpler.

Also:
- Add pre-commit with markdown formatting/linting, as well as hadolint and shellcheck
- Update vllm wheel for B200 fixes


docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -0,0 +1,68 @@
...
+    else
+        echo "Compiling fully from source. Either precompile disabled or wheel not found in index from main."
+        unset VLLM_USE_PRECOMPILED VLLM_PRECOMPILED_WHEEL_LOCATION || true
+        INSTALL_PACKAGES+=(-e /opt/vllm-source)
+        rm /opt/warn-vllm-precompiled.sh
+    fi
+fi
+
+# add nvidia-nccl
+INSTALL_PACKAGES+=("nvidia-nccl-cu12>=2.26.2.post1")
```
smarterclayton: This is no longer necessary after torch 2.8

docker/scripts/cuda/builder/build-nvshmem.sh
```diff
@@ -0,0 +1,55 @@
...
+
+git apply /tmp/patches/cks_nvshmem"${NVSHMEM_VERSION}".patch
+
+mkdir build
+cd build
+
+cmake \
+    -G Ninja \
+    -DNVSHMEM_PREFIX="${NVSHMEM_DIR}" \
+    -DCMAKE_CUDA_ARCHITECTURES="90a;100" \
```
smarterclayton: Would prefer this be an env variable from the top dockerfile

.github/workflows/ci-pr-checks.yaml
```diff
@@ -4,20 +4,41 @@ on:
   push:
     branches:
       - main
-      - dev  
-    paths:
-      - ./**/*.md
-      - .github/workflows/ci-pr-checks.yaml
-      - .github/actions/markdown-link-checker/action.yaml
+      - dev
```
Gregory-Pereira:
```suggestion
```
We don't need `dev` branch

.github/workflows/ci-pr-checks.yaml
```diff
@@ -4,20 +4,41 @@ on:
...
   pull_request:
     branches:
       - main
       - dev
```
Gregory-Pereira:
```suggestion
```
We don't need dev branch

docker/scripts/common/setup-sccache.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# sets up sccache environment variables and starts the server
+# expects USE_SCCACHE env var to be set
```
Gregory-Pereira: If we are abstracting this into its own script we should document all ENV vars that are needed by this beyond just `USE_SCCACHE`

docker/scripts/common/setup-sccache.sh
```diff
@@ -0,0 +1,39 @@
...
+
+if [ "${USE_SCCACHE}" = "true" ]; then
+    # set up AWS credentials if secrets are available
+    if [ -f "/run/secrets/aws_access_key_id" ] && [ -f "/run/secrets/aws_secret_access_key" ]; then
+        AWS_ACCESS_KEY_ID="$(cat /run/secrets/aws_access_key_id)"
+        export AWS_ACCESS_KEY_ID
+        AWS_SECRET_ACCESS_KEY="$(cat /run/secrets/aws_secret_access_key)"
+        export AWS_SECRET_ACCESS_KEY
```
Gregory-Pereira, line 14: Why are these on 2 lines each?

docker/scripts/cuda/runtime/install-vllm.sh
wseaton: nice, ty

docker/scripts/cuda/builder/build-gdrcopy.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# builds and installs gdrcopy from source
+
+# shellcheck source=/dev/null
+source /usr/local/bin/setup-sccache
+
+git clone https://github.com/NVIDIA/gdrcopy.git
+cd gdrcopy
+
+CC="sccache gcc" CXX="sccache g++" PREFIX=/usr/local DESTLIB=/usr/local/lib make lib_install
```
Gregory-Pereira: If were not using SCCache we don't want to wrap our compiler calls in it
```suggestion
if [ "${USE_SCCACHE}" = "true" ]; then
    export CC="sccache gcc" CXX="sccache g++"
fi
PREFIX=/usr/local DESTLIB=/usr/local/lib make lib_install
```

docker/scripts/cuda/builder/build-nvshmem.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+# builds and installs NVSHMEM from source with coreweave patch
+# expects CUDA_MAJOR, NVSHMEM_VERSION, NVSHMEM_DIR, NVSHMEM_CUDA_ARCHITECTURES env vars to be set
```
Gregory-Pereira: Why separate out `NVSHMEM_CUDA_ARCHITECTURES`? Shouldn't this be in sync with the rest of our `TORCH_CUDA_ARCH_LIST`? Also this script needs our `PYTHON_VERSION` set

docker/scripts/cuda/builder/install-base-packages.sh
Gregory-Pereira, line 1: Does this need to be separated out of the `Dockerfile`? I don't think we want to remove everything from the `Dockerfile`, just the complicated bits should be refactored into scripts

docker/scripts/cuda/builder/install-sccache.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+# installs sccache binary from github releases and verifies connectivity
```
Gregory-Pereira, line 4: Doesn't this depend on `openssl` or `libssl` dependeing on if your in `RHEL` or `Ubuntu` based container? Maybe call that out

docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+# installs vllm and dependencies in runtime stage
+# expects VIRTUAL_ENV, CUDA_MAJOR, VLLM_REPO, VLLM_COMMIT_SHA, VLLM_PREBUILT, VLLM_USE_PRECOMPILED env vars
```
Gregory-Pereira: Depends on `jq`

docker/Dockerfile.cuda
```diff
@@ -13,69 +13,27 @@ ARG CUDA_PATCH=1
...
 # Create UV constraint files
+# hadolint ignore=DL3059
 RUN cat > /tmp/build-constraints.txt <<'EOF'
 torch==2.8.0
 EOF
 
+# hadolint ignore=DL3059
 RUN cat > /tmp/constraints.txt <<'EOF'
 torch==2.8.0
 EOF
```
Gregory-Pereira: If hadolint doesnt like this, couldn't we just `COPY` in a constraints file and ditch `HEREDOC`? I don't see a reason not to since were already totally refactoring this

Gregory-Pereira (commented):
Some questions to start

docker/Dockerfile.cuda
wseaton: good point, this was done since the dockerfile was "flat" before, let's move it out

.github/workflows/build-image.yml
```diff
@@ -23,7 +23,39 @@ env:
   REGISTRY: ghcr.io
   IMAGE_NAME: llm-d
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
Gregory-Pereira, line 49: I didn't opt for this because sometimes I push another change after just to update guides or kustomize manifests and I don't need the rebuild. This strategy will mean forcing a new build to complete, but will cut down on the number of jobs we have - not sure where we stand on that tradeoff

.github/workflows/build-image.yml
```diff
@@ -39,17 +71,20 @@ jobs:
...
+              - 'docker/scripts/**'
             aws:
               - 'docker/Dockerfile.aws'
+              - 'docker/scripts/**'
             gke:
               - 'docker/Dockerfile.gke'
 
   cuda-build-llm-d:
-    needs: detect-changes
+    needs: [pre-commit, detect-changes]
```
Gregory-Pereira: Do we really want linting to block image builds?

docker/scripts/common/setup-sccache.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+# sets up sccache environment variables and starts the server
+#
+# Required environment variables:
+# - USE_SCCACHE: whether to configure and start sccache (true/false)
```
Gregory-Pereira, line 5: Not actually required, if unset it will act as false

docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -0,0 +1,70 @@
...
+git -C /opt/vllm-source config --system --add safe.directory /opt/vllm-source
+git -C /opt/vllm-source fetch --depth=1 origin "${VLLM_COMMIT_SHA}" || true
+git -C /opt/vllm-source checkout -q "${VLLM_COMMIT_SHA}"
+
+# detect if prebuilt wheel exists
+WHEEL_URL=$(pip install \
+  --no-cache-dir \
+  --no-index \
+  --no-deps \
+  --find-links "https://wheels.vllm.ai/${VLLM_COMMIT_SHA}/vllm/" \
```
Gregory-Pereira: Should we enable separating the precompiled binaries commit from the source commit like in: https://github.com/llm-d/llm-d/pull/362/files that we use for testing EPLB?

docker/scripts/cuda/builder/build-nvshmem.sh
wseaton: They are a subset basically, nvshmem (as required by deepep and pplx) doesn't require all architectures in vllm's `TORCH_CUDA_ARCH_LIST`, so this is a build optimization.

docker/scripts/cuda/builder/install-base-packages.sh
wseaton: I think it should be split, as in the future to support arm or to do variant OS builds we'll need to refactor this script into more of a "builder", which is going to result in more bash-isms creeping their way into the inline run. (Like how in install vllm we append a bunch of package names dynamically to a list). They are easier to deal with IMHO pulled out into a script

Gregory-Pereira (commented):
Some more questions and comments

docker/scripts/cuda/runtime/install-vllm.sh
wseaton: Yes, this is definitely useful, especially combined with workflow dispatch. Want me to add as scope to this PR?

.github/workflows/build-image.yml
wseaton: Good point, let's disable

.github/workflows/build-image.yml
wseaton: the arm image caching is so good now the builds complete in <3m with no changes, if we can get caching working on the self-hosted runner I think we can have our cake and eat it too here?

Gregory-Pereira: Sounds great

.github/workflows/build-image.yml
```diff
@@ -81,19 +118,27 @@ jobs:
             type=sha
             type=raw,value=latest,enable={{is_default_branch}}
 
-      - name: Build & Push (official action)
-        id: build-and-push
+      - name: Set platform variables
+        id: platform
+        run: |
+          ARCH=$(echo "${{ matrix.platform }}" | cut -d/ -f2)
+          echo "arch=${ARCH}" >> $GITHUB_OUTPUT
```
Gregory-Pereira: Do we need a separate step here, can we just add another variable to that matrix entry for the image name (with arch included)?

.github/workflows/build-image.yml
```diff
@@ -131,7 +238,7 @@ jobs:
           fi
 
   aws-build-llm-d:
-    needs: detect-changes
+    needs: [pre-commit, detect-changes]
```
Gregory-Pereira: Same as before do we want pre-commit to block builds?

.github/workflows/build-image.yml
```diff
@@ -210,7 +317,7 @@ jobs:
           fi
 
   gke-build-llm-d:
-    needs: detect-changes
+    needs: [pre-commit, detect-changes]
```
Gregory-Pereira: Same as before do we want pre-commit to block builds?

.github/workflows/build-image.yml
```diff
@@ -289,7 +396,7 @@ jobs:
           fi
 
   xpu-build-llm-d:
-    needs: detect-changes
+    needs: [pre-commit, detect-changes]
```
Gregory-Pereira: Same as before do we want pre-commit to block builds?

.github/workflows/ci-pr-checks.yaml
```diff
@@ -4,20 +4,39 @@ on:
   push:
     branches:
       - main
-      - dev  
-    paths:
-      - ./**/*.md
-      - .github/workflows/ci-pr-checks.yaml
-      - .github/actions/markdown-link-checker/action.yaml
```
Gregory-Pereira: Should we remove push here? This is an PR CI check for linting, not sure what it gives us for code thats already merged

docker/scripts/cuda/runtime/install-vllm.sh
Gregory-Pereira: 362 is a test PR and not meant to merge, I think it would be a good idea to add it here.

docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -0,0 +1,73 @@
...
+if [ "${VLLM_PREBUILT}" = "1" ]; then
+  if [ -z "${WHEEL_URL}" ]; then
+    echo "VLLM_PREBUILT set but no platform compatible wheel exists for: https://wheels.vllm.ai/${VLLM_COMMIT_SHA}/vllm/"
+    exit 1
+  fi
+  INSTALL_PACKAGES+=("${WHEEL_URL}")
+  rm /opt/warn-vllm-precompiled.sh
+else
+  if [ "${VLLM_USE_PRECOMPILED}" = "1" ] && [ -n "${WHEEL_URL}" ]; then
+    echo "Using precompiled binaries and shared libraries for commit: ${VLLM_COMMIT_SHA}."
```
Gregory-Pereira: If we want to add the separation of precompiled binary commit sha and source sha we should update this here.

docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -0,0 +1,73 @@
...
+    export VLLM_USE_PRECOMPILED=1
+    export VLLM_PRECOMPILED_WHEEL_LOCATION="${WHEEL_URL}"
+    INSTALL_PACKAGES+=(-e /opt/vllm-source)
```
Gregory-Pereira, line 59: Wont this be treated as a package rather than an argument because this will add it to the array? Why don't we just add it to the "${WHEEL_URL}" before installing and move adding that to `INSTALL_PACKAGES` later?

.golangci.yml
Gregory-Pereira, line 1: Delete this file entirely, we have 0 go code in this repo. Andy added this as a template to every repo

scripts/lint-envvars.py
Gregory-Pereira, line 1: I am worried that these Dockerfile + Build script Env Var enforcement and linters will cause more trouble than they are worth. It seems to me that by just requiring people to state their env vars without enforcement in a build script will be enough  and for the comparison across dockerfile and scripts we can just have an ENV block right above the script run to compare + leverage `set -exou pipefail` as much as we can. Not 100% opposed to it, I just feel like this will add mental complexity and overhead to maintain for little value

Gregory-Pereira (commented):
Some more comments and questions

scripts/lint-envvars.py
Gregory-Pereira: Based on slack discussions, we have decided to test out these scripts and see how they do. If they introduce overhead later on we can always remove them later.

.github/workflows/build-image.yml
wseaton: Good call, fixing this here. We move to the matrix.include setup in the arm PR anyhow

Gregory-Pereira: Sounds good, then ignore the comment. Will resolve

docker/scripts/cuda/runtime/install-vllm.sh
wseaton: > Wont this be treated as a package rather than an argument because this will add it to the array? 

the array is going to cause `-e` and `/opt/vllm-source` to be splat into different arguments, which is what we want since it's a kwarg

> Why don't we just add it to the "${WHEEL_URL}" before installing and move adding that to INSTALL_PACKAGES later?

There are two install modes in this script: 

1) install from wheel, non editable
2) install from source as editable w/ precompiled kernels

The first doesn't require `-e` so I think this is correct

docker/scripts/cuda/runtime/install-vllm.sh
wseaton: added workflow dispatch config

Gregory-Pereira:
CUDA Manifest is fixed in the ARM PR, deferring that till then

Gregory-Pereira (approved):
/lgtm

PR #326, opened 2025-10-07
[guides] update modelservice to v0.2.12

poussa:


guides/inference-scheduling/ms-inference-scheduling/values_xpu.yaml
```diff
@@ -70,11 +70,9 @@ decode:
       limits:
         memory: 24Gi
         cpu: "8"
-        gpu.intel.com/i915: "1"
```
yuanwu2017, line 73: Why remove this line？

poussa: Because in v0.2.12, the limits and requests for extended resources comes from the `accelerator.type` automagically (via Help templating).

yuanwu2017:
inference-scheduling CI passed. https://github.com/llm-d/llm-d/actions/runs/18702775156

poussa:
> I have no problem with upgrading these helm chart version, the only issue in accepting this change is that I have to test all of our guides against this, and right now were putting all of our the `v0.3.1` release. Will take a look as soon as that gets out

That's fine. This PR needs to be updated anyway since the model-service helm chart has released a more recent version.

PR #338, opened 2025-10-09
remove requirements for specifying environment when tearing down

Gregory-Pereira:
Another release prep PR. This requirement was removed with how we handle the `helmfile`s and common values files.

### DIFF
--- a/guides/inference-scheduling/README.md
+++ b/guides/inference-scheduling/README.md
@@ -124,7 +124,7 @@ To remove the deployment:
 
 ```bash
 # From examples/inference-scheduling
-helmfile destroy -n ${NAMESPACE} -e <your_environment>
+helmfile destroy -n ${NAMESPACE}
 
 # Or uninstall manually
 helm uninstall infra-inference-scheduling -n ${NAMESPACE}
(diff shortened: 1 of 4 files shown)


guides/inference-scheduling/README.md
```diff
@@ -134,8 +134,6 @@ helm uninstall ms-inference-scheduling -n ${NAMESPACE}
 
 **_NOTE:_** If you set the `$RELEASE_NAME_POSTFIX` environment variable, your release names will be different from the command above: `infra-$RELEASE_NAME_POSTFIX`, `gaie-$RELEASE_NAME_POSTFIX` and `ms-$RELEASE_NAME_POSTFIX`.
 
-**_NOTE:_** You need to specify your `environment` with the `-e <environment>` flag to `helmfile` for removing a installation of the guide when using a non-default option. IE if you deploy with `-e istio` and undeploy `-e istioBench` or vice versa, it may fail. If you encounter this it is recommended to manually uninstall all 3 releases with `helm` as shown above.
```
liu-cong, line 137: did we make and change so the env is not required any more? or it wasn't needed from the beginning?

Gregory-Pereira: Yes, we did need this at one point but its been changed for a few weeks. This is based on pulling common env values. Before it would complain if you used a different env because those values would not exist in the relevant charts for tear down. How we solve this is create the fields we just dont assign them a value, see: https://github.com/llm-d/llm-d/blame/main/guides/prereq/gateway-provider/common-configurations/istio.yaml#L17-L21, which removes the error from `helmfile`'s perspective

PR #351, opened 2025-10-11
[WIP] Remove GCP RDMA config

smarterclayton:
The previous environment has the effect of binding all devices to
the first RDMA NIC, reducing bandwidth to 1/8th of the total
capacity. An upstream change will be made to DeepEP or vLLM to
yield a permanent fix, but in the meantime we should revert to
the default behavior which yields about half the target bandwidth.


Gregory-Pereira:
As with #349, this was a debug PR tagged on v0.3.1 to track the issue which one was not filed for. As such I will remove this from the v0.3.1 release tracker now that #407 has been tagged into the release tracker

Issue #346, opened 2025-10-09
[Feat] Upgrade NIXL to v0.6.1


Gregory-Pereira:
nixl v0.6.1 is out w/ a few critical bugfixes, one of them is stability related: <https://github.com/ai-dynamo/nixl/releases/tag/0.6.1>,

[Slack Message](https://redhat-internal.slack.com/archives/C09HYSPUTKR/p1760017902463839)

diegocastanibm:
NIXL is now at v0.10.0+. Completely superseded by #848

PR #354, opened 2025-10-13
building UCX and NIXL for EFA unification

### LINKED ISSUE #332: [task] Align AWS and CUDA images
Because we are trying to get the v0.3.0 release our the door, we have a temporarily workaround to split our CUDA image that installs NIXL as part of UCX. The AWS image supporting EFA builds UCX and NIXL. At some point we should fix this and aggregate the images back into 1

Gregory-Pereira:
cc @anishasthana @wseaton @smarterclayton 

Resolves: #332 #346


docker/Dockerfile.cuda
```diff
@@ -395,14 +463,14 @@ ARG VLLM_USE_PRECOMPILED=1
...
-# Install all compiled wheels (DeepEP, DeepGEMM, pplx-kernels, LMCache, nixl)
+# Install all compiled wheels (DeepEP, DeepGEMM, pplx-kernels, LMCache)
 # Installs vllm source. Supports three install modes:
     # 1) install vllm from source totally editably - dev option, supports fork and commit swapping
     # 3) install vllm from source with pulling precomiled binaries (shared libraries) from the vllm wheels index - less flexible dev option, may or may not work swapping shas but faster than full editable
     # 2) install vllm from a wheel on the vllm wheels index - prod option, no flexbility
 RUN --mount=type=cache,target=/var/cache/git \
     source /opt/vllm/bin/activate && \
-    uv pip install nixl cuda-python 'huggingface_hub[hf_xet]' && \
+    uv pip install cuda-python 'huggingface_hub[hf_xet]' && \
```
wseaton: nit, can we collapse the pip installs to one line now that we are here? this prevents resolution bugs

guides/pd-disaggregation/ms-pd/values.yaml
```diff
@@ -50,6 +50,10 @@ decode:
       - "--max-model-len"
       - "32000"
     env:
+      - name: UCX_LOG_LEVEL
+        value: INFO
+      - name: UCX_TLS
+        value: "efa,sockcm,rc,sm,self,cuda_copy,cuda_ipc"
```
tlrmchlsmth: I think it's supposed to be `srd` instead of `efa`
https://github.com/openucx/ucx/blob/e16f38603aa11b31856190f613588038c2035ccb/src/ucp/core/ucp_context.c#L619-L640

PR #357, opened 2025-10-14
Use the vLLM image for xpu

yuanwu2017:


guides/inference-scheduling/README.xpu.md
```diff
@@ -0,0 +1,293 @@
+# Intel XPU Inference Scheduling Complete Deployment Guide
```
poussa: Adding a copy comment from #372:

I am not sure this deatiled README is needed. First, this contains a lot of duplication from the general guides documentation and is also a close copy of the README.xpu.md file. Not sure if even that is needed.

I'd like to stay out of detailing how to build images. Instead add CI/CD capability to build the image and reference that. Also, all the `git` and `minikube` sections are not needed.

I think the staring point (prerequisites) of the docs should be that you have an image and suitable cluster. This way we can keep the docs short and readable. I also think we guides have already documentation for that. Thus, I am challenging the usefulness of these README's.

docker/Dockerfile.xpu
```diff
@@ -1,187 +1,84 @@
 ARG ONEAPI_VERSION=2025.1.3-0
+ARG VLLM_VERSION=3cd36660f72f75b888c82a8feac93ea9f17c8e1e
```
Gregory-Pereira: Are we sure we want something ahead of [v0.11.1rc1](https://github.com/vllm-project/vllm/releases/tag/v0.11.1rc1)? I think from a project perspective were still staying on `v0.11.0` release because we are not prepared to pickup torch `2.9.0`. That being said the XPU image is independent so you can definetly chose to base off a commit ahead of the rest of the project, just wanted to call this out.

docker/Dockerfile.xpu
```diff
@@ -1,187 +1,84 @@
...
 
-WORKDIR /workspace
+ENV VLLM_TARGET_DEVICE=xpu
+ENV VLLM_WORKER_MULTIPROC_METHOD=spawn
 
-#RUN mkdir -p /wheels
-# Define commit SHAs as build args to avoid layer invalidation
-#ARG LMCACHE_COMMIT_SHA=c1563bc9c72ea0d71156a3d9a6cd643170828acf
+RUN --mount=type=cache,target=/root/.cache/pip \
+    python3 setup.py install
```
Gregory-Pereira: use `pip install .` as it follows PEP 517/518 (builds a wheel in an isolated env, resolves deps correctly, records metadata so uninstall works). `setup.py install` is deprecated: it bypasses dependency resolution, can skip wheel metadata, and is harder to uninstall/reproduce.

```suggestion
    pip install -v .
```

Gregory-Pereira:
Tested in: https://github.com/llm-d/llm-d/actions/runs/18619883902/job/53089610076?pr=310. Confirmed its working, NIXL and all.

guides/inference-scheduling/README.xpu.md
Gregory-Pereira: Yeah - not something we want to maintain long term. Because we did not have CI working for the v0.3.0 release we thought it would be helpful to have some more prescriptive docs, which is why Yuan drafted this up. It was a helpful stop-gap however we do wan't to deconstruct this doc, and add its relevant components to the rest of the documentation we have in the repos, so +1 please remove this file @yuanwu2017

docker/Dockerfile.xpu
yuanwu2017: Ok

guides/inference-scheduling/README.xpu.md
yuanwu2017: Done.

docker/Dockerfile.xpu
yuanwu2017: Done.

guides/inference-scheduling/README.md
```diff
@@ -18,12 +21,26 @@ This example out of the box requires 2 Nvidia GPUs of any kind (support determin
 - [Create the `llm-d-hf-token` secret in your target namespace with the key `HF_TOKEN` matching a valid HuggingFace token](../prereq/client-setup/README.md#huggingface-token) to pull models.
 - Have the [Monitoring stack](../../docs/monitoring/README.md) installed on your system.
 
+-  Intel GPU Plugin Required (For XPU deployments)
+
+    ```bash
+    # Deploy Intel GPU Device Plugin v0.32.1
+    kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin?ref=v0.32.1'
+
+    ```
```
Gregory-Pereira: I would create an XPU doc in the [accelerators directory](https://github.com/llm-d/llm-d/tree/main/docs/accelerators). That way this can be referenced from any of the examples we want to support XPU in, and factor it out of all guides.

guides/inference-scheduling/README.md
```diff
@@ -18,12 +21,26 @@ This example out of the box requires 2 Nvidia GPUs of any kind (support determin
...
+
+**_NOTE:_**
+**Kubernetes v1.33.0+** is recommended for complete sidecar init container support (restartPolicy: Always). If using Kubernetes v1.28.x or below, pods may get stuck in Init:0/1 state due to incomplete sidecar support.
```
Gregory-Pereira: Would this info be better as an update to: https://github.com/llm-d/llm-d/blob/main/guides/prereq/infrastructure/README.md#llm-d-infrastructure because it would apply to all the examples and cloud providers

yuanwu2017: ok

guides/inference-scheduling/README.md
Gregory-Pereira: Question here - I see you marked this as resolved but I dont't see this added to a document for XPU configuration or any common overlay. Are you asking to defer this until a later date or rejecting the suggestion?

yuanwu2017: Ok. let me modify it.

yuanwu2017: Done.

Gregory-Pereira:
OK, It looks good, lets just rebase in the changes for #310 and we should be good to go!

PR #367, opened 2025-10-16
Add wide ep gke test

rlakhtakia:


.github/workflows/e2e-wide-ep-accelerator-gke.yaml
```diff
@@ -0,0 +1,260 @@
...
+          kubectl apply -k ./manifests/gateway/gke-l7-regional-external-managed -n ${NAMESPACE} | tee ~/wide-ep-deployment.log
+          echo "---------------------------------------" >> ~/wide-ep-deployment.log
+
+      - name: Wait for all pods to be ready
+        run: |
+          kubectl wait pod \
+            --for=condition=Ready \
+            --all \
+            -n "${NAMESPACE}" \
+            --timeout=90m
```
Gregory-Pereira: 90 minutes seems pretty excessive, I have never seen a wide-ep deployment take more than 20 minutes. Maybe lets start with 25?

```suggestion
            --timeout=25m
```

.github/workflows/e2e-wide-ep-accelerator-gke.yaml
```diff
@@ -0,0 +1,260 @@
...
+          kubectl get pods -n "${NAMESPACE}"
+
+      - name: Debug gateway
+        run: |
+          kubectl -n ${NAMESPACE} get gateway
+          kubectl -n ${NAMESPACE} describe gateway
+          kubectl -n ${NAMESPACE} get gateway -o yaml
+          kubectl -n ${NAMESPACE} describe httpRoute
+          kubectl -n ${NAMESPACE} get httpRoute -o yaml
+          kubectl -n ${NAMESPACE} get inferencepool -o yaml
```
Gregory-Pereira: Do we still need this? If so lets move this to part of the upload logs
```suggestion
```

Gregory-Pereira (commented):
/lgtm just 2 comments

.github/workflows/e2e-wide-ep-accelerator-gke.yaml
rlakhtakia: I think we can just remove it since we are logging the status two steps later.

.github/workflows/e2e-wide-ep-accelerator-gke.yaml
rlakhtakia: When I was testing, it was taking closer to 30 minutes. So I updated it to 35.

PR #371, opened 2025-10-16
Arm cuda support

wseaton:
Stacked PR on bash script refactor


.github/workflows/build-image.yml
```diff
@@ -39,29 +71,48 @@ jobs:
...
     strategy:
       fail-fast: false
-    runs-on: vllm-runner
+      matrix:
+        include:
+          - platform: linux/amd64
+            runner: vllm-runner
+          - platform: linux/arm64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
```
Gregory-Pereira: Can't we just use the docker `--platform linux/arm64` to do cross builds? Maybe your storage saving implementation is better than mine, but I remember really struggling for all the space on default runners even after removal

wseaton: I tried it, it does work, but the build is several times slower because of x86 using qemu for emulation. The default runners are actually working fine for now with the free space action, longer term we should try to set up a self-hosted aarch64 runner.

PR #372, opened 2025-10-16
add intel HPU for llm-d P/D disaggregation

ZhengHongming888:
This PR is to add the Intel's HPU support for llm-d P/D framework. 

There are three main modification here -

1. ms-pd/values_hpu.yaml 
    - add the HPU support for llm-d on Prefill/Decode by using the Intel HPU definition (intel_gaudi)
    - intel_gaudi (accelerator type) is defined in another [PR134](https://github.com/llm-d-incubation/llm-d-modelservice/pull/134) in llm-d-modelservice/charts/llm-d-modelservice/values.yaml and llm-d-modelservice/charts/llm-d-modelservice/values.schema.json files.
    - need merge PR134 first and then can run this PR. 

2. helmfile.yaml.gotmpl
    - include the option of ms-pd/values_hpu.yaml for modelservice part.
    -
3. README.hpu.md
   - very detailed guide for how to run the llm-d P/D disaggregation based on Intel's HPU.
   - Step-by-Step instructions

Please check and review it. Welcome for your any comments.

Thanks.

PR #382, opened 2025-10-21
Arm cuda support (from clean branch)

wseaton:


.github/workflows/build-image.yml
```diff
@@ -50,7 +50,6 @@ concurrency:
 
 jobs:
   pre-commit:
-    if: github.event_name == 'pull_request'
```
Gregory-Pereira, line 53: Why are we removing this? If this is a workflow dispatch we don't care about pre-commit do we?

.github/workflows/build-image.yml
```diff
@@ -95,6 +94,7 @@ jobs:
             cuda:
               - 'docker/Dockerfile.cuda'
               - 'docker/scripts/**'
+              - 'docker/scripts/cuda/**/*.json'
```
Gregory-Pereira: Maybe this would be better served as a `docker/packages/` directory? Im thinking intel might re-use some between XPU and HPU coming in [this pr](https://github.com/llm-d/llm-d/pull/372)

.github/workflows/build-image.yml
```diff
@@ -121,19 +121,47 @@ jobs:
     if: ${{ needs.detect-changes.outputs.cuda == 'true' }}
     strategy:
       fail-fast: false
+      # multi-OS build matrix: creates 4 images (2 platforms × 2 OS variants)
+      # - RHEL (ubi9): default/production image (no suffix in final tags)
+      # - Ubuntu (24.04): alternative base for compatibility testing (-ubuntu suffix)
```
Gregory-Pereira, line 128: Will there be any interaction here with upstream vLLM? Don't they specifically build against Ubuntu 20.04 and 22.04?

.github/workflows/build-image.yml
```diff
@@ -153,6 +181,12 @@ jobs:
             type=sha
             type=raw,value=latest,enable={{is_default_branch}}
 
+      - name: Set platform variables
+        id: platform
+        run: |
+          ARCH=$(echo "${{ matrix.platform }}" | cut -d/ -f2)
+          echo "arch=${ARCH}" >> $GITHUB_OUTPUT
+
```
Gregory-Pereira: Again can this be rolled up into the matrix, seems wasteful to have a step for it

.github/workflows/build-image.yml
```diff
@@ -162,16 +196,22 @@ jobs:
...
+            BASE_IMAGE_SUFFIX=${{ matrix.base_image_suffix }}
             VLLM_REPO=${{ inputs.vllm_repo || 'https://github.com/vllm-project/vllm.git' }}
             VLLM_COMMIT_SHA=${{ inputs.vllm_commit_sha || 'b8b302cde434df8c9289a2b465406b47ebab1c2d' }}
             VLLM_PRECOMPILED_WHEEL_COMMIT=${{ inputs.vllm_precompiled_wheel_commit || inputs.vllm_commit_sha || '' }}
+          cache-from: >-
+            ${{ matrix.platform == 'linux/arm64' &&
+            format('type=gha,scope=cuda-{0}-arm64', matrix.os) || '' }}
+          cache-to: >-
+            ${{ matrix.platform == 'linux/arm64' &&
+            format('type=gha,mode=max,scope=cuda-{0}-arm64', matrix.os) || '' }}
```
Gregory-Pereira: I am confused on this, will it just cache for arm64 builds? Do we not want caching for amd64?

.github/workflows/build-image.yml
```diff
@@ -230,22 +290,31 @@ jobs:
...
+          done
+
+          # create manifest and tag it
           docker buildx imagetools create \
-            $(echo "${{ steps.meta.outputs.tags }}" | tr '\n' ' ' | sed 's/ */ /g' | \
-              awk '{for(i=1;i<=NF;i++) printf("-t %s ", $i)}') \
-            $(printf '${{ env.REGISTRY }}/${{ github.repository }}-cuda-dev@sha256:%s ' *)
+            -t ${{ env.REGISTRY }}/${{ github.repository }}-cuda-dev${{ matrix.image_suffix }}:pr-${{
+              github.event.pull_request.number }} \
+            -t ${{ env.REGISTRY }}/${{ github.repository }}-cuda-dev${{ matrix.image_suffix }}:sha-${GITHUB_SHA::7} \
```
Gregory-Pereira: This is tricky - I think the logic we use for the shas are: `$(git rev-parse --short HEAD)` which will produce the smallest unique sha. Sometimes it could be 5 characters, sometimes 12, etc. I am murky on the details but im not certain this will be reliable

.github/workflows/ci-pr-checks.yaml
```diff
@@ -1,6 +1,9 @@
 name: CI - PR Checks
 
 on:
+  push:
+    branches:
+      - main
```
Gregory-Pereira: Remove, we discussed this in the last PR

docker/scripts/cuda/builder/build-compiled-wheels.sh
```diff
@@ -65,12 +56,12 @@ uv build --wheel --no-build-isolation --out-dir /wheels
...
+# build pplx-kernels wheel (skip on ARM64)
+if [ "${TARGETPLATFORM}" != "linux/arm64" ]; then
     git clone "${PPLX_KERNELS_REPO}" pplx-kernels
     cd pplx-kernels
     git checkout "${PPLX_KERNELS_SHA}"
-    NVSHMEM_PREFIX="${NVSHMEM_DIR}" uv build --wheel --out-dir /wheels
+    TORCH_CUDA_ARCH_LIST="9.0a;10.0+PTX" NVSHMEM_PREFIX="${NVSHMEM_DIR}" uv build --wheel --out-dir /wheels
     cd ..
     rm -rf pplx-kernels
 else
```
Gregory-Pereira, line 67: I remember we talked about this but could you remind me what the blocker is for PPLX on ARM>

docker/packages/common/builder-packages.json
Gregory-Pereira, line 1: Is a mapping the best way to do this? Lets start with it but I was thinking an object with 2 arrays, one for each. Alternatively maybe we make them nested objects and track dependencies per component (NIXL, GDR Copy, etc.)?

docker/Dockerfile.aws
```diff
@@ -245,7 +245,7 @@ WORKDIR /workspace
 
 # Define commit SHAs as build args to avoid layer invalidation
 ARG LMCACHE_COMMIT_SHA="0db8ae4746a207a72483d716b5f82545a2ead24b"
-ARG VLLM_COMMIT_SHA="b8b302cde434df8c9289a2b465406b47ebab1c2d"
+ARG VLLM_COMMIT_SHA="f71952c1c49fb86686b0b300b727b26282362bf4"
```
Gregory-Pereira: Revert this, why are we going back to the old v0.11.0 tag commit BEFORE the pipeline fix for Blackwell? `b8b302cde434df8c9289a2b465406b47ebab1c2d` is the current v0.11.0 tag commit
```suggestion
ARG VLLM_COMMIT_SHA="b8b302cde434df8c9289a2b465406b47ebab1c2d"
```

docker/Dockerfile.aws
```diff
@@ -399,7 +399,7 @@ COPY --from=builder /wheels/*.whl /tmp/wheels/
...
```
Gregory-Pereira: Same as above
```suggestion
ARG VLLM_COMMIT_SHA="b8b302cde434df8c9289a2b465406b47ebab1c2d"
```

docker/Dockerfile.cuda
```diff
@@ -42,14 +35,23 @@ ENV LANG=C.UTF-8 \
...
-RUN dnf -q update -y && dnf clean all
-
 # Install base packages and EPEL in single layer
+COPY docker/scripts/cuda/common/package-utils.sh /tmp/package-utils.sh
+COPY docker/scripts/cuda/builder/builder-package-mappings.json /tmp/builder-package-mappings.json
 COPY docker/scripts/cuda/builder/install-base-packages.sh /tmp/install-base-packages.sh
-RUN chmod +x /tmp/install-base-packages.sh && \
-    /tmp/install-base-packages.sh && \
-    rm /tmp/install-base-packages.sh
+RUN if [ "${TARGETOS:-rhel}" = "ubuntu" ]; then apt-get update -qq && apt-get install -y jq; else dnf -q install -y jq; fi && \
```
Gregory-Pereira: I think the installs for the `jq` should be moved to `install-base-packages.sh` as it is a base package. Also in the case of RHEL we still need to call `dnf update -y`?

docker/Dockerfile.cuda
```diff
@@ -59,50 +61,50 @@ ENV LIBRARY_PATH="/usr/local/cuda/lib64:/usr/local/lib:/usr/local/lib64" \
...
-    PATH="/opt/nvshmem-${NVSHMEM_VERSION}/bin:${PATH}" \
-    CPATH="/opt/nvshmem-${NVSHMEM_VERSION}/include:${CPATH}" \
-    LIBRARY_PATH="/opt/nvshmem-${NVSHMEM_VERSION}/lib:${LIBRARY_PATH}"
+# NVSHMEM installed from nvidia repos via libnvshmem3-cuda-${CUDA_MAJOR}-3.4.5-1
+# Libraries on RHEL: /usr/lib64/nvshmem/${CUDA_MAJOR}/, on Ubuntu: /usr/lib/{arch}-linux-gnu/nvshmem/${CUDA_MAJOR}/
+# Headers: /usr/include/nvshmem_${CUDA_MAJOR}/
+ENV NVSHMEM_DIR="/usr/bin/nvshmem_${CUDA_MAJOR}" \
+    PATH="/usr/bin/nvshmem_${CUDA_MAJOR}:${PATH}" \
+    CPATH="/usr/include/nvshmem_${CUDA_MAJOR}:${CPATH}" \
+    LIBRARY_PATH="/usr/lib64/nvshmem/${CUDA_MAJOR}:/usr/lib/x86_64-linux-gnu/nvshmem/${CUDA_MAJOR}:/usr/lib/aarch64-linux-gnu/nvshmem/${CUDA_MAJOR}:${LIBRARY_PATH}"
```
Gregory-Pereira, line 84: Is there some more graceful way we can handle this? Those are Ubuntu only paths, unless I am mistaken. It wouldn't be the end of the world if they sneak into the UBI variant of the image but not ideal.

docker/Dockerfile.cuda
```diff
@@ -181,41 +184,55 @@ ENV LANG=C.UTF-8 \
     UV_TORCH_BACKEND=${UV_TORCH_BACKEND:-cu${CUDA_MAJOR}${CUDA_MINOR}} \
     UV_CONSTRAINT=/tmp/constraints.txt \
     VIRTUAL_ENV=/opt/vllm \
-    NVSHMEM_DIR="/opt/nvshmem-${NVSHMEM_VERSION}" \
+    NVSHMEM_DIR="/usr/bin/nvshmem_${CUDA_MAJOR}" \
     # LD_LIBRARY_PATH needs the torch path to apply proper linkers so as not to produce torch ABI missmatch
-    LD_LIBRARY_PATH="/opt/vllm/lib64/python${PYTHON_VERSION}/site-packages/torch/lib:/usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/local/cuda/lib64:/usr/local/lib:/usr/local/lib64:/opt/nvshmem-${NVSHMEM_VERSION}/lib:${LD_LIBRARY_PATH}" \
-    PATH="/opt/nvshmem-${NVSHMEM_VERSION}/bin:${PATH}" \
-    CPATH="/opt/nvshmem-${NVSHMEM_VERSION}/include:${CPATH}" \
+    LD_LIBRARY_PATH="/opt/vllm/lib64/python${PYTHON_VERSION}/site-packages/torch/lib:/usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/local/cuda/lib64:/usr/local/lib:/usr/local/lib64:/usr/lib64/nvshmem/${CUDA_MAJOR}:/usr/lib/x86_64-linux-gnu/nvshmem/${CUDA_MAJOR}:/usr/lib/aarch64-linux-gnu/nvshmem/${CUDA_MAJOR}:${LD_LIBRARY_PATH}" \
```
Gregory-Pereira, line 203: Same as before - not a blocker but ideally we don't carry both sets of paths in LD_LIBRARY_PATH

Gregory-Pereira (commented):
Some comments but mostly questions

.github/workflows/build-image.yml
wseaton: You are right, I see this comment: https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile#L18-L24. We should consider emulating this approach to get an older version of glibc

.github/workflows/build-image.yml
wseaton: Agree, I had this change somewhere but it got reverted. We should be using matrix.include

docker/Dockerfile.aws
wseaton: We shouldn't be, bad rebase.

docker/Dockerfile.cuda
wseaton: I'll look into this, the linker should fall back gracefully but you are right, I don't like setting it this way

.github/workflows/build-image.yml
wseaton: ack, let's not truncate this then

docker/Dockerfile.cuda
Gregory-Pereira: If it falls back gracefully we can consider it tech debt, just put a comment in there with a TODO and we can move on for now. I want to unblock other builds ASAP, going to have to get dirty with this for #354 rebase

docker/Dockerfile.cuda
Gregory-Pereira: Based on the discussion above just put a comment about fix this later and ignore my point

docker/Dockerfile.cuda
wseaton: done

.github/workflows/build-image.yml
wseaton: we implemented this, resolving

.github/workflows/build-image.yml
wseaton: Added a solution for common packages in [d270c18](https://github.com/llm-d/llm-d/pull/382/commits/d270c184e70bf4351a1b58531445063f5b280b3d), PTAL

.github/workflows/build-image.yml
wseaton: This was fixed

.github/workflows/build-image.yml
```diff
@@ -121,19 +123,47 @@ jobs:
...
+        if: matrix.platform == 'linux/arm64'
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: true
+          docker-images: false
+          swap-storage: true
```
Gregory-Pereira: Do we need this? Now we are using a managed runner for ARM builds we shouldn't need to free space and this will slow build times down

.github/workflows/build-image.yml
```diff
@@ -162,16 +192,20 @@ jobs:
...
           build-args: |
+            TARGETOS=${{ matrix.os }}
+            BUILD_BASE_IMAGE_SUFFIX=${{ matrix.os == 'ubuntu' && 'ubuntu20.04' || matrix.base_image_suffix }}
+            FINAL_BASE_IMAGE_SUFFIX=${{ matrix.base_image_suffix }}
             VLLM_REPO=${{ inputs.vllm_repo || 'https://github.com/vllm-project/vllm.git' }}
             VLLM_COMMIT_SHA=${{ inputs.vllm_commit_sha || 'b8b302cde434df8c9289a2b465406b47ebab1c2d' }}
             VLLM_PRECOMPILED_WHEEL_COMMIT=${{ inputs.vllm_precompiled_wheel_commit || inputs.vllm_commit_sha || '' }}
+          cache-from: type=gha,scope=cuda-${{ matrix.os }}-${{ matrix.platform == 'linux/arm64' && 'arm64' || 'amd64' }}
+          cache-to: >-
+            type=gha,mode=max,scope=cuda-${{ matrix.os }}-${{ matrix.platform == 'linux/arm64' && 'arm64' || 'amd64' }}
```
Gregory-Pereira: These should include the dev tag, as this workflow is for non-release builds:

```suggestion
          cache-from: type=gha,scope=cuda-dev-${{ matrix.os }}-${{ matrix.platform == 'linux/arm64' && 'arm64' || 'amd64' }}
          cache-to: >-
            type=gha,mode=max,scope=cuda-dev-${{ matrix.os }}-${{ matrix.platform == 'linux/arm64' && 'arm64' || 'amd64' }}
```

.github/workflows/build-image.yml
```diff
@@ -189,22 +223,42 @@ jobs:
       - name: Upload digest
         uses: actions/upload-artifact@v4
         with:
-          name: digests-cuda-${{ matrix.arch }}
+          name: digests-cuda-${{ matrix.os }}-${{ matrix.platform == 'linux/arm64' && 'arm64' || 'amd64' }}
```
Gregory-Pereira: dev tag:

```suggestion
          name: digests-cuda-dev-${{ matrix.os }}-${{ matrix.platform == 'linux/arm64' && 'arm64' || 'amd64' }}
```

.github/workflows/build-image.yml
```diff
@@ -189,22 +223,42 @@ jobs:
...
+          # Ubuntu images are tagged with -ubuntu suffix
+          - os: ubuntu
+            image_suffix: "-ubuntu"
     steps:
       - name: Download digests
         uses: actions/download-artifact@v4
         with:
           path: /tmp/digests
-          pattern: digests-cuda-*
+          pattern: digests-cuda-${{ matrix.os }}-*
```
Gregory-Pereira: dev tag:

```suggestion
          pattern: digests-cuda-dev-${{ matrix.os }}-*
```

Gregory-Pereira (commented):
Some more comments and questions. Also noticing that we didn't update the release pipeline as well. I don't want to keep adding scope to this PR because its already big enough but we should be sure to follow up with that

.github/workflows/build-image.yml
wseaton: yep we can remove now!

.github/workflows/build-image.yml
```diff
@@ -162,16 +180,23 @@ jobs:
...
             VLLM_PRECOMPILED_WHEEL_COMMIT=${{ inputs.vllm_precompiled_wheel_commit || inputs.vllm_commit_sha || '' }}
+          cache-from: >-
+            type=gha,scope=cuda-dev-${{ matrix.os }}-${{
+            matrix.platform == 'linux/arm64' && 'arm64' || 'amd64' }}
+          cache-to: >-
+            type=gha,mode=max,scope=cuda-dev-${{ matrix.os }}-${{
+            matrix.platform == 'linux/arm64' && 'arm64' || 'amd64' }}
+          labels: ${{ steps.meta.outputs.labels }}
+          outputs: >-
+            type=image,name=${{ env.REGISTRY }}/${{ github.repository }}-cuda-dev,
```
Gregory-Pereira: This needs to be scoped with the image suffix as well:

```suggestion
            type=image,name=${{ env.REGISTRY }}/${{ github.repository }}-cuda-dev${{matrix.base_image_suffix}},
```

Gregory-Pereira (approved):
Tested every for both Ubuntu and and UBI builds:
- [x] inference scheduling
- [x] inference scheduling with DP and EP, an MOE Qwen, and PPLX backend
- [x] precise-prefix-cache-aware
- [x] pd-dissagregation
    - [x] benchmark with both guide-llm and lm-eval comparing against `ghcr.io/llm-d/llm-d-cuda:v0.3.0` 
- [x] wide-ep-lws
    - [x] benchmark with both guide-llm and lm-eval comparing against `ghcr.io/llm-d/llm-d-cuda:v0.3.0`

I just don't have access to GB200 or another compatible ARM GPU to test against. That being said, this PR fixes builds and we don't currently have ARM support anyway so I think we should merge this regardless

PR #318, opened 2025-10-05
Add cpu offloading example for GKE + LMCache

dannawang0221:
Add cpu offloading example for GKE + LMCache

PR #401, opened 2025-10-29
Add umbrella kv cache offloading well-lit path folder structure

liu-cong:
This PR creates an umbrella folder structure for kv offloading well-lit path in
[proposal](https://docs.google.com/document/d/1HhQiI1FunB_W3WFudQUFkLZNm4Bjsw2HzQIKhGMpV7g/edit?resourcekey=0-Opmz8zL6jcZCZYcmhYotHA&tab=t.5pndxbtjcnd0).

* As we expect multiple kv cache offloading guides, this PR sets an umbrella `kv-cache-offloading` well-lit path folder, and a pattern to create sub-guides.
* Discusses general guidance for CPU offloading and remote storage offloading, as those are the offloading options we consider in 0.4 scope.
* Added placeholder folders for the CPU offloading and Lustre backend that we already planned. People are welcome to add more guides as a follow up.

Once this is accepted, the first [PR](https://github.com/llm-d/llm-d/pull/318/) of adding LMCache based CPU offloading should be moved under the lmcache-connector folder. 

cc @vMaroon @dannawang0221 @saikat-royc @Gregory-Pereira @smarterclayton @ahg-g


guides/kv-cache-offloading/cpu/README.md
```diff
@@ -0,0 +1,37 @@
+# Offloading KV Cache to CPU Memory
```
smarterclayton: Other alternative options (I'm not set against the current one, but want to discuss)

1. Use `prefix-cache` instead of `kv-cache`, unless we have a strong reason (i.e. to be consistent with `prefix cache aware routing` and `precise prefix cache aware routing`)
  a. what future developments may make `prefix cache` inaccurate?
3. Is this about offloading only?  Is a local ssd or peer to peer cache "offloaded"?  a. `remote` doesn't capture peer to peer
  b. `disaggregated` is used in some places today but is long and may be confused with disaggregated serving
3. This creates a kv cache hierarchy (or is the start of it), with many different implementation options
4. I prefer to stay away from `management` or `manager` because those are too generic.

Other wordings that might be relevant?

guides/kv-cache-offloading/README.md
```diff
@@ -0,0 +1,31 @@
+# Well-lit Path: KV Cache Offloading
+
+## Overview
+
+Efficient caching of attention Key & Value (KV) tensors is crucial for boosting Large Language Model (LLM) inference performance such as Time to First Token (TTFT) and overall throughput, as well as reducing the cost. Caching in the accelerator High-Bandwidth Memory (HBM) is a free lunch that state of the art inference engines already take. Yet more tokens and memory in GenAI applications demands more KV caching, driving the need for offloading KV cache from HBM to more storage options. This well-lit path contains multiple sub-guides per the KV storage type, with high level guidance on when to use them.
```
smarterclayton: > ... as well as reducing cost to serve.

Suggest changing from `is a free lunch` to something a bit more formal - perhaps:

> State of the art inference engines already implement native KV cache reuse across requests in accelerator High-Bandwidth Memory (HBM), but in most serving environments HBM is already a constrained resource.  To increase the amount of available memory beyond HBM requires ...

In the opening, I think we should explicitly talk about a cache hierarchy and tiering (since multiple others already implement it) - maybe as a new paragraph

> This well-lit path describes the tiers of a KV cache hierarchy, provides high level guidance on their suitability per workload, and makes recommendations about selecting and configuring a KV cache offloading implementation.

I would also suggest that if we are using `KV Cache Offloading` as our term, we also specifically have a sentence in the first paragraph connecting to a prefix cache to align the two, and then scrub the other guides to make sure they consistently link back to this guide.

smarterclayton:
Good starting point, some structural comments.

guides/kv-cache-offloading/README.md
liu-cong: Tweaked the tiering wording slightly to not feel like it's a must do (although it's probably the most reasonable setup).

guides/kv-cache-offloading/cpu/README.md
liu-cong: For 1, changed it to `prefix-cache`. This is more generic as the SSM models don't have `kv-cache` but a different cache format. `prefix-cache` is a more general term.  
For 2, good point, P2P can be seen as a "Cache Sharing" rather than "offloading". I changed it to `prefix-cache-storage`, this is best I can think of... Another options is we break down by `offloading` vs `sharing`, but that it feels like just another sharding dimension that doesn't directly translate to user value.

guides/prefix-cache-storage/README.md
```diff
@@ -0,0 +1,54 @@
...
+* There are usually more CPU RAM storage available than accelerator HBM on the host offering much larger cache capacity.
+* CPU - accelerator transfer is faster than recomputation for most cases.
+* (WIP) Prefix cache storage tier aware inference scheduling makes smart decisions based on cache tier (accelerator HBM vs. CPU RAM).
+
+In low cache size scenario where HBM is primarily used, async CPU offloading should incur little overhead. In high cache size scenario loading cache from CPU RAM offers significantly higher cache hit and thus better performance than HBM only.
+
+### Local Disk
+
+Utilizing local disk storage can significantly increase the cache capacity. However disks are typically significantly slower than CPU RAM. Consider this when your workload can tolerate the latency overhead, and the cache capacity of local disks is sufficient for your use case. Otherwise we recommend a shared storage because it:
+
```
vMaroon, line 34: Think it would be more consistent with the rest of the doc if the points below this line are transformed to standalone bullets.

guides/kv-cache-offloading/cpu/README.md
vMaroon: P2P offloading could be an interesting concept, technically there would be nothing stopping us from using it that way when the implementation progresses. On the latter, do we want to expand some more on the plan in this doc or leave it completely for coming releases?

We intend to evolve the NIXL connector to a P2P NIXL based one, starting with the GPU block-pool and the newly introduced but not yet abstracted CPU block-pool as backends.

liu-cong: I'd like to get some good insights such as benchmark numbers to give some high level recommendation instead of just adding an option. @vMaroon Feel free to add more details there on P2P (I don't have currently).

liu-cong:
> Could you please clarify the purpose of the guides in this PR? Is it meant to be a tutorial on setting up vLLM with different offloading connectors, or is it focused on configuring llm-d and describing which connectors are supported?

This purpose of this is to set up the high level guidance on choosing the tiered cache structure (it needs more refinement as we gather more insights), and a folder structure so as follow ups we can add the detailed user guide. This PR itself doesn't not add any concrete user guide.

PR #402, opened 2025-10-29
Add the XPU precise-prefix-cache-aware example

yuanwu2017:
Add XPU configuration for precise-prefix-cache-aware inference with support for Istio and kgateway gateway providers.
Test result:

image

### DIFF
--- a/guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
+++ b/guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
@@ -29,7 +29,7 @@ decode:
   containers:
   - name: "vllm"
     # Use Intel latest published xpu image
-    image: ghcr.io/llm-d/llm-d-xpu-dev:pr-380
+    image: ghcr.io/llm-d/llm-d-xpu-dev:pr-592
     modelCommand: custom
     command:
       - /bin/sh
@@ -88,24 +88,24 @@ decode:
       httpGet:
         path: /v1/models
         port: vllm
-      initialDelaySeconds: 15
+      initialDelaySeconds: 30
       periodSeconds: 30
-      timeoutSeconds: 5
+      timeoutSeconds: 10
       failureThreshold: 60
     livenessProbe:
       httpGet:
         path: /health
         port: vllm
-      periodSeconds: 10
-      timeoutSeconds: 5
-      failureThreshold: 3
+      periodSeconds: 30
+      timeoutSeconds: 10
+      failureThreshold: 20
     readinessProbe:
       httpGet:
         path: /v1/models
         port: vllm
-      periodSeconds: 5
-      timeoutSeconds: 2
-      failureThreshold: 3
+      periodSeconds: 10
+      timeoutSeconds: 5
+      failureThreshold: 6
   volumes:
   - name: metrics-volume
     emptyDir: {}


guides/precise-prefix-cache-aware/README.md
```diff
@@ -37,6 +37,18 @@ To see what gateway options are supported refer to our [gateway provider prereq
...
+#### Intel XPU deployment
+
+```bash
+helmfile apply -e xpu -n ${NAMESPACE} # targets istio as gateway provider with Intel XPU hardware
+```
+
+You can also combine Intel XPU hardware with different gateway providers:
+
+```bash
+helmfile apply -e xpu-kgateway -n ${NAMESPACE} # targets kgateway as gateway provider with Intel XPU hardware
```
poussa, line 58: We don't have such environment (xpu-kgateway), do we?

yuanwu2017: I tested it. It can work.

poussa: > I tested it. It can work.

ah, sorry. Indeed the env is there. Sorry for the noise.

guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
```diff
@@ -0,0 +1,145 @@
...
+      - containerPort: 5557
+        protocol: TCP
+      - containerPort: 8200
+        name: metrics
+        protocol: TCP
+    resources:
+      limits:
+        memory: 24Gi
+        cpu: "8"
+        gpu.intel.com/i915: "1"
```
poussa: You don't need to specify the GPU resources here anymore. It is populated automatically based on the accelerator type. That is, if you have "accelerator.type: intel", resource name will be  gpu.intel.com/i915. 

Also, the latest modelservice Helm charts have intel-i915 and intel-xe types which would simplify this and the README section.

Finally, with DRA this would be super simple (since it handles both xe and i915 with single config). Maybe for the next release.

yuanwu2017: Thanks. I will simplify this section after the DRA driver is enabled in XPU CI.

guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
```diff
@@ -0,0 +1,128 @@
...
+      - |
+        set -x
+        exec vllm serve Qwen/Qwen3-0.6B \
+          --port 8000 \
+          --served-model-name "Qwen/Qwen3-0.6B" \
+          --dtype float16 \
+          --disable-sliding-window \
+          --block-size 64 \
+          --prefix-caching-hash-algo sha256_cbor \
+          --kv-transfer-config '{"kv_connector":"NixlConnector", "kv_role":"kv_both", "kv_buffer_device":"cpu"}' \
```
Gregory-Pereira: @yuanwu2017 do we need this? Theres no PD dissagregation in this example so no NIXL transfers that happen, just vLLM publishing the results of its KVEvents API in a way the scheduler can see. 

Going to remove it for now and let me know if I am wrong and we can add it back

yuanwu2017: You are right. We don't need the nixl config. Thanks.

guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
```diff
@@ -0,0 +1,119 @@
...
+
+    ports:
+      - containerPort: 8000
+        name: vllm
+        protocol: TCP
+    resources:
+      limits:
+        memory: 24Gi
+        cpu: "8"
+        gpu.intel.com/i915: "1"
```
poussa: we don't need this line (gpu.intel.com...), it is automatically populated by the model-service helm chart.

guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
```diff
@@ -0,0 +1,119 @@
...
+      requests:
+        cpu: "4"
+        memory: 12Gi
+        gpu.intel.com/i915: "1"
```
poussa: ditto

guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
```diff
@@ -0,0 +1,117 @@
+# Intel XPU configuration for precise prefix cache aware routing with KV-Events
+# This configuration combines XPU-specific settings with KV events for precise cache awareness
+
+multinode: false
+
+modelArtifacts:
+  name: Qwen/Qwen3-0.6B
+  uri: "hf://Qwen/Qwen3-0.6B"
```
diegocastanibm, line 8: @Gregory-Pereira I think we want `Qwen3-32B`, right?

guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
```diff
@@ -0,0 +1,117 @@
...
+    # Use Intel latest published xpu image
+    image: ghcr.io/llm-d/llm-d-xpu-dev:pr-592
+    modelCommand: custom
+    command:
+      - /bin/sh
+      - '-c'
+    args:
+      - |
+        set -x
+        exec vllm serve Qwen/Qwen3-0.6B \
```
diegocastanibm, line 40: `Qwen3-32B`

guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
```diff
@@ -0,0 +1,117 @@
...
+          --port 8000 \
+          --served-model-name "Qwen/Qwen3-0.6B" \
```
diegocastanibm, line 42: `Qwen3-32B`

guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
Gregory-Pereira: This is a good point I forgot to raise with the intel folks. @yuanwu2017 @poussa were planning on upping all of our deployments to be "production" so 16+ gpus and minimum model sizing of Qwen/Qwen3-32B. 

@diegocastanibm were going to make an exception for now. This is because the XPU CI system we have would not support a deployment of that size to my knowledge. We can give XPU, TPU and others a pass for this release.

guides/precise-prefix-cache-aware/ms-kv-events/values_xpu.yaml
Gregory-Pereira: Were making an exception, see above.

PR #391, opened 2025-10-26
EFA support rebased

Gregory-Pereira:
Rebased on top of (depends on): #382 
Consolidates work from: #384 #354 

Changes:

- Rename NVSHMEM_DIR to NVSHMEM_PREFIX to be consistent with how CMAKE install path conventions
- Pulling in double build stage to populate the cache for the different phases
- Support for Entitlement in UBI Cuda builder image
- Installing EFA, Building UCX against it, building nvshmem against it
- Extending NVSHMEM build script to support source from github or developer source download
- Adding `NVSHMEM_UCX_SUPPORT=1` as per: https://redhat-internal.slack.com/archives/C09HYSPUTKR/p1760401520887149?thread_ts=1760399516.357069&cid=C09HYSPUTKR

cc @wseaton @andjsmi

PR #407, opened 2025-10-30
Patch nvshmem to avoid an uninitialized value passed to RoCE

smarterclayton:
NVSHMEM does not clear a stack allocated struct `ibv_ah_attr` which is passed directly to the kernel RDMA functions. On kernels with a patch to validate the value of static_rate the create_ah() function will return EINVAL (sometimes).

When built in environments that automatically zero stack values, the library will always initialize the device to the default rate. Otherwise, it may result in the error `Unable to create ah.` especially on RoCE devices.

### DIFF
--- a/docker/packages/cuda/builder-packages.json
+++ b/docker/packages/cuda/builder-packages.json
@@ -1,7 +1,5 @@
 {
   "rhel_to_ubuntu": {
-    "libnvshmem3-cuda-${CUDA_MAJOR}-3.4.5-1": "libnvshmem3-cuda-${CUDA_MAJOR}=3.4.5-1",
-    "libnvshmem3-devel-cuda-${CUDA_MAJOR}-3.4.5-1": "libnvshmem3-dev-cuda-${CUDA_MAJOR}=3.4.5-1",
-    "libnvshmem3-static-cuda-${CUDA_MAJOR}-3.4.5-1": "libnvshmem3-static-cuda-${CUDA_MAJOR}=3.4.5-1"
+    "python3.9-devel": null
   }
 }
--- a/docker/scripts/cuda/builder/build-nvshmem.sh
+++ b/docker/scripts/cuda/builder/build-nvshmem.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -Eeu
+set -Eeux
 
 # builds and installs NVSHMEM from source with coreweave patch
 #
@@ -23,7 +23,13 @@ tar -xf "nvshmem_src_cuda${CUDA_MAJOR}.tar.gz"
 
 cd nvshmem_src
 
-git apply /tmp/patches/cks_nvshmem"${NVSHMEM_VERSION}".patch
+for i in /tmp/patches/cks_nvshmem"${NVSHMEM_VERSION}".patch /tmp/patches/nvshmem_zero_ibv_ah_attr_"${NVSHMEM_VERSION}".patch; do
+    if [[ -f $i ]]; then
+        git apply $i
+    else
+        echo "Unable to find patch matching nvshmem version ${NVSHMEM_VERSION}: $i"
+    fi
+done
 
 mkdir build
 cd build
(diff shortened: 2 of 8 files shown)


Gregory-Pereira:
@smarterclayton you can pull some pieces out of #391, Ive got a script and changes for the Cmake issue you are hitting, happy to sync on this or help push to this PR as well

docker/packages/cuda/builder-packages.json
```diff
@@ -1,7 +1,5 @@
 {
   "rhel_to_ubuntu": {
-    "libnvshmem3-cuda-${CUDA_MAJOR}-3.4.5-1": "libnvshmem3-cuda-${CUDA_MAJOR}=3.4.5-1",
-    "libnvshmem3-devel-cuda-${CUDA_MAJOR}-3.4.5-1": "libnvshmem3-dev-cuda-${CUDA_MAJOR}=3.4.5-1",
-    "libnvshmem3-static-cuda-${CUDA_MAJOR}-3.4.5-1": "libnvshmem3-static-cuda-${CUDA_MAJOR}=3.4.5-1"
+    "python3.9-devel": null
```
Gregory-Pereira, line 3: Is there a reason why we don't need this in ubuntu? pretty sure it needs it:

```suggestion
    "python3.9-devel": "python3.9-dev"
```

docker/scripts/cuda/builder/build-nvshmem.sh
```diff
@@ -24,6 +24,7 @@ tar -xf "nvshmem_src_cuda${CUDA_MAJOR}.tar.gz"
 cd nvshmem_src
 
 git apply /tmp/patches/cks_nvshmem"${NVSHMEM_VERSION}".patch
+git apply /tmp/patches/nvshmem_zero_ibv_ah_attr_"${NVSHMEM_VERSION}".patch
```
Gregory-Pereira: The dockerfile is meant to make nvshmem version configurable even if we are pinning to 3.3.20 for now, lets add a check to test if the version is 3.3.20 and if so apply these patches:

```suggestion
if [ "${NVSHMEM_VERSION}" = "3.3.9" ] || [ "${NVSHMEM_VERSION}" = "3.3.20" ]; then
    git apply /tmp/patches/cks_nvshmem"${NVSHMEM_VERSION}".patch
    if [ "${NVSHMEM_VERSION}" = "3.3.20" ]; then
        git apply /tmp/patches/nvshmem_zero_ibv_ah_attr_"${NVSHMEM_VERSION}".patch
    fi
fi
```

Gregory-Pereira (changes requested):
Looks great - 2 questions

docker/packages/cuda/builder-packages.json
smarterclayton: ubuntu did not need it to build successfully, because ubuntu does not have `python3.9` in its path (NVSHMEM during build logic searches the PATH for any `pythonX.Y` to find the versions).

docker/scripts/cuda/builder/build-nvshmem.sh
smarterclayton: I have that in the 3.4.5 PR, will bring it back once I know 3.4.5 compiles (#411 )

docker/packages/cuda/builder-packages.json
Gregory-Pereira: We talked about this, we don't think its needed because ubuntu doesn't have the older python still installed in path

Gregory-Pereira (dismissed):
/lgtm

docker/scripts/cuda/builder/build-nvshmem.sh
smarterclayton: Integrated now

smarterclayton: Integrated a version of this, rather

Gregory-Pereira (approved):
/lgtm

smarterclayton:
Spawned https://github.com/llm-d/llm-d/issues/412 for ubuntu.

PR #414, opened 2025-10-30
[Docs] Fix InferencePool version number getting cut off in WideEP guide

tlrmchlsmth:
Sidestep this UI issue which at least on macOS gives no indication that there's more to the line regardless of how big the window is:
 
image


Gregory-Pereira (approved):
/lgtm good call, I hate the scrolling code window

PR #418, opened 2025-10-30
Add JIT directories and model directories

smarterclayton:
Ensure all artifacts that are generated during startup are written to a temporary directory.  Separate models (infrequently change) from JIT caches (frequently changed, may have bugs that require clearing the cache). As a best practice, use the large SSD attached to GPU machines on Coreweave and GCP as an ephemeral store for models and JIT artifacts.

### DIFF
--- a/guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
+++ b/guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
@@ -42,6 +42,10 @@ spec:
             emptyDir:
               medium: Memory
               sizeLimit: 2Gi # roughly 32MB per local DP plus scratch space
+          - name: hf-cache
+            emptyDir: {}
+          - name: jit-cache
+            emptyDir: {}
         containers:
         - name: vllm
           image: ghcr.io/llm-d/llm-d-cuda:v0.3.1
@@ -142,6 +146,21 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: status.podIP
+
+          # Use cache directories from the mounted volume under one easy to mount
+          # root directory.
+          - name: CUDA_CACHE_PATH
+            value: /var/cache/vllm/cuda
+          - name: CCACHE_DIR
+            value: /var/cache/vllm/ccache
+          - name: VLLM_CACHE_ROOT
+            value: /var/cache/vllm/vllm
+          - name: FLASHINFER_WORKSPACE_BASE
+            value: /var/cache/vllm/flashinfer
+          # HuggingFace is likely to be quite a bit larger, give it its own cache
+          - name: HF_HUB_CACHE
+            value: /var/cache/huggingface
+
           ports:
           - containerPort: 8200
             name: metrics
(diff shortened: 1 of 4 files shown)


guides/wide-ep-lws/manifests/modelserver/coreweave/kustomization.yaml
```diff
@@ -8,25 +8,20 @@ patches:
       kind: LeaderWorkerSet
     patch: |-
       # We add the volume for different providers since the hostPath can be different.
-      - op: add
-        path: /spec/leaderWorkerTemplate/workerTemplate/spec/volumes/-
-        value: 
+      - op: replace
```
liu-cong, line 11: For this to work we need to carefully maintain the volume index in the base yaml, IMO this is error prone.

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -142,6 +146,25 @@ spec:
...
+          - name: CCACHE_DIR
+            value: /var/cache/vllm/ccache
+          - name: VLLM_CACHE_ROOT
+            value: /var/cache/vllm/vllm
+          - name: FLASHINFER_WORKSPACE_BASE
+            value: /var/cache/vllm/flashinfer
+          - name: TRITON_CACHE_DIR
+            value: /var/cache/vllm/triton
+          - name: DG_JIT_CACHE_DIR
+            value: /var/cache/vllm/deepgemm
```
tlrmchlsmth: Why separate out the different components of the vllm cache? 

vLLM defaults DG_JIT_CACHE_DIR to `VLLM_CACHE_ROOT/deepgemm` already, and always overrides TRITON_CACHE_DIR. Generally I'm trying to keep all of vLLM's jitted code in the VLLM_CACHE_ROOT directory by default.

Not sure about flashinfer though - @mgoin do you know? I don't see anything when grepping  `FLASHINFER_WORKSPACE_BASE` in vLLM

guides/wide-ep-lws/manifests/modelserver/coreweave/kustomization.yaml
smarterclayton: What's an alternative?

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
smarterclayton: No reason to separate the components, other than I wasn't aware that vLLM was manipulating running env vars.

liu-cong (dismissed):

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
smarterclayton: Stripped those two out.

guides/wide-ep-lws/manifests/modelserver/coreweave/kustomization.yaml
smarterclayton: We agreed we don't have an alternative, because we want the "base" to be useful, AND because we should get validation errors if the index is wrong (can't have a two).

liu-cong (approved):

PR #429, opened 2025-11-03
Set LD_LIBRARY_PATH for nvshmem appropriately

smarterclayton:
Remove nvidia nvshmem packages if we are detected as having built nvshmem.


docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -93,5 +94,10 @@ echo "DEBUG: Installing packages: ${INSTALL_PACKAGES[*]}"
 # install all packages in one command with verbose output to prevent GHA timeouts
 uv pip install -v "${INSTALL_PACKAGES[@]}"
 
+# uninstall the NVSHMEM dependency brought in by vllm if using a compiled NVSHMEM
+if [[ "${NVSHMEM_DIR-}" != "" ]]; then
+  uv pip uninstall nvidia-nvshmem-cu${CUDA_MAJOR} nvshmem4py-cu${CUDA_MAJOR}
```
smarterclayton: It looks like the right nvshmem4py-cu is installed (the one we built) so I will probably remove this before merge

Gregory-Pereira (approved):
/lgtm

PR #439, opened 2025-11-07
guides: Make vLLM log more useful in inference-scheduling

russellb:
This change makes the vLLM logging more user friendly out of the box
for someone using this example to try out llm-d.

- `--disable-uvicorn-access-log`

This disables logging every HTTP request received by vLLM. When this
is on, the log is flooded with many requests to `/metrics` every
second. To disable this message only for the `/metrics`, we could use
a custom logger configuraiton, but it would probably double the size
of this example, so I'm not sure it's worth it.

- Change `VLLM_LOGGING_LEVEL` from `DEBUG` to `INFO`

It's nice to have this in the log to make it easy to change quickly,
but the `DEBUG` logs are a bit much for a typical user. The env var is
still in place to make it very easy to change if someone needs to for
debugging something.

Signed-off-by: Russell Bryant <rbryant@redhat.com>

### DIFF
--- a/guides/inference-scheduling/ms-inference-scheduling/values.yaml
+++ b/guides/inference-scheduling/ms-inference-scheduling/values.yaml
@@ -38,6 +38,7 @@ decode:
     args:
       - "--kv-transfer-config"
       - '{"kv_connector":"NixlConnector", "kv_role":"kv_both"}'
+      - "--disable-uvicorn-access-log"
     env:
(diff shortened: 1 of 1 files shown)


markmc:
> * `--disable-uvicorn-access-log`
> 
> This disables logging every HTTP request received by vLLM. When this is on, the log is flooded with many requests to `/metrics` every second. 

See also vllm-project/vllm#29046

PR #446, opened 2025-11-10
fix: add --no-cache-dir to vllm pip install

starpit:
This shaves about 350 megabytes off the image.


diegocastanibm:
One of the issues that I see disabling the cache is that, if the script is used in the CI/CD pipelines, then building times will increase significantly since the cache can't speed up repeated installations across builds.
Also, it seems to be an issue in the checks that could be solved maybe with a rebase

cc. @Gregory-Pereira

PR #330, opened 2025-10-07
feat: Add updated readiness probe for vLLM containers

rajinator:
Implements model-aware health checking for Kubernetes readiness probes. Ensures pods are marked Ready only when vLLM has loaded models and API is serving requests, preventing premature traffic routing.

Changes:
- Add readiness_probe.sh script with /health and /v1/models validation
- Update Dockerfiles to include probe script in images
- Update example configurations with exec probes
- Add mandatory PORT parameter to prevent misconfigurations
- Add documentation for probe usage

Tested with simulated accelerator deployment on Kind cluster.

- Reopening from upstream, instead of fork

Closes #300

PR #450, opened 2025-11-12
refactor: Optimize vLLM probe timings for faster response

Ayobami-00:
## Description
This PR implements optimized probe timings as suggested by @smarterclayton in PR #330.

### Changes
- Reduced readiness probe frequency from 5s to 1s for faster detection
- Optimized startup probe parameters for more responsive checking
- Made liveness probe timing more consistent
- Updated documentation to reflect optimized values

### Background
These changes address part of the feedback from PR #330. While using `/v1/models` as 
the readiness probe endpoint isn't ideal due to potential overhead with many models/adapters,
the timing optimizations help mitigate some concerns by:

1. Detecting failures faster (1s vs 5s)
2. Allowing faster startup detection
3. Maintaining the same maximum window for model loading

### Future Work
A separate effort will be initiated to work with vLLM upstream on developing a more 
efficient dedicated `/health/ready` endpoint that avoids the overhead issues.

### DIFF
--- a/docs/readiness-probes.md
+++ b/docs/readiness-probes.md
@@ -34,16 +34,15 @@ containers:
   - containerPort: 8000  # or 8200 for decode pods
     protocol: TCP
   
-  # Startup Probe: Wait for model to load during initialization
+  # Startup Probe: Wait for vLLM server to start
   # Protects liveness/readiness probes from firing too early
   startupProbe:
     httpGet:
-      path: /v1/models
+      path: /health        # Using lightweight health endpoint
       port: 8000
-    initialDelaySeconds: 5    # Time before first probe (reduced for faster response)
-    periodSeconds: 5         # How often to probe during startup (increased frequency)
+    periodSeconds: 1        # Check every second for fastest detection
     timeoutSeconds: 3        # HTTP request timeout
-    failureThreshold: 360    # Max attempts (5s * 360 = 30min max startup time)
+    failureThreshold: 360    # Max attempts (1s * 360 = 6min max startup time)
   
   # Liveness Probe: Is the server process alive?
   # Simple health check, restarts container if failing
@@ -123,12 +122,12 @@ $ curl http://localhost:8000/v1/models
 ```
 Container Start
       ↓
-[startupProbe on /v1/models]
-  ↓ (5s intervals, up to 30min)
-  ✓ Model loaded → Startup complete
+[startupProbe on /health]
+  ↓ (1s intervals, up to 6min)
+  ✓ Server running → Startup complete
       ↓
 [livenessProbe on /health] ← Restarts if server crashes (every 5s)
-[readinessProbe on /v1/models] ← Routes traffic when ready (checks every 1s)
+[readinessProbe on /v1/models] ← Routes traffic when model is ready (checks every 1s)
 ```
 
 ## Benefits
@@ -178,10 +177,9 @@ decode:
     
     startupProbe:
       httpGet:
-        path: /v1/models
+        path: /health
         port: 8200
-      initialDelaySeconds: 5
-      periodSeconds: 5
+      periodSeconds: 1
       timeoutSeconds: 3
       failureThreshold: 360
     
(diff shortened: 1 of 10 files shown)


docs/readiness-probes.md
```diff
@@ -40,30 +40,30 @@ containers:
     httpGet:
       path: /v1/models
       port: 8000
-    initialDelaySeconds: 15    # Time before first probe
-    periodSeconds: 30           # How often to probe during startup
-    timeoutSeconds: 5           # HTTP request timeout
-    failureThreshold: 60        # Max attempts (30s * 60 = 30min max startup time)
+    initialDelaySeconds: 5    # Time before first probe (reduced for faster response)
```
smarterclayton: There's no real value to an initialDelaySeconds.  I would simply remove this.

docs/readiness-probes.md
```diff
@@ -40,30 +40,30 @@ containers:
...
+    periodSeconds: 5         # How often to probe during startup (increased frequency)
```
smarterclayton: There's no disadvantage to probing an HTTP endpoint that is cheap every 1 second.

Also, the startup probe should be on the health endpoint because the model server doesn't start unless the models are loaded, and health is cheap.

guides/wide-ep-lws/manifests/modelserver/base/prefill.yaml
```diff
@@ -119,24 +119,23 @@ spec:
...
             httpGet:
               path: /health
-              port: metrics
-            initialDelaySeconds: 0
+              port: 8000
             periodSeconds: 1
-            timeoutSeconds: 5
-            failureThreshold: 1800
+            timeoutSeconds: 3
+            failureThreshold: 360
```
smarterclayton, line 125: This could be in excess of 45 minutes (if not longer).

Please update wide-ep startup probes to take `45 * 60` failureThreshold.

guides/pd-disaggregation/ms-pd/values.yaml
```diff
@@ -164,25 +163,24 @@ prefill:
...
-        path: /v1/models
+        path: /health
         port: 8000
-      initialDelaySeconds: 15
-      periodSeconds: 30
-      timeoutSeconds: 5
-      failureThreshold: 60
+      periodSeconds: 1
+      timeoutSeconds: 3
+      failureThreshold: 360
```
smarterclayton, line 170: This value is likely too low even for llama8b, much less llama 70b.  We need to set the minimum value longer than the majority of environments will take to download from hugging face.

For 70b I would pick a time that is no shorter than 30 minutes, simply because we also may have warmup jitting.  So `30 * 60` for all llama70b model servers in the guides.

guides/inference-scheduling/ms-inference-scheduling/values_tpu.yaml
```diff
@@ -62,25 +62,24 @@ decode:
...
-        path: /v1/models
+        path: /health
         port: 8200
-      initialDelaySeconds: 15
-      periodSeconds: 30
-      timeoutSeconds: 5
-      failureThreshold: 60
+      periodSeconds: 1
+      timeoutSeconds: 3
+      failureThreshold: 360
```
smarterclayton, line 69: Because this is llama70b, needs to be 30m / consistent with other use.

docs/readiness-probes.md
```diff
@@ -34,36 +34,35 @@ containers:
...
-      path: /v1/models
+      path: /health        # Using lightweight health endpoint
       port: 8000
-    initialDelaySeconds: 15    # Time before first probe
-    periodSeconds: 30           # How often to probe during startup
-    timeoutSeconds: 5           # HTTP request timeout
-    failureThreshold: 60        # Max attempts (30s * 60 = 30min max startup time)
+    periodSeconds: 1        # Check every second for fastest detection
+    timeoutSeconds: 3        # HTTP request timeout
+    failureThreshold: 360    # Max attempts (1s * 360 = 6min max startup time)
```
smarterclayton, line 45: This value will vary based on the size of the model and the amount of model startup cost (i.e. prewarming).  Therefore it cannot be a single recommended value.

You should add a section here that describes how to pick a startup probe failure threshold based on the estimated time to start.  I.e. include model pulling, model loading, and model warmup, all of which can vary.  Describe the values we are recommending below (see my comments) in that context (wide ep uses deepseek which is ~600GB, disagg uses llama70b fp8 which is roughly 70GB).  Communicate to the user they should select a value that is longer than the slowest download time that should be considered a "success", and then mention as well that model warmup time may vary by accelerator type.

PR #455, opened 2025-11-19
refactor wideep to use recipes folder.

zetxqx:
Let wide-ep to use the recipes folder as well.
1. I didn't find a clean way to use the model server recipe, because we have prefill and decode models server different setup. Reusing the existing is cleaner IMO.
2. The PR's change is mainly reusing the gateway/httproute in recipe. I don't want to change the gateway name, so this PR changed the gateway to be **llm-d-inference-gateway**, httproute name to be **llm-d-route** and inferencePooll to be **llm-d-infpool**
3. the inferencePool helm value is still using the previous values.


guides/wide-ep-lws/README.md
```diff
@@ -93,9 +93,9 @@ helm install deepseek-r1 \
 <!-- TAB:Istio -->
 #### Istio
 ```bash
-helm install deepseek-r1 \
+helm install llm-d-infpool \
   -n ${NAMESPACE} \
-  -f inferencepool.values.yaml \
+  -f ./manifests/inferencepool/values.yaml \
```
liu-cong: I prefer not adding a folder which doesn't really add any value.

guides/wide-ep-lws/manifests/gateway/gke-l7-regional-external-managed/kustomization.yaml
```diff
@@ -2,33 +2,9 @@
...
-  # Remove port value for the inference pool in the v1 apis
-  patch: |-
-    - op: remove
-      path: /spec/rules/0/backendRefs/0/port
+- ../../../../recipes/gateway/gke-l7-regional-external-managed
 labels:
 - pairs:
     app.kubernetes.io/instance: wide-ep
-    app.kubernetes.io/gateway: wide-ep-inference-gateway
+    app.kubernetes.io/gateway: llm-d-inference-gateway
```
liu-cong: It looks like we keep the kustomization here just for these labels. I think these are for informational purpose and it's OK not having them. So we should be able to get rid of these and use the recipe directly.

guides/wide-ep-lws/README.md
zetxqx: moved to one upper level

guides/wide-ep-lws/manifests/gateway/gke-l7-regional-external-managed/kustomization.yaml
zetxqx: removed and use the gateway recipe directly

liu-cong (approved):

PR #459, opened 2025-11-20
Infra, MS and GAIE bumps + istio change compat

Gregory-Pereira:
relates to: https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1831

### DIFF
--- a/guides/inference-scheduling/helmfile.yaml.gotmpl
+++ b/guides/inference-scheduling/helmfile.yaml.gotmpl
@@ -96,7 +96,7 @@ releases:
   - name: {{ printf "ms-%s" $rn | quote }}
     namespace: {{ $ns }}
     chart: llm-d-modelservice/llm-d-modelservice
-    version: v0.3.6
+    version: v0.3.8
     installed: true
     needs:
       - {{ printf "infra-%s" $rn | quote }}
--- a/guides/pd-disaggregation/helmfile.yaml.gotmpl
+++ b/guides/pd-disaggregation/helmfile.yaml.gotmpl
@@ -91,7 +91,7 @@ releases:
   - name: {{ printf "ms-%s" $rn | quote }}
     namespace: {{ $ns }}
     chart: llm-d-modelservice/llm-d-modelservice
-    version: v0.3.6
+    version: v0.3.8
     installed: true
     needs:
       - {{ printf "infra-%s" $rn | quote }}
--- a/guides/precise-prefix-cache-aware/helmfile.yaml.gotmpl
+++ b/guides/precise-prefix-cache-aware/helmfile.yaml.gotmpl
@@ -82,7 +82,7 @@ releases:
   - name: {{ printf "ms-%s" $rn | quote }}
     namespace: {{ $ns }}
     chart: llm-d-modelservice/llm-d-modelservice
-    version: v0.3.6
+    version: v0.3.8
     installed: true
     needs:
       - {{ printf "infra-%s" $rn | quote }}
--- a/guides/simulated-accelerators/helmfile.yaml.gotmpl
+++ b/guides/simulated-accelerators/helmfile.yaml.gotmpl
@@ -71,7 +71,7 @@ releases:
   - name: {{ printf "ms-%s" $rn | quote }}
     namespace: {{ $ns }}
     chart: llm-d-modelservice/llm-d-modelservice
-    version: v0.3.6
+    version: v0.3.8
     installed: true
     needs:
       - {{ printf "infra-%s" $rn | quote }}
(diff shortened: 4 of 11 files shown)


guides/precise-prefix-cache-aware/helmfile.yaml.gotmpl
```diff
@@ -60,16 +60,13 @@ releases:
...
+    {{- else if or (eq .Environment.Name "istio") (eq .Environment.Name "default") (eq .Environment.Name "istioBench") }}
       - provider:
           name: {{ .Environment.Values.provider.name }}
-      - istio:
-        {{ .Environment.Values.istio | toYaml | nindent 10 }}
-      - istio:
-          destinationRule:
-            host: {{ printf "gaie-%s-epp.%s.svc.cluster.local" $rn $ns | quote }}
+          istio:
+            {{ .Environment.Values.provider.istio | toYaml | nindent 10 }}
```
aneeshkp: I think  the  PR restructures the Istio configuration from `Values.istio` to `Values.provider.istio`, but the template references haven't been fully updated across the codebase.
https://github.com/llm-d/llm-d/blob/3ffd128364c0fc2b4f39184ea29c49e09eb6f60d/guides/simulated-accelerators/helmfile.yaml.gotmpl#L62
should be replaced with new Values.provider.istio since the old one is removed from the values

Gregory-Pereira: At that time the PR was labeled as a WIP, I had only covered 2 guides. Now I have covered all of them

guides/inference-scheduling/helmfile.yaml.gotmpl
```diff
@@ -77,11 +77,12 @@ releases:
...
       - provider:
           name: {{ .Environment.Values.provider.name }}
-      - istio:
-        {{ .Environment.Values.istio | toYaml | nindent 10 }}
-      - istio:
-          destinationRule:
-            host: {{ printf "gaie-%s-epp.%s.svc.cluster.local" $rn $ns | quote }}
+          istio:
+            {{ .Environment.Values.provider.istio | toYaml | nindent 12 }}
+      - proivder:
```
kfswain: typo: `proivder` -> `provider`

Seperately: Is there a reason we are specifying provider twice here? this looks like it would hardcode istio values regardless of provider unless I'm reading this wrong?

guides/pd-disaggregation/helmfile.yaml.gotmpl
```diff
@@ -72,11 +72,12 @@ releases:
...
       - provider:
           name: {{ .Environment.Values.provider.name }}
-      - istio:
-        {{ .Environment.Values.istio | toYaml | nindent 10 }}
-      - istio:
-          destinationRule:
-            host: {{ printf "gaie-%s-epp.%s.svc.cluster.local" $rn $ns | quote }}
+          istio:
+            {{ .Environment.Values.provider.istio | toYaml | nindent 12 }}
+      - proivder:
```
kfswain: Same

guides/precise-prefix-cache-aware/helmfile.yaml.gotmpl
```diff
@@ -60,16 +60,15 @@ releases:
...
       - provider:
           name: {{ .Environment.Values.provider.name }}
-      - istio:
-        {{ .Environment.Values.istio | toYaml | nindent 10 }}
-      - istio:
-          destinationRule:
-            host: {{ printf "gaie-%s-epp.%s.svc.cluster.local" $rn $ns | quote }}
+          istio:
+            {{ .Environment.Values.provider.istio | toYaml | nindent 12 }}
+      - proivder:
```
kfswain: Same

guides/simulated-accelerators/helmfile.yaml.gotmpl
```diff
@@ -58,31 +58,27 @@ releases:
...
       - provider:
           name: {{ .Environment.Values.provider.name }}
-      - istio:
-        {{ .Environment.Values.istio | toYaml | nindent 10 }}
-      - istio:
-          destinationRule:
-            host: {{ printf "gaie-%s-epp.%s.svc.cluster.local" $rn $ns | quote }}
+          istio:
+            {{ .Environment.Values.provider.istio | toYaml | nindent 12 }}
+      - proivder:
```
kfswain: Same

guides/inference-scheduling/helmfile.yaml.gotmpl
Gregory-Pereira: Its defined twice there because otherwise we get the following:
```
err: failed to read helmfile.yaml.gotmpl: reading document at index 1: yaml: unmarshal errors:
  line 63: mapping key "destinationRule" already defined at line 47
changing working directory back to "/Users/gregpereirapereira/Documents/tech/work/red-hat/code/neuralmagic/llm-d/guides/inference-scheduling"
in ./helmfile.yaml.gotmpl: failed to read helmfile.yaml.gotmpl: reading document at index 1: yaml: unmarshal errors:
  line 63: mapping key "destinationRule" already defined at line 47
```
This is a function of how helmfile layers values files - it does not do deepmerge like helm does. Instead the set section will work in chronological order. The reason we have two instances of destinationRule is because for non benchmarking setting we want to provide the basics to just make destinationRule functional:
```
provider:
  name: istio
  istio:
    destinationRule:
      trafficPolicy:
        tls:
          mode: SIMPLE
          insecureSkipVerify: true
```
But with benchmarking we provide other values over the top of this to increase concurrency.

This will eventually go away - its tech debt. The idea is that any deployment should support benchmarking because they are performance examples. Next release we will unify the istioBench and istio envs.

kfswain: Ack, ty for the explanation!

kfswain (approved):

PR #428, opened 2025-11-01
Add CPU for llm-d Inference Scheduling

ZhengHongming888:
This PR is to add the CPU support for llm-d Inference Scheduling framework.

There are two main modification here -

1. ms-inference-scheduling/values_cpu.yaml
- add the CPU support for llm-d on Inference Scheduling

2. helmfile.yaml.gotmpl
- include the option of ms-inference-scheduling/values_cpu.yaml for modelservice part.

Testing results -
image

image

Please check and review it. Welcome for your any comments.

Thanks.

PR #392, opened 2025-10-27
Add CPU for llm-d P/D disaggregation

ZhengHongming888:
This PR is to add the CPU support for llm-d P/D framework.

Part 1:    Functionalities related to this PR -
There are three main modification here -

1. ms-pd/values_cpu.yaml
- add the CPU support for llm-d on Prefill/Decode

2. helmfile.yaml.gotmpl
- include the option of ms-pd/values_cpu.yaml for modelservice part.

3. README.cpu.md

- detailed guide for how to run the llm-d P/D disaggregation in pure CPU environment.

This PR has been validated by vllm p/d side (run_accuracy_test.sh which using toy_proxy_server.py for 1p1d) and also validated in llm-d p/d disaggregation as below. 

image

You can see the kv transfer happened in nixl_connector.py as above results. Also checking with routing_proxy pod as below -

image

All above shows the kv cache transferred between prefill and decode for cpu vllm/llm-d. 

Part II:   Scenario & Performance related -

Based on the PR discussion as below some questions related to where the application scenario under CPU pd case is. We did the testing over the different nonpd vs. pd testing cases under 6 numa CPU machine (Intel(R) Xeon(R) 6972P) as below. There are 6 numas in this machine as -
image

We just take 2 results here -
image
image

From the results above the 2p4d_tp1 can show the better performance than the non-pd tp4 for both two scenario (input length = 3.5k and input length = 10k) under the TPOT requirement (<100ms). 

Here we use cpu numa as the basic p and d granularity and also for different rank as tp. That means CPU also can do different TP and different P/D same as GPU cards.

Please check and review it. Welcome for your any comments.

Thanks.

PR #465, opened 2025-11-20
Create CPU dockefile for PD and Inference Scheduling

ZhengHongming888:
This PR is to create cpu dockerfile for the vllm cpu image used by p/d disaggregation (PR [#392](https://github.com/llm-d/llm-d/pull/392)) and inference scheduling (PR [#428](https://github.com/llm-d/llm-d/pull/428)). 

The usage for docker build as -

`cd llm-d

docker build   --no-cache -f docker/Dockerfile.cpu  --tag vllm-cpu-env .`

After running you will see the CPU vllm docker image. 

Thanks.

### DIFF
--- a/docker/Dockerfile.cpu
+++ b/docker/Dockerfile.cpu
@@ -104,6 +104,8 @@ CMD ["/bin/bash"]
 # ============================================================
 FROM vllm-build
 
+ENV LD_LIBRARY_PATH="/usr/local/lib:/usr/lib:${LD_LIBRARY_PATH}"
+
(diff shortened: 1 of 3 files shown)


Gregory-Pereira:
OK, so were dropping PD support for now (to be clear I mean guide integration rather than from a build perspective by installing NIXL for CPU which does not exist in the context of this PR) until a performance use case can be established. Were happy to include the CPU dockerfile changes and the inference-scheduling example. Will follow up with a review right now

ZhengHongming888:
> /lgtm

Thanks @Gregory-Pereira

PR #493, opened 2025-11-24
v0.4.0 intelligent scheduling plugins

vMaroon:
## Summary

This PR updates the default weighted scorer configurations for the `intelligent-scheduling` and `precise prefix-cache aware scheduling` well-lit paths. 
- The weights are based on benchmarking experiments that were executed throughout the past couple of months
- The new `no-hit-lru-scorer` has been shown to increase the saturation point in standard prefix-cache benchmarks (none published yet). In a simple experiment of 8x Qwen3-32B on 1 H100 node:
      
      target: http://infra-kv-events-inference-gateway.llm-d-precise.svc.cluster.local:80
      model: Qwen/Qwen3-32B
      request_type: text_completions
      profile: constant
      rate: [30, 50]
      max_seconds: 100
      data:
        prefix_buckets:
          - bucket_weight: 750
            prefix_count: 150
            prefix_tokens: 4000
        prompt_tokens: 1500
        output_tokens: 1000

     The scorer keeps the system stable at 50 QPS, while the configuration without it does not.

This requires a v0.4.0 scheduler image - do not merge before then.

cc @Gregory-Pereira @liu-cong

### DIFF
--- a/guides/inference-scheduling/gaie-inference-scheduling/values.yaml
+++ b/guides/inference-scheduling/gaie-inference-scheduling/values.yaml
@@ -13,7 +13,34 @@ inferenceExtension:
     ###################
     pullPolicy: Always
   extProcPort: 9002
-  pluginsConfigFile: "default-plugins.yaml"
+
+  pluginsConfigFile: "intelligent-scheduling-config.yaml"
+  pluginsCustomConfig:
+    intelligent-scheduling-config.yaml: |
+      apiVersion: inference.networking.x-k8s.io/v1alpha1
+      kind: EndpointPickerConfig
+      plugins:
+        - type: single-profile-handler
+        - type: prefix-cache-scorer # in v0.4.0+, automatic parameter tuning is enabled
+        - type: kv-cache-utilization-scorer
+        - type: queue-scorer
+        - type: no-hit-lru-scorer
+          parameters:
+            lruSize: 1000  # 1000 pods for tracking
+        - type: max-score-picker
+      schedulingProfiles:
+        - name: default
+          plugins:
+            - pluginRef: prefix-cache-scorer
+              weight: 3.0
+            - pluginRef: no-hit-lru-scorer
+              weight: 2.0 # effective on cold requests
+            - pluginRef: kv-cache-utilization-scorer
+              weight: 2.0
+            - pluginRef: queue-scorer
+              weight: 2.0
+            - pluginRef: max-score-picker
+
   monitoring:
     interval: "10s"
     # Service account token secret for authentication
--- a/guides/precise-prefix-cache-aware/gaie-kv-events/values.yaml
+++ b/guides/precise-prefix-cache-aware/gaie-kv-events/values.yaml
@@ -50,12 +50,17 @@ inferenceExtension:
                 hashSeed: "42"
         - type: kv-cache-utilization-scorer
         - type: queue-scorer
+        - type: no-hit-lru-scorer
+          parameters:
+            lruSize: 1000  # 1000 pods for tracking
         - type: max-score-picker
       schedulingProfiles:
         - name: default
           plugins:
             - pluginRef: precise-prefix-cache-scorer
               weight: 3.0
+            - pluginRef: no-hit-lru-scorer
+              weight: 2.0 # effective on cold requests
             - pluginRef: kv-cache-utilization-scorer
               weight: 2.0
             - pluginRef: queue-scorer


guides/inference-scheduling/gaie-inference-scheduling/values.yaml
```diff
@@ -13,7 +13,34 @@ inferenceExtension:
...
   extProcPort: 9002
-  pluginsConfigFile: "default-plugins.yaml"
+
+  pluginsConfigFile: "intelligent-scheduling-config.yaml"
+  pluginsCustomConfig:
+    intelligent-scheduling-config.yaml: |
+      apiVersion: inference.networking.x-k8s.io/v1alpha1
+      kind: EndpointPickerConfig
+      plugins:
+        - type: single-profile-handler
```
ahg-g, line 23: this is added by default, not need to set it explicitly.

guides/inference-scheduling/gaie-inference-scheduling/values.yaml
```diff
@@ -13,7 +13,34 @@ inferenceExtension:
...
+          plugins:
+            - pluginRef: prefix-cache-scorer
+              weight: 3.0
+            - pluginRef: no-hit-lru-scorer
+              weight: 2.0 # effective on cold requests
+            - pluginRef: kv-cache-utilization-scorer
+              weight: 2.0
+            - pluginRef: queue-scorer
+              weight: 2.0
+            - pluginRef: max-score-picker
```
ahg-g, line 42: ditto, this is set by default

guides/inference-scheduling/gaie-inference-scheduling/values.yaml
```diff
@@ -13,7 +13,34 @@ inferenceExtension:
...
+  pluginsCustomConfig:
+    intelligent-scheduling-config.yaml: |
+      apiVersion: inference.networking.x-k8s.io/v1alpha1
+      kind: EndpointPickerConfig
+      plugins:
+        - type: single-profile-handler
+        - type: prefix-cache-scorer # in v0.4.0+, automatic parameter tuning is enabled
+        - type: kv-cache-utilization-scorer
+        - type: queue-scorer
+        - type: no-hit-lru-scorer
```
ahg-g, line 27: Did we validate that this has no negative side effects for the cases that it doesn't optimize for?

ahg-g (commented):
if no-hit-lru is safe to enable by default for all workloads, and has no model server dependency, I recommend a followup and upstreaming it into IGW so that it becomes part of `default-plugins.yaml`

guides/inference-scheduling/gaie-inference-scheduling/values.yaml
vMaroon: No, but in low reuse scenarios this practically replaces random scheduling with round-robin scheduling. Other scorers should over-power it on system load. We can decrease its weight to 1 if sounds better.

ahg-g:
also, instead of overriding the whole config, you can add the no-hit-lru using the pluginsCustomConfig helm value: https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/5033e19b78decceba735d416b87cf162c4905ab3/config/charts/inferencepool/templates/epp-config.yaml#L24C34-L24C53

guides/inference-scheduling/gaie-inference-scheduling/values.yaml
liu-cong: We have regression benchmarks in EPP, @rlakhtakia is the POC. I am hesitant of putting this up without running those tests. 

Can we add this once we validate this has no negative impact?  It should be an easy patch given just config updates. And how about the active request scorer? I thought that can help as well?

vMaroon: The `active-request-scorer` can help yes, but I'm not sure it should run alongside the queue-scorer. Considering that the running-queue scorer is close, I feel like the active-request-scorer wont have any light.

vMaroon: @liu-cong can the regression benchmarks be ran ASAP? If it's not a big effort, it would be ideal to include this scorer since it's mentioned in the release.

We can limit it to the precise prefix-cache path, but if not a hassle then it is preferable.

liu-cong: @rlakhtakia has kicked off benchmark runs with the new config, looking forward to the results! Thanks @rlakhtakia

vMaroon: @liu-cong @rlakhtakia any updates?

liu-cong: yeah got some results, though it's a bit confusing. the new config had slightly lower latency  for prefill heavy workloads but lower throughput in a few datapoints... I think this could be due to a bug in the benchmark tool which is the legacy latency profiler generator. That said, I think I trust the latency metric more than the throughput, so the new config looks promising.

I suggest we add this to something like `preview-values.yaml`, and explain the use case it optimizes, so users can try it out, and make it the default after we have more comprehensive benchmark results. @vMaroon how does this sound?

PR #503, opened 2025-11-26
Avoid defining LMCACHE_COMMIT_SHA in multiple places

terrytangyuan:
The commit SHA only needs to be defined once at the top of each Dockerfile, and build stages inherit the value. When we need to update the LMCache commit SHA in the future, we only need to change it in one place per file.


Gregory-Pereira (approved):
/lgtm

PR #512, opened 2025-11-28
fix: EPP configuration issues in precise-prefix-cache-aware guide

yuanwu2017:
**Problem**: When EPP initializes the precise-prefix-cache-scorer plugin, it attempts to download the HuggingFace tokenizer to a read-only directory.

**Error log**:
```
E1128 05:17:49.814731 Failed to get hf tokenizer for precise-prefix-cache-scorer
  mkdir /go: permission denied
```

**Root cause**: The EPP container runs as non-root user and the default HuggingFace cache directory (/go/pkg/mod/.../bin/) is read-only.

**Fix**: Configure `tokenizersCacheDir: "/tmp/tokenizers"` in the pluginsCustomConfig to use a writable directory.


guides/precise-prefix-cache-aware/gaie-kv-events/values.yaml
```diff
@@ -30,10 +30,9 @@ inferenceExtension:
         secretKeyRef:
           name: llm-d-hf-token
           key: HF_TOKEN
-  flags:
+  extraArgs:
     # Log verbosity
-    - name: v
```
liu-cong, line 35: this is being fixed in https://github.com/llm-d/llm-d/pull/509

I am curious where does the extraArgs come from?

yuanwu2017: I asked claude code.  and it seems that it interprets the special extraArgs as a universal one.
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/
image


Let me remove it.

PR #515, opened 2025-12-01
[docs] Improve accelerator documentation

poussa:
* Add Intel Gaudi
* Add sections about device plugins and DRA
* Add links to each vendor's device plugin and DRA resource drivers

### DIFF
--- a/docs/accelerators/README.md
+++ b/docs/accelerators/README.md
@@ -28,15 +28,32 @@ For integration into the well-lit paths our standard for contribution is higher,
 > [!NOTE] 
 > The community can assist but is not responsible for keeping hardware guide variants updated. We reserve the right to remove stale examples and documentation with regard to hardware support.
 
-## Hardware-Specific Setup
+## Accelerator Resource Management
 
-### Intel XPU
+To enable llm-d accelerators to access hardware devices, the devices must be exposed to containers. Kubernetes provides two mechanisms to accomplish this:
 
-Intel XPU deployments require the Intel GPU Device Plugin to be installed in your Kubernetes cluster:
+1. [Device Plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/)
+2. [Dynamic Resource Allocation (DRA)](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)
 
-```bash
-# Deploy Intel GPU Device Plugin v0.32.1
-kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin?ref=v0.32.1'
-```
+Typically, clusters use one mechanism or the other to expose accelerator devices. While it's possible to use both mechanisms simultaneously, this requires special configuration not covered in this document.
 
-This plugin enables Kubernetes to discover and schedule workloads on Intel GPUs. Make sure to install this before deploying any XPU-based inference workloads.
+### Device Plugins
+
+Each vendor provides Device Plugins for their accelerators. The following plugins are available by vendor:
+
+- [AMD ROCm Device Plugin](https://github.com/ROCm/k8s-device-plugin)
+- Google TPU Device Plugin (automatically enabled for TPU instances)
+- [Intel XPU Device Plugin](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/cmd/gpu_plugin/README.md)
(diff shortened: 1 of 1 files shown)


docs/accelerators/README.md
```diff
@@ -28,15 +28,62 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+### Dynamic Resource Allocation
+
+Each vendor provides DRA resource drivers for their accelerators. The following drivers and setup documentation are available by vendor:
+
+- [AMD ROCm Resource Driver](https://github.com/ROCm/k8s-gpu-dra-driver)
+- [Prepare GKE for DRA workloads](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/set-up-dra)
+- [Intel XPU and Gaudi Resource Driver](https://github.com/intel/intel-resource-drivers-for-kubernetes)
+- [NVIDIA GPU Resource Driver](https://github.com/NVIDIA/k8s-dra-driver-gpu)
+
+Since DRA is a newer Kubernetes feature, the following feature gates must be enabled for vanilla v1.34 clusters:
```
byako: I don't know how much is it worth to get into details in this doc, but strictly speaking, only one feature gate _must be_ enabled: `ResourceHealthStatus` - to mitigate kubelet's grpc connection loss to DRA driver's kubelet-plugin. I don't remember if the fix for this is backported into 1.34.2, I'll clarify and come back.

The rest are nice to have or subject to a particular DRA driver's requirements / expectiations, but not a `must`. If it's worth expanding the benefits of having these enabled, here's for instance, how we keep it in our [doc](https://github.com/intel/intel-resource-drivers-for-kubernetes/blob/main/doc/CLUSTER_SETUP.md#useful-and-required-featuregates).

Also, these FeatureGate enable the functionality that is only available in v1beta1 and v1alpha3 resource.k8s.io APIs, so you'll need that option for apiserver section.

docs/accelerators/README.md
```diff
@@ -28,15 +28,62 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+
+```yaml
+apiVersion: kubeadm.k8s.io/v1beta4
+kind: ClusterConfiguration
+kubernetesVersion: v1.34.0
+apiServer:
+  extraArgs:
+  - name: feature-gates
+    value: DRAExtendedResource=true,DRADeviceTaints=true,DRAAdminAccess=true,DRAPartitionableDevices=true,DRAResourceClaimDeviceStatus=true,ResourceHealthStatus=true
+clusterName: kubernetes
```
byako:
```suggestion
  - name: runtime-config
    value: resource.k8s.io/v1beta1=true,resource.k8s.io/v1alpha3=true
clusterName: kubernetes
```

docs/accelerators/README.md
poussa: Maybe this featureGate topic is too much details for this doc. Maybe we just leave it out, or say something like: 

_Since DRA is a newer Kubernetes feature, the some feature gates may be needed. Please check your vendor and/or cluster provider documentation._

WDYT?

byako: LGTM. But runtime-config section is still needed to enable v1beta and v1alpha3 APIs.

docs/accelerators/README.md
poussa: done

docs/accelerators/README.md
poussa: done

docs/accelerators/README.md
```diff
@@ -28,15 +28,46 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+
+```yaml
+apiVersion: kubeadm.k8s.io/v1beta4
+kind: ClusterConfiguration
+kubernetesVersion: v1.34.0
+apiServer:
+  extraArgs:
+  - name: runtime-config
+    value: resource.k8s.io/v1beta1=true,resource.k8s.io/v1alpha3=true
 ```
```
byako:
```suggestion
Since DRA is a newer Kubernetes feature, some feature gates may be required. Consult your vendor and cluster provider documentation for specific requirements.
```

Sorry, what I meant was - these APIs are needed for the FeatureGates, or better to put it this way: functionality that is enabled by the feature gate requires those APIs to be enabled. If enabling feature gates - then enable the alpha and beta apis as well. If not giving an example of how to enable feature gates - then mentioning the APIs can be dropped altogether because every feature gate describes, in the doc, for which components it needs to be enabled, and what APIs it requires.

poussa: fixed (don't overload the doc - drop all feature gate details)

byako (approved):
lgtm

PR #518, opened 2025-12-02
[XPU] Update P/D config to use DRA

poussa:
Migrate Intel XPU configuration from device plugin to DRA format and remove explicit GPU resource specs (not needed).

### DIFF
--- a/guides/pd-disaggregation/ms-pd/values_xpu.yaml
+++ b/guides/pd-disaggregation/ms-pd/values_xpu.yaml
@@ -34,7 +34,7 @@ decode:
       interval: "30s"
   containers:
   - name: "vllm"
-    image: ghcr.io/llm-d/llm-d-xpu:v0.4.0
+    image: ghcr.io/llm-d/llm-d-xpu-dev:pr-592
     modelCommand: vllmServe
     args:
       - "--block-size"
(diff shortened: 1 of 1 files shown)


guides/pd-disaggregation/ms-pd/values_xpu.yaml
```diff
@@ -121,11 +120,9 @@ prefill:
       limits:
         memory: 64Gi
         cpu: "8"
-        gpu.intel.com/i915: "1"
```
tlrmchlsmth: How are the number and type of GPU specified now that it's not in the yaml?

poussa: They coming from the llm-d-modelservice helm chart, [dra](https://github.com/llm-d-incubation/llm-d-modelservice/blob/e6e62419ec1b873c6edd815ea6c5a641580bff8e/charts/llm-d-modelservice/values.yaml#L95-L123) section. The `templates/_dra.tpl` then renders the proper resources to the deployment.

In fact, even if you use the `accelerator` block you don't need to specify the resource names again in the pod's resource request and limits. They are generated the same way.

yuanwu2017 (commented):
Please hold on. XPU CI has not installed DRA driver yet. You can only update the 0.4.0 image.

Issue #519, opened 2025-12-02
[Feat] Sglang support for well-lit path of approximate prefix cache aware scorer


ezrasilvera:
Support sglang as Inference Server in https://github.com/llm-d/llm-d/tree/main/guides/inference-scheduling

Issue #403, opened 2025-10-29
[EPIC] Support sglang


ezrasilvera:
A place holder to hold all needed changes across all repos. 

- [ ] llm-d/llm-d-inference-scheduler#394

robertgshaw2-redhat:
mostly completed

PR #527, opened 2025-12-03
Add SGLang option for inference-scheduling well-lit path

andreyod:
This PR will add an option to deploy SGLang as the inference server instead of vLLM.

This option is added in the "approximate prefix cache aware" profile of the "Intelligent Inference Scheduling" well-lit path with the default hardware (i.e., `GPU` hardware). Issue: https://github.com/llm-d/llm-d/issues/519

We are also planning to explore and address other scenarios like P/D Disaggregation and Precise Prefix. See: https://github.com/llm-d/llm-d/issues/403

### DIFF
--- a/guides/inference-scheduling/README.md
+++ b/guides/inference-scheduling/README.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-This guide deploys the recommended out of the box [scheduling configuration](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md) for most vLLM deployments, reducing tail latency and increasing throughput through load-aware and prefix-cache aware balancing. This can be run on two GPUs that can load [Qwen/Qwen3-32B](https://huggingface.co/Qwen/Qwen3-32B).
+This guide deploys the recommended out of the box [scheduling configuration](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md) for most vLLM and SGLang deployments, reducing tail latency and increasing throughput through load-aware and prefix-cache aware balancing. This can be run on two GPUs that can load [Qwen/Qwen3-32B](https://huggingface.co/Qwen/Qwen3-32B).
 
 This profile defaults to the approximate prefix cache aware scorer, which only observes request traffic to predict prefix cache locality. The [precise prefix cache aware routing feature](../precise-prefix-cache-aware) improves hit rate by introspecting the vLLM instances for cache entries and will become the default in a future release.
 


guides/inference-scheduling/helmfile.yaml.gotmpl
```diff
@@ -27,6 +27,8 @@ environments:
     <<: *I
   digitalocean: &DO
     <<: *I
+  sglang: &SGL
```
liu-cong: sglang is not an environment, it's a component of an environment. An environment contains multiple components and thus multiple decision points (gateway provider, accelerator type, model server, etc.). @Gregory-Pereira is there a better way to organize this? I honestly think it's easier to not automate this, but just add a few tabs for each decision point in the user guide . i.e., choose gateway provider, choose model server.

ezrasilvera: @liu-cong I agree this may not be the optimal way to integrate however I do think we should integrate it into the automation. Eventually I believe we will have regression tests for all well-lit path guides, hopefully integrated into the CICD. I think sglang should be considered as first class citizen for inference engine similar to vLLM and we need to be able to automatically verify it's not broken.

ezrasilvera: BTW, While it's indeed not an environment we choose this approach to minimize code changes and avoid overall refactoring of the helm framework.

guides/inference-scheduling/ms-inference-scheduling/values_sglang.yaml
```diff
@@ -0,0 +1,91 @@
...
+  uri: "hf://Qwen/Qwen3-0.6B"
+  name: "Qwen/Qwen3-0.6B"
+  size: 20Gi
+  authSecretName: "llm-d-hf-token"
+
+routing:
+  servicePort: 8000
+  proxy:
+    image: ghcr.io/llm-d/llm-d-routing-sidecar:v0.4.0-rc.1
+    connector: nixlv2
```
hhk7734:
```suggestion
    connector: sglang
```

Since the sglang connector is available, please switch to this.
https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/sidecar/proxy/connector_sglang.go

hhk7734 (changes requested):

guides/inference-scheduling/ms-inference-scheduling/values_sglang.yaml
andreyod: Thanks. Changed. Although in this scenario (no P/D disaggregation), the routing proxy connector doesn't play its role.

guides/inference-scheduling/helmfile.yaml.gotmpl
Gregory-Pereira: So I think environment is not the ideal place but I am in favor of using it for now. Eventually this will be picked up in the helm to kustomize refactor, so I would rather get this in as an environment for now so that way we can place it in the right place post refactor.

rahulgurnani: I agree with @liu-cong regarding not setting sglang as environment. There are other ways to handle this like setting a environement variable called MODEL_SERVER and having default as vllm.

guides/inference-scheduling/ms-inference-scheduling/values_sglang.yaml
```diff
@@ -0,0 +1,91 @@
...
+  - name: "sglang"
+    image: docker.io/lmsysorg/sglang:v0.5.5.post1
+    command: ["python3", "-m", "sglang.launch_server"]
+    args:
+      - "--model-path"
+      - "Qwen/Qwen3-0.6B"
+      - "--host"
+      - "0.0.0.0"
+      - "--port"
+      - "8200"
```
rahulgurnani: why do we need to use a different port than 8000?

andreyod: 8200 port was used for the vLLM deployment in this well-lit path at the time this PR was created. This is not the case anymore, so we can use 8000.

andreyod:
> @andreyod is this change being still worked upon?

I think it is currently not clear how to proceed with the SGLang support in llm-d.

guides/inference-scheduling/helmfile.yaml.gotmpl
andreyod: Changed to use environment variable called `INFERENCE_SERVER` (to be consistent with the readme)

guides/inference-scheduling/ms-inference-scheduling/values_sglang.yaml
andreyod: Updated to use port 8000

rahulgurnani:
> > @andreyod / @ezrasilvera , anything blocking merge of this? It would be great if we can get this in soon so that we can progress towards supporting SGLang in llm-d 0.6 release. Thanks!
> 
> Hi, Nothing is blocking this. We are waiting for a reviewer to lgtm it.

Okay. I had a comment about the version of llm-d used, otherwise looks good to me.

@smarterclayton can you please help with lgtm/merge this PR? Thanks!

guides/inference-scheduling/ms-inference-scheduling/values_sglang.yaml
```diff
@@ -0,0 +1,92 @@
...
+      portName: "sglang"  # decode service port (from routing.proxy.targetPort)
+      path: "/metrics"
+      interval: "30s"
+  containers:
+  - name: "sglang"
+    image: docker.io/lmsysorg/sglang:v0.5.6.post2
+    command: ["python3", "-m", "sglang.launch_server"]
+    args:
+      - "--model-path"
+      - "Qwen/Qwen3-0.6B"
```
ahg-g: we are using 32B for vllm, can we be consistent here?

ezrasilvera: fixe

guides/inference-scheduling/helmfile.yaml.gotmpl
```diff
@@ -66,10 +66,12 @@ releases:
       - {{ printf "infra-%s" $rn | quote }}
       {{- end }}
     values:
-      {{- if ne .Environment.Name "standalone" }}
-      - gaie-inference-scheduling/values.yaml
-      {{- else }}
+      {{- if eq .Environment.Name "standalone" }}
```
ahg-g, line 69: @capri-xiyue I see some standalone related checks here. Should we clean them up in a follow up PR until we have the standalone chart released?

guides/inference-scheduling/README.md
```diff
@@ -127,6 +127,18 @@ accelerator:
 
 This case expects using 4th Gen Intel Xeon processors (Sapphire Rapids) or later.
 
+### Inference Server Selection
```
smarterclayton, line 140: Please update the overview to say `vLLM and SGLang deployments` (or, if we think this wouldn't be true for most sglang deployments, add a caveat briefly in that section).

ezrasilvera: @smarterclayton Done. (https://github.com/llm-d/llm-d/pull/527/changes/404a5e2428e0867591cb7ddae464b60ee9189bdd).
Please note that currently all our documents refer only to vLLM so it's a wider question whether we indeed want to update all our documentation with maybe a general term of "inference engine" when the details are not vLLM specific. For example, this guide has a link to https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md   where only vLLM is mentioned.

smarterclayton: As soon as we land we can do a a follow up changing the other messaging.  I just wanted this one to be reasonably complete if someone lands on it.

guides/inference-scheduling/helmfile.yaml.gotmpl
smarterclayton: Can be a post merge followup.

smarterclayton (dismissed):

PR #528, opened 2025-12-03
update epp metrics auth to align with upstream changes

sallyom:
GAIE added the option to configure unauthenticated epp metrics endpoint. The new prometheus.auth.enabled flag was not added to llm-d values, so metrics-reader-secret wasn't aligned properly. This fixes the issue. Metrics auth is set to 'true' by default (also the default in GAIE charts), with a comment on how to configure unauthenticated metrics endpoint.


Gregory-Pereira (approved):
/lgtm

PR #529, opened 2025-12-03
Update wide-ep example to 0.4.0 image and tune startup threshold

smarterclayton:
Use the correct log statement for the new sidecar.

DeepSeek B200 is taking about 45m to startup due to both model download
and warmup for flashinfer.

Enable the sampling prefiller option in the sidecar to make local
round robin testing easier.

### DIFF
--- a/guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
+++ b/guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
@@ -24,8 +24,9 @@ spec:
               - --port=8000
               - --vllm-port=8200
               - --connector=nixlv2
-              - -v=1
+              - --zap-log-level=1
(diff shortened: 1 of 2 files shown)


guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -24,8 +24,9 @@ spec:
               - --port=8000
               - --vllm-port=8200
               - --connector=nixlv2
-              - -v=1
+              - --zap-log-level=1
               - --secure-proxy=false
+              - --enable-prefiller-sampling
```
robertgshaw2-redhat, line 29: QQ - what does this do?

robertgshaw2-redhat:
Should we also add this to the 0.4 branch?

PR #531, opened 2025-12-03
Adding Topology Aware Scheduling labels on Wide-EP GKE example

Edwinhr716:
Adding new section to show Topology Aware Scheduling using Kueue + LWS. The GKE manifest will include the labels required to use TAS. This is the LWS side user guide for context https://lws.sigs.k8s.io/docs/examples/tas/

### DIFF
--- a/guides/wide-ep-lws/README.md
+++ b/guides/wide-ep-lws/README.md
@@ -26,7 +26,7 @@ This guide requires 32 Nvidia H200 or B200 GPUs and InfiniBand or RoCE RDMA netw
   * You must have high speed inter-accelerator networking
   * The pods leveraging inter-node EP must be deployed within the same networking domain
   * You have deployed the [LeaderWorkerSet optional controller](../prereq/infrastructure/README.md#optional-install-leaderworkerset-for-multi-host-inference)
-  * Optionally, you have deployed [Kueue and the Kueue Populator](../prereq/infrastructure/README.md#optional-install-kueue-and-kueue-populator-for-topology-aware-scheduling-(tas)-for-multi-host-inference)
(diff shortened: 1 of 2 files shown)


guides/wide-ep-lws/README.md
```diff
@@ -197,3 +197,7 @@ kubectl delete -k ./manifests/gateway/<gke-l7-regional-external-managed|istio|kg
 ## Customization
 
 For information on customizing a guide and tips to build your own, see [our docs](../../docs/customizing-a-guide.md)
+
+## Topology Aware Scheduling (TAS)
+
+For information on how to use topology aware scheduling using Kueue, see [LWS + TAS user guide](https://lws.sigs.k8s.io/docs/examples/tas/). The GKE example already has the required labels.
```
smarterclayton, line 209: I want to have more detail in the guide -the core gap is that there is no gang scheduling and affinity scheduling in Kube is best effort or high toil.  So I think up near the top we need to draw out both of those as sproblems, and present Kueue + TAS today (and grove and then later core kube gang scheduling) as solutions.

A guide is roughly intended to be:

1. Optimization opportunity
2. Tradeoffs / considerations
3. Prereqs to try / choices you have to make
4. Setup
5. Validation / Benchmarking

I think you either want to add those to 2 if they are fundamental to wide ep's tradeoffs, or 3 if tehy are a requiremetns to use wide ep well.

Edwinhr716: 3 makes the most sense to me. Wdyt of adding a new header under Prerequisites called prerequisites to try where we expand on the lack of gang  / topology aware scheduling in Kube and recommend Kueue?

PR #513, opened 2025-12-01
Update Istio in guide to 1.28.1 GA

shmuelk:
The guide and the helmfile includes has the user install an alpha version of Istio 1.28 from July 2025.

This PR updates the version of Istio installed to the 1.28.1 GA version.

It also corrects a small issue in the README.md file for the Gateway installation.

In addition as this new version of Istio does not support the v1alpha2 version of the GIE InferencePool, the references to the v1alpha2 InferencePool have been updated to the v1 version of the InferencePool.

PR #444, opened 2025-11-08
doc(guides): Introduce standardized recipes for Gateway, InferencePool, and vLLM

zetxqx:
This change introduces 
1. a set of reusable and standardized recipes under guides/recipes to simplify the
process of creating new deployment guides. Following: #405
2. It also refactors the CPU Prefix Cache guide to be the first consumer of these new components.

Key changes include:

   * Standardized Naming: Established a consistent naming convention for core components:
       * Gateway: `llm-d-inference-gateway`
       * HTTPRoute: `llm-d-route`
       * InferencePool: `llm-d-infpool`
   * Standardized Labeling: The label `llm-d.ai/inferenceServing: "true"` is now used consistently on model
     server deployments and is the default selector for the InferencePool recipe.
   * Reusable Recipes: Created modular recipes for:
       * Gateway: A base gateway and a GKE-specific overlay.
       * InferencePool: A standard `values.yaml` for the Helm chart.
       * vLLM: A base deployment that can be easily customized.
   * Improved Documentation: Each recipe includes a `README.md` with clear installation and verification
     instructions.
   * Refactored Guide: The `guides/prefix-cache-storage/cpu` guide has been updated to use these new
     recipes, making it much simpler and easier to follow.

PR #534, opened 2025-12-04
default inferencepool is x-k8s until istio support lands

Gregory-Pereira:
cc @liu-cong
Somehow default got swapped back


liu-cong:
Per this comment https://github.com/llm-d/llm-d/pull/444#discussion_r2517278725 Istio 1.28 supports v1 now. Looks like we should get this PR merged instead https://github.com/llm-d/llm-d/pull/513/files

PR #547, opened 2025-12-11
deploy AMD from llm-d guides

Gregory-Pereira:
Guide implementation running IS on AMD GPUs

### DIFF
--- a/guides/inference-scheduling/gaie-inference-scheduling/values.yaml
+++ b/guides/inference-scheduling/gaie-inference-scheduling/values.yaml
@@ -27,7 +27,7 @@ inferenceExtension:
         # To allow unauthenticated /metrics access (e.g., for debugging with curl), set to false
         enabled: true
 inferencePool:
-  apiVersion: inference.networking.x-k8s.io/v1alpha2 # use old API version for inference
+  apiVersion: inference.networking.k8s.io/v1 # use old API version for inference
   targetPortNumber: 8000
   modelServerType: vllm
   modelServers:
--- a/guides/inference-scheduling/helmfile.yaml.gotmpl
+++ b/guides/inference-scheduling/helmfile.yaml.gotmpl
@@ -19,6 +19,8 @@ environments:
     values:
       - provider:
           name: none
+  amd: &AND
+    <<: *I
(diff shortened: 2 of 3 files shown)


guides/inference-scheduling/ms-inference-scheduling/values.yaml
```diff
@@ -24,23 +28,11 @@ decode:
...
   - name: "vllm"
-    image: ghcr.io/llm-d/llm-d-cuda:v0.3.1
+    image: "rocm/vllm-dev:nightly"
     modelCommand: vllmServe
     args:
-      - "--kv-transfer-config"
-      - '{"kv_connector":"NixlConnector", "kv_role":"kv_both"}'
       - "--disable-uvicorn-access-log"
     env:
-      - name: CUDA_VISIBLE_DEVICES
```
robertgshaw2-redhat, line 34: why do we set this in the base config?

guides/inference-scheduling/ms-inference-scheduling/values.yaml
```diff
@@ -24,23 +28,11 @@ decode:
       interval: "30s"
   containers:
   - name: "vllm"
-    image: ghcr.io/llm-d/llm-d-cuda:v0.3.1
+    image: "rocm/vllm-dev:nightly"
     modelCommand: vllmServe
     args:
-      - "--kv-transfer-config"
```
robertgshaw2-redhat, line 30: why do we set this in the baseconfig?

guides/inference-scheduling/ms-inference-scheduling/values.yaml
Gregory-Pereira: CC @tlrmchlsmth, I thought there was some bug that we needed to explicitly set CUDA_VISIBLE_DEVICES to 0, am I mistaken?

guides/inference-scheduling/ms-inference-scheduling/values.yaml
Gregory-Pereira: Not certain - I think we needed to set this back in vllm 0.10.0 but maybe we no longer need to

PR #557, opened 2025-12-18
Resolve Grafana deployment failure and enable automatic dashboard loading

petecheslock:
When running the monitoring stack installation script, the grafana pod enters a `CrashLoopBackOff` state with the error:
```
Datasource provisioning error: datasource.yaml config is invalid. 
Only one datasource per organization can be marked as default
```

### Root Cause
The installation script creates a custom prometheus datasource configuration, but **i think** the `kube-prometheus-stack` helm chart (v80+) is also creating its own default datasource automatically. Both datasources are marked as `isDefault: true`, causing grafana to fail during startup.

I think the helm chart changed its behavior between versions:

- **Earlier versions (v62-70?)**: When you defined custom datasources via `grafana.datasources.datasources.yaml`, the chart would use only your custom configuration
- **Current version (v80.4.1)**: The chart creates its own default datasource ConfigMap **in addition to** any custom datasources you define, resulting in duplicates

### Solution
Disable the Helm chart's automatic datasource creation by adding:
```yaml
grafana:
  sidecar:
    datasources:
      defaultDatasourceEnabled: false
```

This preserves our custom datasource configuration (needed for correct service URLs and TLS support) while preventing the chart from creating its own conflicting datasource.

I've also added in functionality to auto-load the dashboards on deployment, including adding a new dashboard from @deewhyweb.

### DIFF
--- a/docs/monitoring/scripts/install-prometheus-grafana.sh
+++ b/docs/monitoring/scripts/install-prometheus-grafana.sh
@@ -528,12 +528,10 @@ EOF
 
   log_info "⏳ Waiting for Prometheus and Grafana to be ready..."
 
-  # Wait for Prometheus pod to be ready
-  # Note: The Prometheus operator creates the StatefulSet asynchronously after Helm install completes,
-  # so we wait for the pod directly. kubectl wait will wait for the resource to appear.
-  local prometheus_label="operator.prometheus.io/name=${RELEASE_NAME}-kube-prometheus-stack-prometheus"
-  log_info "⏳ Waiting for Prometheus pod to be ready..."
-  $KCMD wait --for=condition=Ready pod -l "${prometheus_label}" -n "${MONITORING_NAMESPACE}" --timeout=300s || log_info "⚠️  Prometheus pod did not become ready within timeout"
+  # Wait for Prometheus custom resource to be available
+  # The Prometheus CR is created by Helm and managed by the Prometheus operator
+  log_info "⏳ Waiting for Prometheus instance to be available..."
+  $KCMD wait --for=condition=Available prometheus "${RELEASE_NAME}-kube-prometheus-stack-prometheus" -n "${MONITORING_NAMESPACE}" --timeout=300s || log_info "⚠️  Prometheus instance did not become available within timeout"
 
   # Wait for Grafana deployment to be available
   log_info "⏳ Waiting for Grafana deployment to be available..."


docs/monitoring/scripts/install-prometheus-grafana.sh
```diff
@@ -411,6 +411,14 @@ grafana:
   service:
     type: ClusterIP
     sessionAffinity: ""
+  sidecar:
+    dashboards:
+      enabled: true
+      searchNamespace: ALL
+      label: grafana_dashboard
+      folderAnnotation: grafana_folder
```
ahadas: you can drop `label` and `folderAnnotation` as they are set with their default value (here and below)

docs/monitoring/scripts/load-llm-d-dashboards.sh
ahadas, line 1: I thought this script is only meant for testing things manually with Grafana but I've noticed it's used by few GitHub actions, and jobs were passing even though the Grafana pod kept crashing (e.g., [here](https://github.com/llm-d/llm-d/actions/runs/20384970291/job/58583931779)), wouldn't it make sense to fail when we reach a timeout on waiting for the Prometheus or Grafana pods?
```diff
- $KCMD wait --for=condition=ready pod -l app.kubernetes.io/name=grafana -n "${MONITORING_NAMESPACE}" --timeout=300s || true
+ $KCMD wait --for=condition=ready pod -l app.kubernetes.io/name=grafana -n "${MONITORING_NAMESPACE}" --timeout=300s
```

docs/monitoring/scripts/install-prometheus-grafana.sh
petecheslock: Fixed - thanks.

docs/monitoring/scripts/load-llm-d-dashboards.sh
petecheslock: Oh that's a great point, I wasn't aware we were including this in the e2e testing, I would agree that if we can't spin up the grafana stack then technically that build should fail so we can go fix it - i'm not sure what a reasonable timeout should be here though, is 300 seconds enough?

ahadas: I think 5 minutes should be enough, and I see that at least "XPU Inference Scheduling Test" is set with workflow_dispatch that enables running it against a PR/branch - I don't have permissions for doing that, @petecheslock are you able to?

and I see I commented on the wrong script, I meant to comment on `install-prometheus-grafana.sh`, but you understood me nevertheless :)

petecheslock: Great point - i'll remove both those timeouts because looking at the logs it seems like it only takes less than 1m to deploy those so if we hit the 5min mark something is up.  We would have caught this issue too months ago as I have found a few different PRs trying to fix this.

petecheslock:
@ahadas oh that's a great idea.  I think I had tried that originally but i must not have had the right conditions to make that work.  The latest commit should greatly cleanup the code and complexity, and removed a race condition i hit as well around checking the statefulset.  Let me know what you think.  

https://github.com/llm-d/llm-d/actions/runs/20786419628/job/59697089781?pr=557#step:11:75

ahadas:
@petecheslock yep, it indeed looks better this way!

for your consideration: it seems like we can further simplify the code (a bit) by replacing the check of prometheus pod(s) with checking the 'Available' condition of the `Prometheus` instance named "llmd-kube-prometheus-stack-prometheus" (with something like: kubectl  -n llm-d-monitoring wait --for=condition=Available prometheus llmd-kube-prometheus-stack-prometheus)

other than that, lgtm

sallyom (approved):
thanks, @petecheslock !
LGTM

PR #558, opened 2025-12-19
update vllm to v0.13; swap to upstream vllm

wseaton:
All of the CUDA cherrypicks for 0.4 have been merged to main, also aggressively bumping the other images. We can asses failure case by case.


Gregory-Pereira (dismissed):
/lgtm

Gregory-Pereira:
Image built successfully, just added QOL fix

Issue #551, opened 2025-12-16
Add benchmarking section to the inference-scheduling guide


ahg-g:
Add benchmarking section to the inference-scheduling guide similar to what we have for the tiered prefix aware scheduling path, and add more details on the default scorers enabled in this path.

diegocastanibm:
Addressed in benchmark

Issue #487, opened 2025-11-24
[RFC] Benchmarking Extensions for Well-Lit Path Guides


vMaroon:
## Motivation

llm-d already has a solid set of well-honed, self-contained “well-lit paths” that make deploying the stack simple and repeatable. We also have llm-d-benchmark, which offers a comprehensive set of tools and scenarios for benchmarking llm-d.

To keep benchmarking and the reproduction of published results aligned with the standards set by our guides, we should add dedicated guidance on running benchmarks in a way that’s both easy and reliable.

## Proposal

Our existing guides work well: each path is self-contained and reproducible. The proposal is to extend every path with a benchmarking section that captures the path’s core benchmark scenario and shows how to reproduce it end-to-end in simple steps.

## Timeline
- TBD, ideally targeting v0.5.0

Authors: Maroon, Marcio

effi-ofer:
Do we also want to include a section or running benchmarking of vllm independent of the full llm-d?

deanlorenz:
This also Relates to: #517 (Reproducible benchmarking - All blogs, guides, and features come with "one line" reproducers).

/assign @deanlorenz @dmitripikus

We are writing a bunch of PR to address this. 

- Base support is in #559 -- this adds a simple benchmark script that is downloadable from [llm-d-benchmark](https://github.com/llm-d/llm-d-benchmark) with no need to clone the `llm-d-benchmark` repo. The benchmark uses a single, simple to edit, `.yaml` file that defines the endpoint, harness, and workload. 

- It is not entirely a "one line" solution, as following the guide may create slightly different `llm-d` stacks. That is, the user must still modify the `.yaml` file to include the correct namespace, endpoint URL, and results PVC. We use `envsubst` and shell snippets to set these up from per-well-lit-path template files. Everything else (e.g., the model name) is hardcoded in each template, corresponding to the well-lit path. There is one `guides/benchmark/README.md` to explain how to set everything. 

- Next PRs (#585, #586, ...) add templates for specific well-lit paths + add benchmarking examples into the README files of the corresponding guides.


## Open points:

1.  The stacks created by the llm-d guides mostly use small models -- it is practically impossible to demonstrate, through benchmarking, that each well-lit feature has any advantage.  For now, we only include the expected results from running the benchmark. 

[...]

PR #559, opened 2025-12-22
Simple benchmarking guide, towards benchmarking well-lit paths

deanlorenz:
connects to #487 #551 

- README markdown file and templates for running benchmark workloads against a stack created by following the guides.
- Utilizes a single benchmarking script from https://github.com/llm-d/llm-d-benchmark that can be downloaded (no need to clone the `llm-d-benchmark` repo).
- The benchmark runs using a single `config.yaml` configuration file. All needed info is explicitly provided in this file, and a user would be able to customize the contents.
- This should work with any inference serving endpoint, including a single vLLM. 
- Then user would easily create the `config.yaml` file from a template that matches the well-lit path used. The template would already be pre-filled with parameters that match the guide.
  - A template for _Inference Scheduling_ with a simple sanity-check [inference-perf](https://github.com/kubernetes-sigs/inference-perf) workload is included. (tested end-to-end)
  - More example templates can be added to show case each well-lit path.
  - NOTE: The included template files for _P/D Disaggregation_ and _Wide EP_ are placeholders to demonstrate how these templates may be referenced from the README file.

### DIFF
--- a/guides/benchmark/README.md
+++ b/guides/benchmark/README.md
@@ -88,6 +88,8 @@ For full, customizable benchmarking, please refer to [llm-d-benchmark](https://g
 >     | head -1
 >   )
 >   export BENCHMARK_TEMPLATE="${BENCH_TEMPLATE_DIR}"/pd_template.yaml
+>   #export BENCHMARK_TEMPLATE="${BENCH_TEMPLATE_DIR}"/pd_vllm_bench_random_concurrent_template.yaml
+>   #export BENCHMARK_TEMPLATE="${BENCH_TEMPLATE_DIR}"/pd_shared_prefix_template.yaml
 >   ```
 > 
 > </details>
@@ -117,6 +119,25 @@ For full, customizable benchmarking, please refer to [llm-d-benchmark](https://g
 > TBD
 > 
 > </details>
+> </tr>
+> <tr>
+> <td>
+> <details>
+> <summary><b>Precise Prefix Caching</b></summary>
+>
+>   ```bash
+>   export GATEWAY_SVC=$(kubectl get svc -n "${NAMESPACE}" \
+>     -l gateway.networking.k8s.io/gateway-name=infra-kv-events-inference-gateway \
+>     --no-headers  -o=custom-columns=:metadata.name \
+>     | head -1
+>   )
+>   export BENCHMARK_TEMPLATE="${BENCH_TEMPLATE_DIR}"/precise_template.yaml
+>   # export BENCHMARK_TEMPLATE="${BENCH_TEMPLATE_DIR}"/precise_guidellm_template.yaml
+>   # export BENCHMARK_TEMPLATE="${BENCH_TEMPLATE_DIR}"/precise_shared_prefix_template.yaml
+>   ```
+>
+> </details>
+> </td>
 > </tr>
 > </table>
 


guides/benchmark/README.md
```diff
@@ -0,0 +1,935 @@
...
+
+</details>
+
+### Output folder
+
+The output files are saved on the benchmark PVC. They are accessible through the launcher pod in the `/requests` folder. Each experiment is saved under its own sub directory. 
+
+This example uses `inference-perf` with a [`shared-prefix`](./inference_scheduling_shared_prefix_template.yaml) workload:
+  ```bash
+  export BENCHMARK_TEMPLATE="${BENCH_TEMPLATE_DIR}"/inference_scheduling_guidellm_template.yaml
```
michalmalka: should the yaml name ve inference-perf and not guidellm?

guides/benchmark/README.md
```diff
@@ -0,0 +1,935 @@
...
+
+You can access the results PVC through the benchmark launcher pod.
+  ```bash
+  export HARNESS_POD=$(kubectl get pods -n ${NAMESPACE} -l app --show-labels | awk -v p='lmdbench-.*-launcher' '$0~p {print $1; exit}')
+  kubectl exec $HARNESS_POD -n $NAMESPACE -- ls /requests
+  ```
+
+To copy a results directory to your local machine use:
+  ```bash
+  kubectl cp ${NAMESPACE}/${HARNESS_POD}:/requests/<results-folder> <destination-path>
```
michalmalka, line 170: Need to explain better what should be the results-folder (or point to the explanation on it... )

docs/getting-started-inferencing.md
```diff
@@ -2,6 +2,9 @@
 
 This document show you how to interact with the model server and inference scheduler.
 
+> [!TIP] 
+> To run a performance test against the llm-d stack checkout our [benchmark doc](../guides/benchmark/README.md).
```
michalmalka, line 6: this link should be at the well-lit path document

guides/benchmark/README.md
dmitripikus: Fixed. Thanks!

docs/getting-started-inferencing.md
dmitripikus: The link is added to well-lit path doc.
Thanks!

guides/benchmark/README.md
dmitripikus: We are working on additional PR that will contain several sets of benchmarking results - one set per specific well-lit path configuration, with better explanation of the results

deanlorenz: > Need to explain better what should be the results-folder (or point to the explanation on it... )

Just to add to @dmitripikus,

- We want to have at least one example per well-lit path, with results and the exact parameters needed to repeat them.
- The README already includes an explanation of the expected results in the # Results Examples (note the collapsed sections). Do we want to have a longer explanation than is already there? I feel that if we do, we should either refer to the `llm-d-benchmark` detailed READMEs or split into a more detailed results README

deanlorenz:
> This is a great!
> 
> A few questions / suggestions
> 
> 1. Does it make sense to have a templates subdirectory under guides/benchmark. Future blog posts may enter a new template like `guides/benchmark/templates/EXAMPLE-BLOG/` with a readme back-referencing the post / any relevant benchmark or template information and the template.yaml. This readme may also link to stored results from the post?
> 2. Are we able to add an example fully no-clone, referencing a template via it's github link?
> 3. (Perhaps future enhancement) A remote storage (bucket) option instead of requiring a pvc.

Agree on point 1. The motivation for the self-contained templates is to be able to reproduce a specific scenario, like the blogs. Ideally, we should have a template linked to each blog. As a first step, we are working on an example in each well-lit-path README that provides sample results and the templates used to create them.

PR #568, opened 2026-01-02
chore: pre-commint/lint and github workflow for build

zdtsw:
- Update pre-commit tool version and docs
- fix lint found by yaml,markdown,hadolint
- threat model link is broken, it has been removed recently
- use lychee.toml for github and markdownlint-cli for local, add config files
- remove .markdown-lint-check and its config file
- set all github script to use pipefail
- remove strategy which is not used for xpu and cpu
- add more path filters to trigger build
- use github action build-push-action than make target for xpu and cpu build
- create new stage 'set-tag' to capture tags on docker image
- remove sccache.config.toml, which is not in use

### DIFF
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -276,7 +276,7 @@ jobs:
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@master
         with:
-          image-ref: ${{ env.REGISTRY }}/${{ github.repository }}-aws:${{ steps.build-and-push.outputs.digest }}
+          image-ref: ${{ env.REGISTRY }}/${{ github.repository }}-aws@${{ steps.build-and-push.outputs.digest }}
           format: 'sarif'
           output: 'trivy-results.sarif'
           severity: 'CRITICAL,HIGH,MEDIUM'


zdtsw:
> Thanks @zdtsw for the contribution. I allowed the workflows to run so you can test them out in this PR - also one note that there are a couple other broken links, if you want to just cherry-pick the commit ([f0c603b](https://github.com/llm-d/llm-d/commit/f0c603b72312dfd6a1d90d34a194da9606963971)) in this PR (#569) over to get the other ones i can close that separate PR.

thanks @petecheslock! 
sure i can get your f0c603b into this PR

petecheslock:
@zdtsw just one small conflict because that file got fixed in a separate PR if you can fix that - also @Gregory-Pereira any thoughts?

zdtsw:
> @zdtsw just one small conflict because that file got fixed in a separate PR if you can fix that - also @Gregory-Pereira any thoughts?

sure thanks! saw the change from Sally will do a rebase soon

zdtsw:
@petecheslock if you konw anyone else can help review this PR as well?

.github/workflows/ci-release.yaml
```diff
@@ -278,7 +276,7 @@ jobs:
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@master
         with:
-          image-ref: ${{ env.REGISTRY }}/${{ github.repository }}-aws@${{ steps.build-and-push.outputs.digest }}
+          image-ref: ${{ env.REGISTRY }}/${{ github.repository }}-aws:${{ steps.build-and-push.outputs.digest }}
```
petecheslock: @zdtsw Ok - I think this might be the only issue with this PR - if the `steps.build-and-push.outputs.digest` is an actual digest then this should use the `@` - the last time this ran it used the `@` for the digest since its not a tag/branch. https://github.com/llm-d/llm-d/actions/runs/19716372634/job/56489401636#step:7:496. Possible this was just copy/pasted from one of the other jobs you updated?

zdtsw: You are absolutely right! 
It must be introduced when i copy-pasted from somethere else, let me fix it

PR #572, opened 2026-01-04
[infra] add Python output suppression flag for vLLM installation

amito:


docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -102,7 +102,11 @@ echo "DEBUG: Installing packages: ${INSTALL_PACKAGES[*]}"
 # install all packages in one command with verbose output to prevent GHA timeouts
 # use flashinfer wheel index for jit-cache pre-built binaries
 CUDA_SHORT_VERSION="cu${CUDA_MAJOR}${CUDA_MINOR}"
-uv pip install -v "${INSTALL_PACKAGES[@]}" \
+VERBOSE_FLAG=""
+if [ -z "${SUPPRESS_PYTHON_OUTPUT:-}" ]; then
```
zdtsw: does it mean, any value  set for `SUPPRESS_PYTHON_OUTPUT` to  "docker build"  will all lead to "-v" get skipped?
e.g 1, 0, true, false, foobar

amito: Yep.
What do we want the behavior to be when someone passes a bogus value (_e.g._, `SUPPRESS_PYTHON_OUTPUT=bla`)? The flag is still there, the value is rubbish. Do we want verbosity anyway, in this case, and just support `0`, `1`, `true` and `false`?
Thanks

Makefile
```diff
@@ -110,7 +110,9 @@ buildah-build: check-builder ## Build and push image (multi-arch if supported)
 .PHONY:	image-build
 image-build: check-container-tool ## Build Docker image using $(CONTAINER_TOOL)
 	@printf "\033[33;1m==== Building Docker image $(IMG) ====\033[0m\n"
```
zdtsw, line 115: should we add
`SUPPRESS_PYTHON_OUTPUT ?=`

Issue #521, opened 2025-12-02
[Feat] Sglang support for well-lit path of Prefill/Decode Disaggregation


ezrasilvera:
Add support for sglang inference server for https://github.com/llm-d/llm-d/blob/main/guides/pd-disaggregation/README.md

PR #580, opened 2026-01-08
Add SGLang option to PD disaggregation well-lit path

andreyod:
This PR will add an option to deploy SGLang as the inference server instead of vLLM.

This option is added in the "P/D disaggregation" well-lit path with the default hardware (i.e., GPU hardware). NIXL is used as the data transfer backend.

Related issue: https://github.com/llm-d/llm-d/issues/521

### DIFF
--- a/guides/pd-disaggregation/gaie-pd/values_sglang.yaml
+++ b/guides/pd-disaggregation/gaie-pd/values_sglang.yaml
@@ -3,7 +3,7 @@ inferenceExtension:
   image:
     name: llm-d-inference-scheduler
     hub: ghcr.io/llm-d
-    tag: v0.5.0
+    tag: v0.6.0
     pullPolicy: Always
   extProcPort: 9002
   pluginsConfigFile: "pd-config.yaml"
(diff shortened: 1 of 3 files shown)


guides/pd-disaggregation/gaie-pd/values_sglang.yaml
```diff
@@ -0,0 +1,65 @@
+inferenceExtension:
+  replicas: 1
+  image:
+    name: llm-d-inference-scheduler
+    hub: ghcr.io/llm-d
+    tag: v0.4.0
```
rahulgurnani: please use latest version

ezrasilvera:
> > And added the RDMA/IB device but no logs to indicate thats working
> 
> Do the logs indicate that TCP fallback is still happening? Maybe try to increase the log level also for UCX: `export UCX_LOG_LEVEL=debug` Besides the logs, try to inspect the IB port counters; these counters should increase during KV transfer.
> 
> ```
> sudo cat /sys/class/infiniband/mlx5_0/ports/1/counters/port_xmit_data
> sudo cat /sys/class/infiniband/mlx5_0/ports/1/counters/port_rcv_data
> ```

@Gregory-Pereira (cc: @andreyod ) If everything else is working besides RDMA over IB, can we just merge it with the TCP and follow up with a fix later on for IB? 
Note, that in any case this PR is defined as an initial enablement and we don't claim that this is the optimal configuration for best performance using SGLang (that's true for also for all other parameters)

rahulgurnani:
Hi @chcost 

Let get this change in. I discussed this offline and we have an RDMA config for GKE that works and is a good starting point who wants to deploy PD on GKE. I can check those changes in once this PR gets in. Thanks!

Gregory-Pereira (approved):
/lgtm

PR #594, opened 2026-01-14
feat: add support for weka

zdtsw:
# Description
- add support to deploy weka backed storage by
1. ship a new probe script into final runtime image to do check on models loading and cufile.json existence
2. create another initConatiner in deployment to ensure nvidia models are loaded --this uses the same llm-d image which should not be wasting time to pull in to the cluster 
3. install extra packages during image build time
4. install new runtime package:  nvdia-gds(core which includes nvidia_fs gds-tools, libcufile etc) libibverbs libffi iproute infiniband-diags (for debug)
- add new guide for weka under "tiered-prefix-cache"
- use kustomize to deploy resources: Gateway, HTTPRoute, InferencePool, Deployment for decode and prefill
- support two overlays: host and pvc which require customization value.
- keep using Helm for InferencePool, DestinationRule and inference-scheduler deployment.
- update for the existing document

ref: https://issues.redhat.com/browse/INFERENG-3986

# Changes
creds to @Gregory-Pereira , cherry-pick his Draft:
- https://github.com/llm-d/llm-d/pull/208
- https://github.com/llm-d/llm-d/pull/246

# Note
this is a dupe of old PR https://github.com/llm-d/llm-d/pull/588

### DIFF
--- a/docker/Dockerfile.cuda
+++ b/docker/Dockerfile.cuda
@@ -46,8 +46,8 @@ ARG NIXL_VERSION="v1.0.0"
 ARG INFINISTORE_REPO="https://github.com/bytedance/InfiniStore.git"
 ARG INFINISTORE_VERSION="0.2.33"
 
-ARG LMCACHE_REPO="https://github.com/LMCache/LMCache.git"
-ARG LMCACHE_VERSION="v0.4.1"
+ARG LMCACHE_REPO="https://github.com/vinaychannagiri23/LMCache.git"
+ARG LMCACHE_VERSION="operation-timeout-error-handling"
 
 # Kernels
 # NM fork of Deepep for duplicate nvshmem IBGDA definitions, see: https://github.com/deepseek-ai/DeepEP/pull/581.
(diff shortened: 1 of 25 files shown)


kfirtoledo:
@zdtsw, I think this guide should be under a tiered prefix cache guide, similar to the llm-d FS backend guide:
https://github.com/llm-d/llm-d/pull/609.
With that, you could also remove some of the redundant YAMLs related to the llm-d deployment in general (for example, the Gateway).

Gregory-Pereira:
> @Gregory-Pereira do you recall why we need [set service to ClusterIP](https://github.com/llm-d/llm-d/blob/a6d70805c3681197e76316a056913beb1bfcbd80/guides/weka/manifests/gateway/istio/gateway.yaml#L8) for Gateway?  Or if we can skip it and keep using [LoadBalancer](https://github.com/llm-d/llm-d/blob/a6d70805c3681197e76316a056913beb1bfcbd80/guides/weka/manifests/gateway/istio/configmap.yaml#L33)

Load balancer is not supported on every kubernetes environment, it depends on the platform. For this reason we made it cluster ip and port forward the service but we have gone back and forth on this. Either one is fine

PR #596, opened 2026-01-14
PredictedLatency support for PD

RishabhSaini:
Blocked on: 
1) https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2145
2) https://github.com/llm-d/llm-d-inference-scheduler/pull/564


guides/pd-disaggregation/gaie-pd/values-slo.yaml
```diff
@@ -0,0 +1,177 @@
+# This file demonstrates how to enable SLO prediction with prefill-decode disaggregation
```
ahg-g: pls use the term predicted-latency instead of slo, the file name as well

PR #598, opened 2026-01-15
[DO NOT MERGE] POC create kustomize patches with helm

liu-cong:
This POC demonstrates using helm to generate parameterized patches on top of a Kustomize based well-lit path guide. Benchmark tool can easily parameterize the config via overriding the values. It's very flexible as we don't need to publish the chart, and only use helm as a templating tool.

The code is generated in Antigravity with Gemini.

This looks pretty promising. Let me know your thoughts.   @maugustosilva @smarterclayton @Gregory-Pereira @seanhorgan


guides/wide-ep-lws/sweeps/charts/base/kustomization.yaml
```diff
@@ -0,0 +1,110 @@
...
+        path: /metadata/name
+        value: {{ .Values.name }}-prefill
+      - op: replace
+        path: /spec/replicas
+        value: {{ .Values.replicas }}
+      - op: replace
+        path: /spec/leaderWorkerTemplate/size
+        value: {{ .Values.lwsSize }}
+  
+  # Env Var Patch (Strategic Merge)
```
smarterclayton, line 25: Did you encounter any issues with strategic merge patch?  I was previously having issues with it and LWS.

liu-cong: Actually strategic merge doesn't work for lists - it simply replaces the entire list. Removed this example

liu-cong:
@maugustosilva Thanks for looking. Directly using a helm chart for benchmarking is another option. However using helm in user facing guides creates a lot of challenges, this [doc](https://docs.google.com/document/d/1HdPPcs1HamKHFt6eqzVJDpp9W8_U12GitCG85EJTW_U/edit?resourcekey=0-5QY-K4EfLmQKrCWUflIeZg&tab=t.0#bookmark=id.rumotquw942u) captured a few. So I am pushing for migrating user guides to kustomize based, composable patches. And the motivation of this PR is to prove that we can lay a helm layer on top of the kustomize based user guides for advanced sweeps.

PR #599, opened 2026-01-15
refactor: build + renaming files

zdtsw:
- rename bild-constraints.txt to cuda-build-constrants.txt
- rename contraints.txt to cuda-constrants.txt
- move setup-sccache.sh into cuda/builder as it is only for cuda with sccache
- move war-vllm-precompiled.sh into cuda/runtime as it only need to be in runtime image


docker/cuda-build-constraints.txt
Gregory-Pereira: Build constraints is not specific to cuda but vLLM, so it should go in the common direcotry i think

Gregory-Pereira (dismissed):
/lgtm

PR #610, opened 2026-01-18
Benchmarking for inference scheduling scenario

dmitripikus:
Benchmarking for inference scheduling scenario

### DIFF
--- a/guides/inference-scheduling/README.md
+++ b/guides/inference-scheduling/README.md
@@ -154,10 +154,11 @@ kubectl apply -f httproute.yaml -n ${NAMESPACE}
 
 ```bash
 helm list -n ${NAMESPACE}
-NAME                        NAMESPACE                 REVISION  UPDATED                               STATUS    CHART                     APP VERSION
-gaie-inference-scheduling   llm-d-inference-scheduler 1         2025-08-24 11:24:53.231918 -0700 PDT  deployed  inferencepool-v1.2.0-rc.1 v1.2.0-rc.1
-infra-inference-scheduling  llm-d-inference-scheduler 1         2025-08-24 11:24:49.551591 -0700 PDT  deployed  llm-d-infra-v1.3.4        v0.3.0
-ms-inference-scheduling     llm-d-inference-scheduler 1         2025-08-24 11:24:58.360173 -0700 PDT  deployed  llm-d-modelservice-v0.3.8 v0.3.0
+
+NAME                            NAMESPACE                      REVISION        UPDATED                                 STATUS          CHART                           APP VERSION
+gaie-inference-scheduling       llm-d-inference-scheduler      1               2026-01-25 13:33:26.952352635 +0200 IST deployed        inferencepool-v1.2.1            v1.2.1 
+infra-inference-scheduling      llm-d-inference-scheduler      1               2026-01-25 13:33:19.761480371 +0200 IST deployed        llm-d-infra-v1.3.5              v0.3.0 
+ms-inference-scheduling         llm-d-inference-scheduler      1               2026-01-25 13:33:39.984445259 +0200 IST deployed        llm-d-modelservice-v0.3.8       v0.3.0
 ```
 
 - Out of the box with this example you should have the following resources:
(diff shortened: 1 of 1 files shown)


guides/inference-scheduling/README.md
```diff
@@ -269,3 +269,291 @@ kubectl delete -f httproute.yaml -n ${NAMESPACE}
...
 
 For information on customizing a guide and tips to build your own, see [our docs](../../docs/customizing-a-guide.md)
+
+## Benchmarking
+
+To run benchmarks against the installed llm-d stack, you need [run_only.sh](https://github.com/llm-d/llm-d-benchmark/blob/main/existing_stack/run_only.sh), a template file from [guides/benchmark](../benchmark/), and a Persistent Volume Claim (PVC) to store the results. Follow the instructions in the [benchmark doc](../benchmark/README.md). 
+
+### Example
+
+This example uses [run_only.sh](https://github.com/llm-d/llm-d-benchmark/blob/main/existing_stack/run_only.sh) with the template [precise_guidellm_template.yaml](../benchmark/inference_scheduling_shared_prefix_template.yaml).
```
zetxqx, line 238: The benchmark workload here is using https://github.com/llm-d/llm-d/blob/5de6cb7b1e329d084098713dfa22b777adee619f/guides/benchmark/precise_shared_prefix_template.yaml#L61-L67 . And the benchmark is based on Qwen0.6B. 

Do we want to 
1. use a workload similar to `[Scenario C](https://docs.google.com/document/d/1vppgD4xCQepdnkEDlAlYzrK18Ar34gTdIEeE2i612OE/edit?resourcekey=0-7AOx_vzjcB6Mi-gfVhRrfg&tab=t.0#heading=h.ue2q2knfpgrf)` 
2. use a larger model and compared with k8s service as the baseline

@BenjaminBraunDev @kaushikmitr

guides/inference-scheduling/README.md
diegocastanibm: @Gregory-Pereira I do not see the `LoadBalancer` when I follow this `README.md` file out-of-the-box. Am I missing anything?

guides/inference-scheduling/README.md
```diff
@@ -269,3 +269,291 @@ kubectl delete -f httproute.yaml -n ${NAMESPACE}
 ## Customization
 
 For information on customizing a guide and tips to build your own, see [our docs](../../docs/customizing-a-guide.md)
+
+## Benchmarking
```
diegocastanibm, line 232: I would place this whole `Benchmarking` section between `Using the stack` and `Cleanup`. It does not make sense to me to run a benchmark after cleaning up the namespace

guides/inference-scheduling/README.md
diegocastanibm, line 238: Since this `README.md` corresponds to `guides/inference-scheduling`, wouldn't make sense to run the example with the `inference_scheduling_template.yaml`?

diegocastanibm (commented):
Thank you for taking the time to add benchmark information into the `README` file. Some comments below

guides/inference-scheduling/README.md
BenjaminBraunDev: I think we should use the same model used in the benchmark and compare to k8s baseline, as well as non-latency prediction setup. I think both of those datapoints are useful, though the comparison with non-latency prediction IGW is most important.

guides/inference-scheduling/README.md
Gregory-Pereira: We used to have the default value for the service type of gateways be LoadBalancer. Since this is dependent on what K8s provider you have and not all support loadbalancer integration, we dropped it for consistency. This indicates the guide readme is outdated

guides/inference-scheduling/README.md
deanlorenz: This PR is based on the current stack configuration (`values.yaml`), which is a small `Qwen3-.6B` model. This is useful for people who want to try `llm-d` or want to test an end-to-end scenario.

We are working on a new PR with a more realistic stack configuration, using `Qwen3-32B` and 20 GPUs. We are testing it with a workload similar to scenario C (without multi-round, which seems broken in `inference-perf`). 

The guide can keep both options, with either of these being the default. 

Comparison to k8s baseline is relevant only for the more realistic scenario.

deanlorenz: We have several sample templates. `inference_scheduling_template.yaml` is a sanity check, running a toy, single point workload. The `inference_scheduling_shared_prefix_template.yaml` template runs a few stages. 

From a naming PoV, we chose the sanity one to be the default, but it does not have very interesting output. We can rename/add templates as needed (no strong opinion on this). The bash snippet uses `select` to let the user choose from all available ones. We could/should move the templates under `inference-scheduling/templates` folder and have shorter template names (e.g., `sanity.yaml`, `shared_perfix.yaml`).

guides/inference-scheduling/README.md
deanlorenz: Updated expected resources (small model).

guides/inference-scheduling/README.md
deanlorenz: DONE.

guides/inference-scheduling/README.md
```diff
@@ -228,6 +229,294 @@ replicaset.apps/ms-inference-scheduling-llm-d-modelservice-decode-8ff7fd5b8    2
...
+            summary: true
+          request_lifecycle:
+            per_request: true
+            per_stage: true
+            summary: true
+        server:
+          api_key: null
+          base_url: http://infra-inference-scheduling-inference-gateway-istio.dpikus-intel-inf.svc.cluster.local:80
+          ignore_eos: true
+          model_name: Qwen/Qwen3-0.6B
```
Gregory-Pereira, line 500: Upgrading to qwen 32b:

```suggestion
          model_name: Qwen/Qwen3-32B
```

guides/inference-scheduling/README.md
```diff
@@ -228,6 +229,294 @@ replicaset.apps/ms-inference-scheduling-llm-d-modelservice-decode-8ff7fd5b8    2
...
+          type: vllm
+        storage:
+          google_cloud_storage: null
+          local_storage:
+            path: /requests/inference-perf_1768395442_shared_prefix_synthetic_inference-scheduling-Qwen3-0.6B
+            report_file_prefix: null
+          simple_storage_service: null
+        tokenizer:
+          pretrained_model_name_or_path: Qwen/Qwen3-0.6B
```
Gregory-Pereira, line 509: Same as above:

```suggestion
          pretrained_model_name_or_path: Qwen/Qwen3-32B
```

guides/inference-scheduling/README.md
ahg-g: just to make sure we are aligned here, the idea is to move away from Qwen3-.6B model completely and focus on the 32B one. The guide already reflects that (https://github.com/llm-d/llm-d/pull/619 made the change). The benchmarking section is expected to show results for the 32B model only.

Gregory-Pereira: This is also my understanding

deanlorenz: New PR #638 updates this to Qwen3-32B.
We may want to keep this one for a small model guide to be added later.

PR #611, opened 2026-01-19
build(cpu): optimize docker build

zdtsw:
# Changes
- remove duplicated apt-get packages
- reduce final image layers and intermediate build stages
- change to use ENTRYPOINT than CMD to align with cuda image
~~- include fix in https://github.com/llm-d/llm-d/pull/612~~

# Breaking changes
take llama 3.2 as example
- pre this PR with `CMD ["/bin/bash"] `
>docker run -it ghcr.io/llm-d/llm-d-cpu-dev:old python -m vllm.entrypoints.openai.api_server --model meta-llama/Llama-3.2-3B
- after this PR with `ENTRYPOINT ["python", "-m", "vllm.entrypoints.openai.api_server"]`
>docker run -it ghcr.io/llm-d/llm-d-cpu-dev:new --model meta-llama/Llama-3.2-3B

# Build
```
>make image-build DEVICE=cpu VERSION=revert-back-22.04 BUILD_TYPE=dev PLATFORMS=linux/amd64
==== Building Docker image quay.io/wenzhou/llm-d/llm-d-cpu-dev:revert-back-22.04 ====
podman build --progress=plain --platform linux/amd64 -t quay.io/wenzhou/llm-d/llm-d-cpu-dev:revert-back-22.04 -f docker/Dockerfile.cpu .
...
```

# PR Test
*updated test from commit 7ee5c9dc9161c07eaac01cad13f9ae584341532e *

```
>podman run -d --rm -p 18080:8000 --ipc=host "ghcr.io/llm-d/llm-d-cpu-dev:sha-7ee5c9d" --model facebook/opt-125m --max-model-len 2048
>curl -s http://127.0.0.1:18080/v1/completions -H "Content-Type: application/json" \
        -d '{
            "model": "facebook/opt-125m",
            "prompt": "Stockholm is a",
            "max_tokens": 100
        }' | python3 -m json.tool
{
    "id": "cmpl-af7ca63766c46f39",
    "object": "text_completion",
    "created": 1769077833,
    "model": "facebook/opt-125m",
    "choices": [
        {
            "index": 0,
            "text": " rocket ship. If they wipe it down she'll be her wallpaper\ni dont know if it's Swedish, if it was I wouldn't be around anymore\nIt's Norwegian for shits and giggles.",
            "logprobs": null,
            "finish_reason": "stop",
            "stop_reason": null,
            "token_ids": null,
            "prompt_logprobs": null,
            "prompt_token_ids": null
        }
    ],
    "service_tier": null,
[...]

PR #612, opened 2026-01-19
fix(GHA): ignore lint for markdown file with line number + build for XPU and CPU with "docker/build-push-action"

zdtsw:
# Description
When made change in [PR](https://github.com/llm-d/llm-d/pull/568), there is a bug which blocked build image for XPU and CPU. This was found when test [PR ](https://github.com/llm-d/llm-d/pull/611)
see: https://github.com/llm-d/llm-d/actions/runs/21134003416/job/60771639084
and lint is failing because of file line number is used
see: https://github.com/llm-d/llm-d/actions/runs/21133599789/job/60770296742

# Changes
This PR is to address above two issues, by
- fix github action for both build-images and ci-release
- ignore checks on the line in lychee


Gregory-Pereira (approved):
Good catch,
/lgtm

PR #613, opened 2026-01-19
Fixing wide ep decode port name

Edwinhr716:
Current Decode yaml in the Wide-EP yaml is broken since the port name is duplicate. Changed to match the port name in the prefill yaml


Edwinhr716:
@Gregory-Pereira Can you take a look?

Gregory-Pereira (approved):
Good catch, thank you
/lgtm

Issue #615, opened 2026-01-20
[Bug] WideEP does not create DestinationRule


zdtsw:
# What happend

On a OCP cluster, following guide https://github.com/llm-d/llm-d/tree/main/guides/wide-ep-lws#istio, it does not create DestinationRule.
InferencePool and other resources are created.

# Version
0.4.0

 # Step to reproduce
```
>helm uninstall llm-d-infpool   -n llm-d-wide-ep   -f ./manifests/inferencepool.values.yaml   --set "provider.name=istio"   --set "inferenceExtension.monitoring.prometheus.enable=true"   oci://us-central1-docker.pkg.dev/k8s-staging-images/gateway-api-inference-extension/charts/inferencepool   --version v1.2.0
>k get dr -n llm-d-wide-ep
No resources found in llm-d-wide-ep namespace.
```

PR #616, opened 2026-01-20
fix: DestinationRule creation from Helm for wideEP

zdtsw:
# Description
Fix https://github.com/llm-d/llm-d/issues/615

# Changes 
- wrong nested providers for istio to enable creation
- fix wrong host value, there is no deepseek-r1-epp, the name is generated by helm release name or can leave it unset to generate FQDN

# Test
`>helm upgrade llm-d-infpool   -n llm-d-wide-ep   -f ./manifests/inferencepool.values.yaml   --set "provider.name=istio"   --set "inferenceExtension.monitoring.prometheus.enable=true"   oci://us-central1-docker.pkg.dev/k8s-staging-images/gateway-api-inference-extension/charts/inferencepool   --version v1.2.0`
```
>k get dr -n llm-d-wide-ep
NAME                HOST                AGE
llm-d-infpool-epp   llm-d-infpool-epp   40m
>k get svc -n llm-d-wide-ep
NAME                TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)             AGE
llm-d-infpool-epp   ClusterIP   172.30.216.230   <none>        9002/TCP,9090/TCP   40m
```

### DIFF
--- a/guides/wide-ep-lws/manifests/inferencepool.values.yaml
+++ b/guides/wide-ep-lws/manifests/inferencepool.values.yaml
@@ -24,7 +24,7 @@ inferenceExtension:
   pluginsCustomConfig:
     custom-plugins.yaml: |
       # ALWAYS DO PD IN THIS EXAMPLE (THRESHOLD 0)
-      # This example uses random routing 
+      # This example uses random routing
       # since it's not yet possible to route to individual DP ranks
       apiVersion: inference.networking.x-k8s.io/v1alpha1
       kind: EndpointPickerConfig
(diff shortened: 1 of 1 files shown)


Gregory-Pereira (dismissed):
/lgtm

guides/wide-ep-lws/manifests/inferencepool.values.yaml
```diff
@@ -54,21 +54,23 @@ inferencePool:
...
-            tcp:
-              maxConnections: 256000
-              maxConnectionDuration: "1800s"
-              connectTimeout: "900s"
+
+provider:
+  name: istio
+  istio:
+    destinationRule:
+      host: llm-d-infpool-epp
```
Gregory-Pereira, line 62: Wrong host / epp service name
```suggestion
      host: deepseek-r1-epp
```

zdtsw: think the old value `deepseek-r1-epp` is not correct.
when using [llm-d-infpool](https://github.com/llm-d/llm-d/tree/main/guides/wide-ep-lws#istio) it creates infpool `"llm-d-infpool" `and the deployment+svc for scheduler is called` "llm-d-infpool-epp"`
see the "Test" part in the PR description.

Gregory-Pereira: Ah good call, you are correct.

PR #619, opened 2026-01-20
Update intellegent inference scheduling well lit path and values.yaml

BenjaminBraunDev:
This PR updates the Intelligent Inference Scheduling guide and deployment configuration to replicate the setup described in the Predicted Latency Aware Scheduling blog post.

  ## Changes
   - Deployment config (`values.yaml`):
     - Updated the model from Qwen/Qwen3-0.6B to Qwen/Qwen3-32B to match the blog post's hardware profile.
     - Scaled the deployment to 10 replicas.
     - Replica to use 2 GPUs with tensor parallelism enabled (--tensor-parallel-size=2), for 2x H100 per server
   - Updated inference scheduling `README.md`

### DIFF
--- a/guides/inference-scheduling/README.md
+++ b/guides/inference-scheduling/README.md
@@ -8,7 +8,7 @@ This profile defaults to the approximate prefix cache aware scorer, which only o
 
 ## Hardware Requirements
 
-This example out of the box uses 20 GPUs (10 replicas x 2 GPUs each) of any supported kind, though fewer can be used so long as `values.yaml` is also updated accordingly:
+This example out of the box uses 16 GPUs (8 replicas x 2 GPUs each) of any supported kind, though fewer can be used so long as `values.yaml` is also updated accordingly:
 
 - **NVIDIA GPUs**: Any NVIDIA GPU (support determined by the inferencing image used)
 - **Intel XPU/GPUs**: Intel Data Center GPU Max 1550 or compatible Intel XPU device
--- a/guides/inference-scheduling/ms-inference-scheduling/values.yaml
+++ b/guides/inference-scheduling/ms-inference-scheduling/values.yaml
@@ -16,7 +16,7 @@ accelerator: # controlls specialized hardware Type
 
 decode:
   create: true
-  replicas: 10
+  replicas: 8
   monitoring:
     podmonitor:
       enabled: true
@@ -30,19 +30,6 @@ decode:
     args:
       - "--disable-uvicorn-access-log"
       - "--tensor-parallel-size=2"
-    env:
-      - name: CUDA_VISIBLE_DEVICES
-        value: "0"
-      - name: UCX_TLS
-        value: "cuda_ipc,cuda_copy,tcp"
-      - name: VLLM_NIXL_SIDE_CHANNEL_HOST
-        valueFrom:
-          fieldRef:
-            fieldPath: status.podIP
-      - name: VLLM_NIXL_SIDE_CHANNEL_PORT
-        value: "5557"
-      - name: VLLM_LOGGING_LEVEL
-        value: INFO
     ports:
       - containerPort: 8000
         name: vllm


guides/inference-scheduling/ms-inference-scheduling/values.yaml
```diff
@@ -1,43 +1,60 @@
...
+    image: ghcr.io/llm-d/llm-d-cuda:v0.3.1
     modelCommand: vllmServe
     args:
+      - "--kv-transfer-config"
+      - '{"kv_connector":"NixlConnector", "kv_role":"kv_both"}'
       - "--disable-uvicorn-access-log"
+      - "--tensor-parallel-size=2"
+    env:
+      - name: CUDA_VISIBLE_DEVICES
+        value: "0,1"
```
Gregory-Pereira: We can remove this

vMaroon: Let's drop this from all guides +1

guides/inference-scheduling/ms-inference-scheduling/values.yaml
```diff
@@ -1,43 +1,60 @@
...
-        name: vllm
+      - containerPort: 5557
+        protocol: TCP
+      - containerPort: 8200
+        name: metrics
         protocol: TCP
     resources:
       limits:
-        nvidia.com/gpu: "1"
+        nvidia.com/gpu: "2"
```
vMaroon, line 39: Pls verify that this is effective - if I remember correctly, the modelservice helm will overwrite this unless this is added:

```
accelerator:
  # Type of accelerator to use
  type: nvidia

decode:
  parallelism:
    tensor: 2
```

I could be wrong, a confirmation of running locally is sufficient.

BenjaminBraunDev: Okay, I'll verify locally

guides/inference-scheduling/README.md
```diff
@@ -2,13 +2,13 @@
 
 ## Overview
 
-This guide deploys the recommended out of the box [scheduling configuration](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md) for most vLLM deployments, reducing tail latency and increasing throughput through load-aware and prefix-cache aware balancing. This can be run on a single GPU that can load [Qwen/Qwen3-0.6B](https://huggingface.co/Qwen/Qwen3-0.6B).
+This guide deploys the recommended out of the box [scheduling configuration](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md) for most vLLM deployments, reducing tail latency and increasing throughput through load-aware and prefix-cache aware balancing. This configuration mirrors the setup used in our Predicted Latency Aware Scheduling blog post using [Qwen/Qwen3-32B](https://huggingface.co/Qwen/Qwen3-32B).
```
vMaroon: I think this is an excessive plug. We should converge into a deployment / benchmark generally, and this is one proposal - but should not "invade" the guide.

guides/inference-scheduling/ms-inference-scheduling/values.yaml
```diff
@@ -1,43 +1,60 @@
...
-accelerator: # controlls specialized hardware Type
-  type: nvidia 
+    image: ghcr.io/llm-d/llm-d-routing-sidecar:v0.4.0-rc.1
+    connector: nixlv2
+    secure: false
 
 decode:
   create: true
-  replicas: 2
+  replicas: 10
```
vMaroon: Can we do 8 replicas? Generally on-prem users / fixed hardware users have GPU nodes that carry 8 devices each. I think multiples of 8 are preferable where possible.

guides/inference-scheduling/README.md
```diff
@@ -2,13 +2,13 @@
...
 
 This profile defaults to the approximate prefix cache aware scorer, which only observes request traffic to predict prefix cache locality. The [precise prefix cache aware routing feature](../precise-prefix-cache-aware) improves hit rate by introspecting the vLLM instances for cache entries and will become the default in a future release.
 
 ## Hardware Requirements
 
-This example out of the box requires 2 GPUs of any supported kind:
+This example out of the box requires 20 GPUs (10 replicas x 2 GPUs each) of any supported kind:
```
vMaroon: A point on how to make it require less is ideal for devs / functional use-cases.

guides/inference-scheduling/ms-inference-scheduling/values.yaml
BenjaminBraunDev: I set this to 10 to match blog, since we use that many for benchmarking senarios A-D in blog comparison. Do you think it's okay to deviate from the blog to better match node GPU counts? I fine with doing that.

@kaushikmitr do you have an opinions on this?

guides/inference-scheduling/ms-inference-scheduling/values.yaml
Gregory-Pereira: Either works here, in MS v0.3.17 we added support for allowing people to explicitly specify resource count to override parallelism: https://github.com/llm-d-incubation/llm-d-modelservice/commit/6de2a1772d76316c41e7755b557211b384bf17c8

guides/inference-scheduling/ms-inference-scheduling/values.yaml
Gregory-Pereira: I reviewed the blog, and it looks like experiments were ran for 4 different cases with these numbers. Increments of 8 GPUs are preferred but not required, so this should come down to how much work it is to redo the experiments with 8 replicas

guides/inference-scheduling/README.md
BenjaminBraunDev: Requires is a strong word here. The benchmark uses this many, but it will work just fine with less. I will adjust the wording to make that clear.

BenjaminBraunDev: done

guides/inference-scheduling/README.md
BenjaminBraunDev: adjusted

guides/inference-scheduling/ms-inference-scheduling/values.yaml
BenjaminBraunDev: done

guides/inference-scheduling/ms-inference-scheduling/values.yaml
Gregory-Pereira: Spoke to kaushik last night and he agreed - we could be fine with 8 replicas. I think it comes down to how much work will it take to re-capture the scenarios  A-D with 8 replicas. If this is additional work that we won't have ready in time for 0.5 then leave as is

guides/inference-scheduling/ms-inference-scheduling/values.yaml
```diff
@@ -29,15 +29,29 @@ decode:
...
+      - name: UCX_TLS
+        value: "cuda_ipc,cuda_copy,tcp"
+      - name: VLLM_NIXL_SIDE_CHANNEL_HOST
+        valueFrom:
+          fieldRef:
+            fieldPath: status.podIP
+      - name: VLLM_NIXL_SIDE_CHANNEL_PORT
+        value: "5557"
+      - name: VLLM_LOGGING_LEVEL
+        value: INFO
```
Gregory-Pereira: Remove. Inference scheduling does not use P/D, and we are not intializing the NIXL connector so there is no reason to set the sidechannel address and the port. Also INFO is the default vLLM log level, I would only include this if we were setting something non-standard.
```suggestion
```

guides/inference-scheduling/ms-inference-scheduling/values.yaml
vMaroon: The workload in the blog, for which we are switching to 10 replicas, does not seem to be suitable for this path. Team might fallback to reproducing the kv-cache-wins-you-can-see blog benchmarks with 8 replicas.

cc @ahg-g

BenjaminBraunDev: Changed to 8 for now

guides/inference-scheduling/ms-inference-scheduling/values.yaml
BenjaminBraunDev: done, I realize this was a rebase artifact as these values were removed in a commit that was merged as I was working

PR #628, opened 2026-01-24
wide-ep dp-aware scheduling

Gregory-Pereira:
cc @vMaroon @tlrmchlsmth @shmuelk @smarterclayton @robertgshaw2-redhat @yangligt2 @diegocastanibm

Please provide color or comments where needed

### DIFF
--- a/guides/wide-ep-lws/README.md
+++ b/guides/wide-ep-lws/README.md
@@ -132,6 +132,46 @@ You can also customize your gateway, for more information on how to do that see 
 
 As with PD, the `wide-ep-lws` guide supports selective PD. For information on this refer to [this section of the PD docs](../pd-disaggregation/README.md#tuning-selective-pd).
 
+## Experimental: Data Parallel (DP) Aware Scheduling
+
+This deployment uses **DP-aware scheduling**, where instead of letting vLLM automatically handle data parallelism internally, we explicitly launch separate vLLM server instances for each data parallel rank. This enables the inference scheduler to route requests directly to specific DP ranks, improving KV cache routing efficiency.
+
+### How It Works
+
+**Traditional Data Parallelism:**
+- vLLM launches a single server process that internally manages DP=16 across GPUs
+- External clients see one endpoint per pod
+- vLLM handles internal load balancing across DP ranks
+
+**DP-Aware Scheduling (This Deployment):**
+- Each pod explicitly launches 8 separate vLLM server instances (one per GPU)
+- Each instance runs on a unique port (8000-8007 for prefill, 8200-8207 for decode)
+- Each instance is assigned an explicit DP rank using `--data-parallel-rank`
+- The inference scheduler can route to specific DP ranks based on request characteristics
+
+**Key Configuration Changes:**
+
+1. **InferencePool**: Declares 8 target ports (8000-8007) to expose all local DP ranks
+2. **Routing Proxy**: Configured with `--data-parallel-size=8` to understand the DP topology
+3. **vLLM Containers**: Launch 8 parallel processes per pod with:
+   - `CUDA_VISIBLE_DEVICES` pinning each process to a specific GPU (0-7)
+   - Unique ports per rank
+   - Explicit `--data-parallel-rank` assignment
+   - Global `--data-parallel-size=16` (8 ranks × 2 pods)
+   - Local `--data-parallel-size-local=8`
+
+**Architecture:**
+
+Each LeaderWorkerSet has 2 pods, each running 8 vLLM instances:
+- **Prefill Pod 0**: DP ranks 0-7 on ports 8000-8007
+- **Prefill Pod 1**: DP ranks 8-15 on ports 8000-8007
+- **Decode Pod 0**: DP ranks 0-7 on ports 8200-8207
+- **Decode Pod 1**: DP ranks 8-15 on ports 8200-8207
+
+The scheduler routes to specific ranks using (pod IP, port) tuples, enabling fine-grained request distribution.
+
+For more information on vLLM's data parallelism features, see the [vLLM documentation](https://docs.vllm.ai/en/latest/serving/data_parallel_deployment/#external-load-balancing).
+
 ## Verifying the installation
 
 * Firstly, you should be able to list all helm releases installed into your chosen namespace:
--- a/guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
+++ b/guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
@@ -30,14 +30,35 @@ spec:
               - --port=8000
               - --vllm-port=8200
               - --connector=nixlv2
-              - --zap-log-level=1
+              - --zap-log-level=4
               - --secure-proxy=false
               - --enable-prefiller-sampling
             image: ghcr.io/llm-d/llm-d-routing-sidecar:v0.5.0
             imagePullPolicy: Always
             ports:
               - containerPort: 8000
-                name: sidecar
+                name: rank0
+                protocol: TCP
+              - containerPort: 8001
+                name: rank1
+                protocol: TCP
+              - containerPort: 8002
+                name: rank2
+                protocol: TCP
+              - containerPort: 8003
+                name: rank3
+                protocol: TCP
+              - containerPort: 8004
+                name: rank4
+                protocol: TCP
+              - containerPort: 8005
+                name: rank5
+                protocol: TCP
+              - containerPort: 8006
+                name: rank6
+                protocol: TCP
+              - containerPort: 8007
+                name: rank7
                 protocol: TCP
             resources: {}
             restartPolicy: Always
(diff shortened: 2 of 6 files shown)


guides/wide-ep-lws/README.md
```diff
@@ -133,6 +133,54 @@ You can also customize your gateway, for more information on how to do that see
 
 As with PD, the `wide-ep-lws` guide supports selective PD. For information on this refer to [this section of the PD docs](../pd-disaggregation/README.md#tuning-selective-pd).
 
+## Experimental: Data Parallel (DP) Aware Scheduling
+
+This deployment uses **DP-aware scheduling**, where instead of letting vLLM automatically handle data parallelism internally, we explicitly launch separate vLLM server instances for each data parallel rank. This enables the inference scheduler to route requests directly to specific DP ranks, improving KV cache routing efficiency.
+
+> **Note**: DP-aware scheduling currently requires H200 GPUs. (@vMaroon - can you clarify the specific hardware requirements and limitations?)
```
vMaroon: This is inherited directly from the guide, which is based on tested deployments. (cc @tlrmchlsmth)

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -68,42 +90,47 @@ spec:
...
+              # --enable-eplb: Expert-parallel load balancing reduces EP load imbalance by replicating heavily-used experts
+
+              START_RANK=$(( ${LWS_WORKER_INDEX:-0} * 8 ))
+
+              # Common vLLM flag arguments shared across all decode ranks
+              COMMON_VLLM_ARGS=(
+                --trust-remote-code
+                --disable-uvicorn-access-log
+                --tensor-parallel-size=1
+                --data-parallel-size=16
```
diegocastanibm: Shouldn't this be 8?

```suggestion
                --data-parallel-size=8
```

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -68,42 +90,47 @@ spec:
...
+              CUDA_VISIBLE_DEVICES="2" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8202 --data-parallel-rank=$((START_RANK + 2)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="3" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8203 --data-parallel-rank=$((START_RANK + 3)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="4" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8204 --data-parallel-rank=$((START_RANK + 4)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="5" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8205 --data-parallel-rank=$((START_RANK + 5)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="6" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8206 --data-parallel-rank=$((START_RANK + 6)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="7" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8207 --data-parallel-rank=$((START_RANK + 7)) "${COMMON_VLLM_ARGS[@]}" &
```
diegocastanibm: I thought we want to use `Qwen/Qwen3-32B` everywhere. Is `deepseek-ai/DeepSeek-R1-0528` an exception here?

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
```diff
@@ -163,32 +190,59 @@ spec:
...
             timeoutSeconds: 5
             failureThreshold: 3
           readinessProbe:
             httpGet:
               path: /v1/models
-              port: vllm
+              port: rank0
             periodSeconds: 10
             timeoutSeconds: 5
             failureThreshold: 3
```
diegocastanibm: maybe we need to increase them since the model is going to be bigger?

guides/wide-ep-lws/manifests/modelserver/base/prefill.yaml
```diff
@@ -48,42 +48,46 @@ spec:
...
+              CUDA_VISIBLE_DEVICES="2" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8002 --data-parallel-rank=$((START_RANK + 2)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="3" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8003 --data-parallel-rank=$((START_RANK + 3)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="4" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8004 --data-parallel-rank=$((START_RANK + 4)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="5" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8005 --data-parallel-rank=$((START_RANK + 5)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="6" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8006 --data-parallel-rank=$((START_RANK + 6)) "${COMMON_VLLM_ARGS[@]}" &
+              CUDA_VISIBLE_DEVICES="7" vllm serve deepseek-ai/DeepSeek-R1-0528 --host=0.0.0.0 --port=8007 --data-parallel-rank=$((START_RANK + 7)) "${COMMON_VLLM_ARGS[@]}" &
```
diegocastanibm: Same thing about the model here. Should we use `Qwen/Qwen3-32B`

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
diegocastanibm: Oh, wait a sec, in the readme, I see this:
```
   - Global `--data-parallel-size=16` (8 ranks × 2 pods)
   - Local `--data-parallel-size-local=8`
```
so, if `--data-parallel-size=16` is correct in this case, why the `--data-parallel-size-local=1` in the line below instead `--data-parallel-size-local=8`?

guides/wide-ep-lws/README.md
```diff
@@ -133,6 +133,54 @@ You can also customize your gateway, for more information on how to do that see
...
+   - Global `--data-parallel-size=16` (8 ranks × 2 pods)
+   - Local `--data-parallel-size-local=8`
+
+**Architecture:**
+
+Each LeaderWorkerSet has 2 pods, each running 8 vLLM instances:
+- **Prefill Pod 0**: DP ranks 0-7 on ports 8000-8007
+- **Prefill Pod 1**: DP ranks 8-15 on ports 8000-8007
+- **Decode Pod 0**: DP ranks 0-7 on ports 8200-8207
+- **Decode Pod 1**: DP ranks 8-15 on ports 8200-8207
```
diegocastanibm, line 169: I'm confused. I do not see `ranks 8-15` in the yaml files. Is this correct?

guides/wide-ep-lws/manifests/modelserver/base/prefill.yaml
Gregory-Pereira: No this should be deepseekr1. Basic examples (IS + precise) will use Qwen3-32B, PD uses a RHAI sparse quantized verison of llama70b, and Wide-ep uses deepseek-r1

guides/wide-ep-lws/README.md
yangligt2: The guides also have B200. So can this be tested on B200, or strictly only on H200?

Gregory-Pereira: It should work on any of the above - I misunderstood Maroons original comment

guides/wide-ep-lws/README.md
tlrmchlsmth: This is right - ranks 0-7 will be on worker 1, and ranks 8-15 will be in a different pod with `LWS_WORKER_INDEX==1`

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
Gregory-Pereira: Not sure, its been working for me so far, and if anything the startup of launching explicit processes manually has made it much faster.

guides/wide-ep-lws/manifests/modelserver/base/decode.yaml
Gregory-Pereira: No I think that flag is for tracking the data parallel world size rather than at a"per pod" level. This means it should be 16 when aggregated across the two LWSs. I wanted to specify the `--data-parallel-size-local` which in this case would actually be 1 (remember were using the 1 pod to explicitly spawn 8 different vLLM servers with different ranks, and this flag is getting passed to each startup command)

PR #630, opened 2026-01-25
install OTel packages in all images

Gregory-Pereira:
Related to #621 

cc @sallyom testing if this can work for us with the existing build setup


wseaton (approved):
LGTM, these are very unlikely to cause issues wrt constriants.

Issue #620, opened 2026-01-21
XPU resources requirement is overwritten by nvidia.com/gpu


VincyZhang:
**Problem:**
    When deploy llm-d on Intel XPU, the resources should be like:
            ```requests:
              cpu: "16"
              memory: 64Gi
              gpu.intel.com/i915: "1"```
   But actually got ```requests:
              cpu: "16"
              memory: 64Gi
              nvidia.com/gpu: "1"```
   The resource requirement for Intel XPU is overwritten by NV GPU. Thus we can't deploy llm-d on Intel XPU.

<img width="278" height="139" alt="Image" src="https://github.com/user-attachments/assets/8308a323-73fa-486f-a564-6f783349d275" />

<img width="248" height="136" alt="Image" src="https://github.com/user-attachments/assets/2bfba4f0-7933-4e8f-ab8d-2c0a31b8774c" />

**Test Config:**
HW: gpu.intel.com/i915
llm-d installed version: 33acf18e05
vllm installed version: v0.1.dev11429+g9a8f3a379.d20260119
**How to Reproduce:**
(1) Install prerequisites in `llm-d/guides/prereq/gateway-provider`
(2) Create namespace in k8s
(3) cd llm-d/guides/pd-disaggregation
(4) helmfile -e xpu -n $NAMESPACE template > test.yaml
(5) cat test.yaml

yuanwu2017:
[PR518](https://github.com/llm-d/llm-d/pull/518) and [PR380](https://github.com/llm-d/llm-d/pull/380) can fix it.

PR #380, opened 2025-10-21
Add the XPU CI test

yuanwu2017:
This patch series adds CI/CD support for Intel XPU (GPU) workloads.

### 1. CI/CD Infrastructure (`ci-pr-test.yaml`)
- Implements automated pull request testing for Intel XPU-related changes
- Orchestrates end-to-end testing pipeline with image building and deployment validation
- Supports both PR review-triggered and manual workflow dispatch modes
- Provides custom image tag override capability for testing specific builds

### 2. E2E Test Workflows

**Inference Scheduling** (`e2e-inference-scheduling-xpu.yaml`)
**PD Disaggregation** (`e2e-pd-xpu.yaml`)

Both workflows now support:
- `workflow_call` trigger for CI integration
- Custom image tag injection via input parameters
-  Three-stage deployment (infra → gaie → model service) to apply image overrides only to decode/prefill containers that need custom images
- 
### 3. Image Build Workflow Updates (`build-image.yml`)

- Enhanced to support `workflow_call` trigger
- Refined path filters to prevent unnecessary builds
- Integrates with CI pipeline for automated image building on PR events

### 5. Documentation Fixes

- Fixed markdown link checker error in OpenShift documentation (escaped underscore in URL)

### 6. Dockerfile.xpu
0.12.0  OK
### 7. Transitioned to out-of-band HTTPRoute deployment approach
- HTTPRoute deployed separately after Helm releases complete
- Dynamic name substitution based on environment variables in CI workflow.
- Added **important warnings** about HTTPRoute customization when using `RELEASE_NAME_POSTFIX`

PR #636, opened 2026-01-26
fix: add accelerator type so it can set proper env variable for vllm container to use intel xpu

zdtsw:
# Description

part of the issue https://github.com/llm-d/llm-d/issues/620 has been resolved by PR https://github.com/llm-d/llm-d/pull/518 when enable dra in values_xpu.yaml
there are left part for the accelerator need to config.
According to [helpers.tpl](https://github.com/llm-d-incubation/llm-d-modelservice/blob/llm-d-modelservice-v0.3.17/charts/llm-d-modelservice/templates/_helpers.tpl#L249) if do not set accelerator.type to "cpu" then  it fall back to use `nvidia` as type but then it wont set [env](https://github.com/llm-d-incubation/llm-d-modelservice/blob/llm-d-modelservice-v0.3.17/charts/llm-d-modelservice/values.yaml#L63) for `VLLM_USE_V1` `TORCH_LLḾ_ALLREDUCE` `VLLM_WORKER_MULTIPROC_METHOD` in vllm container

# Changes
- DRA has been enabled in xpu but still missing accelerator type which set the gpu.intel.com:1 in `ResourceClaimTemplate` but not the env variables needed for vLLM. this change is to add accelerator.type in value
- also fix lint on the yaml file

### DIFF
--- a/guides/pd-disaggregation/ms-pd/values_xpu.yaml
+++ b/guides/pd-disaggregation/ms-pd/values_xpu.yaml
@@ -8,6 +8,9 @@ dra:
   enabled: true
   type: intel
 
+accelerator:
+  type: intel-i915
+
 modelArtifacts:
   uri: "hf://deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B"
   size: 10Gi  # Smaller model size for Intel XPU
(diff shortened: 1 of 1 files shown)


yuanwu2017:
[PR518](https://github.com/llm-d/llm-d/pull/518) and [PR380]( https://github.com/llm-d/llm-d/pull/380 ) can fix [issue620](https://github.com/llm-d/llm-d/issues/620).

guides/pd-disaggregation/ms-pd/values_xpu.yaml
```diff
@@ -8,6 +8,9 @@ dra:
   enabled: true
   type: intel
 
+accelerator:
```
poussa, line 11: The `accelerator` object is not needed, it is for device plugins The `dra` object above is used instead (when `dra.enabled: true`).

guides/pd-disaggregation/ms-pd/values_xpu.yaml
```diff
@@ -29,49 +32,49 @@ decode:
   monitoring:
     podmonitor:
       enabled: true
-      portName: "vllm" # decode vLLM service port (from routing.proxy.targetPort)
+      portName: "vllm"  # decode vLLM service port (from routing.proxy.targetPort)
```
poussa, line 35: why all these whitespaces?

zdtsw: I've pre-commit enabled and got all the lint failure locally.
the fixes (format) is automatically done by https://github.com/llm-d/llm-d/blob/main/.pre-commit-config.yaml#L61

guides/pd-disaggregation/ms-pd/values_xpu.yaml
zdtsw: let me close this PR first and open a new one, it is better uplift chart to 0.4.0 with your changes done in acclerator+dra.
wdyt?

poussa: > let me close this PR first and open a new one, it is better uplift chart to 0.4.0 with your changes done in acclerator+dra. wdyt?

Both are fine (new or update this one).

PR #640, opened 2026-01-27
install storage offloading connector

Gregory-Pereira:
cc @kfirtoledo @vitabortnikov @nilig

### DIFF
--- a/docker/common-versions
+++ b/docker/common-versions
@@ -22,7 +22,7 @@ CUDA_PATCH=1
 
 ### Offloading connector
 
-LLM_D_OFFLOADING_CONNECTOR_VERSION=0.17.1
+LLM_D_OFFLOADING_CONNECTOR_VERSION=0.15.1
 
 # ============================================================================
 # XPU specific settings


docker/Dockerfile.cuda
```diff
@@ -392,6 +392,9 @@ RUN export UV_INSTALL_DIR=/usr/local/bin && \
 # Copy compiled wheels
 COPY --from=builder /wheels/*.whl /tmp/wheels/
 
+# Copy in storage offloading connector wheel
```
zdtsw: is this one only for amd64? or arm64 will try to install it during build as well?

kfirtoledo: This one is only for amd64

Gregory-Pereira: Is there a different arm one? If so we can do conditional install logic but keep this in the Dockerfile. If not we need to build this into one of the scripts

docker/Dockerfile.cpu
```diff
@@ -103,6 +103,9 @@ FROM vllm-build
 
 ENV LD_LIBRARY_PATH="/usr/local/lib:/usr/lib:${LD_LIBRARY_PATH}"
 
+# Copy in storage offloading connector wheel
+COPY docker/packages/wheels/packages/wheels/llmd_fs_connector-0.1.0-cp312-cp312-linux_x86_64.whl  dist/llmd_fs_connector-0.1.0-cp312-cp312-linux_x86_64.whl
```
Gregory-Pereira: Note from kfir, this is based on v0.14.1 of vLLM so we gotta update this for v0.15.0 which were picking

docker/Dockerfile.cuda
Gregory-Pereira: OK this has been refactored to a script to account for this. What remains to be seen is if the same connector will work in both our ubuntu and rhel images. Think it should be fine but will see

Gregory-Pereira:
depends on: #977

docker/scripts/common/install-offloading-connector.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+if [ -z "${wheel}" ]; then
+    echo "No matching wheel found for arch=${arch} in ${wheels_src}" >&2
+    ls -la "${wheels_src}" >&2
+    exit 1
+fi
+
+cp "${wheel}" "${WHEELS_DIR}/"
+ls -lah "${WHEELS_DIR}/$(basename "${wheel}")"
```
diegocastanibm, line 64: The wheel is copied in the tmp dir but not installed in this scrip. Is that the intention?

docker/scripts/common/install-offloading-connector.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+set -Eeuox pipefail
```
diegocastanibm, line 2: I would consider to add `-x` just for DEBUG mode to avoid noisy logs... but it is just a recommendation

PR #645, opened 2026-01-28
fix[lint]: various kinds of lint error+warning

zdtsw:
# Changes
- markdown, one fo the fix has been included in https://github.com/llm-d/llm-d/pull/643, some others will be done once WVA 0.5.0 is released.
- dockerfile
- yaml
- shell + generic error warnings


guides/pd-disaggregation/ms-pd/values.yaml
```diff
@@ -21,75 +21,75 @@ decode:
...
-        # note: GPU resources get controlled by parallelism + accelerators above
-        rdma/ib: 1
-      requests:
-        memory: 64Gi
-        cpu: "16"
-        # note: GPU resources get controlled by parallelism + accelerators above
-        rdma/ib: 1
-    mountModelVolume: true
-    volumeMounts:
+    - name: "vllm"
```
petecheslock, line 28: This (and the other values files) are just indentations right? Do we need to check the Helm templates consuming these values handle this style? I'm assuming so but how can we confirm?

.markdownlint.json
```diff
@@ -11,8 +11,6 @@
     "strict": false,
     "stern": false
   },
-  "MD033": {
-    "allowed_elements": ["details", "summary", "img", "br"]
-  },
+  "MD033": false,
```
petecheslock: I'm not sure we want to straight up disable MD033, some of the readme's and such get [pull from this repo by the website.](https://github.com/llm-d/llm-d.github.io/blob/main/README.md#-remote-content-system)  We do some content parsing to clean up files but i've ran into issues before where some html elements get parsed and break the website build.   

Was there something specific you ran into here?

guides/pd-disaggregation/ms-pd/values.yaml
zdtsw: yep, indentation from 2 to 4.

what can be tested is to make a dry-run by kubectl which should be enough to verify the syntax of these yaml if they work with k8s or not. 
or we could do the real run on all of these if needed.

.markdownlint.json
zdtsw: it is elements like `<b><i>` in some of the markdown files.
i can revert this and put these (5-6 if I recall)  intoe the allowed list.

petecheslock: Ahh yea i ran it locally and saw a bunch of things - what about this?  
```
    "allowed_elements": ["details", "summary", "img", "br", "picture", "source", "p", "table", "tr", "td", "th", "h2", "code", "b", "i"]
```

This will then still block some of the worst offenders.

guides/pd-disaggregation/ms-pd/values.yaml
petecheslock: @zdtsw thanks - i rolled thru with a test suite to validate and things looked good with a kubectl --dry-run

.markdownlint.json
zdtsw: updated

PR #585, opened 2026-01-11
Configuration template for precise-prefix-cache-aware well-lit path benchmarking

dmitripikus:
Configuration template for precise-prefix-cache-aware well-lit path benchmarking

This PR depends on:
- #559 (must be merged first)

PR #652, opened 2026-01-29
Benchmarking of precise well-lit path

dmitripikus:
Work towards #487 
Replaces PR #585 (which used smaller model)
* Benchmarking section in precise-prefix README
* Updates in stack configuration `value` files (for lager model + larger test scale)
* Updates to `precise-prefix` benchmark templates
* Comparison to k8s baseline test

### DIFF
--- a/guides/precise-prefix-cache-aware/ms-kv-events/values.yaml
+++ b/guides/precise-prefix-cache-aware/ms-kv-events/values.yaml
@@ -30,6 +30,7 @@ decode:
     image: vllm/vllm-openai:v0.14.1-cu130
     modelCommand: vllmServe
     args:
+      - "--block-size=64"
       - "--kv-events-config"
       - |- 
         {


guides/precise-prefix-cache-aware/ms-kv-events/values.yaml
```diff
@@ -24,49 +27,61 @@ decode:
...
-        vllm serve Qwen/Qwen3-0.6B \
-        --host 0.0.0.0 \
-        --port 8000 \
-        --block-size 64 \
-        --prefix-caching-hash-algo sha256_cbor \
-        --kv-events-config "{\"enable_kv_cache_events\":true,\"publisher\":\"zmq\",\"endpoint\":\"tcp://gaie-${GAIE_RELEASE_NAME_POSTFIX}-epp.${NAMESPACE}.svc.cluster.local:5557\",\"topic\":\"kv@${POD_IP}@Qwen/Qwen3-0.6B\"}"
+#      - "--host=0.0.0.0"
+#      - "--port=8000"
+      - "--block-size=64"
+      - "--prefix-caching-hash-algo=sha256_cbor"
```
vMaroon: This can be dropped now - no longer a necessity.

dmitripikus: Done. Thank you!

guides/precise-prefix-cache-aware/ms-kv-events/values.yaml
```diff
@@ -24,49 +27,57 @@ decode:
...
-      - '-c'
+    image: vllm/vllm-openai:v0.14.1-cu130
+    modelCommand: vllmServe
     args:
-      - |
-        vllm serve Qwen/Qwen3-0.6B \
-        --host 0.0.0.0 \
-        --port 8000 \
-        --block-size 64 \
-        --prefix-caching-hash-algo sha256_cbor \
```
sagearc, line 38: @vMaroon While we don’t reproduce the block hashes from the EPP, shouldn’t the hashing algorithm be deterministic across all vllm instances? For a given prompt, I’d expect the same block hashes from all vllm pods, so our engine-key <-> request-key mapping is one-to-one.

vMaroon: The removal of this line does not affect hash consistency on the vLLM side - it only defaults to whatever hashing function vLLM has set.

On the other hand, the removal of `PYTHONHASHSEED` env var does indeed make it so that each vLLM replica has a different parent hash for block sequences. But if I'm not mistaken, this would still work (we do intend for it to work at the very least). In our index implementation:
```
    // data holds the mapping of requestKeys to sets of pod identifiers.
	data *lru.Cache[BlockHash, *PodCache]
	// engineToRequestKeys holds the mapping of engineKeys to requestKeys.
	engineToRequestKeys *lru.Cache[BlockHash, BlockHash]
```

`data` holds request-keys to pod-identifiers. Request keys are consistent given an input. `enginteToRequestKeys` can map different vLLM keys to the same request key.

So overall different replicas should safely produce different block hashes for the same sequences, and the indexer would still handle them properly.

Did I miss anything?

vMaroon: We should explicitly verify this, but IMO should not be a blocker for this PR. What do you think?

sagearc: Good points.

> `data` holds request-keys to pod-identifiers. Request keys are consistent given an input. `enginteToRequestKeys` can map different vLLM keys to the same request key.

I just have one concern regarding `BlockRemoved` events. If we don't explicitly clean up `engineToRequestKeys` during eviction, we might end up with orphaned engine keys in the map.

PR #655, opened 2026-01-29
[Bugfix] Fix patches

tlrmchlsmth:
Revert changes from https://github.com/llm-d/llm-d/pull/645  to the nvshmem patches, as the changes caused them to no longer apply

### DIFF
--- a/docker/scripts/common/setup-sccache.sh
+++ b/docker/scripts/common/setup-sccache.sh
@@ -37,7 +37,7 @@ if [ "${USE_SCCACHE}" = "true" ]; then
         echo "Warning: sccache failed to start, continuing without cache" >&2
         unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
         # Remove sccache binary so meson/cmake can't accidentally try to use it
-        rm -f /usr/local/bin/sccache
+        rm -f /usr/local/bin/sccache || true
         return 0
     fi
 
@@ -46,7 +46,7 @@ if [ "${USE_SCCACHE}" = "true" ]; then
         /usr/local/bin/sccache --stop-server 2>/dev/null || true
         unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
         # Remove sccache binary so meson/cmake can't accidentally try to use it
-        rm -f /usr/local/bin/sccache
+        rm -f /usr/local/bin/sccache || true
         return 0
     fi
 


Gregory-Pereira (approved):
/lgtm

docker/scripts/common/setup-sccache.sh
```diff
@@ -36,14 +36,14 @@ if [ "${USE_SCCACHE}" = "true" ]; then
     if ! /usr/local/bin/sccache --start-server; then
         echo "Warning: sccache failed to start, continuing without cache" >&2
         unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
-        return 1
+        return 0
```
tlrmchlsmth, line 41: Added this because the build was failing due to an sccache timeout. 

Here we're supposed to continue without using sccache but the caller's `set -e` causes a crash

docker/scripts/common/setup-sccache.sh
```diff
@@ -36,14 +36,18 @@ if [ "${USE_SCCACHE}" = "true" ]; then
     if ! /usr/local/bin/sccache --start-server; then
         echo "Warning: sccache failed to start, continuing without cache" >&2
         unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
-        return 1
+        # Remove sccache binary so meson/cmake can't accidentally try to use it
+        rm -f /usr/local/bin/sccache
```
Gregory-Pereira: should add || true for idempotency
```suggestion
        rm -f /usr/local/bin/sccache || true
```

docker/scripts/common/setup-sccache.sh
```diff
@@ -36,14 +36,18 @@ if [ "${USE_SCCACHE}" = "true" ]; then
...
+        return 0
     fi
 
     if ! /usr/local/bin/sccache --show-stats >/dev/null 2>&1; then
         echo "Warning: sccache not responding properly, disabling cache" >&2
         /usr/local/bin/sccache --stop-server 2>/dev/null || true
         unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
-        return 1
+        # Remove sccache binary so meson/cmake can't accidentally try to use it
+        rm -f /usr/local/bin/sccache
```
Gregory-Pereira: Same as above:

```suggestion
        rm -f /usr/local/bin/sccache || true
```

Gregory-Pereira:
Tracking https://github.com/llm-d/llm-d/actions/runs/21497318672 as an idicator of success, I just added idempotency which will not affect builds in any way

docker/scripts/common/setup-sccache.sh
tlrmchlsmth: good call

PR #662, opened 2026-01-29
Fix LMCache non crashing build failure

Gregory-Pereira:
LMcache is failing to build but not dying, debugging why


Gregory-Pereira:
Successful builds but were hitting runtime error referenced in #663

PR #665, opened 2026-01-30
Add wide wp benchmark results on B200

liu-cong:


Gregory-Pereira (dismissed):
/lgtm

guides/wide-ep-lws/README.md
```diff
@@ -189,6 +189,47 @@ For instructions on getting started making inference requests see [our docs](../
...
+
+The `GATEWAY_IP` environment variable will be used in the [benchmark template](../benchmark/wide_ep_random_inference_perf.yaml).
+
+2. Follow the [benchmark guide](../../guides/benchmark/README.md) to deploy the benchmark tool and analyze the benchmark results.
+
+### Results
+
+![alt text](benchmark-result.png)
+
+At request rate 250, we achieved the following performance:
```
ahg-g: can we show concurrency as well and the various latencies?

liu-cong: Added latency graph. This is a QPS based run, not concurrency based, so I don't have concurrency info. I think concurrency is best for interactivity oriented use cases that represents the # online users, while this benchmark is targeting a throughput oriented use case.

Gregory-Pereira: Leaving review for this to @ahg-g looks fine from me

ahg-g: > Added latency graph. This is a QPS based run, not concurrency based, so I don't have concurrency info. I think concurrency is best for interactivity oriented use cases that represents the # online users, while this benchmark is targeting a throughput oriented use case.

I think it is best if we can present the results in a way aligned with inferenceMax, probably users will find that more informative.

Gregory-Pereira (approved):

PR #668, opened 2026-01-30
fix: port for NIXL should be updated to 5600

zdtsw:
# Description
- it was using 5557 for NIXL but since vLLM 0.11.1 it updated to 5600

# Ref
- https://github.com/vllm-project/vllm/pull/25056
- https://github.com/vllm-project/vllm/releases/tag/v0.11.1


zdtsw:
cross fix in https://github.com/llm-d/llm-d-inference-scheduler/pull/593

Gregory-Pereira (approved):
/lgtm

PR #674, opened 2026-02-01
SGLang: Well-lit Path for Intelligent Inference Scheduling for approximate prefix cache aware routing

rahulgurnani:
This change contains changes for well-lit path for approximate prefix cache aware routing. The change has been tested on GKE and should work for all gateway providers.

### DIFF
--- a/guides/inference-scheduling/README.md
+++ b/guides/inference-scheduling/README.md
@@ -127,10 +127,10 @@ helmfile apply -e cpu  -n ${NAMESPACE} # targets istio as gateway provider with 
 
 This case expects using 4th Gen Intel Xeon processors (Sapphire Rapids) or later.
 
-### Inference Server Selection
+### Model Server Selection
 
-By default, this well-lit path uses vLLM as the inference server for AI model serving.
-In case you want to deploy SGLang as the inference server, use:
+By default, this well-lit path uses vLLM as the model server.
+In case you want to set MODEL_SERVER environment variable to sglang as the model server, use:
 
 ```bash
 export MODEL_SERVER=sglang


guides/inference-scheduling/ms-inference-scheduling/values_sglang.yaml
```diff
@@ -0,0 +1,91 @@
+multinode: false
+
+modelArtifacts:
+  uri: "hf://Qwen/Qwen3-0.6B"
+  name: "Qwen/Qwen3-0.6B"
```
ahg-g, line 5: we moved to use 32B

guides/inference-scheduling/helmfile.yaml.gotmpl
```diff
@@ -57,17 +57,19 @@ releases:
...
       {{- if ne .Environment.Name "standalone" }}
       - {{ printf "infra-%s" $rn | quote }}
       {{- end }}
     values:
-      {{- if ne .Environment.Name "standalone" }}
-      - gaie-inference-scheduling/values.yaml
-      {{- else }}
+      {{- if eq .Environment.Name "standalone" }}
       - standalone-inference-scheduling/values.yaml
+      {{- else if eq (env "MODEL_SERVER") "sglang" }}
```
ahg-g, line 69: why as an env var instead of explicit flag?

ezrasilvera: @ahg-g Maybe it will be better to converge the discussion into one place. We have an open PR for this at #527 Please have a look and feel feel to review.  Originally we did use explicit flag (-e) but due to the PR comments/feedback we switched to using env variable aligned with the proposal from @rahulgurnani . 
BTW, we already updated the PR for the PD well lit path (#580 ) in the same way (from a flag to env var), so once we all agree on an approach we will update both

ahg-g: That is fine, I was just curious, thanks for linking to the previous discussion

PR #676, opened 2026-02-02
[xpu] Use DRA for inference scheduling

poussa:
We would like to use DRA for all Intel accelerators. For some reason, this was left out. If possible, can we get this one to 0.5?

### DIFF
--- a/guides/inference-scheduling/ms-inference-scheduling/values_xpu.yaml
+++ b/guides/inference-scheduling/ms-inference-scheduling/values_xpu.yaml
@@ -28,7 +28,7 @@ decode:
   containers:
     - name: "vllm"
       # Use Intel latest published xpu image
-      image: ghcr.io/llm-d/llm-d-xpu:v0.4.0
+      image: ghcr.io/llm-d/llm-d-xpu-dev:pr-592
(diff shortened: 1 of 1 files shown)


yuanwu2017 (approved):

poussa:
> i think the ongoing prep release 0.5.0 is to uplift modelservice to 0.4.5 which has the new spec of the accelerator+dra. we need change https://github.com/llm-d/llm-d/pull/617/files accordingly @Gregory-Pereira

The 0.4 modelservice accelerator API is still under [debate / review](https://github.com/llm-d-incubation/llm-d-modelservice/pull/191). Me and @yuanwu2017 want to unify XPU DRA resource names (`accelerator.type`) to use `intel` (instead of intel-xe and intel-i915). Thus, my suggestion is to do llm-d v0.5 with modelservice v0.3. Then stabilize the modelservice accelerator API and update it after llm-d v0.5.

Gregory-Pereira:
We are using v0.4.5 for the MS for the base examples. What we are doing is using the same pattern you shared above in all XPU and HPU examples for this release, so we should be OK.

poussa:
Closing since the #696 took care of this.

PR #678, opened 2026-02-02
[DO_NOT_MERGE] try building EFA first

Gregory-Pereira:


Gregory-Pereira:
Dropping in favor of #741

PR #684, opened 2026-02-03
feat: add support to build for CPU on AMX

zdtsw:
# Description
- AVX2 is old one for Intel, we already have support for AVX512 the new version
- this change is only to add more support on AMX alongwith AVX512

# Ref
- original support in vllm in https://github.com/vllm-project/vllm/pull/27954
- request in https://issues.redhat.com/browse/RHAIRFE-28


Gregory-Pereira (approved):
/lgtm

cc @poussa @yuanwu2017

PR #693, opened 2026-02-03
needs manifest aggregation logic to releas workflows

Gregory-Pereira:
cc @diegocastanibm @wseaton @zdtsw


Gregory-Pereira:
pushing this through for next RC check

PR #695, opened 2026-02-04
adding setuptools and setuptools-scm for LMCache version metadata

Gregory-Pereira:
LMcache requires setuptools and setuptools-scm for proper build metadata, otherwhise the wheel gets listed as version 0.0.0


Gregory-Pereira:
Verified this works, its registering as 3.13 here

PR #697, opened 2026-02-04
fix gke monitoring in helmfile

zetxqx:
the gaie extension has changed the way to enable gke monitoring. https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/e1cbdd1f76c58ba144fe7abe368a77241b78be24/config/charts/inference-extension/templates/_gke.yaml#L2-L3

now gke monitoring is enabled when the following conditions all saitisfied
 1. if provider is gke 
 2. if `inferenceExtension.monitoring.prometheus.enabled==true` 
 3. if `inferenceExtension.monitoring.prometheus.auth.enabled` is true (this is enabled by [default](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/e1cbdd1f76c58ba144fe7abe368a77241b78be24/config/charts/inferencepool/values.yaml#L57-L58))


guides/inference-scheduling/helmfile.yaml.gotmpl
```diff
@@ -75,10 +75,8 @@ releases:
           name: {{ .Environment.Values.provider.name }}
         inferenceExtension:
           monitoring:
-            gke:
-              enabled: true
             prometheus:
-              enabled: false
+              enabled: true
```
ahg-g, line 81: @Gregory-Pereira this (and the changes to the other gotmpl files below) creates monitoring resources for all providers, are we ok with that?

ahg-g:
/lgtm

@Gregory-Pereira pls take a look

guides/inference-scheduling/helmfile.yaml.gotmpl
Gregory-Pereira: Yes, that is ideal. We want to be opt-in by default and allow opt out

guides/recipes/inferencepool/values.yaml
```diff
@@ -9,7 +9,5 @@ inferenceExtension:
     # See https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1905.
     kv-cache-usage-percentage-metric: "vllm:kv_cache_usage_perc"
   monitoring:
-    gke:
-      enable: false
     prometheus:
       enable: false
```
Gregory-Pereira: This should be removed too yeah? So that it can be pulled from our default helmfile? Basically we should pick 1 thing either helmfile or GIE values to controll this behaviour

guides/tiered-prefix-cache/cpu/manifests/inferencepool/values.yaml
```diff
@@ -9,8 +9,6 @@ inferenceExtension:
     # See https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1905.
     kv-cache-usage-percentage-metric: "vllm:kv_cache_usage_perc"
   monitoring:
-    gke:
-      enable: false
     prometheus:
       enable: false
```
Gregory-Pereira: same as above

guides/wide-ep-lws/manifests/inferencepool.values.yaml
```diff
@@ -18,8 +18,6 @@ inferenceExtension:
     # Prometheus ServiceMonitor will be created when enabled for EPP metrics collection
     prometheus:
       enabled: false
```
Gregory-Pereira: same as above

Gregory-Pereira (commented):
It looks fine, just want to confirm:
1. Our goal is to turn on monitoring by default
2. What do we want to own pass of values to the charts? Ideally I think we do this in the individual values files, because this removes templating logic from the helmfile but either is fine honestly as long as its consistent

guides/recipes/inferencepool/values.yaml
zetxqx: The guide using recipes is installing using helm not helmfile ([e.g.](https://github.com/llm-d/llm-d/blob/3278f6ab0b12095055b81a8e5d5354800a3a9c5e/guides/recipes/inferencepool/README.md?plain=1#L17-L23)). So I changed the value to `enabled: true` and remove the command line explicitly set the value.

guides/tiered-prefix-cache/cpu/manifests/inferencepool/values.yaml
zetxqx: updated

guides/wide-ep-lws/manifests/inferencepool.values.yaml
zetxqx: updated

Gregory-Pereira (dismissed):
/lgtm

Gregory-Pereira (approved):

PR #707, opened 2026-02-06
[BUGFIX] Comment caching to avoid error

diegocastanibm:
**Description:**
In the `ci-release.yaml` the builder layer caches from the builder image and it's going to cache to the builder image and then the final layer caches from the builder layer and then it caches to the final layer. Somehow, the `build-push-action` is letting stuff through.

**Action:**
We're just going to disable build from or cache from cache to temporarily. We don't delete it because the caching is correct.


diegocastanibm:
Close this PR. Created a new PR from upstream branch in https://github.com/llm-d/llm-d/pull/719

PR #722, opened 2026-02-09
update the result for inference-scheduling and precise-prefix-cache-aware

zetxqx:
1. picked the request rate at 60 for comparison in the table, and use mean metrics for both guides
2. added some graph


guides/inference-scheduling/README.md
```diff
@@ -534,48 +534,27 @@ There is a report for each stage.
...
+| Metric | k8s (Mean) | llm-d (Mean) | Δ (llm-d - k8s) | Δ% vs k8s |
+| :--- | :--- | :--- | :--- | :--- |
+| Requests/sec | 5.7306 | 14.4123 | +8.6817 | +151.5% |
+| Input tokens/sec | 43,417.86 | 109,192.83 | +65,774.97 | +151.5% |
+| Output tokens/sec | 5,362.16 | 13,574.37 | +8,212.21 | +153.2% |
+| Total tokens/sec | 48,780.02 | 122,767.19 | +73,987.17 | +151.7% |
+| Request latency (s) | 105.4133 | 67.8649 | -37.5484 | -35.6% |
+| TTFT (s) | 34.9145 | 0.1203 | -34.7942 | -99.66% |
+| Time/output token (ms) | 70.42 | 67.66 | -2.76 | -3.9% |
+| Inter-token latency (ms) | 70.42 | 67.66 | -2.76 | -3.9% |
```
ahg-g, line 558: the last two are duplicates, TPOT and ITL are the same, right? I would remove the TPOT one. Same for the precise guide.

ahg-g (commented):
Great, thanks @zetxqx !

guides/inference-scheduling/README.md
zetxqx: yes, using mean, TPOT and ITL are basically the same. I removed the TPOT.

guides/precise-prefix-cache-aware/README.md
```diff
@@ -585,25 +587,26 @@ There is a report for each stage.
...
-| Time/output token (ms)                                            | 46.27       | 52.30       | 6.03           | 13.0%     |
-| Inter-token latency (ms)                                          | 30.57       | 41.46       | 10.89          | 35.6%     |
+The following graphs illustrate the relationship between latency, throughput, and QPS, as generated by the `inference-perf --analyze`. For benchmarking, we compared our results against a standard Kubernetes (k8s) service endpoint that routes traffic directly to vLLM pods.
+
+<img src="./benchmark-results/latency_vs_qps.png" width="900" alt="Throughput vs QPS">
+<img src="./benchmark-results/throughput_vs_qps.png" width="450" alt="Throughput vs Latency">
+
+The following data illustrates the performance of the stage at a fixed request rate of **40**. At this level, the llm-d reaches its saturation point, where increasing the request rate no longer improves throughput. These results are also benchmarked against a standard K8s service.
+
+- **Throughput**: Requests/sec **+132.2%** ; Output tokens/sec **+135.1%**
```
vMaroon: I dont think a placement induced spike should be picked as the representative point. I think either a different one should be chosen or even better: rerun the experiment.

zetxqx: I removed the warmup [stage](https://github.com/llm-d/llm-d/blob/4413af1f5efa436ee8451c629b0f991827bdeb68/guides/benchmark/inference_scheduling_guide_template.yaml#L35-L37) in the graph. from qps vs throughput graph, around QPS=40 is truly the saturation point, what spike are you referring to?

vMaroon: TTFT graph spikes at 40 QPS then settles back. This benchmark can be reproduced in a manner in which the line is always flat - it's a matter of early placement, which is why we need some of the active/running request scorers in the guides sooner than later.

zetxqx: I see, let me rerun using the current guide. if there is still some spike, I'll probably just use QPS=60 for now.  Agree with `In general we should introduce a throughput-oriented benchmark as well/instead later on.`

zetxqx: Image I did two extra run but I see spike at around QPS=40 for both. Do you know the reason?

liu-cong: not sure why it spiked at 40 and came down, @vMaroon . 

given the current result I suggest we show 3 request rates - 30, 40 60. if the active request scorer helps reducing the spike, we will update later and show the improvements.

Let's use the best results we have now and not block.

maugustosilva (dismissed):

vMaroon (approved):

PR #725, opened 2026-02-10
feat: add support to build for CPU on AMX

zdtsw:
- AVX2 is old one for Intel, we already have support for AVX512 the new version
- this change is only to add more support on AMX alongwith AVX512

# Notes
origin PR in https://github.com/llm-d/llm-d/pull/684


Gregory-Pereira (approved):
/lgtm

PR #726, opened 2026-02-10
pd + kv-cache-aware

Gregory-Pereira:
cc @sallyom @vMaroon @robertgshaw2-redhat 

This is for pytorch demo

### DIFF
--- a/guides/sally-pd-with-kv-aware/gaie-pd-kv/values.yaml
+++ b/guides/sally-pd-with-kv-aware/gaie-pd-kv/values.yaml
@@ -31,14 +31,26 @@ inferenceExtension:
       apiVersion: inference.networking.x-k8s.io/v1alpha1
       kind: EndpointPickerConfig
       plugins:
-        - type: single-profile-handler
+        - type: prefill-header-handler
+        - type: prefill-filter
+        - type: decode-filter
+        - type: max-score-picker
+        - type: queue-scorer
+          parameters:
+            hashBlockSize: 5
+            maxPrefixBlocksToMatch: 256
+            lruCapacityPerServer: 31250
+        - type: pd-profile-handler
+          parameters:
+            threshold: 0
+            hashBlockSize: 5
         - type: precise-prefix-cache-scorer
           parameters:
             tokenProcessorConfig:
               blockSize: 64
             indexerConfig:
               tokenizersPoolConfig:
-                modelName: "Qwen/Qwen3-32B"
+                modelName: "openai/gpt-oss-120b"
                 hf:
                   tokenizersCacheDir: "/tmp/tokenizers"
             kvEventsConfig:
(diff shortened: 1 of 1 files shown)


guides/sally-pd-with-kv-aware/gaie-pd-kv/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    pd-kv-aware.yaml: |
+      apiVersion: inference.networking.x-k8s.io/v1alpha1
+      kind: EndpointPickerConfig
+      plugins:
+        - type: prefill-header-handler
+        - type: prefill-filter
+        - type: decode-filter
+        - type: max-score-picker
+        - type: queue-scorer
+          parameters:
```
vMaroon, line 39: This should be removed (also where it was copied from).

guides/sally-pd-with-kv-aware/gaie-pd-kv/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+            hashBlockSize: 5
+            maxPrefixBlocksToMatch: 256
+            lruCapacityPerServer: 31250
+        - type: pd-profile-handler
+          parameters:
+            threshold: 0
```
vMaroon, line 45: FYI today the `precise-prefix-cache-scorer` only works with this threshold, it is not integrated into the pd-profile-handler like the `prefix-cache-scorer` is. @mayabar is working on a refactoring of the profile and UX.

guides/sally-pd-with-kv-aware/gaie-pd-kv/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+        plugins:
+        - pluginRef: prefill-filter
+        - pluginRef: precise-prefix-cache-scorer
+          weight: 3.0
+        - pluginRef: kv-cache-utilization-scorer
+          weight: 2.0
+        - pluginRef: queue-scorer
+          weight: 2.0
+        - pluginRef: max-score-picker
+      - name: decode
```
vMaroon, line 73: I think we need the `active-request-scorer` or the `running-requests-size-scorer` in this profile.

guides/sally-pd-with-kv-aware/gaie-pd-kv/values.yaml
Gregory-Pereira: So do that mean that this wont work for now?

guides/sally-pd-with-kv-aware/gaie-pd-kv/values.yaml
Gregory-Pereira: We dont have configs for this in any of the guides, we should implement in others too

PR #730, opened 2026-02-10
just deprecate queue scorrer params

Gregory-Pereira:
cc @vMaroon @ahg-g

### DIFF
--- a/guides/pd-disaggregation/gaie-pd/values.yaml
+++ b/guides/pd-disaggregation/gaie-pd/values.yaml
@@ -20,7 +20,6 @@ inferenceExtension:
       - type: prefill-header-handler
       - type: prefill-filter
       - type: decode-filter
-      - type: max-score-picker
       - type: queue-scorer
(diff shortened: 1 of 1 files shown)


guides/pd-disaggregation/gaie-pd/values.yaml
ahg-g: we can remove those two lines too, ditto to the decode profile below

ahg-g: lets remove the `max-score-picker` only (it is added by default), we can keep the weight for readability.

Gregory-Pereira: Does this commit: https://github.com/llm-d/llm-d/pull/730/changes/f5e86b5c9c4d7b708412731d8381f1e356f74c69 address your concern here? If so Ill do the same for the other guides

ahg-g: yes, thanks

PR #737, opened 2026-02-11
[lint] fixing main for linting

Gregory-Pereira:
cc @diegocastanibm just a linting PR. Nothing here should be destructive


zdtsw (commented):
/LGTM

PR #740, opened 2026-02-12
feat: add support to use kv-cache UDS tokenizer sidecar in guide

zdtsw:
# Chanages
- remove config fiedls which have been removed
- add new env, ports, probes and volumes used to do gRPC between EPP and UDS
- update image released by llm-d
- cherry-pick https://github.com/llm-d/llm-d/pull/799/ 

# Ref
https://github.com/llm-d/llm-d-inference-scheduler/pull/609
depend on https://github.com/llm-d/llm-d-kv-cache/pull/307 or have an official image release on v0.5.0

### DIFF
--- a/guides/precise-prefix-cache-aware/gaie-kv-events/values.yaml
+++ b/guides/precise-prefix-cache-aware/gaie-kv-events/values.yaml
@@ -37,7 +37,7 @@ inferenceExtension:
           key: HF_TOKEN
   sidecar:
     enabled: true
-    image: ghcr.io/llm-d/llm-d-uds-tokenizer:v0.5.0
+    image: ghcr.io/llm-d/llm-d-uds-tokenizer:v0.5.0-rc1
     imagePullPolicy: IfNotPresent
     name: tokenizer-uds
     # Both HF_HOME and TOKENIZERS_DIR are required to point to the same mounted directory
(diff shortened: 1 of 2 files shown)


guides/precise-prefix-cache-aware/gaie-kv-events/values.yaml
```diff
@@ -35,6 +35,27 @@ inferenceExtension:
         secretKeyRef:
           name: llm-d-hf-token
           key: HF_TOKEN
+  sidecar:
+    enabled: true
+    image: ghcr.io/llm-d/llm-d-uds-tokenizer:v0.5.0
+    imagePullPolicy: IfNotPresent
+    name: tokenizer-uds
+    env:
+      - name: TOKENIZERS_DIR
```
vMaroon, line 45: I didn't get to testing whether this work properly, might need changing. I did get it running and have benchmarked it with a similar setting, but need to verify this exact configuration.

guides/precise-prefix-cache-aware/gaie-kv-events/values.yaml
```diff
@@ -35,6 +35,27 @@ inferenceExtension:
         secretKeyRef:
           name: llm-d-hf-token
           key: HF_TOKEN
+  sidecar:
+    enabled: true
+    image: ghcr.io/llm-d/llm-d-uds-tokenizer:v0.5.0
```
vMaroon: `ghcr.io/llm-d/llm-d-uds-tokenizer:v0.5.1-rc1` was cut - we can switch and upgrade later.

guides/precise-prefix-cache-aware/gaie-kv-events/values.yaml
vMaroon: OK - tested a working config, must duplicate this with HF_HOME key (have both keys pointing to the mounted dir).

PR #743, opened 2026-02-12
Fix broken slack links in documentation

yuanwu2017:
Replace direct Slack archive links with the general workspace link to avoid Lychee link checker rate limiting (429 Too Many Requests).

### DIFF
--- a/ONBOARDING.md
+++ b/ONBOARDING.md
@@ -25,7 +25,7 @@ The purpose of this policy is to ensure a smooth on-boarding and off-boarding pr
 
 ### 2.3. Getting Help
 
-- The organization's maintainers are here to help contributors be efficient and confident in their collaboration effort. If you need help you can reach out to the maintainers on slack at the [llm-d-dev channel](https://llm-d.slack.com/archives/C08SH9K8JGK).
+- The organization's maintainers are here to help contributors be efficient and confident in their collaboration effort. If you need help you can reach out to the maintainers on slack in the #llm-d-dev channel (llm-d Slack workspace: https://llm-d.ai/slack).
 - Be sure to join the [llm-d-dev Google Group](https://groups.google.com/g/llm-d-contributors) to get access to important artifacts like proposals, diagrams, and meeting invitations.
 
 ### 2.4. Orientation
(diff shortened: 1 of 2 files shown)


ONBOARDING.md
```diff
@@ -25,7 +25,7 @@ The purpose of this policy is to ensure a smooth on-boarding and off-boarding pr
 
 ### 2.3. Getting Help
 
-- The organization's maintainers are here to help contributors be efficient and confident in their collaboration effort. If you need help you can reach out to the maintainers on slack at the [llm-d-dev channel](https://llm-d.slack.com/archives/C08SH9K8JGK).
+- The organization's maintainers are here to help contributors be efficient and confident in their collaboration effort. If you need help you can reach out to the maintainers on slack in the #llm-d-dev channel (llm-d Slack workspace: https://llm-d.ai/slack).
```
zdtsw, line 28: i think this is intended to direct ppl to llm-d-dev channel than the slack
unless you wanna change it to be `you can reach out to the maintainers on https://llm-d.ai/slack.`

SIGS.md
```diff
@@ -71,7 +71,7 @@ SIGs operate within the broader llm-d project governance framework defined in [P
 
 **💬 Communication**:
 
-- **Slack Channel**: [#sig-inference-scheduler](https://llm-d.slack.com/archives/C08SBNRRSBD)
+- **Slack Channel**: #sig-inference-scheduler (llm-d Slack workspace: https://llm-d.ai/slack)
```
zdtsw, line 74: similar comments on all changes below

ONBOARDING.md
yuanwu2017: You are right. But if directed to llm-d-dev, the frequent updates to the llm-d-dev channel may cause the lychee test to fail probabilistically.  Is it possible to omit the lychee test for this?

zdtsw: > dates to the llm-d-dev channel may cause the lychee test to fail probabilistically. Is it possible to omit the lychee test for this?

was lychee failing due to network untstability (re-run the job can fix it) or the URL does get changes from time to time? 
I will be a bit surprised if it is the later case. 
but if you really see this very often, one thing i can think is: something like
`if you need help you can reach out to the maintainers from [llm-d slack](https://llm-d.ai/slack) on the #llm-d-dev channel`
to ignore this in lychee config might point to a invalid URL to the user

PR #776, opened 2026-02-16
Enables AMD prefill/decode disaggregation well-lit path

vcave:
- `values_amd.yaml` to use GPU+NIC resources.
- `values_amd_host_network.yaml` to use `hostNetwork` for debugging.
- KV transfers are handled by RIXL

### DIFF
--- a/guides/pd-disaggregation/README.amd.md
+++ b/guides/pd-disaggregation/README.amd.md
@@ -192,7 +192,7 @@ For instructions on getting started making inference requests see [our docs](../
 
 ## Tuning Selective PD
 
-Selective PD is a feature in the `inference-scheduler` within the context of prefill-decode disaggregation, although it is disabled by default. This features enables routing to just decode even with the P/D deployed. To enable it, you will need to set `threshold` value for the `pd-profile-handler` plugin, in the [GAIE values file](./gaie-pd/values.yaml). You can see the value of this here:
+Selective PD is a feature in the `inference-scheduler` within the context of prefill-decode disaggregation, although it is disabled by default. This feature enables routing to just decode even with the P/D deployed. To enable it, you will need to set `threshold` value for the `pd-profile-handler` plugin, in the [GAIE values file](./gaie-pd/values.yaml). You can see the value of this here:
 
 ```bash
 cat gaie-pd/values.yaml | yq '.inferenceExtension.pluginsCustomConfig."pd-config.yaml"' | yq '.plugins[] | select(.type == "pd-profile-handler")'
@@ -204,7 +204,7 @@ parameters:
 
 Some examples in which you might want to do selective PD might include:
 
-* When the prompt is short enough that the amount of work split inference into prefill and decode phases, and then open a kv transfer between those two GPUs is greater than the amount of work to do both phases on the same decode inference worker.
+* When the prompt is short enough, the overhead of splitting inference into prefill and decode phases and transferring the KV cache between GPUs becomes larger than simply running both phases on a single decode inference worker.
 * When Prefill units are at full capacity.
 
 For information on this plugin, see our [`pd-profile-handler` docs in the inference-scheduler](https://github.com/llm-d/llm-d-inference-scheduler/blob/v0.3.0/docs/architecture.md?plain=1#L205-L210)


guides/pd-disaggregation/ms-pd/values_amd.yaml
```diff
@@ -0,0 +1,206 @@
...
+      path: "/metrics"
+      interval: "30s"
+  containers:
+    - name: "vllm"
+      image: ghcr.io/llm-d/llm-d-rocm:v0.5.0
+      modelCommand: vllmServe
+      args:
+        # Keep tensor-parallelism as the first set of arguments
+        - "--tensor-parallel-size"
+        - "4"
```
Gregory-Pereira: Remove. Either use `.decode.parallelism.tensor` or set the tensor-parallel-size flag like you do here, otherwise templating will duplicate this flag.

```suggestion
```

guides/pd-disaggregation/README.amd.md
Gregory-Pereira, line 1: cc @smarterclayton on docs review. I think the goal is to not have providers have their own README in line for guides, but guidance should be split up - infra / hardware discussion should go in the core docs, and we should integrate a small subsection of the PD readme talking about stuff unique to using the ROCM image.

Gregory-Pereira (commented):
I didnt get all the way through the readme content will come back and do another pass tomorrow but looks good overall

guides/pd-disaggregation/ms-pd/values_amd.yaml
```diff
@@ -0,0 +1,203 @@
...
+    llm-d.ai/accelerator-vendor: "amd"
+    llm-d.ai/model: "Llama-3.3-70B-Instruct-FP8-KV"
+
+accelerator:
+  type: amd
+
+routing:
+  servicePort: 8000
+  proxy:
+    image: ghcr.io/llm-d/llm-d-routing-sidecar:v0.5.0
```
diegocastanibm, line 21: I do not see a `v0.5.0` for the sidecar. The lastest is `v0.3.0`:
https://github.com/llm-d/llm-d-routing-sidecar/releases

guides/pd-disaggregation/ms-pd/values_amd_host_network.yaml
```diff
@@ -0,0 +1,251 @@
...
+        #  value: "self,sm,rocm,rc"
+        # Customize network devices ; see docs
+        #- name: UCX_NET_DEVICES
+        #  value: "mlx5_0:1,mlx5_2:1,mlx5_3:1,mlx5_4:1,mlx5_5:1,mlx5_7:1,mlx5_8:1,mlx5_9:1"
+        # Customize RDMA GID ; see docs
+        #- name: UCX_IB_GID_INDEX
+        #  value: "3"
+        # Customize KV-transfer traffic class ; see docs
+        #- name: UCX_IB_TRAFFIC_CLASS
+        #  value: "41"
```
diegocastanibm: Remove comment section

guides/pd-disaggregation/README.amd.md
```diff
@@ -0,0 +1,271 @@
...
+
+1. GPU Support
+
+* [AMD GPU Operator](https://instinct.docs.amd.com/projects/gpu-operator/en/latest/index.html): Required to expose GPU devices to containers.
+
+2. Network Support (NIC)
+
+* A network operator is required to expose NIC devices. Choose the operator that matches your hardware:
+** [AMD Network Operator](https://instinct.docs.amd.com/projects/network-operator/en/main/overview.html): For AMD AINIC hardware.
+** [NVIDIA Network Operator](https://docs.nvidia.com/networking/display/kubernetes2410/nvidia+network+operator): For NVIDIA NIC hardware.
```
diegocastanibm: ** is not valid markdown for nested list items. This will render as bold text rather than a sub-list.
```suggestion
- [AMD Network Operator](https://instinct.docs.amd.com/projects/network-operator/en/main/overview.html): For AMD AINIC hardware.
- [NVIDIA Network Operator](https://docs.nvidia.com/networking/display/kubernetes2410/nvidia+network+operator): For NVIDIA NIC hardware.
```

guides/pd-disaggregation/README.amd.md
```diff
@@ -0,0 +1,271 @@
...
+
+```bash
+# Remove the model services
+helmfile destroy -n ${NAMESPACE}
+
+# Remove the infrastructure
+helm uninstall ms-pd -n ${NAMESPACE}
+helm uninstall gaie-pd -n ${NAMESPACE}
+helm uninstall infra-pd -n ${NAMESPACE}
+```
```
diegocastanibm, line 218: My understanding is that `helmfile destroy` already removes all releases managed by the `helmfile` (`ms-pd`, `gaie-pd`, `infra-pd`). The subsequent `helm uninstall` commands would fail since those releases would already be gone. Should be one or the other maybe?

guides/pd-disaggregation/helmfile.yaml.gotmpl
```diff
@@ -96,7 +98,13 @@ releases:
       - {{ printf "infra-%s" $rn | quote }}
       - {{ printf "gaie-%s" $rn | quote }}
     values:
-    {{- if eq .Environment.Name "xpu" }}
+    {{- if eq .Environment.Name "amd" }}
+      {{- if env "USE_HOST_NETWORK" }}
```
diegocastanibm: use an explicit check like `eq (env "USE_HOST_NETWORK") "yes"`

guides/pd-disaggregation/ms-pd/values_amd.yaml
vcave: This repo is deprecated, this one has the newer image https://github.com/llm-d/llm-d-inference-scheduler/tags

guides/pd-disaggregation/ms-pd/values_amd_host_network.yaml
vcave: I was leaving it here for reference but can fold it into the documentation instead. Would still like to leave one comment to point to the documentation in case people missed it.

diegocastanibm: After a conversation with @Gregory-Pereira , you can leave it, but I suggest that you remove the duplication since you have `UCX_PROTO_INFO` twice:
```suggestion
        # UCX Logging and debugging
        #- name: UCX_PROTO_INFO
        #  value: "y"
        #- name: UCX_LOG_LEVEL
        #  value: "info"
        # Customize transport layers ; see docs
        #- name: UCX_TLS
        #  value: "self,sm,rocm,rc"
        # Customize network devices ; see docs
        #- name: UCX_NET_DEVICES
        #  value: "mlx5_0:1,mlx5_2:1,mlx5_3:1,mlx5_4:1,mlx5_5:1,mlx5_7:1,mlx5_8:1,mlx5_9:1"
        # Customize RDMA GID ; see docs
        #- name: UCX_IB_GID_INDEX
        #  value: "3"
        # Customize KV-transfer traffic class ; see docs
        #- name: UCX_IB_TRAFFIC_CLASS
        #  value: "41"
```

guides/pd-disaggregation/README.amd.md
diegocastanibm: @clubanderson , Will the `helm uninstall`'s make the nightly's fail?

Gregory-Pereira: helmfile destroy calls helm uninstall. Typically in our readmes we have that as one or the other

guides/pd-disaggregation/ms-pd/values_amd_host_network.yaml
```diff
@@ -0,0 +1,253 @@
...
```
tlrmchlsmth: We need these for running on OCI - I know we use helm instead of kustomize for the pd example but should we add a values_rocm_oci.yaml with these envs in place?

vcave: This was more for debug but I understand it can get confusing if people start using it without understanding the consequences. Removed.

guides/pd-disaggregation/README.amd.md
```diff
@@ -0,0 +1,234 @@
...
+
+**_NOTE:_** This assumes no other guide deployments in your given `${NAMESPACE}` and you have not changed the default release names via the `${RELEASE_NAME}` environment variable.
+
+## Using the stack
+
+For instructions on getting started making inference requests see [our docs](../../docs/getting-started-inferencing.md)
+
+## Tuning Selective PD
+
+Selective PD is a feature in the `inference-scheduler` within the context of prefill-decode disaggregation, although it is disabled by default. This features enables routing to just decode even with the P/D deployed. To enable it, you will need to set `threshold` value for the `pd-profile-handler` plugin, in the [GAIE values file](./gaie-pd/values.yaml). You can see the value of this here:
```
diegocastanibm:
```suggestion
Selective PD is a feature in the `inference-scheduler` within the context of prefill-decode disaggregation, although it is disabled by default. This feature enables routing to just decode even with the P/D deployed. To enable it, you will need to set `threshold` value for the `pd-profile-handler` plugin, in the [GAIE values file](./gaie-pd/values.yaml). You can see the value of this here:
```

guides/pd-disaggregation/README.amd.md
```diff
@@ -0,0 +1,234 @@
...
+cat gaie-pd/values.yaml | yq '.inferenceExtension.pluginsCustomConfig."pd-config.yaml"' | yq '.plugins[] | select(.type == "pd-profile-handler")'
+type: pd-profile-handler
+parameters:
+  threshold: 0 # update this
+  hashBlockSize: 5
+```
+
+Some examples in which you might want to do selective PD might include:
+
+* When the prompt is short enough that the amount of work split inference into prefill and decode phases, and then open a kv transfer between those two GPUs is greater than the amount of work to do both phases on the same decode inference worker.
```
diegocastanibm: This sentence is not clear to me. I suggest the sentence:
```suggestion
* When the prompt is short enough, the overhead of splitting inference into prefill and decode phases and transferring the KV cache between GPUs becomes larger than simply running both phases on a single decode inference worker.
```

diegocastanibm (approved):
Just some small comments to improve the text. Otherwise /lgtm

PR #783, opened 2026-02-17
✨ Fix dev image latest tag + nightly build + image_override for all E2E nightlies

clubanderson:
## Summary
- **Fix `latest` tagging** in `build-image.yml`: CUDA and AWS manifest jobs now apply a `latest` tag when building from the default branch (not PRs). Previously only `sha-SHORT_SHA` and `pr-NUMBER` tags were created.
- **Add `force_build` input** to `build-image.yml`: bypasses change detection to build all platforms, enabling a nightly caller.
- **New `nightly-build-image.yaml`**: runs at 04:00 UTC daily, calls `build-image.yml` with `force_build: true` to ensure fresh dev images before E2E nightlies (06:00+ UTC).
- **Add `image_override`** to all 12 nightly E2E callers: passes `ghcr.io/llm-d/llm-d-cuda-dev:latest` to the reusable workflows so nightlies test the latest dev image from main.
- For callers with inline manifests (wide-ep-lws on all platforms, tiered-prefix-cache on OCP), adds `sed`-based image override to replace hardcoded release images.

## Motivation
Nightly E2E workflows deployed hardcoded release images (`ghcr.io/llm-d/llm-d-cuda:v0.5.0`) that didn't include recent fixes from main (e.g. CUDA compat fix from PR #727 merged after the v0.5.0 release). This caused CUDA Error 803 failures on CKS nightlies. The nightlies should test the latest code from main, not pinned releases.

## Depends on
- llm-d/llm-d-infra#28 (adds `image_override` input to all 3 reusable nightly workflows)

## Test plan
- [ ] Verify `latest` tag logic only fires on default branch, never PRs
- [ ] `force_build` defaults to `false` — no impact on existing `ci-pr-test.yaml` caller
- [ ] After both PRs merge: trigger `nightly-build-image.yaml`, verify `latest` tag appears on GHCR
- [ ] Trigger any nightly, verify pods use `ghcr.io/llm-d/llm-d-cuda-dev:latest`

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -259,6 +259,17 @@ jobs:
           if-no-files-found: error
           retention-days: 1
 
+      - name: Tag latest on default branch
+        if: |
+          steps.build-debug.outputs.digest != '' &&
+          github.ref == format('refs/heads/{0}', github.event.repository.default_branch) &&
+          github.event_name != 'pull_request'
+        run: |
+          IMAGE="${{ env.REGISTRY }}/${{ github.repository }}-cuda${{ matrix.image_suffix }}"
+          docker buildx imagetools create \
+            -t "${IMAGE}:latest" \
+            "${IMAGE}@${{ steps.build-debug.outputs.digest }}"
+
   cuda-build-llm-d:
     needs: [detect-changes]
     if: ${{ needs.detect-changes.outputs.cuda == 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-image.yml
Gregory-Pereira, line 148: Add conditional logic here since were not routing override through path filter:

```suggestion
    if: ${{ needs.detect-changes.outputs.cuda == 'true' || steps.force.outputs.cuda == 'true' }}
```

clubanderson: Good catch on the intent, but this is already handled\! The `detect-changes` job outputs at lines 80-85 already fold in the force step:

```yaml
cuda: ${{ steps.force.outputs.cuda || steps.filter.outputs.cuda || steps.dispatch.outputs.cuda }}
```

So `needs.detect-changes.outputs.cuda` already includes the force override — no change needed here. (Also, `steps.force` wouldn't be accessible from a downstream job since it's a step inside `detect-changes`.)

Gregory-Pereira, line 1: I see you added the changes of latest for CUDA and AWS, we should also do this for cuda-debug, cpu and xpu

clubanderson: Done — added `latest` tagging for all three in cc1225e:

- **cuda-debug**: `imagetools create -t :latest` step after the push-by-digest build
- **xpu**: conditional `latest` tag in `build-push-action` tags field
- **cpu**: same pattern as xpu

Gregory-Pereira (approved):
/lgtm

PR #788, opened 2026-02-17
🌱 OCP H100 nodeSelector + CKS GPU preemption for nightly E2E

clubanderson:
## Summary
- **OCP callers**: Adds H100 nodeSelector (`nvidia.com/gpu.product=NVIDIA-H100-80GB-HBM3`) to inference-scheduling and precise-prefix-cache OCP nightlies via `pre_deploy_script`
- **CKS callers**: Enables `allow_gpu_preemption: true` on all 4 CKS nightlies (IS, PD, WVA, wide-ep-lws) so they can proceed when `hpc-verification` (priority -1) is consuming GPUs

## Context
**OCP**: pok-prod cluster has H100 nodes. Without explicit nodeSelector, pods may land on nodes without the expected GPU type.

**CKS**: `hpc-verification` runs every hour for ~59 minutes at priority -1, consuming all 64 GPUs. Our nightly pods (default priority 0) can preempt these, but the GPU check hard-fails before pods are created. With `allow_gpu_preemption=true` + llm-d-infra PR #31 (merged), the check only proceeds if blocking pods have negative priority.

## Test plan
- [ ] Trigger IS-OCP or PPC-OCP nightly — verify nodeSelector appears in deployed values
- [ ] Trigger any CKS nightly while hpc-verification runs — verify it proceeds with preemption warning

### DIFF
--- a/.github/workflows/nightly-e2e-inference-scheduling-cks.yaml
+++ b/.github/workflows/nightly-e2e-inference-scheduling-cks.yaml
@@ -42,5 +42,6 @@ jobs:
       pod_wait_timeout: '30m'
       pod_readiness_delay: 180
       image_override: 'ghcr.io/llm-d/llm-d-cuda-dev:latest'
+      allow_gpu_preemption: true
       skip_cleanup: ${{ github.event.inputs.skip_cleanup == 'true' }}
     secrets: inherit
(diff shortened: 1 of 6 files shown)


Gregory-Pereira (approved):
/lgtm

PR #791, opened 2026-02-18
Image tagging improvements

Gregory-Pereira:
These improvements should increase the UX of consuming the images.
- Capture short sha early so it can't change in between build + output
- Add long sha for easier referencing
- Enable matrix builds everywhere but still only build default (amd64 on rhel) for AWS and cuda debug


Gregory-Pereira:
cc @diegocastanibm @clubanderson

maugustosilva (approved):

PR #795, opened 2026-02-18
refactor: install-vllm.sh

zdtsw:
# Changes
- make it easy to follow 3 different modes
- add more comments for env variable/flags
- only doing "git clone" conditionally

### DIFF
--- a/docker/scripts/cuda/runtime/install-vllm.sh
+++ b/docker/scripts/cuda/runtime/install-vllm.sh
@@ -29,8 +29,11 @@ set -euo pipefail
 # CUDA_MINOR                      - CUDA minor version (e.g., 9)
 # FLASHINFER_VERSION              - FlashInfer version for flashinfer-cubin and flashinfer-jit-cache (e.g., v0.5.2)
 # BUILD_NIXL_FROM_SOURCE          - Whether NIXL was built from source ("true"/"false")
+# SUPPRESS_PYTHON_OUTPUT          - Suppress verbose pip output ("true"/"1", optional)
 #
 # ============================================================================
+
+: "${SUPPRESS_PYTHON_OUTPUT:=}"
 
 . /opt/vllm/bin/activate
 
@@ -100,7 +103,7 @@ else
 
   # MODE 2
   if [ "${VLLM_USE_PRECOMPILED}" = "1" ] && [ -n "${WHEEL_URL}" ]; then
-    echo "Using precompiled binaries and shared libraries from source code commit: ${VLLM_PRECOMPILED_WHEEL_COMMIT} (source: ${VLLM_COMMIT_SHA})."
+    echo "Using precompiled binaries from commit: ${VLLM_PRECOMPILED_WHEEL_COMMIT} (with source code from: ${VLLM_COMMIT_SHA})."
     export VLLM_USE_PRECOMPILED=1      # Do not really need set it here as it is done in vllm envs.py by VLLM_PRECOMPILED_WHEEL_LOCATION
     export VLLM_PRECOMPILED_WHEEL_LOCATION="${WHEEL_URL}"
     INSTALL_PACKAGES+=(-e /opt/vllm-source)
--- a/docker/scripts/cuda/runtime/install-vllm.sh
+++ b/docker/scripts/cuda/runtime/install-vllm.sh
@@ -30,6 +30,7 @@ set -euo pipefail
 # FLASHINFER_VERSION              - FlashInfer version for flashinfer-cubin and flashinfer-jit-cache (e.g., v0.5.2)
 # BUILD_NIXL_FROM_SOURCE          - Whether NIXL was built from source ("true"/"false")
 # SUPPRESS_PYTHON_OUTPUT          - Suppress verbose pip output ("true"/"1", optional)
+# NVSHMEM_DIR                      - Path to compiled NVSHMEM installation (optional; triggers uninstall of vLLM's NVSHMEM wheel)
 #
 # ============================================================================
 
@@ -104,7 +105,7 @@ else
   # MODE 2
   if [ "${VLLM_USE_PRECOMPILED}" = "1" ] && [ -n "${WHEEL_URL}" ]; then
     echo "Using precompiled binaries from commit: ${VLLM_PRECOMPILED_WHEEL_COMMIT} (with source code from: ${VLLM_COMMIT_SHA})."
-    export VLLM_USE_PRECOMPILED=1      # Do not really need set it here as it is done in vllm envs.py by VLLM_PRECOMPILED_WHEEL_LOCATION
+    export VLLM_USE_PRECOMPILED=1      # Do not really need to set it here as it is done in vllm envs.py by VLLM_PRECOMPILED_WHEEL_LOCATION
     export VLLM_PRECOMPILED_WHEEL_LOCATION="${WHEEL_URL}"
     INSTALL_PACKAGES+=(-e /opt/vllm-source)
 
@@ -131,7 +132,7 @@ uv pip install ${VERBOSE_FLAG} "${INSTALL_PACKAGES[@]}" \
 # Uninstall the NVSHMEM dependency brought in by vLLM if using a compiled NVSHMEM
 # We built our own NVSHMEM in the dockerfile builder stage, so we don't need the one from vLLM as dependency
 if [[ "${NVSHMEM_DIR-}" != "" ]]; then
-  uv pip uninstall -y nvidia-nvshmem-cu${CUDA_MAJOR}
+  uv pip uninstall "nvidia-nvshmem-cu${CUDA_MAJOR}"
 fi
 
 # Clean up
--- a/docker/scripts/cuda/runtime/install-vllm.sh
+++ b/docker/scripts/cuda/runtime/install-vllm.sh
@@ -70,7 +70,7 @@ else
   esac
 
   # If no wheel found, set to empty string and fall to WHEEL_URL=""
-  WHEEL_FILENAME=$(echo "${WHEEL_INDEX_HTML}" | grep -oE "vllm-[^\"]+${PLATFORM_TAG}\.whl" | head -1 || echo "")
+  WHEEL_FILENAME=$(echo "${WHEEL_INDEX_HTML}" | { grep -oE "vllm-[^\"]+${PLATFORM_TAG}\.whl" || true; } | head -1)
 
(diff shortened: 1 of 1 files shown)


docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -1,121 +1,140 @@
...
-# install all packages in one command; enable verbose output by default to help prevent GHA timeouts (can be suppressed via SUPPRESS_PYTHON_OUTPUT)
-# use flashinfer wheel index for jit-cache pre-built binaries
 CUDA_SHORT_VERSION="cu${CUDA_MAJOR}${CUDA_MINOR}"
 VERBOSE_FLAG="-v"
 if [ "${SUPPRESS_PYTHON_OUTPUT}" = "true" ] || [ "${SUPPRESS_PYTHON_OUTPUT}" = "1" ]; then
   VERBOSE_FLAG=""
 fi
-uv pip install ${VERBOSE_FLAG} "${INSTALL_PACKAGES[@]}" \
+uv pip install "${VERBOSE_FLAG}" "${INSTALL_PACKAGES[@]}" \
   --extra-index-url "https://flashinfer.ai/whl/${CUDA_SHORT_VERSION}"
```
Gregory-Pereira, line 133: Quoting `VERBOSE_FLAG` in uv pip install "${VERBOSE_FLAG}" means an empty string is still passed as an argument when suppression is enabled. Previously the empty expansion disappeared entirely. I think this should stay unquoted, or be handled by conditionally building the command/args array instead.

docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -1,121 +1,140 @@
...
+  WHEEL_URL=""
 else
+  echo "DEBUG: Architecture: $(uname -m), Python: $(python3 --version)"
+  MACHINE=$(uname -m)
+  case "${MACHINE}" in
+    x86_64|amd64) PLATFORM_TAG="manylinux_2_31_x86_64" ;;
+    aarch64|arm64) PLATFORM_TAG="manylinux_2_31_aarch64" ;;
+    *) echo "ERROR: Unsupported architecture: ${MACHINE}"; exit 1 ;;
+  esac
   WHEEL_FILENAME=$(echo "${WHEEL_INDEX_HTML}" | grep -oE "vllm-[^\"]+${PLATFORM_TAG}\.whl" | head -1)
```
Gregory-Pereira: `set -euo pipefail` above now interacts badly with the wheel scrape pipeline. The script is supposed to tolerate “no matching wheel found” and fall back to source build, but with `pipefail` enabled this line will exit the script when `grep` finds no match. Either drop the strict pipefail above or refactor this

Gregory-Pereira (changes requested):
Good refactor and can help us tighten our images / build process. 2 potential issues I want to call out

docker/scripts/cuda/runtime/install-vllm.sh
zdtsw: updated

docker/scripts/cuda/runtime/install-vllm.sh
zdtsw: updated

diegocastanibm: This hasn't been fix I suppose, right? I still see the `pipefail`.

docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -1,121 +1,143 @@
...
-  # URL-encode the + sign in the wheel filename
-  WHEEL_URL="https://wheels.vllm.ai/${VLLM_PRECOMPILED_WHEEL_COMMIT}/${WHEEL_FILENAME}"
-  WHEEL_URL=$(echo "${WHEEL_URL}" | sed -E 's/\+/%2B/g')
-  echo "DEBUG: Found wheel: ${WHEEL_FILENAME}"
-  echo "DEBUG: Wheel URL: ${WHEEL_URL}"
-else
-  WHEEL_URL=""
-  echo "DEBUG: No wheel found for platform: ${PLATFORM_TAG}"
+  # If no wheel found, set to empty string and fall to WHEEL_URL=""
+  WHEEL_FILENAME=$(echo "${WHEEL_INDEX_HTML}" | grep -oE "vllm-[^\"]+${PLATFORM_TAG}\.whl" | head -1 || echo "")
```
diegocastanibm: The `|| echo ""` only catches head failing. 
I asked Claude and the solution that is proposing is:
`WHEEL_FILENAME=$(echo "${WHEEL_INDEX_HTML}" | { grep -oE "vllm-[^\"]+${PLATFORM_TAG}\.whl" || true; } | head -1)`

docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -1,121 +1,143 @@
...
+# shellcheck disable=SC2086
 uv pip install ${VERBOSE_FLAG} "${INSTALL_PACKAGES[@]}" \
   --extra-index-url "https://flashinfer.ai/whl/${CUDA_SHORT_VERSION}"
 
-# uninstall the NVSHMEM dependency brought in by vllm if using a compiled NVSHMEM
+# Uninstall the NVSHMEM dependency brought in by vLLM if using a compiled NVSHMEM
+# We built our own NVSHMEM in the dockerfile builder stage, so we don't need the one from vLLM as dependency
 if [[ "${NVSHMEM_DIR-}" != "" ]]; then
-  uv pip uninstall nvidia-nvshmem-cu${CUDA_MAJOR}
+  uv pip uninstall "nvidia-nvshmem-cu${CUDA_MAJOR}"
```
diegocastanibm: maybe adding the `-y` flag to avoid prompt for confirmation?

docker/scripts/cuda/runtime/install-vllm.sh
zdtsw: to keep `pipefai`: if wheel failed to be found, it set `WHEEL_FILENAME` to "" and then `WHEEL_URL`="" then trigger build from source.
it should be ok to keep it. 
currently we do not have anything in `install-vllm.sh` required for pipefail, 
i am fine, if we want to take it away.

diegocastanibm: we can keep it, but then, we need to adapt 73, am I right?
Check what I commented here:
https://github.com/llm-d/llm-d/pull/795#discussion_r2997325067

Is the `|| echo ""` only catching head failing? I'm struggling to understand that part . If the theory behind your decision is that `head -1` fails on an empty string, that is not true. I have checked it in my terminal using:
```
$ echo "" | grep -oE "vllm-[^\"]+${PLATFORM_TAG}\.whl" | head -1 || echo "blabla"
```
This should show `blablabla` but is showing empty string. Could you clarify this part for me? (either in this comment or in the comment below).
Thanks

zdtsw: sorry i did not follow
i have updated it to be
`WHEEL_FILENAME=$(echo "${WHEEL_INDEX_HTML}" | { grep -oE "vllm-[^\"]+${PLATFORM_TAG}\.whl" || true; } | head -1)` + keep `pipefail`
should it work, or?

as for test,  positive case, we use wheel
```
[wenzhou@ (⎈|N/A:N/A)]$ WHEEL_FILENAME=$(echo "abc" | grep -oE "abc" | head -1 || echo "blabla")
[wenzhou@ (⎈|N/A:N/A)]$ echo $WHEEL_FILENAME
abc
```
negative case 1 if `WHEEL_INDEX_HTML` has a value but not match, and we build from source
```
[wenzhou@ (⎈|N/A:N/A)]$ WHEEL_FILENAME=$(echo "def" | grep -oE "abc" | head -1 || echo "balaba")
[wenzhou@ (⎈|N/A:N/A)]$ echo $WHEEL_FILENAME

[wenzhou@ (⎈|N/A:N/A)]$ 
```
negative case 2 when `curl -sf  `failed and  `WHEEL_INDEX_HTML` assigned to "" and we build from source
```
[wenzhou@ (⎈|N/A:N/A)]$ WHEEL_FILENAME=$(echo "" | grep -oE "abc" | head -1 || echo "balabla")
[wenzhou@ (⎈|N/A:N/A)]$ echo $WHEEL_FILENAME
```

diegocastanibm: Sorry to insist but I need to understand this; In the case of the negative case 1 that you are posting above:
```
negative case 1 if WHEEL_INDEX_HTML has a value but not match, and we build from source

[wenzhou@ (⎈|N/A:N/A)]$ WHEEL_FILENAME=$(echo "def" | grep -oE "abc" | head -1 || echo "balaba")
[wenzhou@ (⎈|N/A:N/A)]$ echo $WHEEL_FILENAME

[wenzhou@ (⎈|N/A:N/A)]$ 
```
don't you want to see "blablabla" ("" in the original code)? That means that the `|| echo "blablabla"` is not running

zdtsw: if i see "blabala" then it goes to "echo "DEBUG: Found wheel:..."
but we do want to build from source when something is wrong, right? 

basically in this case, it does not matter echo "" or  echo "balabala" , it almost not possible to get a exit non-zero before || to trigger the echo part.

if we take it step by step
so you get a `WHEEL_URL` return , either valid html or empty string,
then grep with the wheel we want, either return the first line with the real name, or empty string
to have `echo` in the end is for a rare case if above step failed (not likely) and still we have something to feed `WHELL_FILENAME` so it can trigger the build from source.

docker/scripts/cuda/runtime/install-vllm.sh
diegocastanibm: This comment has not been applied yet

zdtsw: ah, thanks, 
i finally understand the problem.
i push the branch to my [forked](https://github.com/zdtsw/llm-d/commits/chore_install_vllm/) one and this PR is in the upstream branch. thats why we were talking about different code

diegocastanibm (dismissed):
We need to pass the tests, but ok otherwise. Thanks @zdtsw

diegocastanibm (approved):
Please, do not merge until all tests are passed

zdtsw:
i lost permission to fix conflicts to chore_install_vllm branch so open a new one https://github.com/llm-d/llm-d/pull/1401

PR #796, opened 2026-02-18
Proposal: Prism - Performance analysis for distributed inference systems

seanhorgan:
Community-driven standard & reference for visualizing performance benchmarks from various sources

### DIFF
--- a/docs/proposals/prism-performance-analysis.md
+++ b/docs/proposals/prism-performance-analysis.md
@@ -8,9 +8,10 @@ Currently, AI Platform Engineers and ML Engineers face significant challenges as
 
 We are proposing to donate the Prism codebase (currently at [https://github.com/seanhorgan/prism](https://github.com/seanhorgan/prism)) to the `llm-d` organization to:
 
-1.  Establish a community-driven standard & reference for visualizing performance of distributed inference systems.
+1.  Establish a community-driven reference for visualizing performance of distributed inference systems.
 2.  Allow the community to contribute high-quality benchmarks from various sources, e.g. results that conform to the standard `llm-d` benchmark format.
-3.  Make it easier to validate benchmarks and deploy optimized `llm-d` stacks.
+3.  Assist `llm-d` users in visualizing results of their benchmark sweeps
+4.  Make it easier to validate benchmarks and deploy optimized `llm-d` stacks.
 
 ## Goals
 


docs/proposals/prism-performance-analysis.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Proposal
+
+We propose the following:
+
+1.  Create a new repository `llm-d/prism` to host the source code for the Prism application.
+2.  Deploy Prism as a publicly accessible application that is linked to from the `llm-d.ai` website.
+3.  Create a process for the full lifecycle of benchmarks (e.g. ingesting, validating, publishing, archiving) new benchmarks from `llm-d` feature developers and community members, and making them accessible through Prism.
+
+## User Stories (CUJs)
```
smarterclayton, line 35: I think there should be a self deployed prism mode for people who want to visualize their own results.

docs/proposals/prism-performance-analysis.md
```diff
@@ -0,0 +1,110 @@
...
+
+**Prism** aims to solve this by providing a streamlined, intuitive interface for discovering, comparing, and reproducing benchmarks for state-of-the-art distributed inference systems that scale from single-node to complex multi-node disaggregated environments and agentic & RL systems.
+
+We are proposing to donate the Prism codebase (currently at [https://github.com/seanhorgan/prism](https://github.com/seanhorgan/prism)) to the `llm-d` organization to:
+
+1.  Establish a community-driven standard & reference for visualizing performance of distributed inference systems.
+2.  Allow the community to contribute high-quality benchmarks from various sources, e.g. results that conform to the standard `llm-d` benchmark format.
+3.  Make it easier to validate benchmarks and deploy optimized `llm-d` stacks.
```
smarterclayton: I'd add one more "Assist llm-d users in visualizing results of their benchmark sweeps"

smarterclayton (changes requested):

docs/proposals/prism-performance-analysis.md
```diff
@@ -0,0 +1,110 @@
...
+Prism is architected as a high-performance Single Page Application (SPA) backed by a lightweight Node.js proxy.
+
+*   **Frontend:** React-based dashboard featuring a high-density filtering system and a WebGL-accelerated charting engine.
+*   **BFF Proxy (Backend for Frontend):** A Node.js service that handles authentication and proxies requests to external APIs (like Google Cloud). It supports Application Default Credentials (ADC) for seamless access to organization-wide storage buckets without exposing tokens to the client.
+*   **Persistence:** User preferences, active data connections, and filter states are persisted in local storage, allowing for a tailored, stateful work session.
+
+### Data Ingestion: Catalog-First Architecture
+Prism utilizes a "Catalog-First" approach to manage disparate benchmark sources via a unified ingestion layer. This allows users to mix-and-match official baselines with private experimental data.
+
+*   **Connectors:**
```
wseaton, line 82: Is the scope of prism just benchmark results? 

Similar infrastructure could be useful for the project for behavior testing based on analysis of the metrics bundle generated during a run (ex. epp, and vllm pods). I've been thinking about this mainly in the context of regression testing, like how do we prove that "router stickiness" is fixed with certain plugin configurations.

@smarterclayton if this is not in scope for prism, we should think about how we organize a shared community testing approach around this as well. Mainly standardization around metrics bundle format and collection methodology.

smarterclayton: In previous experience with this, something more general than prism is probably necessary.  Could have some similarities, but prism is really optimized for the core experience around sweeps and a latency throughput curve.  Probably wouldn't be as useful outside / be hard to bridge the use cases.

docs/proposals/prism-performance-analysis.md
```diff
@@ -0,0 +1,110 @@
+# Proposal: Prism - Performance analysis for distributed inference systems
+
+## Motivation
+
+Currently, AI Platform Engineers and ML Engineers face significant challenges assembling the full end-to-end inference serving stack for their applications, leading to lengthy, manual evaluation cycles, suboptimal performance, and unnecessarily high friction & costs. While many benchmarks and tools exist, the data follows different methodologies, formats, and is often scattered across disconnected docs, spreadsheets, or vendor-specific sites.
+
+**Prism** aims to solve this by providing a streamlined, intuitive interface for discovering, comparing, and reproducing benchmarks for state-of-the-art distributed inference systems that scale from single-node to complex multi-node disaggregated environments and agentic & RL systems.
+
+We are proposing to donate the Prism codebase (currently at [https://github.com/seanhorgan/prism](https://github.com/seanhorgan/prism)) to the `llm-d` organization to:
```
natoscott, line 9: Seems a bit of a stretch to say this is a "Community-driven standard" when neither the tool nor the source code is available for community review and evaluation?  We seem to have the horse before the cart here.

But other than that, thanks in advance - looking forward to it - sounds like an interesting contribution!

docs/proposals/prism-performance-analysis.md
seanhorgan: Agreed - supporting users that want to visualize their own results and compare them with published benchmarks is important. Right now a user can paste output from inference-perf into prism or add a GCS bucket with results. A user can also deploy their own version of the app.

docs/proposals/prism-performance-analysis.md
seanhorgan: Right now prism is focused on comparing benchmarks around sweeps & baselines focused on performance metrics along with cost & quality (on the horizon). There's a gray area with broader real-time observability use cases and you are right to call out the importance of standardization around metrics. The llm-d benchmark sig has put forward this standard: https://github.com/llm-d/llm-d-benchmark/blob/main/benchmark_report/br_v0_2_json_schema.json.

docs/proposals/prism-performance-analysis.md
seanhorgan: Thanks for the comment. Sorry, I shared the code with some of the llm-d leads but wasn't sure what the process was for sharing the code publicly as part of the proposal.

In terms of standards, the more important one here is the benchmark results format as that makes it quite easy to visualize and compare with others: 

https://github.com/llm-d/llm-d-benchmark/blob/main/benchmark_report/br_v0_2_json_schema.json

docs/proposals/prism-performance-analysis.md
seanhorgan: added.

docs/proposals/prism-performance-analysis.md
natoscott: No problem and thanks for following up.  I was mainly keen to see what Prism is all about (beyond this proposal write-up), to contrast it to comparable tools / services like https://github.com/SemiAnalysisAI/InferenceX - without the github repo it's difficult.

> what the process was for sharing the code publicly

Is it to be released under an open source license?  Assuming yes, assuming Apache like rest of llm-d - I recommend opening up your referenced github repo for public access so this proposal can be considered in its entirety by the community.

chcost: @natoscott thanks for the comment, and @seanhorgan for following up. Yes, this would be added as a new repo under llm-d org and under the same license. This proposal template is the first step in our contribution process, with the creation of the repo for code transfer and comment by the community next. Coming up next, stay tuned!

smarterclayton (approved):

PR #819, opened 2026-02-19
Release doc

diegocastanibm:
**Description:**
`RELEASE.md` describes the release process followed by the release engineers.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,12 +6,46 @@ This document describes the release process for llm-d. The release dates should 
 
 ### 1. Feature freeze
 
-This section describes the general process for tracking work for a release, which involves attending lead syncs and SIG meetings. There is also a milestone release tracker that needs to be kept up to date. The release tracker can be found in the [milestones](https://github.com/llm-d/llm-d/milestones) section of the repository.
+  The feature freeze phase ensures that all planned work for a release is identified, tracked, and finalized before the pre-release phase begins.
 
-The release tracker is created 1-2 weeks into the sprint. 4-weeks later, the release tracker is frozen and it should contain all the features that will be in the release. Each feature should have a specific example: the example should be production-ready, meaning that the feature should show an example that is realistic and that can be used in production. Toy-examples should be avoided.
+#### Release tracker
 
-> [!NOTE]
-> The release tracker is not necessarily accurate. However, in the future, we want to keep this tracker up to date and as accurate as possible.
+  Each release is tracked using a GitHub [milestone](https://github.com/llm-d/llm-d/milestones). The milestone serves as the single source of truth for all features, bug fixes, and tasks targeted for the release.
+
+  **Timeline:**
+
+  1. **Sprint start (weeks 1-2):** The release leaders create the milestone for the upcoming release. SIG leads and project maintainers begin nominating issues and features for inclusion.
+  2. **Feature collection (weeks 2-4):** SIG leads propose features from their respective areas during the weekly project standup (Wednesdays 12:30 PM ET) and SIG meetings. Each proposed feature must have a corresponding GitHub issue linked to the milestone.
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+
+In the `Component Summary` section, we need to generate a matrix that shows each component name, the component version, the previous version of the component, and the type.
+
+Also in this phase, we need to include all the CI/CD changes that are required for the release based on the components that we are releasing.
+
+### 3. Release work
+
+The final release work involves creating a tag in the llm-d repo, which triggers a release [workflow](https://github.com/llm-d/llm-d/blob/main/.github/workflows/ci-release.yaml) to build images. We currently use GHCR for image storage. The prep work involves bumping to an image tag that does not exist yet, and then tagging the repo creates the necessary images.
+
+There are two different packages: main and `dev`. The `dev` packages are created by the PRs, while the main packages are created by the release process. For example, see [llm-d-cpu](https://github.com/llm-d/llm-d/packages?q=llm-d-cpu) packages.
```
Gregory-Pereira: This is a bit unclear I would provide examples in line

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+## Phases
+
+### 1. Feature freeze
+
+This section describes the general process for tracking work for a release, which involves attending lead syncs and SIG meetings. There is also a milestone release tracker that needs to be kept up to date. The release tracker can be found in the [milestones](https://github.com/llm-d/llm-d/milestones) section of the repository.
+
+The release tracker is created 1-2 weeks into the sprint. 4-weeks later, the release tracker is frozen and it should contain all the features that will be in the release. Each feature should have a specific example: the example should be production-ready, meaning that the feature should show an example that is realistic and that can be used in production. Toy-examples should be avoided.
+
+> [!NOTE]
+> The release tracker is not necessarily accurate. However, in the future, we want to keep this tracker up to date and as accurate as possible.
```
Gregory-Pereira: We should expand and formalize this

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+## Example of a dry-run of the release
+
+We create a new tag for the dry run:
+
+```bash
+git fetch upstream --tags
+git tag v0.5.0-rc.1
+git push upstream --tags
+```
+
```
Gregory-Pereira, line 100: We need to also mention that people might want to cleanup the auto-generated tags. EX if you make and push: `v0.5.0-rc.1` GH will create 5 tags: `v0` (gets updated - should not be this is a prod tag), `v0.5` (gets updated or created - should not be this is a prod tag), `v0.5.0` (gets updated or created - should not be thats a prod tag), `v0.5.0-rc` (to cleanup), `v0.5.0-rc.1` (the actual tag). Its possible theres a better way to do this for RCs maybe we should look into that

RELEASE.md
diegocastanibm: I tried to make it clear

RELEASE.md
diegocastanibm: I have expanded these lines

RELEASE.md
diegocastanibm: if I understand this part correctly, we want `v0`, `v0.5` and `v0.5.0` pointing to the same commit. However, if we create a release candidate, the new `v0`, `v0.5` and `v0.5.0` will point to the last commit that the RC is pointing. I propose, to avoid this, to use a different convention for the RC.

I propose to use a non-semver tag format: Use a format like rc-v0.5.0-1 instead of v0.5.0-rc.1 to avoid triggering GHCR's semver parsing. Once the `ci-release` has past all the test, then we generate the final `v0.5.0` tag. If we like this solution, we would need to change the `ci-release` to execute also with `rc` tags. What do you think?

In any case, I have added an extended text to explain it in this release doc

Gregory-Pereira: Yeah this makes total sense. We will should probably cleanup the leftover rc's then later on but im on board with this approach

Gregory-Pereira (approved):
I think this is a good start to merge. We should get the bulk of this in so that we can refine it over the course of this next release

/lgtm

PR #820, opened 2026-02-19
[Benchmark Guide] [WVA + IS] Document and Provide a WVA + IS Benchmark Guide

Vezio:
# Summary

Provides documentation and sample workload that leverages `llm-d-benchmark` to benchmark Workload Variant Autoscaler + Inference Scheduling scenario.

## Details

- Added `guides/benchmark/wva_inference_scheduling_guidellm_template.yaml`
  - Provides a sample workload that is configure to demonstrate observed scale up and scale down under sufficient load. 
- Updated `guides/benchmark/README.md`
  - Provides documentation on how to run the above workload .

## Notes

- `WVA`, as of current 0.5.0, _only_ supports the _inference scheduling_ well-lit-path, hence the single configuration.

### DIFF
--- a/guides/benchmark/wva_inference_scheduling_guidellm_template.yaml
+++ b/guides/benchmark/wva_inference_scheduling_guidellm_template.yaml
@@ -12,7 +12,7 @@
 # > kubectl get hpa -w
 # 
 # # In another terminal, in the namespace containing the llm-d stack, run:
-# > oc get po -w
+# > kubectl get po -w
 # 
 # After sufficient work spawned, you will observe HPA replica targets increase
 # as well as decode pods increase. Once workload has decreased sufficiently, 


guides/benchmark/README.md
```diff
@@ -140,6 +140,32 @@ For full, customizable benchmarking, please refer to [llm-d-benchmark](https://g
...
+> <td>
+> <details>
+> <summary><b>Workload Variant Autoscaling (WVA) + Intelligent Inference Scheduling</b></summary>
+>
+> ```bash
+> # To obtain the Gateway SVC, please find your inference gateway pod first, then obtain the name.
+> #
+> # For example:
+> #
+> # > oc describe po <INFERENCE-GATEWAY-POD> | grep gateway.networking.k8s.io/gateway-name
```
Gregory-Pereira: We should be opinionated about CLI
```suggestion
> # > kubectl describe po <INFERENCE-GATEWAY-POD> | grep gateway.networking.k8s.io/gateway-name
```

guides/benchmark/wva_inference_scheduling_guidellm_template.yaml
```diff
@@ -0,0 +1,65 @@
+# 
+# This an example guide that leverages the guidellm harness to demonstrates Workload Variant Autoscaler (WVA)
+# From running this experiment, it will be observed that decode pod replicas will autoscale as traffic rate increases 
+# and decreases over time.
+# 
+# WVA currently only supports the Inference Scheduling well-lit-path.
```
Gregory-Pereira, line 6: Can we clarify this? I think this is not the case where we can actually use autoscaling in a variety of cases, we just don't have them integrated into the guide path. I would just clarify the language here

guides/benchmark/wva_inference_scheduling_guidellm_template.yaml
```diff
@@ -0,0 +1,65 @@
...
+# 
+# To see scale up and scale down occur while this workload is running
+# follow the below steps...
+# 
+# # In one terminal, in the namespace containing the llm-d stack, run:
+# > oc get hpa -w
```
Gregory-Pereira:
```suggestion
# > kubectl get hpa -w
```

guides/benchmark/wva_inference_scheduling_guidellm_template.yaml
```diff
@@ -0,0 +1,65 @@
...
+# 
+# # In another terminal, in the namespace containing the llm-d stack, run:
+# > oc get po -w
```
Gregory-Pereira:
```suggestion
# > kubectl get po -w
```

Gregory-Pereira (commented):
looks good just a few comments

guides/benchmark/wva_inference_scheduling_guidellm_template.yaml
Vezio: > Can we clarify this? I

Yes - before opening this PR - I looked into this and I will CC @lionelvillard for additional context.

@lionelvillard @Gregory-Pereira, it may be helpful to document what is officially supported and what is not. Adding this to the WVA repository would likely be sufficient, I took a quick look and didn't find it there up front.

> I think this is not a case where we can actually use autoscaling in a variety of cases; we just don’t have them integrated into the guide path.

To clarify - “unsupported” - in this context means that no internal or external tests have validated code paths beyond the `inference scheduling` well-lit path.

While it may work with other well-lit paths, in theory, that would not constitute _official support._ At this stage, using it outside the validated path would effectively be “use at your own risk.”

I’ve CC’d @lionelvillard here for any further clarification as well as if there are any plans of expanding testing / support beyond `inference scheduling`

Good question @Gregory-Pereira

Gregory-Pereira: Great context here thank you for tracking that down!

PR #839, opened 2026-02-24
Update README for clarity

smarterclayton:


README.md
```diff
@@ -14,60 +14,34 @@ Achieve SOTA Inference Performance On Any Accelerator
...
+2. [Disaggregated Serving (prefill/decode disaggregation)](./guides/pd-disaggregation/README.md) - Reduce time to first token (TTFT) and get more predictable time per output token (TPOT) by splitting inference into prefill servers handling prompts and decode servers handling responses, primarily on large models such as Llama-70B and when processing very long prompts.
+3. [Wide Expert-Parallelism](./guides/wide-ep-lws/README.md) - Deploy very large Mixture-of-Experts (MoE) models like [DeepSeek-R1](https://github.com/vllm-project/vllm/issues/16037) for much higher throughput for RL and latency-insensitive workloads, using [Data Parallelism and Expert Parallelism](https://docs.vllm.ai/en/latest/serving/data_parallel_deployment.html) over fast accelerator networks.
+4. [Tiered KV Prefix Caching with CPU and Storage Offload](./guides/tiered-prefix-cache/README.md) - Improve prefix cache hit rate by offloading KV-cache entries to CPU memory, local SSD, and remote attached high-performance storage (Lustre).
```
vMaroon: Think it might be misleading to highlight Lustre since we support others/any POSIX (high-performance) storage like IBM Storage Scale.

README.md
```diff
@@ -14,60 +14,34 @@ Achieve SOTA Inference Performance On Any Accelerator
...
-<!-- - [2025-08] Read more about the [intelligent inference scheduler](https://llm-d.ai/blog/intelligent-inference-scheduling-with-llm-d), including a deep dive on how different balancing techniques are composed to improve throughput without overloading replicas. -->
+## Key Capabilities
 
-## 📄 About
+1. [Intelligent Inference Scheduling](./guides/inference-scheduling/README.md) - Deploy [vLLM](https://docs.vllm.ai) behind a smart Envoy load balancer enhanced with an [inference scheduler](https://github.com/llm-d/llm-d-inference-scheduler) to decrease serving latency and increase throughput with [prefix-cache aware routing](./guides/precise-prefix-cache-aware/README.md), utilization-based load balancing, and [predicted latency balancing (experimental)](./guides/predicted-latency-based-scheduling).
+2. [Disaggregated Serving (prefill/decode disaggregation)](./guides/pd-disaggregation/README.md) - Reduce time to first token (TTFT) and get more predictable time per output token (TPOT) by splitting inference into prefill servers handling prompts and decode servers handling responses, primarily on large models such as Llama-70B and when processing very long prompts.
```
vMaroon: Our P/D example now showcases `openai/gpt-oss-120b` - perhaps we should highlight that instead of llama-70b.

README.md
smarterclayton: We want to list the ones we support here so people know what we are talking about.  Realistically POSIX is not "high performance storage", so is there a better category we can put over this?

We're not recommending people use CephFS for this, are we?

README.md
smarterclayton: thanks

README.md
chcost: agree. Perhaps we should list Lustre, Scale and WekaFS, the three we have been more directly engaged. Something like: "high-performance network storage (e.g. Lustre, IBM Storage Scale, WekaFS)."

dannyharnik: In reality we support also local storage and less performant file systems. The appeal for these is lower but may be useful at times. 
I think the suggestion by @chcost is good, but maybe even remove the "high performance".

smarterclayton: The readme is about production serving, so useful at times needs to be "useful for production".  While we support other options, those only exist to support production.

smarterclayton: I reduced this to "remote high-performance filesystem storage"

README.md
```diff
@@ -14,60 +14,36 @@ Achieve SOTA Inference Performance On Any Accelerator
...
+3. [Wide Expert-Parallelism](./guides/wide-ep-lws/README.md) - Deploy very large Mixture-of-Experts (MoE) models like [DeepSeek-R1](https://github.com/vllm-project/vllm/issues/16037) for much higher throughput for RL and latency-insensitive workloads, using [Data Parallelism and Expert Parallelism](https://docs.vllm.ai/en/latest/serving/data_parallel_deployment.html) over fast accelerator networks.
+4. [Tiered KV Prefix Caching with CPU and Storage Offload](./guides/tiered-prefix-cache/README.md) - Improve prefix cache hit rate by offloading KV-cache entries to CPU memory, local SSD, and remote high-performance filesystem storage.
+5. [Dynamic Multi-Workload Autoscaling](./guides/dynamic-multi-workload-autoscaling/README.md) - Autoscale multiple model workloads on a shared hardware pool without violating service level objectives.
```
robertgshaw2-redhat, line 27: can do it in a follow up, but I think we should also talk about "FAIRNESS" here --- related to multiple tiers of users and interleaving these onto the same set of resources. i.e. the FLOW CONTROL features

README.md
```diff
@@ -76,50 +52,42 @@ llm-d accelerates distributed inference by integrating industry-standard open te
...
 
-## 🚀 Getting Started
+- [**vLLM Native CPU Offloading**](https://docs.vllm.ai/en/latest/examples/offline_inference/basic/#cpu-offload) and [**llm-d filesystem backend**:](https://github.com/llm-d/llm-d-kv-cache/tree/main/kv_connectors/llmd_fs_backend) llm-d uses vLLM's KVConnector abstraction to configure a pluggable KV cache hierarchy, including offloading KVs to host, remote storage, and systems like LMCache, Mooncake, and KVBM. 
 
-`llm-d` can be installed as a full solution, customizing enabled features, or through its individual components for experimentation.
+- [**Variant Autoscaling over Hardware, Workload, and Traffic**](https://github.com/llm-d-incubation/ig-wva): A traffic- and hardware-aware autoscaler that (a) measures the capacity of each model server instance, (b) derive a load function that takes into account different request shapes and QoS, and (c) assesses recent traffic mix (QPS, QoS, and shapes) to calculate the optimal mix of instances to handle prefill, decode, and latency-tolerant requests, enabling use of HPA for SLO-level efficiency.
 
-### Pre-requisites
+For more details of architecture see the [project proposal](./docs/proposals/llm-d.md).
```
robertgshaw2-redhat, line 67: the other thing to add here would be integrated telemetry. Especially with the tracing items sally is working on

README.md
smarterclayton: Are we at guide stage?  If not, we can mention it in the intelligent inference scheduling as a feature.

README.md
smarterclayton: Is tracing complete?  The monitoring doc doesn't quite provide a really punchy feature (arguably all of what's in the doc is tablestakes).  I would assume we'd have a "production serving metrics" doc or guide that talks about the key metrics and shows how the dashboards should be used / maps to running safely - which isn't quite there yet.

smarterclayton: Agreed in the meeting to add tracing as a follow up when we land the guide.

chcost: yes, agreed. Add tracing after we land the corresponding guide.

PR #849, opened 2026-02-26
Update values_xpu.yaml

VincyZhang:


guides/pd-disaggregation/ms-pd/values_xpu.yaml
```diff
@@ -3,225 +3,197 @@
...
 multinode: false
 
-# Configure accelerator type for Intel XPU via DRA
-dra:
-  enabled: true
+# Configure accelerator type for Intel XPU
+accelerator:
   type: intel
+  resources:
+    intel: "gpu.intel.com/xe"  # Add gpu.intel.com/xe
```
poussa, line 10: WHy are we doing this change? We have just enabled DRA for all Intel devices. This goes against it.

poussa (changes requested):

guides/pd-disaggregation/ms-pd/values_xpu.yaml
VincyZhang: Sorry for the misunderstanding, this is only for test. Please don't merge.

PR #855, opened 2026-02-26
Remove pplx All2All backend

tlrmchlsmth:
Remove the pplx all2all backend from the llm-d image and documentation as the kernels have been has been deprecated in favor of https://github.com/perplexityai/pplx-garden, and the all2all backend has been removed from vLLM in https://github.com/vllm-project/vllm/pull/33724.


Gregory-Pereira (approved):
/lgtm

PR #856, opened 2026-02-27
Use an argument instead of an env for the All2All backend

tlrmchlsmth:
See https://github.com/vllm-project/vllm/pull/33535, vLLM no longer uses environment variables to specify the all2all backend. Instead it must now be passed as an argument to `vllm serve`


Gregory-Pereira (approved):
/lgtm

PR #861, opened 2026-02-27
Move routing-proxy from initContainers to standard containers

roytman:
The routing-proxy container must coexist with the main decode container throughout the pod's lifecycle, not just during initialization.

Additionally, because it is a regular container, Kubernetes can run Liveness and Readiness probes on the proxy. If the proxy hangs but the vLLM engine is fine, Kubernetes knows the pod is no longer "Ready" to serve traffic. initContainers do not support these continuous health checks. We can add it in future releases.


roytman:
routing-sidecar runs as a sidecar container and not as an initContainer.

PR #421, opened 2025-10-31
Refresh support for kgateway agentgateway GIE and GW API

danehans:
- update gateway-provider assets for the deprecated `kgateway` path to use the `kgateway-dev` `agentgateway` charts at `v2.2.1`
- add first-class [agentgateway](https://agentgateway.dev/) v1.0.0 support across provider assets, guides, recipes, and CI inputs
- mark `kgateway` paths deprecated for future removal
- add OpenShift-specific `AgentgatewayParameters` recipes for `agentgateway` and the legacy `kgateway` migration path
- bump GAIE to v1.4.0 and GW API to v1.5.1

Fixes: #399

PR #868, opened 2026-02-28
lora-serving

Gregory-Pereira:
cc @nainaz


guides/lora-serving/inference-pool-values.yaml
```diff
@@ -0,0 +1,13 @@
+inferencePool:
+  modelServers:
+    matchLabels:
+      llm-d.ai/inference-serving: "true"
+      llm-d.ai/guide: "lora-serving"
+inferenceExtension:
+  flags:
+    # in vLLM 10.0+, the metric is renamed while upstream GAIE is still using the old name as default.
+    # See https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1905.
```
ahg-g: It is adjusted in 1.3 I think

guides/lora-serving/inference-pool-values.yaml
```diff
@@ -0,0 +1,13 @@
+inferencePool:
+  modelServers:
+    matchLabels:
+      llm-d.ai/inference-serving: "true"
+      llm-d.ai/guide: "lora-serving"
+inferenceExtension:
```
ahg-g: Don't we want to schedule based on LoRA affinity? I don't think the scorer is enabled by default now. May be we should do that in IGW as part of the default-plugins

chcost: Nice guide, @Gregory-Pereira! A few things from my side:
                                                                                                                                                                                                                     
- should guides/recipes/vllm/standard/kustomization.yaml be in this PR? Wouldn't changing the image tag to latest break reproducibility for all guides using the standard recipe, and adding nvidia.com/gpu resource limits to the shared base breaks non-NVIDIA deployments? 
- The top-level kustomization.yaml has preloaded commented out and runtime-loaded active, but the README marks Preloaded as the default tab. Users following the one-command preloaded instructions might get the wrong config.

Fixes:
- +1 to @ahg-g's point on LoRA affinity scoring. Without it, requests can land on pods that don't have the adapter loaded, especially in runtime-loaded mode with multiple replicas.
- register-adapters.sh interpolates variables directly into JSON strings. We should use jq to build the payload to avoid malformed JSON.

Some other potential issues, not sure: 
- I also think we should verify whether the kv-cache-usage-percentage-metric flag is still needed with GAIE v1.3 (per @ahg-g's comment).
- Consider adding a trap in the script to clean up orphaned port-forward processes on exit
- Hardcoded replicas: 4 in both overlays might be worth documenting as adjustable

guides/lora-serving/register-adapters.sh
```diff
@@ -0,0 +1,159 @@
...
+#   -h, --help    Show this help message
+#
+# Adapters are defined in the ADAPTERS array below — edit it to suit your needs.
+
+set -euo pipefail
+
+# ── Adapters ────────────────────────────────────────────────────────────
+# Each entry is "adapter_name=huggingface_repo_id".
+ADAPTERS=(
+  "topic-control=nvidia/llama-3.1-nemoguard-8b-topic-control"
```
dmitripikus, line 22: Hi Greg @Gregory-Pereira , Thanks for cool guide!!

Just a short comment about the choice of LoRA adapters.
Only `nvidia/llama-3.1-nemoguard-8b-topic-control` has `meta-llama/Llama-3.1-8B-Instruct` as a base model name in its `adapter_config.json` (that is actually a base model used)
Other two adapters have different base model names.
From technical perspective of loading process, there is no issue when adapters are loaded statically, or dynamically using Load/Unload API.
But if dynamic loading from Persistent Volume is used (I am looking at it now), where loading is not done explicitly, but implicitly when service request arrives, I see that if base model is not correct in `adapter_config.json`, adapter is not found (=not loaded).
So when config for loading from Persistent Volume is added, we may need to change the adapters used in demo.
Thank you!

guides/recipes/vllm/cuda/kustomization.yaml
```diff
@@ -20,3 +20,11 @@ patches:
...
               llm-d.ai/hardware-variant: "gpu"
               llm-d.ai/accelerator-vendor: "nvidia"
+          spec:
+            containers:
+              - name: vllm
+                resources:
+                  limits:
+                    nvidia.com/gpu: 1
+                  requests:
+                    nvidia.com/gpu: 1
```
diegocastanibm, line 30: Just as a note, this will break non-nvidia deployments (HPU, GPU)

guides/lora-serving/vllm/overlays/runtime-loaded-from-pv/copy_lora_structure.sh
Gregory-Pereira, line 1: @dmitripikus is this meant to run inside the llm-d image? If so it needs to be added to the build scripts

guides/recipes/vllm/cuda/kustomization.yaml
Gregory-Pereira: They would be broken to begin with because its a cuda only image above

guides/lora-serving/inference-pool-values.yaml
Gregory-Pereira: updated

guides/lora-serving/inference-pool-values.yaml
Gregory-Pereira: - should guides/recipes/vllm/standard/kustomization.yaml - Yes, I changed the image to latest because of access to the vLLM + llm-d offloading conenctors - AFAIK nothing from previous releases have that. What I have done is swap to `ghcr.io/llm-d/llm-d-cuda:latest` which only gets created from releases so to your point we will now have something more stable. I have tried to fix the other paths where possible but will leave that out of scope, but have adjusted my implementation to not affect 
- Top level kustomization has been adjusted to use preloaded which should be the default for a production ready deployment.
- Added the lora scorer plus the other defaults (please comment on the weights @ahg-g @nirrozenbaum @kfswain). 
- kv-cache-usage-percentage-metric flag is definetly not needed, #421 has most of this but I can remove it here as well (which I have tried to do)
- Commented on the hardcoded replicas.
- Added trap support for port forwards

guides/recipes/vllm/cuda/kustomization.yaml
Gregory-Pereira: I have clarified this with the base and cuda build directories for standard recipes - everything should be working now. Should be noted it was previously broken where it might not have requested cuda GPUS but its image was tied to cuda

guides/lora-serving/register-adapters.sh
Gregory-Pereira: Happy to change these adapters to anything these were just the first ones I could pull.

guides/lora-serving/vllm/overlays/runtime-loaded-from-pv/copy_lora_structure.sh
dmitripikus: Hi Greg @Gregory-Pereira, thanks for the note and for the refactoring/fixing this!!

`copy_lora_structure.sh` is an auxiliary script that was supposed to run in the cluster (pod).
This script is not very convenient to use, so I build another script that gets base model as a parameter, identify relevant LoRA adapters compatible to it, and copies them to PersistentVolume (all of the adapters or only part of them).
I'll replace `copy_lora_structure.sh` with new script soon.

guides/lora-serving/register-adapters.sh
dmitripikus: Thank you Greg @Gregory-Pereira !
I am writing script that gets base model as a parameter and automatically identifies compatible LoRA adapters at HuggingFace.
This will make choosing adapters for demo more easy.

PR #899, opened 2026-03-04
Add Intel XPU RDMA deployment support

VincyZhang:
Add xpu_rdma deployment configuration with:
- values_xpu_rdma.yaml with RDMA/DRA settings for PD disaggregation
- rdma-resource-claims.yaml for ResourceClaimTemplate definitions
- helmfile.yaml.gotmpl and README.xpu.md updates for deployment documentation

Features:
- Prefill-decode disaggregation on Intel Data Center GPU Max
- Dynamic Resource Allocation (DRA) for GPU assignment
- RDMA network device support via UCX transport
- Separate decode and prefill deployments with optimized parallelism
- Comprehensive startup/liveness/readiness probes for stability


guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,276 @@
...
+        capabilities:
+          add: ["IPC_LOCK","NET_ADMIN","SYS_PTRACE","SYS_ADMIN","SYS_RAWIO"]
+      env:
+        - name: UCX_TLS
+          value: "ib,rc,ze_copy"
+        - name: VLLM_NIXL_SIDE_CHANNEL_HOST
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+        - name: VLLM_NIXL_SIDE_CHANNEL_PORT
```
zdtsw, line 73: any reason to set the decode pod to 5500?

VincyZhang: No particular reason, it's okay to set back 5600, as other values_*.yaml

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,276 @@
...
+    - name: rdma-net-interface
+      resourceClaimTemplateName: rdma-net-template
+  containers:
+    - name: "vllm"
+      image: ghcr.io/llm-d/llm-d-xpu:v0.5.0
+      modelCommand: custom
+      command: ["/bin/bash", "-c"]
+      args:
+        - |
+          export UCX_NET_DEVICES="$(rdma link | grep netdev | awk '{print $2}' | cut -d'/' -f1):1"
```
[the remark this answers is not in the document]
VincyZhang: In rdma-resource-claim-template, only 1 RDMA netdev is required, so it's not possible when multiple RDMA netdevs exist. But we can change to `export UCX_NET_DEVICES="$(rdma link | awk '/netdev/ { split($2,a,"/"); dev=a[1]":1"; if (list=="") list=dev; else list=list","dev } END { print list }')"`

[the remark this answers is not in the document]
VincyZhang: To apply PD with GDR path, we need use a customized image based on ghcr.io/llm-d/llm-d-xpu:v0.5.0. Will create a patch later for Dockerfile.xpu

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
VincyZhang: Set to 5600, and verified Okay

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,276 @@
...
+    - name: rdma-net-interface
+      resourceClaimTemplateName: rdma-net-template
+  containers:
+    - name: "vllm"
+      image: ghcr.io/llm-d/llm-d-xpu:v0.5.0
+      modelCommand: custom
+      command: ["/bin/bash", "-c"]
+      args:
+        - |
+          export UCX_NET_DEVICES="$(rdma link | awk '/netdev/ { split($2,a,"/"); dev=a[1]":1"; if (list=="") list=dev; else list=list","dev } END { print list }')"
```
poussa: can we put this into script like the gke_rdma file is doing?

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,276 @@
...
+          exec vllm serve Qwen/Qwen3-0.6B \
```
poussa, line 173: You don't need the exec, do you?

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,276 @@
...
+            --port 8000 \
+            --served-model-name "Qwen/Qwen3-0.6B" \
```
poussa, line 175: Can we use bigger model? Reduce the context len to get more room to a bigger model. 32K is quite big for 12-16 GB VRAM.

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,276 @@
...
+      ports:
+        - containerPort: 8000
+          name: vllm
+          protocol: TCP
+        - containerPort: 5600
+          name: nixl
+          protocol: TCP
+      resources:
+        limits:
+          memory: 64Gi
```
poussa: reduce the memory requirement, not that much CPU memory is required. Only torch, etc run there.

guides/pd-disaggregation/README.xpu.md
```diff
@@ -16,6 +16,28 @@ This document provides complete steps for deploying Intel XPU PD (Prefill-Decode
 * Intel GPU Plugin deployed
 * kubectl access with cluster-admin privileges
 
+### Additional Requirements for RDMA Deployment (`-e xpu_rdma`)
```
poussa, line 21: Don't we need to deploy DRANET as well? If so, add it to the readme.

poussa (changes requested):
A

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
VincyZhang: The script is mounted from host server, do you suggest this way?
https://github.com/llm-d/llm-d/blob/main/guides/pd-disaggregation/ms-pd/values_gke_rdma.yaml#L68-L69

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
VincyZhang: Using `exec` is recommended, because without `exec`, signal handling becomes unreliable and shutdowns may time out or require SIGKILL.
1. Clean shutdowns: `vllm` becomes PID 1 and receives SIGTERM directly when the Pod stops, ensuring graceful termination
2. Correct exit codes: The container exit status reflects actual result, not the shell's
3. No zombie processes: Avoids leaving a shell wrapper that could fail to reap children
4. Best practice: Standard pattern for container `entrypoints` with long-running processes

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
yuanwu2017: > Can we use bigger model? Reduce the context len to get more room to a bigger model. 32K is quite big for 12-16 GB VRAM.

This patch is for RDMA.  Let's keep it sample.  All the models will be replaced in another patches. However, our hardware and CI machine configurations cannot reach the same level as the NV machines in LLMD. Therefore, we will configure the largest possible model according to the capabilities of our hardware, currently focusing on the size suitable for BMG.

guides/pd-disaggregation/README.xpu.md
VincyZhang: add installation guide and how-to-check

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
poussa: Ah, no, not from the host. I was thinking if we have the script in the image itself, it would look cleaner. Anyway, this is minor nit.

poussa (approved):

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,270 @@
...
+      enabled: true
+      portName: "vllm"  # decode vLLM service port (from routing.proxy.targetPort)
+      path: "/metrics"
+      interval: "30s"
+  resourceClaims:
+    - name: rdma-net-interface
+      resourceClaimTemplateName: rdma-net-template
+  containers:
+    - name: "vllm"
+      image: ghcr.io/llm-d/llm-d-xpu:v0.5.0
```
Gregory-Pereira:
```suggestion
      image: ghcr.io/llm-d/llm-d-xpu:v0.5.1
```

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,270 @@
...
+      enabled: true
+      portName: "vllm"  # prefill vLLM service port (from routing.servicePort)
+      path: "/metrics"
+      interval: "30s"
+  resourceClaims:
+    - name: rdma-net-interface
+      resourceClaimTemplateName: rdma-net-template
+  containers:
+    - name: "vllm"
+      image: ghcr.io/llm-d/llm-d-xpu:v0.5.0
```
Gregory-Pereira:
```suggestion
      image: ghcr.io/llm-d/llm-d-xpu:v0.5.1
```

Gregory-Pereira (dismissed):
/lgtm just the image upgrades. Honestly can take or leave those - can always bump all versions together when we release prep so ill approve and you can make the call

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
VincyZhang: Thanks! I think we'd better follow your release pattern to make things easier.

Gregory-Pereira (dismissed):
/ltgm

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
Gregory-Pereira: Mounted from host server is fine, also if we wanted to embed this into the image thats fine too.

yuanwu2017 (dismissed):

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,270 @@
...
+    - name: rdma-net-interface
+      resourceClaimTemplateName: rdma-net-template
+  containers:
+    - name: "vllm"
+      image: ghcr.io/llm-d/llm-d-xpu:v0.5.1
+      modelCommand: custom
+      command: ["/bin/bash", "-c"]
+      args:
+        - |
+          export UCX_NET_DEVICES="$(rdma link | awk '/netdev/ { split($2,a,"/"); dev=a[1]":1"; if (list=="") list=dev; else list=list","dev } END { print list }')"
```
poussa: I think this line works in RoCE/eth environments. On IB, the `rdma link`output is different (no netdev):

```
rdma link
link mlx5_0/1 subnet_prefix fe80:0000:0000:0000 lid 6 sm_lid 1 lmc 0 state ACTIVE physical_state LINK_UP
```

The following will work for both:
```
 export UCX_NET_DEVICES="$(rdma link | awk '/^link/ { split($2,a,"/"); dev=a[1]":1"; if (list=="") list=dev; else list=list","dev } END { print list }')"
```

guides/pd-disaggregation/README.xpu.md
```diff
@@ -16,6 +16,45 @@ This document provides complete steps for deploying Intel XPU PD (Prefill-Decode
...
+The RDMA path in this guide applies `ms-pd/rdma-resource-claims.yaml`, which uses:
+
+* `apiVersion: resource.k8s.io/v1`
+* `ResourceClaimTemplate` (DRA)
+
+Before using RDMA deployment, ensure all of the following are true:
+
+* Kubernetes version is **v1.34.0+** (or another version in your distribution that serves `resource.k8s.io/v1`).
+* Dynamic Resource Allocation (DRA) is enabled in your cluster control plane/node components for your distro.
+* `rdma-dranet` `DeviceClass` already exists in the target cluster.
```
poussa, line 32: I think it is better to assume there is no `DeviceClass`in the cluster since the DRANET deployment does not install it. Thus, I would suggest we add the `DeviceClass`creation to the docs.

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
poussa: If we want to drop the dependency to the rdma core in the xpu image we can use the sysfs for the device discovery:

```
UCX_NET_DEVICES="$(ls /sys/class/infiniband/ | awk '{dev=$0":1"; list=(list=="")?dev:list","dev} END{print list}')"
```

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
```diff
@@ -0,0 +1,270 @@
...
+          exec vllm serve Qwen/Qwen3-0.6B \
+            --port 8200 \
+            --served-model-name "Qwen/Qwen3-0.6B" \
+            --block-size 64 \
+            --kv-transfer-config '{"kv_connector":"NixlConnector", "kv_role":"kv_both", "kv_buffer_device":"xpu"}' \
+            --disable-uvicorn-access-log \
+            --max-model-len 32000 \
+            --no-enable-prefix-caching
+      securityContext:
+        privileged: true
```
poussa, line 61: this is not good. Need a way to run without it.

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
poussa: Actually, I think we need to use the `sysfs`approach since now the vllm container has securityContext.privileged: true, which is not good.

guides/pd-disaggregation/helmfile.yaml.gotmpl
```diff
@@ -104,6 +106,8 @@ releases:
       - ms-pd/values_amd.yaml
     {{- else if eq .Environment.Name "xpu" }}
       - ms-pd/values_xpu.yaml
+    {{- else if eq .Environment.Name "xpu_rdma" }}
```
poussa, line 115: You can also automate the resourceclaimtemplate apply/delete with the following (placement is not here):

```diff
 +    {{- if eq .Environment.Name "xpu_rdma" }}
 +    hooks:
 +      - events: ["presync"]
 +        command: "kubectl"
 +        args: ["apply", "-f", "ms-pd/rdma-resource-claims.yaml", "-n", "{{ $ns }}"]
 +      - events: ["postuninstall"]
 +        command: "kubectl"
 +        args: ["delete", "-f", "ms-pd/rdma-resource-claims.yaml", "-n", "{{ $ns }}", "--ignore-not-found"]
 +    {{- end }}
```

poussa (changes requested):
Based on my findings, I would suggest changes before merge.

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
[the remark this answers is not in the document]
poussa: hmm, I think we need to work the other way round. Create an image which can be used in the PR. What is the point of merging a config which nobody can use?

maugustosilva (dismissed):

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
VincyZhang: Actually `privileged: true` is necessary to access the RDMA devices, so we need to keep it until we find out a better way.

guides/pd-disaggregation/README.xpu.md
VincyZhang: Added in Readme

guides/pd-disaggregation/ms-pd/values_xpu_rdma.yaml
VincyZhang: We choose your first suggestion. Because we don't always mount /sys/class/infiniband into the container, but the rdma-core is installed inside vLLM image as default.

yuanwu2017 (approved):

guides/pd-disaggregation/helmfile.yaml.gotmpl
VincyZhang: This is a good point. Actually we have a long-term plan that we can just change the modelservice chart template (https://github.com/llm-d-incubation/llm-d-modelservice/pull/245) , and we don't need to claim the RDMA resource manually anymore.

poussa:
@VincyZhang we should build this PR with proper kustomize layout, not just single monolithic .yaml.

PR #911, opened 2026-03-05
chore: update xpu image build inputs

VincyZhang:
### Title
Update XPU image build inputs

### Summary
- Adjust XPU build workflow inputs
- Refresh XPU Dockerfile and NIXL install script
- Update vLLM version pin for XPU image

### Changes
- .github/workflows/build-image.yml
- docker/Dockerfile.xpu
- docker/scripts/xpu/install_nixl_from_source_ubuntu.py
- docker/vllm-version


.github/workflows/ci-pr-test.yaml
```diff
@@ -60,22 +60,19 @@ jobs:
             xpu-inference-scheduling:
               - .github/workflows/ci-pr-test.yaml
               - .github/workflows/e2e-inference-scheduling-xpu.yaml
-              - docker/Dockerfile.xpu
```
yuanwu2017, line 63: If removing the monitor of dockerfile,  How to trigger workflow test for new image? Maybe we can include the vllm-version file?

.github/workflows/ci-pr-test.yaml
```diff
@@ -60,22 +60,19 @@ jobs:
...
               - guides/**/values_xpu.yaml
               - guides/inference-scheduling/ms-inference-scheduling/values_xpu.yaml
               - guides/inference-scheduling/helmfile.yaml.gotmpl
             xpu-pd-disaggregation:
               - .github/workflows/ci-pr-test.yaml
               - .github/workflows/e2e-pd-xpu.yaml
-              - docker/Dockerfile.xpu
```
yuanwu2017, line 70: same above

.github/workflows/ci-pr-test.yaml
```diff
@@ -60,22 +60,19 @@ jobs:
...
               - guides/**/values_xpu.yaml
               - guides/pd-disaggregation/ms-pd/values_xpu.yaml
               - guides/pd-disaggregation/README.xpu.md
               - guides/pd-disaggregation/helmfile.yaml.gotmpl
             xpu-prefix-cache:
               - .github/workflows/ci-pr-test.yaml
               - .github/workflows/e2e-prefix-cache-xpu.yaml
-              - docker/Dockerfile.xpu
```
yuanwu2017, line 78: same above

poussa (approved):

docker/vllm-version
```diff
@@ -1,7 +1,8 @@
 VLLM_REPO=https://github.com/vllm-project/vllm.git
-VLLM_COMMIT_SHA=1892993bc18e243e2c05841314c5e9c06a80c70d
+VLLM_COMMIT_SHA=b31e9326a7d9394aab8c767f8ebe225c65594b60
```
Gregory-Pereira: Revert, were not bumping the version. If you want to use the shared version file you can, but this will be controlled by the project. We pick a specific vLLM version based on the features were support each release. We try to take latest but that does not work sometimes.

```suggestion
VLLM_COMMIT_SHA=1892993bc18e243e2c05841314c5e9c06a80c70d
```

docker/vllm-version
```diff
@@ -1,7 +1,8 @@
...
 
 # Cuda configurations only
-VLLM_PRECOMPILED_WHEEL_COMMIT=1892993bc18e243e2c05841314c5e9c06a80c70d
+VLLM_PRECOMPILED_WHEEL_COMMIT=b31e9326a7d9394aab8c767f8ebe225c65594b60
```
Gregory-Pereira: Same as above, revert:
```suggestion
VLLM_PRECOMPILED_WHEEL_COMMIT=1892993bc18e243e2c05841314c5e9c06a80c70d
```

guides/pd-disaggregation/README.xpu.md
```diff
@@ -37,24 +37,30 @@ If you need to customize the vLLM version or build the image from source, you ca
...
 
 #### Intel Corporation Battlemage G21
 
 ```shell
-# Build with default vLLM version (v0.11.0)
+# Build directly from the upstream vLLM Dockerfile.xpu
 git clone https://github.com/vllm-project/vllm.git
-git checkout v0.11.0
+cd vllm
+git checkout 0.17.0
```
Gregory-Pereira: v0.15.1. We will be bumping to something more updated for the v0.6.0 llm-d release but I want to do all that in another PR outside the scope of refactor, otherwise well be bug fixing EFA, and Cuda issues in this PR which will bring bloat

```suggestion
git checkout v0.15.1
```

Makefile
```diff
@@ -1,9 +1,15 @@
 SHELL := /usr/bin/env bash
 
+include docker/vllm-version
+
 # Defaults
 PROJECT_NAME ?= llm-d
 DOCKERFILE_DIR = docker
+XPU_BUILD_DIR ?= .cache/vllm-src
+XPU_DOCKERFILE_URL ?= https://raw.githubusercontent.com/vllm-project/vllm/main/docker/Dockerfile.xpu
```
Gregory-Pereira: Is this right? I think this will pull the vLLM dockerfile from `main` even if you are trying to build a specific version

docker/vllm-version
VincyZhang: reverted.

VincyZhang:
> @VincyZhang Image was built failed in #912.

please try again. @yuanwu2017

Makefile
VincyZhang: You're correct, add `XPU_VLLM_TAG` for vLLM XPU version only.
```
XPU_VLLM_TAG ?= v0.17.0
XPU_DOCKERFILE_URL ?= https://raw.githubusercontent.com/vllm-project/vllm/$(XPU_VLLM_TAG)/docker/Dockerfile.xpu
```

docker/vllm-version
VincyZhang: reverted

VincyZhang:
> A lot of this looks good. My biggest question here is do you guys want to use the vLLM version shared across the project? We are more than fine with this if that is the direction you want to go, but that means you cannot pick a specific version of vLLM - you will be tied to whatever the project will use. If that sounds OK then it would definetly make our lives easier if you use the common vLLM versions

Thanks for explanation. Considering the current situation, we need to assign a specific vLLM version for XPU, actually higher than https://github.com/vllm-project/vllm/tree/v0.17.0 will be acceptable.
How about add a `VLLM_XPU_COMMIT_SHA` in docker/vllm-version, just similar to what you did for Cuda.
```
# XPU configurations only
VLLM_XPU_COMMIT_SHA=b31e9326a7d9394aab8c767f8ebe225c65594b60
```

yuanwu2017 (approved):

PR #921, opened 2026-03-08
docs: Add benchmark template purpose matrix to guides/benchmark/README.md

lisperz:
## What this PR does

Adds a **Benchmark Template Purpose Matrix** section to `guides/benchmark/README.md` to help users quickly identify which benchmark template to use for their stack type and testing goals.

## Why is this needed

The `guides/benchmark/` directory contains 10 templates with similar names but no documentation explaining their differences. New users have no clear guidance on which template to choose, leading to confusion and trial-and-error.

## Changes

- Adds a comparison table of all 10 benchmark templates covering:
  - Main purpose
  - Workload pattern
  - Key configuration parameters
  - When to use each template
- Adds reference sections explaining:
  - Template naming convention (`<stack-type>_[variant]_template.yaml`)
  - Load pattern types (`constant`, `poisson`, `concurrent`)
  - Data pattern types (`random`, `shared_prefix`)
  - Harness types (`inference-perf`, `guidellm`)

## Type of change

- [x] Documentation update (no functional changes)

## Testing

No functional changes. Content was verified against all 10 template files in `guides/benchmark/`.

## Related Issue(s)

N/A


Gregory-Pereira (commented):
For this one we have a work in progress to move each benchmarking configuration to the applicable guide directory. However this does seem valuable to instead serve as a guide for people to create their own benchmarking config / listing why the different guide benchmark configs are different.

I suggest we hold this until we have done that template migration but conceptually I think it would still be a good change

PR #965, opened 2026-03-13
Fix the HPU failed issue and enable HPU PR checks in CI

yuanwu2017:
The existing HPU e2e test workflows (e2e-inference-scheduling-hpu.yaml, e2e-pd-hpu.yaml) were not wired into ci-pr-test.yaml, so HPU tests never ran as part of PR validation.

Changes
build-image.yml: Expose hpu output from workflow_call so downstream jobs can detect whether an HPU image was built. ci-pr-test.yaml:
Add hpu-inference-scheduling and hpu-pd-disaggregation path filters to detect-changes. Add test-hpu-inference-scheduling job calling e2e-inference-scheduling-hpu.yaml. Add test-hpu-pd job calling e2e-pd-hpu.yaml.
Both jobs follow the same gating pattern as the existing XPU jobs (change detection + build gate + workflow_dispatch override).


Gregory-Pereira (approved):
awesome, thank you!
/lgtm

PR #983, opened 2026-03-19
Upgrade to CUDA13 and support GB200

tlrmchlsmth:
### Purpose

This branch upgrades the build infrastructure to CUDA 13.0, targeting GB200/Blackwell GPUs. It also updates vLLM to approximately `v0.18.0` (plus a few commits).

It adds an extra layer to the image to choose between two versions of DeepEP: The `runtime-gb200` stage is specifically for GB200 MNNVL systems.

It also upgrades the Ubuntu image to 24.04 across the board. Reasoning: Ubuntu 20.04 is EoL and there isn't a CUDA13 version to use. llm-d doesn't produce manylinux wheels so doesn't need to worry about glibc versioning, so the reason that vLLM uses 20.04 in its build stage doesn't apply here.

NVSHMEM no longer built from source. It now uses the `nvidia-nvshmem-cu13` pip package. I could not get the build from source working, I think related to the fact that NVSHMEM removed the static-only build https://github.com/NVIDIA/nvshmem/releases/tag/v3.5.19-1.

Also includes many various fixes.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -23,7 +23,7 @@ on:
   workflow_dispatch:
     inputs:
       platforms:
-        description: "Platforms to build (comma-separated: cuda,aws,xpu,cpu,hpu)"
+        description: "Platforms to build (comma-separated: cuda,aws,gb200,xpu,cpu,hpu)"
         required: false
         default: "cuda"
         type: string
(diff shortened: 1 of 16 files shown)


tlrmchlsmth:
Reopening from fork with squashed history.

Issue #979, opened 2026-03-17
Proposal: Inference Resilience Operator (IRO): automated hardware fault recovery for wide-EP deployments


aishukamal:
# Inference Resilience Operator (IRO): automated hardware fault recovery for wide-EP deployments

## Request

We are proposing IRO as a new incubating component in the llm-d ecosystem.
We are seeking community review of this proposal and approval to begin development
under the llm-d-incubation org.

## Summary

Add a new Kubernetes-native component to llm-d - the Inference Resilience
Operator (IRO) - that automatically coordinates hardware fault events with the
inference engine, sequencing the right engine-side response and
infrastructure-side recovery action to minimize serving interruption and restore
full capacity without manual intervention.

## Motivation

Production llm-d deployments running wide Expert Parallelism across multiple
nodes face a class of hardware failures that no existing component handles
end-to-end. When a hardware fault occurs - an accelerator error, a NIC failure,
a kernel panic - two things must happen in a specific order: the inference engine
must be told to stop routing traffic to the affected rank and drain in-flight
requests, and the infrastructure layer must execute the appropriate recovery
action (reset the device, reboot the node, or replace the node entirely).

These two concerns are completely decoupled today. Infrastructure fault detection
agents can cordon and drain nodes but have no awareness of the inference engine
running on them. The inference engine detects internal failures and exposes
recovery APIs but has no awareness of what the infrastructure layer is doing. No
component coordinates between them.

The consequence is a choice between two bad outcomes: the engine crashes and the
entire instance restarts with minutes of full downtime and all in-flight requests
dropped, or an operator is paged to manually sequence the recovery steps.

Every vLLM fault tolerance RFC published in the past few months - [#27866](https://github.com/vllm-project/vllm/issues/27866),
[#27774](https://github.com/vllm-project/vllm/issues/27774), [#27908](https://github.com/vllm-project/vllm/issues/27908), [#20323](https://github.com/vllm-project/vllm/issues/20323), and [#28243](https://github.com/vllm-project/vllm/issues/28243) - explicitly assumes an external orchestrator
will own this coordination. None of them build it. IRO is that component.

### Goals

- Automatically coordinate engine-side response and infrastructure-side recovery
  for hardware faults, without manual operator intervention.
- Match recovery action to fault severity - a transient device error must not
  trigger the same engine response as a node replacement.
- Delegate recovery action decisions to the infrastructure agent that detected
  the fault. IRO trusts the agent's recommended action by default and allows
[...]

PR #984, opened 2026-03-19
Proposal doc for Inference Resilience Operator

aishukamal:
This PR adds a proposal in docs/proposals/ for Inference Resilience Operator (IRO) for automated hardware fault recovery for wide-EP deployments.

See https://github.com/llm-d/llm-d/issues/979 for details.

### DIFF
--- a/docs/proposals/inference-resilience-operator.md
+++ b/docs/proposals/inference-resilience-operator.md
@@ -66,16 +66,55 @@ will own this coordination. None of them build it. IRO is that component.
 ## Proposal
 
 IRO sits between the infrastructure layer and the inference engine. When a
-hardware fault occurs, the infrastructure recovery controller creates a
-`RecoveryRequest` CRD carrying the resolved recovery action. IRO watches
-`RecoveryRequest`, coordinates the engine-side response, and restores serving
-capacity once infrastructure recovery completes.
+hardware fault occurs, the "infrastructure recovery controller" creates a
+`RecoveryRequest` CRD (a new CRD introduced by this proposal) carrying the
+resolved recovery action. IRO watches `RecoveryRequest`, coordinates the
+engine-side response, and restores serving capacity once infrastructure
+recovery completes.
 
-The infrastructure recovery controller owns the fault-to-action mapping and all
-operator override logic. IRO trusts `requestedAction` on `RecoveryRequest` as
-the final resolved decision.
+```
+┌──────────────────────────────────────────────────────────────────────┐
+│                        Infrastructure layer                          │
+│     Cloud agents · on-prem monitors · hardware fault detectors       │
+│                  Infrastructure recovery controller                  │
+└───────────────┬──────────────────────────▲───────────────────────────┘
+                │ RecoveryRequest CRD      │ RecoveryRequest status
+                │ (write)                  │ (read)
+┌───────────────▼──────────────────────────┴───────────────────────────┐
+│                 Inference Resilience Operator (IRO)                  │
+│                                                                      │
+│  ┌──────────────────────────────┐  ┌─────────────────────────────┐   │
+│  │     InferenceReconciler      │  │      Rank topology map      │   │
+│  │       state machine          │  │  nodeName + deviceID → rank │   │
+│  └──────────────────────────────┘  └─────────────────────────────┘   │
+│                                                                      │
+│                                                                      │
+│  ┌────────────────────────────────────────────────────────────────┐  │
+│  │        vLLM EngineAdapter (implements EngineAdapter interface) │  │
+│  │                  Discovers vLLM API server                     │  │
+│  │            HTTP client → vLLM service endpoint                 │  │
+│  │               ZMQ SUB → vLLM fault notify port                 │  │
+│  └───────────────────┬───────────────────────▲────────────────────┘  │
+└──────────────────────┼───────────────────────┼───────────────────────┘
+                       │ HTTP to engine        │ ZMQ fault events
+                       │ API server            │
+┌──────────────────────▼───────────────────────┴───────────────────────┐
+│                          Inference engine                            │
+│                  API server · ClientSentinel · EngineCore            │
+└──────────────────────────────────────────────────────────────────────┘
+```
 
-Engine recovery tracks are determined entirely by `requestedAction`:
+The "infrastructure recovery controller" in this context is a cloud provider
+specific or vendor specific component (e.g., a GKE node repair daemon, or a custom
+controller) that watches for hardware errors, then creates a `RecoveryRequest` CR in
+Kubernetes. The infrastructure recovery controller owns the fault-to-action
+mapping and all operator override logic for the infrastructure layer. IRO trusts
+`requestedAction` on `RecoveryRequest` as the final decision and coordinates the
+engine-side response accordingly. IRO will ship a reference implementation of the
+infrastructure recovery controller but the interface is open to any Kubernetes
+provider.
+
+Engine recovery tracks are inferred from `requestedAction` on `RecoveryRequest`:
 
 - **RESET_DEVICE → Track A**: pause engine, reset device, resume engine (seconds)
 - **REBOOT_NODE → Track B**: pause engine, reboot node, resume engine (minutes).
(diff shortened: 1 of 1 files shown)


docs/proposals/inference-resilience-operator.md
```diff
@@ -0,0 +1,213 @@
...
+  operator override logic. IRO trusts the resolved action it receives.
+- Restore full serving capacity automatically after infrastructure recovery
+  completes.
+- Work identically across GKE, EKS, bare-metal, and on-premises deployments.
+- Work with any inference engine (initially focusing on vLLM) via swappable adapters.
+
+### Non-Goals
+
+- IRO does not perform fault detection. It consumes signals from existing
+  infrastructure agents.
```
ahg-g, line 55: How does it consume it from infra agents? can you give a concrete example in the design section?

docs/proposals/inference-resilience-operator.md
```diff
@@ -0,0 +1,213 @@
...
+
+- Automatically coordinate engine-side response and infrastructure-side recovery
+  for hardware faults, without manual operator intervention.
+- Match recovery action to fault severity - a transient device error must not
+  trigger the same engine response as a node replacement.
+- IRO's sole responsibility is engine coordination, with a strong sequencing
+  guarantee: IRO acts on the engine before or in parallel with infrastructure
+  recovery, and resumes the engine only once infrastructure recovery is confirmed
+  complete. The sequencing contract between the two layers is the core value.
+- The infrastructure recovery controller owns all fault-to-action mapping and
```
ahg-g, line 45: what do you mean by "infrastructure recovery controller", what is it exactly? can you give an example?

docs/proposals/inference-resilience-operator.md
```diff
@@ -0,0 +1,213 @@
...
+  the action.
+- IRO does not handle planned node maintenance or lifecycle events in v1.
+- IRO does not implement general cluster autoscaling. IRO scales the inference
+  serving group only in response to hardware faults, never in response to load.
+
+## Proposal
+
+IRO sits between the infrastructure layer and the inference engine. When a
+hardware fault occurs, the infrastructure recovery controller creates a
+`RecoveryRequest` CRD carrying the resolved recovery action. IRO watches
```
ahg-g: is this a new CRD that is part of this proposal? if so, I suggest to say that explicitly here.

docs/proposals/inference-resilience-operator.md
```diff
@@ -0,0 +1,213 @@
...
+  coordinates the engine while the infrastructure recovery controller executes
+  the action.
+- IRO does not handle planned node maintenance or lifecycle events in v1.
+- IRO does not implement general cluster autoscaling. IRO scales the inference
+  serving group only in response to hardware faults, never in response to load.
+
+## Proposal
+
+IRO sits between the infrastructure layer and the inference engine. When a
+hardware fault occurs, the infrastructure recovery controller creates a
```
ahg-g: I assume "infrastructure recovery controller" is a cloud provider specific component, if so, are we proposing that they integrate with RecoveryRequest and create instances of it?

docs/proposals/inference-resilience-operator.md
```diff
@@ -0,0 +1,213 @@
...
+  does not provide an operator policy override mechanism. Both belong to the
+  infrastructure recovery controller.
+- IRO does not execute cloud-provider-specific recovery actions directly. It
+  coordinates the engine while the infrastructure recovery controller executes
+  the action.
+- IRO does not handle planned node maintenance or lifecycle events in v1.
+- IRO does not implement general cluster autoscaling. IRO scales the inference
+  serving group only in response to hardware faults, never in response to load.
+
+## Proposal
```
ahg-g, line 66: reading the proposal, the sequence of events are not clear to me. Can we add a sequence diagram?

docs/proposals/inference-resilience-operator.md
```diff
@@ -0,0 +1,213 @@
...
+
+I can integrate with IRO by implementing the published `EngineAdapter` interface,
+without modifying IRO's core recovery logic. My engine's specific API surface is
+entirely encapsulated in the adapter.
+
+## Design Details
+
+### CRD
+
+IRO is defined by one CRD. The infrastructure recovery controller creates
```
ahg-g: we need to define some protocol on when the RecoveryRequest is expected to be created by the "infrastructure recovery controller" and its lifecycle.

docs/proposals/inference-resilience-operator.md
```diff
@@ -0,0 +1,213 @@
...
+- `nodeName`
+- `deviceID` (optional)
+- `errorCode` (optional, carried for observability — IRO does not interpret it)
+- `requestedAction` (RESET_DEVICE | REBOOT_NODE | REPLACE_NODE — resolved by the
+  infrastructure recovery controller before creation)
+- `status.phase` (Pending | InProgress | Completed | Failed — written by the
+  infrastructure recovery controller; IRO watches for Completed to resume the engine)
+- `status.conditions[EngineReadyForRecovery]` — optional; see open question below
+
+### EngineAdapter interface
```
ahg-g: it is not clear how the vllm engine will get the action, are we expecting vllm to implement something? like expose a REST API that IRO will call into? If so, we should define such interface as "engine protocol".

docs/proposals/inference-resilience-operator.md
aishukamal: "Infrastructure recovery controller" is any cloud provider specific component that detects hardware faults and automates recovery at the infrastructure layer. Think of it as the component that powers the "node auto-repair" feature in GKE or EKS today. I added a definition for this in the "Proposal" section.

docs/proposals/inference-resilience-operator.md
aishukamal: It consumes signals through the `RecoveryRequest` CRD. The interface between infra and IRO is this CRD. Made this clear in the "Design Details" section.

docs/proposals/inference-resilience-operator.md
aishukamal: Done.

docs/proposals/inference-resilience-operator.md
aishukamal: Yes. Made it explicit.

docs/proposals/inference-resilience-operator.md
aishukamal: Yes, that is the ask. IRO could ship a reference implementation of this controller based on open source monitoring agents such as https://github.com/aws/eks-node-monitoring-agent but the ideal state is for cloud providers to implement this themselves as this controller will need to call cloud specific APIs.

docs/proposals/inference-resilience-operator.md
aishukamal: Please take a look at the sequence diagram and let me know if that answers this question. One thing that has not been covered is who destroys the RecoveryRequest. One approach is for the producer (infrastructure recovery controller) to own destruction but gated on a Finalizer in the CRD that IRO will update once it has complete engine recovery.

docs/proposals/inference-resilience-operator.md
aishukamal: IRO calls the engine via an HTTP/REST API that the engine exposes. For vLLM specifically, the adapter targets the fault tolerance REST APIs being developed in PR #34833 and RFC #28243 (e.g., POST /fault_tolerance/apply {pause}, POST /fault_tolerance/apply {retry}).

docs/proposals/inference-resilience-operator.md
```diff
@@ -0,0 +1,281 @@
...
+│     Cloud agents · on-prem monitors · hardware fault detectors       │
+│                  Infrastructure recovery controller                  │
+└───────────────┬──────────────────────────▲───────────────────────────┘
+                │ RecoveryRequest CRD      │ RecoveryRequest status
+                │ (write)                  │ (read)
+┌───────────────▼──────────────────────────┴───────────────────────────┐
+│                 Inference Resilience Operator (IRO)                  │
+│                                                                      │
+│  ┌──────────────────────────────┐  ┌─────────────────────────────┐   │
+│  │     InferenceReconciler      │  │      Rank topology map      │   │
```
hexfusion, line 87: can you clarify where the rank topology map (nodeName + deviceID -> vLLM rank) comes from is there an existing component or API that maintains this mapping, or is it something IRO needs to build and keep in sync across scaling events?

aishukamal: I'm not aware of an existing component that provides this mapping readily. But IRO might be able to deduce this mapping based on information available in the K8s pod spec. This is still an open design task at this point, so I don't have all the details yet.

hexfusion: That's fair, we can add it as an open question if that feels appropriate.

chcost:
also good with me

ahg-g:
@clubanderson  can we please an incubation repo named `llm-d-resiliency-operator` to host this project?

PR #990, opened 2026-03-21
cleaning up consistent cache paths and user permissions in images

Gregory-Pereira:
cc @yuanwu2017 @diegocastanibm @ZhengHongming888 @vcave @zdtsw 

This PR:
- ensures consistent opinionated paths for downloading models, and HF operations
- Creates user `llm-d` (previously `vllm`) with user 2000 as part of the `root` group. This will support running on any environment where you cannot run as `root`, but not inhibit people from running as `root` if they so choose
- Ensures that paths are writeable by either user `root` or `llm-d`
- Creates consistent opinionated default cache locations for runtime libraries

### DIFF
--- a/docker/Dockerfile.cuda
+++ b/docker/Dockerfile.cuda
@@ -444,7 +444,7 @@ RUN useradd --uid 2000 --gid 0 -m llm-d && \
     chown -R root:0 /opt/vllm-source /home/llm-d /var/lib/llm-d && \
     chmod -R g+rwX /opt/vllm-source /home/llm-d /var/lib/llm-d && \
     find /opt/vllm-source /home/llm-d /var/lib/llm-d -type d -exec chmod g+s {} \; && \
-    ln -snf /var/lib/llm-d/models /models
+    ln -snf "$LLM_D_MODELS_DIR" /models
 
 # Configure default cache locations for libraries that write transient data at runtime.
 # These are pointed to /tmp to ensure they are always writable regardless of the
@@ -452,6 +452,9 @@ RUN useradd --uid 2000 --gid 0 -m llm-d && \
 #
 # This avoids reliance on $HOME (which may not be set or writable) and ensures
 # consistent behavior across CPU/GPU backends and container platforms.
+#
+# NUMBA_CACHE_DIR and TRITON_CACHE_DIR are only relevant for CUDA (JIT compilation);
+# other backends do not use them.
 ENV OUTLINES_CACHE_DIR=/tmp/outlines \
     NUMBA_CACHE_DIR=/tmp/numba \
     TRITON_CACHE_DIR=/tmp/triton \
@@ -527,7 +530,6 @@ ENV PATH="${VIRTUAL_ENV}/bin:/usr/local/nvidia/bin:${PATH}" \
     TORCH_NCCL_HEARTBEAT_TIMEOUT_SEC=15 \
     TORCH_NCCL_DUMP_ON_TIMEOUT=0 \
     VLLM_SKIP_P2P_CHECK=1 \
-    VLLM_CACHE_ROOT=/tmp/vllm \
     UCX_MEM_MMAP_HOOK_MODE=none
 
 USER 2000
--- a/docker/Dockerfile.cuda
+++ b/docker/Dockerfile.cuda
@@ -430,17 +430,38 @@ RUN export UV_INSTALL_DIR=/usr/local/bin && \
 # Copy compiled wheels
 COPY --from=builder /wheels/*.whl /tmp/wheels/
 
-# Create the vllm user
-RUN useradd --uid 2000 --gid 0 -m vllm && \
-    touch /home/vllm/.bashrc
+# default opinionated env var for HF_HOME, over-writeable
+ENV LLM_D_MODELS_DIR=/var/lib/llm-d/models \
+    HF_HOME=/var/lib/llm-d/.hf \
+    HF_HUB_CACHE=/var/lib/llm-d/.hf/hub
 
-# Create the vllm workspace with permissions to swap commits and remotes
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.cuda
```diff
@@ -430,17 +430,38 @@ RUN export UV_INSTALL_DIR=/usr/local/bin && \
...
+# Create a non-root runtime user with primary group 0 so application-owned
+# directories that are root:0 and group-writable remain writable by both the
+# default runtime user and root across container platforms.
+RUN useradd --uid 2000 --gid 0 -m llm-d && \
+    touch /home/llm-d/.bashrc && \
+    mkdir -p /opt/vllm-source "$LLM_D_MODELS_DIR" "$HF_HOME" "$HF_HUB_CACHE" && \
+    chown -R root:0 /opt/vllm-source /home/llm-d /var/lib/llm-d && \
+    chmod -R g+rwX /opt/vllm-source /home/llm-d /var/lib/llm-d && \
+    find /opt/vllm-source /home/llm-d /var/lib/llm-d -type d -exec chmod g+s {} \; && \
+    ln -snf /var/lib/llm-d/models /models
```
zdtsw:
```suggestion
    ln -snf "$LLM_D_MODELS_DIR" /models
```

docker/Dockerfile.cuda
```diff
@@ -430,17 +430,38 @@ RUN export UV_INSTALL_DIR=/usr/local/bin && \
...
+# Configure default cache locations for libraries that write transient data at runtime.
+# These are pointed to /tmp to ensure they are always writable regardless of the
+# runtime user (root, non-root, or arbitrary UID in Kubernetes environments).
+#
+# This avoids reliance on $HOME (which may not be set or writable) and ensures
+# consistent behavior across CPU/GPU backends and container platforms.
+ENV OUTLINES_CACHE_DIR=/tmp/outlines \
+    NUMBA_CACHE_DIR=/tmp/numba \
+    TRITON_CACHE_DIR=/tmp/triton \
+    VLLM_CACHE_ROOT=/tmp/vllm
```
zdtsw, line 461: we do not need line 530 then?

docker/Dockerfile.cuda
```diff
@@ -430,17 +430,38 @@ RUN export UV_INSTALL_DIR=/usr/local/bin && \
...
+    ln -snf /var/lib/llm-d/models /models
+
+# Configure default cache locations for libraries that write transient data at runtime.
+# These are pointed to /tmp to ensure they are always writable regardless of the
+# runtime user (root, non-root, or arbitrary UID in Kubernetes environments).
+#
+# This avoids reliance on $HOME (which may not be set or writable) and ensures
+# consistent behavior across CPU/GPU backends and container platforms.
+ENV OUTLINES_CACHE_DIR=/tmp/outlines \
+    NUMBA_CACHE_DIR=/tmp/numba \
```
zdtsw, line 459: maybe in the comments, that only cuda need JIT not others

zdtsw (commented):
some nits

docker/Dockerfile.cuda
Gregory-Pereira: Comment added: https://github.com/llm-d/llm-d/pull/990/changes#diff-aed8cad212e1551c709b65927fda3650042bbaab1f8618b74bb9cfbb7ed440fdR456-R457

zdtsw (commented):
/LGTM

docker/Dockerfile.rocm
```diff
@@ -154,4 +154,31 @@ RUN --mount=type=bind,from=build_rixl,src=/export/wheels,target=/export/wheels \
...
+    ln -snf /var/lib/llm-d/models /models
+
+# Default cache locations for runtime libraries. Use /tmp to guarantee writability
+# across root, non-root, and arbitrary UID execution without relying on $HOME.
+ENV TRITON_CACHE_DIR=/tmp/triton \
+    VLLM_CACHE_ROOT=/tmp/vllm
+
+# Ensure cache dirs exist and /tmp is universally writable.
+RUN mkdir -p "$TRITON_CACHE_DIR" "$VLLM_CACHE_ROOT" && \
+    chmod 1777 /tmp
```
diegocastanibm, line 180: Why in the rest of the Dockerfiles you have used `chmod -R 1777 /tmp` and here you are not using the recursive flag? Maybe, for security reasons, we want to remove the -R from the other Dockerfiles? (not sure about this one).

docker/Dockerfile.hpu
```diff
@@ -80,4 +80,31 @@ RUN ln -s /workspace/vllm/tests /workspace/tests \
     && ln -s /workspace/vllm/examples /workspace/examples \
     && ln -s /workspace/vllm/benchmarks /workspace/benchmarks
 
+# default opinionated env var for HF_HOME, over-writeable
+ENV LLM_D_MODELS_DIR=/var/lib/llm-d/models \
+    HF_HOME=/var/lib/llm-d/.hf \
+    HF_HUB_CACHE=/var/lib/llm-d/.hf/hub
+
+# Create a non-root runtime user with primary group 0 so application-owned
+# directories that are root:0 and group-writable remain writable by both the
```
diegocastanibm, line 89:
```suggestion
# Create a non-root runtime user with primary group 0 so application-owned
# directories that are root:0 and group-writable remain writable by both the
# default runtime user and root across container platforms.
```

diegocastanibm (commented):
One comment worth to note from Claude:
```
scripts/warn-vllm-precompiled.sh still references old user:
  This script references /home/vllm/.bashrc and chown vllm:root. With the user
  renamed to llm-d, this script will silently fail (the path/user won't exist).
  Should be updated or confirmed as intentionally left for backward compat.
```

PR #1025, opened 2026-03-25
Add P/D Disaggregation with RoCE/GDR for OpenShift

maugustosilva:
This required upgrading `modelservice` to release v0.4.9

### DIFF
--- a/guides/pd-disaggregation/ms-pd/values_ocp_rdma.yaml
+++ b/guides/pd-disaggregation/ms-pd/values_ocp_rdma.yaml
@@ -28,7 +28,7 @@ decode:
   replicas: 2
   initContainers:
     - name: preprocess
-      image: "icr.io/vopo/llm-d-benchmark:latest"
+      image: ghcr.io/llm-d/llm-d-benchmark:v0.5.1
       imagePullPolicy: Always
       command: ["set_llmdbench_environment.py", "-e", "/shared-config/llmdbench_env.sh", "-i"]
       env:
@@ -131,7 +131,6 @@ decode:
         runAsGroup: 0
         runAsUser: 0
       imagePullPolicy: Always
-    mountModelVolume: true
     volumeMounts:
       - name: metrics-volume
         mountPath: /.config
@@ -157,7 +156,7 @@ prefill:
   replicas: 1
   initContainers:
     - name: preprocess
-      image: "icr.io/vopo/llm-d-benchmark:latest"
+      image: ghcr.io/llm-d/llm-d-benchmark:v0.5.1
       imagePullPolicy: Always
       command: ["set_llmdbench_environment.py", "-e", "/shared-config/llmdbench_env.sh", "-i"]
       env:


guides/pd-disaggregation/ms-pd/values_ocp_rdma.yaml
```diff
@@ -0,0 +1,283 @@
...
+    secure: false
+
+accelerator:  # controls specialized hardware Type
+  type: nvidia
+
+decode:
+  replicas: 2
+  initContainers:
+    - name: preprocess
+      image: "icr.io/vopo/llm-d-benchmark:latest"
```
diegocastanibm: Two comments about this line:
- Since you now very very well the main maintainer of `llm-d-benchmark` I guess you are fine with the tag, but just in case, are you sure you want to use a mutable tag?
- Not related with this PR directly, but why are we using IBM Cloud Registry for `llm-d-benchmark`, and not `ghcr.io/llm-d/`?

guides/pd-disaggregation/ms-pd/values_ocp_rdma.yaml
```diff
@@ -0,0 +1,283 @@
...
+      securityContext:
+        capabilities:
+          add:
+          - "IPC_LOCK"
+          - "SYS_RAWIO"
+          - "NET_ADMIN"
+          - "NET_RAW"
+        runAsGroup: 0
+        runAsUser: 0
+      imagePullPolicy: Always
```
diegocastanibm, line 262: Don't you want `mountModelVolume: true` as in the `Decode`? Just asking because almost all the code between Prefill and Decode is identical. I suppose that if it is `true` by default in the Chart is fine, but it is worth to check

guides/pd-disaggregation/ms-pd/values_ocp_rdma.yaml
maugustosilva: Ooops. WRONG TAG, thanks! Will fix it!

guides/pd-disaggregation/ms-pd/values_ocp_rdma.yaml
maugustosilva: I have removed any references to `mountModelVolume: true`, given that we are using `uri: "hf://Qwen/Qwen3-8B"` (we should **not** mount any PVCs on this test)

diegocastanibm: Something that came to my mind: if you are removing the `mountModelVolume: true`, then we are not caching the model. Consecutive tests will be slower?

maugustosilva: They will, but this is the standard behavior for the guides, since we neither create a `pvc` nor download the model. This is yet another reason why I consider the guides to be the **wrong** vehicle to test it, and why I am advocating for an `llm-d-deployer`/`llm-d-installer`

PR #1060, opened 2026-04-02
Vllm 0.18.0 upgrade

Gregory-Pereira:
Dupe of #1059 but against an upstream branch - bug in cancel + re-run


Gregory-Pereira:
Dropping this, were already on 0.19.1

PR #1099, opened 2026-04-07
GKE SGlang RDMA config for P/D serving

rahulgurnani:
Testing this change presently, sending it out early for visibility.


rahulgurnani:
closing this for now since we have migrated to kustomize. Will raise a separate change.

PR #1126, opened 2026-04-10
fix: split vLLM command array and fill in quickstart TODOs

madhugoutham:
## Summary

  - Fix vLLM container command from single string to array of args the single string `["vllm serve openai/gpt-oss-20b"]` causes an OCI runtime error (`exec: no such file or directory`) on all clusters
  - Fill in 3 TODOs with verified kubectl output and working curl command

  ## Testing

  | Test | Result |
  |------|--------|
  | GPU broken command `["vllm serve ..."]` | `RunContainerError` exit code 128 |
  | GPU fixed command `["vllm", "serve", "..."]` | 1/1 Running, model loaded |
  | Simulator (CPU) | 2/2 Running, full request/response works |
  | End-to-end request (curl → EPP → model server) | JSON response returned |


robertgshaw2-redhat (approved):
LGTM! Thanks for validating it

PR #1131, opened 2026-04-11
[Refactor Install]: Helm -> Kustomize for IIS

Gregory-Pereira:
## Summary

This PR refactors the model server deployment from Helm values files to Kustomize, restructures the guide documentation, and cleans up stale docs.

### Why Kustomize for model servers?

The previous approach used Helm values files (`values.yaml`, `values_amd.yaml`, `values_cpu.yaml`, etc.) to configure model server deployments. This had several problems:

- **Opaque abstraction**: Users had to understand the Helm chart's template logic to know what Kubernetes resources their values would produce. Debugging required `helm template` round-trips.
- **Rigid structure**: Adding a new hardware backend or serving configuration meant duplicating an entire values file with most fields unchanged.
- **Poor composability**: Features like PD disaggregation, monitoring, or multi-host serving couldn't be toggled independently — they were baked into the values structure.

Kustomize solves these by working directly with Kubernetes manifests:

- **Transparent**: The base manifests are the actual resources that get deployed. Overlays show exactly what changes.
- **Composable**: Features are modeled as independent axes that can be combined:
  - **Base axis**: Single-host (`Deployment`) vs multi-host (`LeaderWorkerSet`) — determines the workload type
  - **Component axis**: PD disaggregation is a Kustomize Component that can be layered onto either base, adding the routing-proxy sidecar, a separate prefill workload, port changes, and prefill PodMonitor
  - **Overlay axis**: Hardware/server-specific patches (args, env, resources, images) are clean overlays on top
- **Discoverable**: `kustomize build` shows exactly what will be applied — no template rendering needed.

The scheduler continues to use Helm (via the upstream `gateway-api-inference-extension` charts) since it benefits from Helm's release management and the chart is maintained externally.

### Model server Kustomize architecture

```
recipes/modelserver/
├── base
│   ├── multi-host
│   │   ├── default
│   │   │   ├── monitoring/
│   │   └── pd
│   │       ├── monitoring/
│   └── single-host
│       ├── default
│       │   └── monitoring/
│       └── pd
│           ├── monitoring/
└── common
    ├── kustomization.yaml
    └── sa.yaml

modelserver
├── amd
│   └── vllm
├── cpu
│   └── vllm
├── cuda
│   ├── sglang
│   └── vllm
├── hpu
│   └── vllm
├── tpu-v6
│   └── vllm
├── tpu-v7
│   └── vllm
└── xpu
    └── vllm
```

Each overlay follows the pattern: `kustomize build guides/<guide>/modelserver/<accelerator>/<server>/`

### Key design decisions

- **Labels are scoped via `fields`**: Workload labels (model, accelerator, role) are applied only to Deployments and PodMonitors using Kustomize `fields` selectors. The ServiceAccount gets only the guide label, keeping it clean for RBAC.
- **PD disagg uses JSON patch for renames**: Kustomize SMP can't rename resources (the name field is a match key). The PD component uses JSON patch (`op: replace`) to rename the base workload to `decode` and change the modelserver port to 8200, then SMP for additive changes like role labels.
- **Monitoring is optional**: The PodMonitor is provided by a nested component under `common/monitoring/`. Users without monitoring CRDs can comment out the component reference.
- **Probes and ports live in base**: Overlays don't need to repeat liveness/readiness/startup probes or port definitions — they come from the base skeleton and work via named port references.

### Other changes

- **Docs cleanup**: Removed stale `wip-docs-new/` directory, old getting-started doc, and unused asset SVGs
- **Guide restructuring**: Added `01_installing_a_guide.md` as the central installation doc explaining deployment choices (standalone vs proxy, parallelism strategies, model architecture considerations)
- **Scheduler recipes**: Moved to a layered values pattern (`base.values.yaml` + `features/*.values.yaml` + guide-specific values)
- **Multi-node serving**: Added `prereq/multi-node-serving/` with LWS installation script
- **Inference-scheduling README**: Simplified to reference the new structure rather than embedding inline installation steps
- **Accelerator docs**: Updated `docs/accelerators/README.md`
- **vLLM logging args**: All vLLM overlays now use `--disable-access-log-for-endpoints=/health,/metrics,/v1/models` and `--disable-log-stats` (replaces deprecated `--disable-uvicorn-access-log`)

## Test plan

- [ ] `kustomize build` succeeds for all 10 inference-scheduling overlays (7 vLLM + 3 SGLang)
- [ ] `kustomize build` succeeds for PD disagg test overlays (single-host and multi-host)
- [ ] Labels are correctly scoped (SA gets only guide label, workloads get full label set)
- [ ] PD component correctly adds routing-proxy sidecar, renames workload to `decode`, creates separate `prefill` workload
- [ ] `namePrefix` correctly propagates to `serviceAccountName` in both Deployment and LeaderWorkerSet
- [ ] No stale `--disable-uvicorn-access-log` references remain in vLLM patches

🤖 PR description Generated with [Claude Code](https://claude.com/claude-code)


guides/optimized-baseline/scheduler/optimized-baseline.values.yaml
```diff
@@ -0,0 +1,7 @@
+## Inference-scheduling guide overrides for the scheduler.
+## Layer on top of: recipes/scheduler/base.values.yaml
+
+inferencePool:
```
robertgshaw2-redhat, line 4: I think it would be better if we explicitly configured the scheduler here, so that users can see how to customize it as needed

guides/inference-scheduling/modelserver/cuda/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,56 @@
...
+  replicas: 8
+  template:
+    spec:
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          # Include all args in overlay patch for consistency
+          args:
+            - "Qwen/Qwen3-32B"
+            - "--port 8000"
```
chcost: I think this will break at runtime. With exec form (`command: ["vllm", "serve"]`), each args entry is a separate `sys.argv` token — `"--port 8000"` gets passed as one string and argparse won't parse it. Needs to be `--port=8000` (or two separate items). Same for `--served-model-name`. This pattern is repeated across all 10 patch files.

guides/01_installing_a_guide.md
```diff
@@ -0,0 +1,113 @@
+# Installing a Guide
+
+## Introduction
+
+An llm-d [well-lit path](../docs/wip-docs-new/getting-started/README.md#well-lit-paths) is a tested, benchmarked deployment recipe — not a one-liner install command, but a series of technical choices tuned for a specific production pattern. Each guide provides sensible defaults, but users should review and [customize](04_customizing_a_guide.md) the configuration for their own models, hardware, and traffic patterns.
```
chcost, line 5: We'll need to update these links after we finish the docs revamp. The `wip-docs-new` references point to a directory which doesn't seem to exist on main. Also `customizing-your-gateway.md` is deleted in this PR, and `recipes/gateway/README.md` doesn't seem to exist either.

guides/recipes/modelserver/base/single-host/pd/monitoring/prefill-deployment.yaml
```diff
@@ -0,0 +1,28 @@
...
+  template:
+    spec:
+      serviceAccountName: sa
+      containers:
+        - name: modelserver
+          image: REPLACE_MODEL_SERVER_IMAGE
+          command: []
+          args: []
+          env: []
+          livenessProbe:
```
chcost, line 30: Missing `port` on all three probes here. Will fail validation at apply time.

guides/inference-scheduling/modelserver/cpu/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,63 @@
...
+  name: modelserver
+spec:
+  replicas: 2
+  template:
+    spec:
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          args:
+            - "Qwen/Qwen3-32B"
```
chcost: Was this intentional? Old config used `meta-llama/Llama-3.2-3B-Instruct` (3B) for CPU. Qwen3-32B on 64 CPU cores seems unlikely to work.

guides/inference-scheduling/modelserver/tpu-v7/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,63 @@
...
+            #   value: "otlp"
+            # - name: OTEL_TRACES_SAMPLER
+            #   value: "parentbased_traceidratio"
+            # - name: OTEL_TRACES_SAMPLER_ARG
+            #   value: "0.1"
+          resources:
+            limits:
+              memory: 64Gi
+              cpu: "16"
+              google.com/tpu: 8
```
chcost, line 51: Old values had `google.com/tpu: 4`. Intentional change?

guides/03_benchmarking_a_guide.md
```diff
@@ -0,0 +1,38 @@
+# Benchmarking a Guide - EPHEMERAL DOC
```
chcost, line 1: Should either finalize or remove before merge. Same for `scheduler/README.md` ("REMOVE THIS README LATER").

guides/optimized-baseline/scheduler/optimized-baseline.values.yaml
chcost, line 4: +1 to Robert's comment. At minimum show the commonly-tuned knobs as commented-out defaults so users know what's available without reading a different file.

guides/recipes/modelserver/common/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+resources:
+  - sa.yaml
+
+components:
+  # Optional: comment out to remove PodMonitoring
+  # Note - will throw issues if your cluster does not have the monitoring CRDs
+  - monitoring
```
chcost: This makes monitoring CRDs a hard requirement for all builds. If someone doesn't have the monitoring stack, `kustomize build` will fail. Worth calling out more prominently or making it opt-in at the overlay level.

guides/optimized-baseline/README.md
```diff
@@ -4,346 +4,44 @@
...
-- **Intel XPU/GPUs**: Intel Data Center GPU Max 1550 or compatible Intel XPU device
-- **Intel Gaudi (HPU)**: Gaudi 1, Gaudi 2, or Gaudi 3 with DRA support
-- **TPUs**: Google Cloud TPUs (when using GKE TPU configuration)
+### Supported Hardware Backends
 
-**Using fewer accelerators**: Fewer accelerators can be used by modifying the `values.yaml` corresponding to your deployment. For example, to use only 2 GPUs with the default NVIDIA GPU deployment, update `replicas: 2` in [ms-inference-scheduling/values.yaml](./ms-inference-scheduling/values.yaml#L17-L36).
+This guide includes configurations for the following accelerators:
 
-**Alternative CPU Deployment**: For CPU-only deployment (no GPUs required), see the [Hardware Backends](#hardware-backends) section for CPU-specific deployment instructions. CPU deployment requires Intel/AMD CPUs with 64 cores and 64GB RAM per replica.
+| Backend | Directory | Notes |
```
chcost, line 25: What happened to `digitalocean` and `gke` (the one with the LD_LIBRARY_PATH workaround)? Intentionally dropped?

guides/inference-scheduling/modelserver/tpu-v7/vllm/patch-vllm.yaml
Gregory-Pereira: This one was - the old values had 4 TPU as you point out but the TP 8 argument IIRC

guides/recipes/modelserver/common/kustomization.yaml
Gregory-Pereira: It should be opt out but easily opt-out-able from the overlays, so they dont need to modify common. Will look into it

guides/optimized-baseline/README.md
Gregory-Pereira: Digital ocean we can support without any changes based on the values file: https://github.com/llm-d/llm-d/blob/main/guides/inference-scheduling/ms-inference-scheduling/digitalocean-values.yaml

GKE should be fixed in the container image itself

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
 > [!NOTE]
 > The community can assist but is not responsible for keeping hardware guide variants updated. We reserve the right to remove stale examples and documentation with regard to hardware support.
 
+## NVIDIA CUDA
```
ahg-g:
```suggestion
## NVIDIA GPU
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+
+NVIDIA GPUs are the default accelerator for all llm-d guides. Any NVIDIA GPU is supported, with the specific capabilities determined by the inference container image used. No special cluster configuration is required beyond the NVIDIA device plugin or DRA driver.
+
+### CUDA Runtime and Driver Requirements
```
ahg-g:
```suggestion
**CUDA Runtime and Driver Requirements**
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+For the full CUDA/driver compatibility matrix, see the [CUDA Toolkit Release Notes](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html).
+
+## AMD ROCm
+
+AMD GPUs are supported via ROCm. The specific GPU models supported are determined by the inference container image. See the AMD device plugin or DRA driver below for cluster setup.
+
 ## Intel XPU
 
-Intel accelerators are supported via the well-lit paths (see the **Intel** row in the table above).
+Intel Data Center GPU Max 1550 and Intel BMG GPUs (Battlemage G21) are supported. Intel XPU deployments use Dynamic Resource Allocation (DRA) with a unified accelerator type that automatically handles driver selection for both i915 and xe drivers.
```
ahg-g:
```suggestion
Intel Data Center GPU Max 1550 and Intel BMG GPUs (Battlemage G21) are supported. Intel XPU deployments use DRA with a unified accelerator type that automatically handles driver selection for both i915 and xe drivers.
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+
+For cluster prerequisites, ensure you have the [Intel Resource Drivers for Kubernetes](https://github.com/intel/intel-resource-drivers-for-kubernetes) installed.
+
+## Intel Gaudi (HPU)
+
+Intel Gaudi 1, Gaudi 2, and Gaudi 3 accelerators are supported via Dynamic Resource Allocation (DRA). Ensure you have the [Intel Resource Drivers for Kubernetes](https://github.com/intel/intel-resource-drivers-for-kubernetes) installed on your cluster before deploying HPU guide variants.
```
ahg-g:
```suggestion
Intel Gaudi 1, Gaudi 2, and Gaudi 3 accelerators are supported via DRA. Ensure you have the [Intel Resource Drivers for Kubernetes](https://github.com/intel/intel-resource-drivers-for-kubernetes) installed on your cluster before deploying HPU guide variants.
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+
+## Google TPU
```
ahg-g: I suggest to put the TPU section right after the nvidia GPUs

guides/optimized-baseline/modelserver/nvidia-gpu/sglang/kustomization.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
ahg-g, line 1: why is the path named cuda instead of gpu?

guides/recipes/scheduler/README.md
```diff
@@ -0,0 +1,53 @@
+# Scheduler Recipes
+
+llm-d uses the [Gateway API Inference Extension](https://github.com/kubernetes-sigs/gateway-api-inference-extension) scheduler to make intelligent routing decisions for inference requests. There are two deployment modes:
```
ahg-g:
```suggestion
llm-d uses the [Inference Scheduler](https://github.com/llm-d/llm-d-inference-scheduler) to make intelligent request scheduling decisions for inference requests. There are two deployment modes:
```

guides/recipes/scheduler/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+```
+base.values.yaml                              # shared defaults (this directory)
+  + features/monitoring.values.yaml           # optional feature toggles
+  + <guide>/scheduler/<guide>.values.yaml     # guide-specific overrides
+```
+
+## Standalone (Default)
+
+Use this when you **do not** want to deploy an external proxy. The standalone chart deploys the scheduler with an embedded Envoy sidecar that handles traffic directly.
```
ahg-g:
```suggestion
Use this when you **do not** want to deploy a proxy via Kubernetes Gateway APIs. The standalone chart deploys the scheduler EPP with an Envoy sidecar to proxy the traffic directly.
```

guides/recipes/scheduler/README.md
```diff
@@ -0,0 +1,53 @@
...
+  oci://registry.k8s.io/gateway-api-inference-extension/charts/standalone \
+  -f guides/recipes/scheduler/base.values.yaml \
+  -f guides/recipes/scheduler/features/monitoring.values.yaml \
+  -f guides/<your-guide>/scheduler/<your-guide>.values.yaml \
+  --set provider.name=<gke|istio|none> \
+  -n ${NAMESPACE} \
+  --version v1.4.0
+```
+
+## With Proxy
```
ahg-g:
```suggestion
## With Kubernetes Gateway API
```

guides/recipes/scheduler/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Use this when you want to route traffic through an external proxy (Istio, Agentgateway, etc.) via the Kubernetes Gateway API. This requires:
+
+1. A proxy control plane installed (see [prereq/gateway-provider](../../prereq/gateway-provider/README.md))
```
ahg-g:
```suggestion
1. A Gateway control plane installed (see [prereq/gateway-provider](../../prereq/gateway-provider/README.md))
```

guides/recipes/scheduler/README.md
```diff
@@ -0,0 +1,53 @@
...
+  -f guides/recipes/scheduler/features/monitoring.values.yaml \
+  -f guides/<your-guide>/scheduler/<your-guide>.values.yaml \
+  --set provider.name=<gke|istio|none> \
+  -n ${NAMESPACE} \
+  --version v1.4.0
+```
+
+## With Proxy
+
+Use this when you want to route traffic through an external proxy (Istio, Agentgateway, etc.) via the Kubernetes Gateway API. This requires:
```
ahg-g:
```suggestion
Use this when you want to route traffic through a proxy managed by the Kubernetes Gateway API  (e.g., GKE Gateway, Istio, Agentgateway) . This requires:
```

guides/recipes/scheduler/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+1. A proxy control plane installed (see [prereq/gateway-provider](../../prereq/gateway-provider/README.md))
+2. A Gateway resource created (see [recipes/gateway](../gateway/))
+3. The inferencepool chart deployed (below)
```
ahg-g:
```suggestion
2. Creating a Gateway resource (see [recipes/gateway](../gateway/))
3. Deploying the inferencepool chart (below)
```

guides/recipes/scheduler/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The inferencepool chart deploys the scheduler as an Endpoint Picker Pod (EPP). The proxy communicates with the EPP via gRPC ext-proc to determine request routing.
```
ahg-g:
```suggestion
The inferencepool chart deploys the scheduler as an Endpoint Picker Pod (EPP). The proxy communicates with the EPP via gRPC ext-proc to determine request scheduling.
```

guides/recipes/scheduler/README.md
ahg-g: We need to keep either `../../prereq/gateway-provider/` or https://github.com/llm-d/llm-d/tree/main/docs/wip-docs-new/guides/gateways, right?

guides/recipes/modelserver/base/multi-host/default/decode-leaderworkerset.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: leaderworkerset.x-k8s.io/v1
```
ahg-g, line 1: I suggest to name it decode-leaderworkerset.yaml to mirror the prefill-leaderworkerset.yaml

guides/recipes/modelserver/base/multi-host/pd/monitoring/decode-leaderworkerset.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: leaderworkerset.x-k8s.io/v1
```
ahg-g: ditto on prefixing the name with `decode-`

guides/recipes/modelserver/base/multi-host/pd/decode-leaderworkerset.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: leaderworkerset.x-k8s.io/v1
```
ahg-g: ditto on prefixing the name with `decode-`

guides/recipes/modelserver/base/multi-host/default/monitoring/decode-leaderworkerset.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: leaderworkerset.x-k8s.io/v1
```
ahg-g: ditto on prefixing the name with `decode-`

guides/recipes/modelserver/base/multi-host/default/monitoring/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
ahg-g: Trying to understand the difference between this and the one in one level up. This directory doesn't layer on top of the other, it is an independent recipe but with monitoring? are we not concerned that we have to keep them in sync?

guides/recipes/modelserver/base/single-host/default/decode-deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+              path: /health
+              port: modelserver
+          readinessProbe:
+            httpGet:
+              path: /v1/models
+              port: modelserver
+          startupProbe:
+            httpGet:
+              path: /v1/models
+              port: modelserver
```
diegocastanibm, line 38:
```suggestion
          startupProbe:
            httpGet:
              path: /v1/models
              port: modelserver
            failureThreshold: 60
            periodSeconds: 10
```

guides/inference-scheduling/modelserver/cuda/vllm/kustomization.yaml
```diff
@@ -0,0 +1,56 @@
...
+images:
+  - name: REPLACE_MODEL_SERVER_IMAGE
+    newName: ghcr.io/llm-d/llm-d-cuda
+    newTag: v0.6.0
+
+labels:
+  - pairs:
+      llm-d.ai/guide: inference-scheduling
+  - pairs:
+      llm-d.ai/inference-serving: "true"
```
diegocastanibm: One solution to fix bug 2 is to move `llm-d.ai/guide: inference-scheduling` to the next block:

```suggestion
      
  - pairs:
      llm-d.ai/guide: inference-scheduling
      llm-d.ai/inference-serving: "true"
```

diegocastanibm:
# Testing Report: PR #1131 — Helm to Kustomize Refactor for Model Server

**Date:** 2026-04-17
**Status:** Complete
**PR:** https://github.com/llm-d/llm-d/pull/1131
**Branch:** `is-kustomize-refactor`

## Summary

End-to-end testing of the Kustomize refactor for the `inference-scheduling` guide. The PR replaces Helm values files with Kustomize overlays for model server deployment and restructures guide documentation.

## Test Environment

| Component | Details |
|-----------|---------|
| Cluster | OpenShift 4.19 (v1.32.9), 14 worker nodes with NVIDIA H100-80GB (8 GPUs each) |
| Tools | kustomize v5.8.0, helm v3.18.3 |
| Namespace | `diego-kustomize-test` |
| Guide tested | `inference-scheduling` with `cuda/vllm` overlay |
| Model | Qwen/Qwen3-32B, TP=2, 6 replicas (reduced from 8 due to GPU availability) |
| Scheduler | Standalone chart v1.4.0 with monitoring enabled |

## Phase 0 — Dry-Run Validation (No GPU Required)

Ran `kustomize build` against all 8 inference-scheduling overlays to verify they produce valid YAML without deploying anything.

| Check | Result |
|-------|--------|
| `kustomize build` for all 8 overlays (cuda/vllm, cuda/sglang, amd, cpu, hpu, xpu, tpu-v6, tpu-v7) | All OK |
| Deprecated `--disable-uvicorn-access-log` flag absent | Confirmed — all overlays use `--disable-access-log-for-endpoints` |
| `namePrefix` propagation to `serviceAccountName` | Correct in all 8 overlays |

## Phase 1 — Live Deployment

### Scheduler (Helm)

Deployed using the new layered values pattern:

```bash
helm install inference-scheduling-scheduler \
  oci://registry.k8s.io/gateway-api-inference-extension/charts/standalone \
  -f guides/recipes/scheduler/base.values.yaml \
  -f guides/recipes/scheduler/features/monitoring.values.yaml \
  -f guides/inference-scheduling/scheduler/inference-scheduling.values.yaml \
  -n diego-kustomize-test --version v1.4.0
```

Result: Scheduler pod running (2/2 containers — EPP + Envoy sidecar). No issues.

### Model Server (Kustomize)

Deployed using:

```bash
kustomize build guides/inference-scheduling/modelserver/cuda/vllm/ | kubectl apply -n diego-kustomize-test -f -
```

Result: **Two critical bugs prevented out-of-the-box deployment.** Both required manual patches to unblock testing. See Bugs section below.

## Phase 2 — Functional Verification

After applying manual patches for bugs #1 and #2:

| Check | Result |
|-------|--------|
| Inference request (chat/completions via scheduler) | OK — correct response from Qwen3-32B |
| Multiple sequential requests (5x) | 5/5 successful with unique response IDs |
| EPP routing (ext-proc) | OK — logs confirm request → EPP → proxy → model server flow |
| EPP pod discovery | OK — EPP detected model server pods and started metrics refreshers |
| PodMonitor selector matches pods | OK — matchLabels align with pod labels |
| Labels on Deployment | Correct (all expected labels present) |
| Labels on pod template | Correct (post-patch — see bug #2) |

## Phase 3 — Cleanup

| Check | Result |
|-------|--------|
| `kustomize build ... \| kubectl delete` | Clean — SA, Deployment, PodMonitor all deleted |
| `helm uninstall` | Clean — all scheduler resources removed |
| Orphaned resources check | None — only default OpenShift SAs remaining |

## Bugs Found

### Bug 1 (Critical) — `startupProbe.failureThreshold` too low for model loading

- **File:** `guides/recipes/modelserver/base/single-host/default/deployment.yaml`
- **Impact:** All model server pods enter CrashLoopBackOff. The deployment is non-functional out of the box for any non-trivial model.
- **Root cause:** The base Deployment does not set `failureThreshold` or `periodSeconds` on the `startupProbe`. Kubernetes defaults to `failureThreshold: 3` and `periodSeconds: 10`, giving the container only 30 seconds to start. Qwen3-32B with TP=2 takes 4-5 minutes to load into GPU memory. vLLM receives a SIGTERM during tokenizer initialization and crashes.
- **Workaround applied:** `kubectl patch` to set `failureThreshold: 60` (10 minutes).
- **Suggested fix:** Add explicit `failureThreshold: 60` and `periodSeconds: 10` to the startupProbe in the base Deployment skeleton. This affects all overlays (single-host default, single-host pd, multi-host default, multi-host pd).

### Bug 2 (Critical) — Label `llm-d.ai/guide` missing from pod templates

- **File:** `guides/inference-scheduling/modelserver/cuda/vllm/kustomization.yaml` (and likely all other overlays)
- **Impact:** The scheduler (InferencePool) cannot find any model server pods, returning `ServiceUnavailable - failed to find candidate pods` for every inference request. The deployment appears healthy but is completely non-functional.
- **Root cause:** The `labels` section in the kustomization has two blocks. The first block (line 16-17) defines `llm-d.ai/guide: inference-scheduling` but does NOT have `includeTemplates: true`, so this label only reaches `metadata.labels` of the top-level resources (Deployment, SA, PodMonitor) — not the pod template. The InferencePool's `matchLabels` selector requires `llm-d.ai/guide: inference-scheduling` on the pods themselves.
- **Workaround applied:** `kubectl patch` to add the label to the pod template.
- **Suggested fix:** Either move `llm-d.ai/guide` into the second labels block (which has `includeTemplates: true`), or add `includeTemplates: true` to the first block. This likely affects all 8 overlays since they follow the same pattern.

### Bug 3 (Minor / Design Clarification) — ServiceAccount receives all labels

- **File:** `guides/inference-scheduling/modelserver/cuda/vllm/kustomization.yaml`
- **Impact:** None functionally. Design question only.
- **Description:** The PR description states that "The ServiceAccount gets only the guide label, keeping it clean for RBAC." However, in the generated output, the SA receives all labels (guide, model, accelerator-variant, accelerator-vendor, inference-serving). The `fields` selectors in the second labels block include `kind: ServiceAccount` at `metadata/labels`, which causes the full label set to be applied.
- **Suggested action:** Confirm with the PR author whether this is intentional or if the SA should be excluded from the second labels block's `fields` list.

cc. @Gregory-Pereira @maugustosilva

guides/optimized-baseline/modelserver/nvidia-gpu/sglang/kustomization.yaml
Gregory-Pereira: Because AMD has GPUs as well. So sounds like we should either name them `cuda` vs `rocm` or `nvidia` vs `amd`

guides/recipes/scheduler/README.md
Gregory-Pereira: Yeah, one of those, I think wonce the Wip-docs-new "graduates" it will replace that but they will be the same path

guides/optimized-baseline/modelserver/nvidia-gpu/sglang/kustomization.yaml
ahg-g: `Nvidia GPU` then? cuda is not the name of the hardware.

guides/recipes/modelserver/base/multi-host/default/monitoring/kustomization.yaml
Gregory-Pereira: Its just a kustomize security / build thing that it can't build one layer. So this just gives us a place to build from. 

>  are we not concerned that we have to keep them in sync?

My understanding is that nothing will change in base across releases unless we change the architecture of llm-d like getting rid of pd sidecar for example

docs/accelerators/README.md
ahg-g: the reason I am suggesting this is because a single sub header looks strange, it also confuses the reader about the level of this header (the formatting does not distinguish it from the parent one very clearly)

guides/inference-scheduling/modelserver/cuda/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,69 @@
...
+  template:
+    spec:
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          # Include all args in overlay patch for consistency
+          args:
+            - "Qwen/Qwen3-32B"
+            - "--port=8000"
+            - "--served-model-name=Qwen/Qwen3-32B"
```
robertgshaw2-redhat: `--served-model-name` is not needed

guides/inference-scheduling/modelserver/cuda/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,69 @@
...
+            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
+            - "--gpu-memory-utilization=0.95"
+            - "--tensor-parallel-size=2"
+            - "--disable-log-stats"
```
robertgshaw2-redhat: `--disable-log-stats` should not be set

guides/optimized-baseline/modelserver/nvidia-gpu/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,69 @@
...
+              memory: 128Gi
+              nvidia.com/gpu: 2
+            requests:
+              cpu: '8'
+              memory: 96Gi
+              nvidia.com/gpu: 2
+          volumeMounts:
+            - mountPath: /dev/shm
+              name: shm
+            - mountPath: /.cache
```
robertgshaw2-redhat, line 44: these compile caches point to an emptydir, we can remove this to reduce complexity

guides/inference-scheduling/modelserver/cuda/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,69 @@
...
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          # Include all args in overlay patch for consistency
+          args:
+            - "Qwen/Qwen3-32B"
+            - "--port=8000"
+            - "--served-model-name=Qwen/Qwen3-32B"
+            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
+            - "--gpu-memory-utilization=0.95"
```
robertgshaw2-redhat: I would suggest leaving this as the default, reduces complexity

guides/inference-scheduling/modelserver/cuda/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,69 @@
...
+  replicas: 8
+  template:
+    spec:
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          # Include all args in overlay patch for consistency
+          args:
+            - "Qwen/Qwen3-32B"
+            - "--port=8000"
```
robertgshaw2-redhat: default port is 8000, this is not needed

guides/inference-scheduling/modelserver/cuda/sglang/patch-sglang.yaml
```diff
@@ -0,0 +1,75 @@
...
+            - "--host=0.0.0.0"
+            - "--port=8000"
+            - "--enable-metrics"
+            # Uncomment for distributed tracing
+            # - "--enable-trace"
+            # - "--otlp-traces-endpoint=http://otel-collector:4317"
+          env:
+            - name: TRITON_CACHE_DIR
+              value: "/.cache"
+            - name: CUDA_VISIBLE_DEVICES
```
robertgshaw2-redhat: I doubt you need to explicitly set `CUDA_VISIBLE_DEVICES`

guides/inference-scheduling/modelserver/cuda/sglang/patch-sglang.yaml
```diff
@@ -0,0 +1,75 @@
...
+            timeoutSeconds: 5
+            failureThreshold: 3
+          readinessProbe:
+            periodSeconds: 5
+            timeoutSeconds: 2
+            failureThreshold: 3
+      volumes:
+        - name: metrics-volume
+          emptyDir: {}
+        - name: torch-compile-cache
```
robertgshaw2-redhat: this is not being used

guides/inference-scheduling/modelserver/cuda/sglang/patch-sglang.yaml
```diff
@@ -0,0 +1,75 @@
...
+            - "--enable-metrics"
+            # Uncomment for distributed tracing
+            # - "--enable-trace"
+            # - "--otlp-traces-endpoint=http://otel-collector:4317"
+          env:
+            - name: TRITON_CACHE_DIR
+              value: "/.cache"
+            - name: CUDA_VISIBLE_DEVICES
+              value: "0"
+            - name: UCX_TLS
```
robertgshaw2-redhat: I dont think you need to set `UCX_TLS`, since UCX is not being used unless NIXL is being used

guides/inference-scheduling/modelserver/cuda/sglang/patch-sglang.yaml
```diff
@@ -0,0 +1,75 @@
...
+          args:
+            - "--model-path=Qwen/Qwen3-32B"
+            - "--host=0.0.0.0"
+            - "--port=8000"
+            - "--enable-metrics"
+            # Uncomment for distributed tracing
+            # - "--enable-trace"
+            # - "--otlp-traces-endpoint=http://otel-collector:4317"
+          env:
+            - name: TRITON_CACHE_DIR
```
robertgshaw2-redhat: Since this is set to an emptydir, im not sure this is needed

robertgshaw2-redhat (commented):
I left a few comments about superflous stuff in the manifests, can address after merge

docs/accelerators/README.md
Gregory-Pereira: Makes sense, will include

guides/optimized-baseline/modelserver/nvidia-gpu/sglang/kustomization.yaml
Gregory-Pereira: Seems a bit awkward but lets go with that for now, can always rename later

guides/recipes/modelserver/base/single-host/default/decode-deployment.yaml
Gregory-Pereira: Startup probe timing has been added to the overlay because it is influenced by deployment in each guide - ie heftier guides will have different calculations than the base ones. The configuration about what methodology is used for the probe (what API Route / port is being used) lives in base because those will never change

guides/optimized-baseline/modelserver/nvidia-gpu/vllm/patch-vllm.yaml
Gregory-Pereira: Are we sure? I think the issue is that we use strict linux file permissioning in the image itself, and so when we attempt to write to the `.cache` directory in root it will get angry. I think this is something specific to our images and not upstream vLLM ones but please push back if I am thinking about this incorrectly, or if we want to change this in our images

guides/inference-scheduling/modelserver/cuda/vllm/patch-vllm.yaml
Gregory-Pereira: I will remove for now. It might be helpful to keep this around regardless to give users an easy way they could set this to be whatever they want

guides/inference-scheduling/modelserver/cuda/sglang/patch-sglang.yaml
Gregory-Pereira: I agree, was left in because it was copied 1-1 from their values: https://github.com/llm-d/llm-d/blob/main/guides/inference-scheduling/ms-inference-scheduling/values_sglang.yaml#L48-L49. Will remove for now

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
 > [!NOTE]
 > The community can assist but is not responsible for keeping hardware guide variants updated. We reserve the right to remove stale examples and documentation with regard to hardware support.
```
ahg-g, line 31:
```suggestion
> The community can assist but is not responsible for keeping hardware guide variants updated. We reserve the right to remove stale examples and documentation with regard to hardware support.
```

guides/inference-scheduling/modelserver/nvidia-gpu/sglang/kustomization.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../../../recipes/modelserver/base/single-host/default
+
+# To disable monitoring comment the following component out
+components:
+  - ../../../../recipes/modelserver/components/monitoring
+
+namePrefix: inference-scheduling-cuda-sglang-
```
ahg-g:
```suggestion
namePrefix: inference-scheduling-nvidia-gpu-sglang-
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
 
+## NVIDIA GPUs
```
ahg-g, line 33:
```suggestion
### NVIDIA GPUs
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+| CUDA 12.9.1 (current) | 525.60.13 | < 580 |
+| CUDA 13.0.2 (planned) | 580.65.06 | N/A |
+
+> **Recommended driver version: 575.x** for current llm-d releases using CUDA 12.9.1. This provides the latest features and fixes within the CUDA 12.x compatible driver range.
+>
+> When llm-d moves to CUDA 13.0.2, the minimum driver version will become **580.65.06**. Users should plan to upgrade their node drivers to 580+ ahead of this transition.
+
+For the full CUDA/driver compatibility matrix, see the [CUDA Toolkit Release Notes](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html).
+
+## Google TPU
```
ahg-g, line 54:
```suggestion
### Google TPU
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+
+Google Cloud TPUs (v6e, v7) are supported when running on GKE. See the [GKE infrastructure provider docs](../infra-providers/gke/README.md) for cluster setup.
+
+## AMD ROCm
```
ahg-g, line 58:
```suggestion
### AMD ROCm
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+
+AMD GPUs are supported via ROCm. The specific GPU models supported are determined by the inference container image. See the AMD device plugin or DRA driver below for cluster setup.
+
 ## Intel XPU
```
ahg-g, line 62:
```suggestion
### Intel XPU
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
 
-Intel accelerators are supported via the well-lit paths (see the **Intel** row in the table above).
+Intel Data Center GPU Max 1550 and Intel BMG GPUs (Battlemage G21) are supported. Intel XPU deployments use DRA with a unified accelerator type that automatically handles driver selection for both i915 and xe drivers.
+
+For cluster prerequisites, ensure you have the [Intel Resource Drivers for Kubernetes](https://github.com/intel/intel-resource-drivers-for-kubernetes) installed.
+
+## Intel Gaudi (HPU)
```
ahg-g, line 68:
```suggestion
### Intel Gaudi (HPU)
```

docs/accelerators/README.md
```diff
@@ -30,9 +30,48 @@ For integration into the well-lit paths our standard for contribution is higher,
...
+
+Intel Gaudi 1, Gaudi 2, and Gaudi 3 accelerators are supported via DRA. Ensure you have the [Intel Resource Drivers for Kubernetes](https://github.com/intel/intel-resource-drivers-for-kubernetes) installed on your cluster before deploying HPU guide variants.
+
+## CPU Inferencing
```
ahg-g, line 72:
```suggestion
### CPU Inferencing
```

guides/optimized-baseline/modelserver/amd-gpu/vllm/kustomization.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../../../recipes/modelserver/base/single-host/default
+
+# To disable monitoring comment the following component out
+components:
+  - ../../../../recipes/modelserver/components/monitoring
+
+namePrefix: inference-scheduling-rocm-vllm-
```
ahg-g: couple of questions:
1) shouldn't the path name be intelligent scheduling, not inference scheduling?
2) both intelligent-scheduling and inference-scheduling are long as prefixes, we may need to shorten it
3) should we have llm-d in the name?
4) should we add the model name?

I know 3 and 4 contradicts 2 on prefix length :)

ahg-g:
@Gregory-Pereira good to go, but tests are failing.

guides/optimized-baseline/modelserver/amd-gpu/vllm/kustomization.yaml
ahg-g: making the suggestion here, I recommend to name the path `optimized-baseline` or `optimized-default`

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -20,7 +20,7 @@ body:
       label: Feature Area
       description: Which area of llm-d does this feature relate to?
       options:
-        - Inference Scheduling
+        - optimized baseline
```
ahg-g:
```suggestion
        - Optimized Baseline
```

guides/01_installing_a_guide.md
Gregory-Pereira: Agreed, lets continue to track this as our docs evolve

ahg-g:
Not all commits are signed

Gregory-Pereira:
I don't have access to my ssh key on this device right now.

Issue #927, opened 2026-03-09
Adopt vLLM's --shutdown-timeout for graceful Model Service upgrades


markmc:
## Description

As of vllm-project/vllm#34730, `vllm serve` now has a `--shutdown-timeout` parameter that will be included in the v0.18.0 release.

llm-d should adopt this parameter for vLLM Model Service instances to ensure graceful shutdowns during upgrades.

## Proposal

1. **Add `--shutdown-timeout` to vLLM Model Service configuration**
   - Set the timeout to an appropriately long interval to allow most requests to complete
   - Make this configurable via helm charts

2. **Update pod `terminationGracePeriodSeconds`**
   - Set this value greater than the `--shutdown-timeout` value
   - Ensures Kubernetes doesn't force-kill the pod before vLLM's graceful shutdown completes

## Benefits

- Minimizes user traffic disruption during Model Service upgrades
- Allows in-flight requests to complete gracefully
- Provides operators with predictable upgrade behavior

## Implementation Notes

The relevant helm charts are in https://github.com/llm-d-incubation/llm-d-modelservice

## References

- vLLM PR: https://github.com/vllm-project/vllm/pull/34730
- Available in: vLLM v0.18.0+

markmc:
The recommended approach for this is now documented in the vLLM K8s deployment guide (vllm-project/vllm#43208), based on the pattern implemented in KServe (kserve/kserve#5485, kserve/kserve#5496):

1. Add a `preStop` hook (`/bin/sleep 15`) to allow endpoint removal before SIGTERM
2. Set `--shutdown-timeout` to `terminationGracePeriodSeconds - 15` (default: 45s)
3. Raise `terminationGracePeriodSeconds` to 60s

llm-d should adopt the same pattern in its Model Service configuration.

Issue #499, opened 2025-11-25
Avoid Prefill/Decode version/configuration mismatch


markmc:
### Summary

The Endpoint Picker (EPP) will pair Prefill and Decode instances with mismatched vLLM versions or configurations during rolling updates, potentially causing crashes or other undefined behavior due to KV cache format incompatibilities, KV transfer metadata mismatches, and other version-specific issues. Further investigation is needed to understand all problematic scenarios.

### Background

In the P/D well-lit path, an `InferencePool` object defines which pods are eligible for selection via label selectors. The EPP then independently selects a pair of Decode and Prefill workers from the pool based on:
- Pod labels (`llm-d.ai/role: prefill` or `decode`)
- Load metrics
- Prefix cache hit probability

However, Prefill and Decode instances are deployed as separate Kubernetes Deployments , each with their own image and configuration specifications.

### Problem

The EPP selection logic does not verify that paired Prefill/Decode instances have matching:
- vLLM version/image digest
- Attention backend configuration (FlashAttention, etc.)
- KV cache configuration
- Other runtime parameters

This creates two high-risk scenarios:

**1. Version Upgrades**: When performing a rolling update of the vLLM image on Decode pods while Prefill pods remain on an older version (or vice versa), the EPP will pair instances with mismatched versions.

**2. Configuration Changes**: Similarly, changing configuration parameters (e.g., `--attention-backend`) on one role but not the other can create incompatible pairings.

### Suggested Approach

The EPP should only consider P/D pairs with matching version and configuration. Possible implementations:

1. **Version Label Matching**: Add a version/configuration hash label (e.g., `llm-d.ai/config-version`) to both Prefill and Decode pods. The EPP filters to only consider pairs with matching values.
2. **InferencePool per Configuration**: For each rollout, create a new InferencePool with unique label selectors (e.g., `llm-d.ai/config-hash: abc123`) that match both the new Prefill and Decode deployments. 
3. **Pod Annotation Verification**: The routing sidecar or EPP could verify compatibility by comparing pod annotations containing version/config metadata before forwarding requests.

Whichever approach is taken, the rollout of Prefill and Decode needs to be coordinated such that the new:old ratio of both matches - e.g. given 4 Prefill workers and 1 Decode workers, there would be no point in upgrading D from new to old before starting to upgrade P, and vice-versa.

[...]

PR #1138, opened 2026-04-12
[Docs] [3/N] Add Disaggregation Architecture Docs

robertgshaw2-redhat:
SUMMARY:
* add disagg overview doc
* add operations doc


robertgshaw2-redhat:
cc @tlrmchlsmth @NickLucche @markmc

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
+# Disaggregated Serving
+
+## Functionality
+
+Disaggregated serving separates the **prefill** and **decode** stages of LLM inference onto different workers, enabling:
```
markmc: Maybe it's too embedded into llm-d now, but this "worker" term is confusing vs the per-accelerator vLLM process we call a "worker"

I'd say "prefill and decode model service instances"

Probably good to at least define "worker == model service instance" at the outset

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+- **Specialization of P and D workers** - For example, using a larger TP for the memory-bound decoding phase while a smaller TP for the computation-bound prefill phase allows both phases to be computed efficiently.
+
+- **Avoidance of Request Interference** - For long context requests, long-context prefills can create head-of-line blocking challenges that slow down processing of existing requests in the decode phase. Separating the prefill phase of these long requests into dedicated prefill engines allows the ongoing decoding requests to be efficiently processed without being blocked by these long prefills, improving quality-of-service.
```
markmc: "Prefill engine" - if we're calling it a worker, let's stick to that term

I don't love "head-of-line blocking" jargon (I don't think it's directly analogous to the networking problem), but perhaps it's common

I find a simple diagram very instructive on this interference problem

```
Legend:  [##] = prefill    [--] = decode

          t0   t1   t2   t3   t4   t5   t6   t7
          |    |    |    |    |    |    |    |
Request A: [##] [--] [------------] [--] [--] [--]
                                  ^
                                  decode delay

Request B:           [############] [--] [--] [--]
                     ^
             prefill interruption
```

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+- **Request Flow Orchestration** - select and route the requests to the correct prefill and decode pods
+
+- **Efficient KV Transfer** - transfer the KV cache from the P instance to the D instance. llm-d leverages NIXL integration in vLLM and SGLang for RDMA.
+
+> [!NOTE]
+> Disaggregated Serving requires high performance (RDMA) interconnects between nodes for Efficient KV transfer. Without RDMA, NIXL falls back to TCP for transfer which is not efficient and should only be used for testing and development.
+
+### Request Flow Orchestration
+
+llm-d's EPP natively supports the concept of P/D disaggregation, enabling the following request flow:
```
markmc: I'd maybe mention "by selecting a prefill and decode worker pair" since that's key to this "EPP natively supports" point

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+    PD-Profile-Handler-->>Prefill-Profile: Request
+    Prefill-Profile-->>Prefill-Profile: Filter, Score, Pick
+    Prefill-Profile-->>PD-Profile-Handler: P Endpoint
+    PD-Profile-Handler-->>Proxy: D, P Endpoint
+```
+
+In this way, llm-d's disaggregated serving functionality composes neatly with the existing set of scheduling functionality, enabling use of the existing set of scorers for prefix and load aware routing in the disaggregated setting.
+
+Note that both the prefill and decode endpoints are part of 1 `InferencePool`. The `decode-profile` and `prefill-profile` are responsible for selecting only D workers or P workers via the `filter` step. By default, llm-d uses the label key `llm-d.ai/role` with the following values:
```
markmc: nit: "one" instead of "1!

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+vLLM and SGLang use slightly different protocols for KV Transfer between the P and D workers, accepting additional parameters in the body of the requests. The Routing Sidecar supports both of these:
+
+- **vLLM** (`nixlv2`, default) — A two-phase sequential protocol. The sidecar generates a UUID, sends a prefill request with `kv_transfer_params` containing remote-decode metadata, and `max_tokens=1` to suppress output. It captures the KV transfer parameters from the prefiller's response and injects them into the decode request before forwarding it to the local decoder. If the prefiller returns a server error, the sidecar falls back to decoder-only mode (client errors are not retried).
+
+**SGLang** (`sglang`) — Uses a concurrent prefill/decode model. Instead of waiting for prefill to complete, the sidecar injects bootstrap coordination parameters (`bootstrap_host`, `bootstrap_port`, `bootstrap_room`) into both requests, fires the prefill asynchronously in a goroutine (with `context.WithoutCancel` to prevent premature cancellation), and immediately sends the decode request synchronously. The decoder and prefiller coordinate KV transfer out-of-band via the bootstrap room.
```
markmc: Missing bullet point

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+1. **Prefiller → Decoder (P/D)** — The standard disaggregated path. The request is sent to a remote prefiller, KV transfer metadata is collected, and the enriched request is forwarded to the local decoder.
+2. **Decoder-only** — No routing headers present; the request is proxied directly to the local model server instance.
+
+All non-completion routes (`GET /health`, and any other path) pass through to the decoder unchanged.
+
+##### KV Transfer Protocol
+
+vLLM and SGLang use slightly different protocols for KV Transfer between the P and D workers, accepting additional parameters in the body of the requests. The Routing Sidecar supports both of these:
+
+- **vLLM** (`nixlv2`, default) — A two-phase sequential protocol. The sidecar generates a UUID, sends a prefill request with `kv_transfer_params` containing remote-decode metadata, and `max_tokens=1` to suppress output. It captures the KV transfer parameters from the prefiller's response and injects them into the decode request before forwarding it to the local decoder. If the prefiller returns a server error, the sidecar falls back to decoder-only mode (client errors are not retried).
```
markmc: The UUID thing ... maybe it's too subtle, but the `X-Request-Id` generated by the sidecar is only for tracing and log correlation. The ID that D should use when notifying P is sent back by P in `kv_transfer_params["remote_request_id"]`

other connectors whose proxy sends to P and D concurrently AFAIR allocate an ID upfront and send to both via `kv_transfer_params["transfer_id"]` that is separate from `X-Request-Id`

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
+# Disaggregated Serving: Operations vLLM
+
+While Disaggregated serving offers superior performance for high scale inference, it introduces additional operational complexity, including:
+- [Dynamic Connections](#dynamic-connections) - how to add or remove P and D workers on the fly when workers require point-to-point RDMA connections
+- [Request Cancellation](#request-cancellation) - how to free KV caches from the engines when requests stop, in a distributed setting
+- [Fault Tolerance](#fault-tolerance) - how to ensure crashes do not create cascading failures and that resources are cleaned up
```
markmc, line 6: Upgrade rollout. See #499 and #927

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+In production enviornments it is common for pods to be created and destroyed - either recovering from crashes or dynamically adjusting capacity alongside load. In a P/D enviornment, the ability to dynamically add/remove replicas from the deployment is complicated by the need to establish/destroy connections between P and D workers on the fly and the need to fee
+
+vLLM leverages NIXL for KV transfer. A key feature of NIXL is support for dynamically adding and removing connections.
+
+### Scale-Up - Creating New Connections
+
+To create new P/D connections, vLLM executes a "NIXL Handshake" between the D and P worker to setup the RDMA connection. This is a relatively expensive operation (~5s) that is done once per engine pair, with all subsequent requests leveraging the existing connection. llm-d uses a "dynamic lazy" roll-out strategy, avoiding the need for a centralized bootstrap server maintaining global state.
+
+
+It works like this:
```
markmc, line 21: Diagram is broken?

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+
+```mermaid
+sequenceDiagram
+    participant R as Routing Proxy
+    participant P as Prefill Engine
+    participant D as Decode Engine
+
+    R->>P: Request with do_remote_decode=True
+    P-->>P: Run prefill
+    P->>R: Response with KVTransferParams including remote_host, remote_port, and remote_kv_blocks
```
markmc: `remote_request_id` too

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+        D-->>D: Bootstrap RDMA connection
+        deactivate D
+    end
+
+    D-->>P: NIXL READ: pull KV cache blocks via RDMA
+    D-->>D: Run decodes
+    D->>R: Response
+```
+
+- Prefill workers run a side channel server over ZeroMQ. When the P worker finishes processing, it constructs the `KVTransferParams` with `remote_host=VLLM_SIDE_CHANNEL_HOST` (usually the pod IP) and `remote_port=VLLM_SIDE_CHANNEL_PORT` in the response body.
```
markmc: This "side channel" term comes out of nowhere for the reader.

"runs a server thread to handle requests for NIXLMetadata" ?

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+
+llm-d's Kubernetes-native design simplifies these operational practices.
+
+## Dynamic Connections
+
+In production enviornments it is common for pods to be created and destroyed - either recovering from crashes or dynamically adjusting capacity alongside load. In a P/D enviornment, the ability to dynamically add/remove replicas from the deployment is complicated by the need to establish/destroy connections between P and D workers on the fly and the need to fee
+
+vLLM leverages NIXL for KV transfer. A key feature of NIXL is support for dynamically adding and removing connections.
+
+### Scale-Up - Creating New Connections
```
markmc: This is all very interesting internal details, but the operations implication isn't really jumping out at me

I think what's missing here is being more direct that this "lazy" architecture was chosen to allow scale-up?

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
+# Disaggregated Serving: Operations vLLM
```
markmc: I sort of forgot that this is an architecture document - you'd documented aspects of the architectural design that address operations challenges. That's maybe worth restating

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+    D-->>D: Run decodes
+    D->>R: Response
+```
+
+- Prefill workers run a side channel server over ZeroMQ. When the P worker finishes processing, it constructs the `KVTransferParams` with `remote_host=VLLM_SIDE_CHANNEL_HOST` (usually the pod IP) and `remote_port=VLLM_SIDE_CHANNEL_PORT` in the response body.
+- Decode worker receives the request with the `KVTransferParams`; if there is not yet a connection to the remote P worker, it runs a background thread to fetch the `NIXLMetadata` and create the RDMA connection. This action does not block core engine execution, enabling other requests to proceed as usually.
+
+#### Discovery
+
+Since workers are added to an `InferencePool` via standard Kuberentes selectors and labels, new prefill and decode workers are automatically added when the pod lifecycle switches to `status: Running` in Kuberentes - there is no need for a special control plane!
```
markmc: "new prefill and decode workers are added" -> "new prefill and decode workers are discovered automatically when their model service pods status becomes ..."

markmc: "no need for a special control plane" deserves to be explained further ... we can't assume the reader understands what a control-plane architecture would need to be if there is no auto-discovery?

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+    PWorker->>DSidecar: Response with KVTransferParams
+    DSidecar->>DWorker: Request with KVTransferParams and do_remote_prefill=True
+    DWorker-->>PWorker: Pull KV Cache (NIXL RDMA)
+    PWorker-->>PWorker: Release KV Cache
+    DWorker-->>DWorker: Run decodes
+    DWorker->>DSidecar: Response
+    DSidecar->>Proxy: Response
+```
+
+We will discuss the design of the EPP and Routing Sidecar.
```
ahg-g:
```suggestion
Next we will discuss the design of the EPP and Routing Sidecar for disaggregated serving.
```

ahg-g: btw, we should call it something like "disagg sidecar", not "routing sidecar".

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+    D->>P: NIXL.send_notif
+    P->>P: Free KVs
+```
+
+> [!NOTE]
+> There is a small window in which request cancellation will not trigger KV freeing on the P instance. If the request is disconnected after it is completed on the P worker but before it reaches the D worker's scheduler (for example, if it disconnects while the request is inside Routing Proxy), the D worker never knows about the request and therefore is unable to free the remote blocks on the P worker. As a result, the KV blocks are stranded on the P worker until the timeout `VLLM_NIXL_ABORT_REQUEST_TIMEOUT`, which defaults to 480s. We are currently working on a lease-extension strategy that will dramatically shorten the timeout window.
+
+## Fault Tolerance
+
+llm-d leverages dynamic xPyD configuration for disaggregated serving, meaning within an `InferencePool`, all D workers are connected to all P workers. This creates a critical operational risk - crashes in workers have the potential for cascading failures if the system is not tolerant of failures.
```
markmc: "xPyD" just means the number of P and D workers is configurable? Seems like unnecessary jargon

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### EPP
+
+The llm-d EPP supports disaggregation via the `pd-profile-handler`.
```
ahg-g, line 50: this will be deprecated in favor of a new handler named `disagg-profile-handler` which also covers EPD

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+> [!NOTE]
+> Rather than hardcoding a single scheduling algorithm, the EPP delegates execution to one or more `Profile Handlers`, each of which represents a complete scheduling strategy with its own set of scorers and pickers. They can be thought of as "the dispatcher", which maps each incoming inference request to the right scheduling strategy before the scorers and pickers do their work of selecting the actual endpoint. By default, llm-d uses the `single-profile-handler` for simple aggregated serving.
+
+When configured with `pd-profile-handler`, the EPP processes requests in the following steps:
+- The proxy forwards request metadata to the EPP.
+- The `pd-profile-handler` first runs the `decode-profile`, which runs the `filter`, `score`, `pick` lifecycle to select the D endpoint.
```
ahg-g:
```suggestion
- The `pd-profile-handler` first runs the `decode-profile`, which runs the `filter`, `score`, `pick` scheduler profile to select the D endpoint.
```

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+- The `pd-profile-handler` then consults the `decider` — given how much of the prompt is cached on the D instance, should this request run disagg?
+- If `no`: the `pd-profile-handler` returns only the D endpoint to the proxy
+- If `yes` (large uncached suffix), the `pd-profile-handler` also runs the `prefill-profile`, which runs the `filter`, `score`, `pick` lifecycle to select the P endpoint and returns both the P and D endpoints to the proxy.
```
ahg-g:
```suggestion
- If `yes` (large uncached suffix), the `pd-profile-handler` also runs the `prefill-profile`, which runs the `filter`, `score`, `pick` scheduler profile to select the P endpoint and returns both the P and D endpoints to the proxy.
```

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
markmc: This sort of reads like "if we had more static configuration profiles like "2P1D" (or whatever), all D workers would not need to be connected to all P workers, and there would not be critical operation risk

I think I'd just state that all P and D workers are connected, therefore fault tolerance is critical - we cannot allow a failure in one worker to bring down the entire worker group

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Prefill Worker Failure
+
+Prefill Worker failures is most critical failure mode in disaggregated deployments, as vLLM's NIXL integration uses a READ-centric KV Transfer protocol. When a P worker crashes, the D workers will attempt to execute an RDMA READ without first checking that the P worker is still running.
```
markmc: "the most critical failure mode" has me pondering ... "why is it the most critical?"

Maybe "a critical failure mode" is sufficient

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+- **vLLM** (`nixlv2`, default) — A two-phase sequential protocol. The sidecar generates a UUID, sends a prefill request with `kv_transfer_params` containing remote-decode metadata, and `max_tokens=1` to suppress output. It captures the KV transfer parameters from the prefiller's response and injects them into the decode request before forwarding it to the local decoder. If the prefiller returns a server error, the sidecar falls back to decoder-only mode (client errors are not retried).
+
+**SGLang** (`sglang`) — Uses a concurrent prefill/decode model. Instead of waiting for prefill to complete, the sidecar injects bootstrap coordination parameters (`bootstrap_host`, `bootstrap_port`, `bootstrap_room`) into both requests, fires the prefill asynchronously in a goroutine (with `context.WithoutCancel` to prevent premature cancellation), and immediately sends the decode request synchronously. The decoder and prefiller coordinate KV transfer out-of-band via the bootstrap room.
+
+
+### Efficient KV Transfer
```
ahg-g: why is this left empty?

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
```diff
@@ -0,0 +1,143 @@
...
+- Facilitate the multi-step inference request
+- Mutate the requests to follow each model server's KV transfer protocol
+
+##### Request Flow
+
+When a request arrives, the sidecar inspects a routing header set by the proxy:
+
+| Header | Purpose |
+|---|---|
+| `x-prefiller-host-port` | One or more prefill pod addresses (comma-separated or multi-value) |
```
ahg-g, line 113: @elevran I suggest to rename this to an llm-d prefixed key, like `x-llm-d-prefill-endpoint`

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
markmc: "READ-centric KV transfer" gives a tantalizing hint that there is an alternative, which is kind of frustrating ... maybe we need to explain somewhere that the NIXL connector architecture is for D to read KV from P into D memory, but an alternative architecture is for P to write KV directly into D memory

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+
+vLLM handles Prefill Worker failure by building on top of NIXL's error handling functionality. When a NIXL_READ is attempted and fails (including to a remote engine that has crashed), NIXL returns an error code such as `NIXL_ERR_BACKEND`. vLLM catches this error and handle it according to the [`kv_load_failure_policy`](https://docs.vllm.ai/en/stable/features/nixl_connector_usage/?h=nixl#kv-load-failure-policy):
```
markmc: Maybe worth mentioning that this creates a unique scenario in vLLM - where a scheduled prefill fails ... there's really no other scenario where we handle a failed computation step?

markmc (commented):
Nice work, HTH

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+vLLM leverages NIXL for KV transfer. A key feature of NIXL is support for dynamically adding and removing connections.
+
+### Scale-Up - Creating New Connections
+
+To create new P/D connections, vLLM executes a "NIXL Handshake" between the D and P worker to setup the RDMA connection. This is a relatively expensive operation (~5s) that is done once per engine pair, with all subsequent requests leveraging the existing connection. llm-d uses a "dynamic lazy" roll-out strategy, avoiding the need for a centralized bootstrap server maintaining global state.
+
+
+It works like this:
+
+```mermaid
```
ahg-g, line 23: Not rendered correctly, see https://github.com/robertgshaw2-redhat/llm-d/blob/a1684a0940132aa8cd09afc9ada84caf7fae9c37/docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
ahg-g: we should link to this from some section in the main README.md file

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+
+    R->>D: Request (do_remote_prefill=True)
+    note over D: D crashes 💥
+
+    P->>P: Wait `VLLM_NIXL_ABORT_REQUEST_TIMEOUT`
+    P->>P: Free KV Blocks
+```
+
+
+> [!WARNING] Robustness against Decode Worker failure is currently a major weakness of the design, since the `VLLM_NIXL_ABORT_REQUEST_TIMEOUT` defaults to a long timeout (`480s` to avoid early-free when D engines are backed up). We recommend that production users consider reducing this timeout, especially if they can ensure Decode Workers do not result in significant queuing. We are currently working on a "lease-extension" system, which will dramatically improve this situation and avoid the tradeoff.
```
ahg-g:
```suggestion
> [!WARNING] 
> Robustness against Decode Worker failure is currently a major weakness of the design, since the `VLLM_NIXL_ABORT_REQUEST_TIMEOUT` defaults to a long timeout (`480s` to avoid early-free when D engines are backed up). We recommend that production users consider reducing this timeout, especially if they can ensure Decode Workers do not result in significant queuing. We are currently working on a "lease-extension" system, which will dramatically improve this situation and avoid the tradeoff.
```

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+
+> [!NOTE]
+> Documentation for graceful scale down is currently work in progress.
+
+## Request Cancellation
+
+Given the compute intensity and duration of inference requests, model servers like vLLM support "Request Cancellation", where currently in-progress requests are freed when the client disconnects.
+
+In a P/D disaggregation setup, this feature is more complicated, because the resources associated with an inference request are spread across multiple servers (as the P worker holds onto the KV caches until they have been retrieved by the D worker). As a result, if the request is canceled while it is still "in-flight" on the D worker but before the KV transfer occurs, we need to ensure that the resources on the P worker are properly cleaned up. llm-d accomplishes this functionality by building on top of vLLM's existing request cancellation infrastructure:
+- When requests are disconnected in vLLM, it triggers the `abort` codepath, which cleans up running resources. When request with `do_remote_prefill=True` are aborted, vLLM sends a NIXL side channel message (`send_notif`), instructing the remote prefill engine to free the KV cache for the cancelled request.
```
NickLucche: not sure `do_remote_prefilll=True` is going to be clearer than just saying "on D" from the user pov

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
NickLucche: I like xPyD :3 should we add a glossary ?

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+    end
+```
+
+Failed Prefill Worker pods are automatically moved to [`status: Terminated`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-state-terminated) state as part of the standard Pod lifecycle. Since llm-d leverages the Kubernetes API Server for service discovery, no additional traffic will be routed to the failed worker until the pod has been restarted and returns to the `status: Running` state.
+
+In this way, `llm-d` gracefully isolates Prefill Worker failure.
+
+### Decode Worker Failure
+
+While D Worker failures are unlikely to result in P worker crashes (since P workers never initiate a READ or WRITE to D workers), there is a challenge around ensuring that KV cache memory on the P instance is not stranded since the P worker holds onto the KV cache until it has been explicitly pulled from the D worker.
```
NickLucche: I don't think we ever mentioned WRITE so far, perhaps we could just omit to make sure it's not mistaken

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+While Disaggregated serving offers superior performance for high scale inference, it introduces additional operational complexity, including:
+- [Dynamic Connections](#dynamic-connections) - how to add or remove P and D workers on the fly when workers require point-to-point RDMA connections
+- [Request Cancellation](#request-cancellation) - how to free KV caches from the engines when requests stop, in a distributed setting
+- [Fault Tolerance](#fault-tolerance) - how to ensure crashes do not create cascading failures and that resources are cleaned up 
+
+llm-d's Kubernetes-native design simplifies these operational practices.
+
+## Dynamic Connections
+
+In production enviornments it is common for pods to be created and destroyed - either recovering from crashes or dynamically adjusting capacity alongside load. In a P/D enviornment, the ability to dynamically add/remove replicas from the deployment is complicated by the need to establish/destroy connections between P and D workers on the fly and the need to fee
```
hexfusion: minor: s/enviornments/environments/ and sentence was trucated.   `between P and D workers on the fly and the need to fee...`

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+    P->>R: Response (w/ KVTransferParams)
+
+    R->>D: Request (do_remote_prefill=True)
+    D->>D: Request disconnected, calls abort
+    D->>P: NIXL.send_notif
+    P->>P: Free KVs
+```
+
+> [!NOTE]
+> There is a small window in which request cancellation will not trigger KV freeing on the P instance. If the request is disconnected after it is completed on the P worker but before it reaches the D worker's scheduler (for example, if it disconnects while the request is inside Routing Proxy), the D worker never knows about the request and therefore is unable to free the remote blocks on the P worker. As a result, the KV blocks are stranded on the P worker until the timeout `VLLM_NIXL_ABORT_REQUEST_TIMEOUT`, which defaults to 480s. We are currently working on a lease-extension strategy that will dramatically shorten the timeout window.
```
hexfusion: 👍 great ctx here this could be pretty unexpected.

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
```diff
@@ -0,0 +1,148 @@
...
+
+Failed Prefill Worker pods are automatically moved to [`status: Terminated`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-state-terminated) state as part of the standard Pod lifecycle. Since llm-d leverages the Kubernetes API Server for service discovery, no additional traffic will be routed to the failed worker until the pod has been restarted and returns to the `status: Running` state.
+
+In this way, `llm-d` gracefully isolates Prefill Worker failure.
+
+### Decode Worker Failure
+
+While D Worker failures are unlikely to result in P worker crashes (since P workers never initiate a READ or WRITE to D workers), there is a challenge around ensuring that KV cache memory on the P instance is not stranded since the P worker holds onto the KV cache until it has been explicitly pulled from the D worker.
+
+vLLM avoids permanent KV cache stranding by introducing a timeout on the P worker side `VLLM_NIXL_ABORT_REQUEST_TIMEOUT` (default `480s`). After `VLLM_NIXL_ABORT_REQUEST_TIMEOUT` elapses, the P worker will free the KV caches from any requests that have not been READ yet, eventually cleaning up the resources.
```
hexfusion: this is sort of a dupe of the note on line 82 do we need both?

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
robertgshaw2-redhat: I agree. I will change this

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
robertgshaw2-redhat: thanks, this is a good idea

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
robertgshaw2-redhat: WIP

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
robertgshaw2-redhat: I think I will just remove this

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
robertgshaw2-redhat: are you suggesting we add comments about the WIP work?

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
robertgshaw2-redhat: yeah. I think all of these could have full guides attached to them. but im intending this to be a design doc

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
robertgshaw2-redhat: I guess the point is "you dont have to do anything" --- and explaining why / how it works architecurally

I could probably have a better name for the doc. Operations isnt quite right

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
robertgshaw2-redhat: tweaked a bit to make it more about the implication

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
robertgshaw2-redhat: simplified

docs/wip-docs-new/architecture/advanced/disaggregation/operations-vllm.md
robertgshaw2-redhat: yeah. I plan to cover this in a different section

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
robertgshaw2-redhat: im just aligning to what its called in the code/yamls

I agree this namee is better

docs/wip-docs-new/architecture/advanced/disaggregation/README.md
robertgshaw2-redhat: ok should i make this change now?

ahg-g: @elevran should we? the next release will use the new handler, right?

PR #1159, opened 2026-04-14
Add agentgateway guide

danehans:
## Summary
- add a wip agentgateway gateway guide that mirrors the new Istio gateway guide structure
- document the current self-installed agentgateway InferencePool configuration using `provider.name=none`
- include cleanup and troubleshooting notes for the self-installed agentgateway path

## Testing
- validated the guide flow against the existing GKE cluster
- created a temporary agentgateway `Gateway`, `InferencePool`, and `HTTPRoute` in `llmd-is`
- sent a successful external `/v1/completions` request through the temporary gateway path

### DIFF
--- a/docs/wip-docs-new/guides/gateways/agentgateway.md
+++ b/docs/wip-docs-new/guides/gateways/agentgateway.md
@@ -1,0 +1,333 @@
+# Agentgateway
+
+This guide shows how to deploy llm-d with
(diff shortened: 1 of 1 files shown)


robertgshaw2-redhat (approved):
LGTM, thanks for the quick turn. I will update the Istio doc to match this one

PR #1163, opened 2026-04-14
[Docs] Guides Directory Cleanup

robertgshaw2-redhat:
* remove `QUICKSTART.md`, which is not actually a quickstart document
* update `README.md` to be more current


guides/QUICKSTART.md
```diff
@@ -1,99 +0,0 @@
-# llm-d Quick Start
```
robertgshaw2-redhat, line 1: note: each guide recreates these instructions, so this is unnecessary

Gregory-Pereira (dismissed):
/lgtm

PR #1206, opened 2026-04-20
[Proposal]: Add GenAI payload events proposal

nickaggarwal:
## Summary

Design proposal for capturing LLM prompts and completions as OpenTelemetry payload
attributes, following the upstream
[GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) —
specifically `gen_ai.input.messages`, `gen_ai.output.messages`, and
`gen_ai.system_instructions` (these are `Opt-In` / `Development` stability upstream and
are tracked as attribute names stabilise).

Closes #1108.

---

### What's in this PR

| File | Purpose |
| --- | --- |
| `docs/proposals/genai-payload-events.md` | Full proposal following `PROPOSAL_TEMPLATE.md`: summary, motivation, proposal, design details, alternatives. Covers attribute spec, multimodal/non-text payload handling, capture layers, `PayloadStore` interface (ownership + execution plan), redaction pipeline, threat model, and phased implementation plan |
| `guides/recipes/observability/tracing/otel-collector.yaml` | The existing OTel Collector recipe, extended with an **opt-in, commented** `traces/payloads` pipeline (filter + redaction processors). The default `traces` pipeline is unchanged |

This is a **design-only** PR — no implementation, no `values.yaml` stanzas. The
operator opt-in surface (canonical field names defined in the proposal) will land
alongside the in-flight Helm → Kustomize migration in a follow-up Phase-1 implementation
PR. No `guides/**/values.yaml` files are touched here.

> **Update (review feedback):** Migrated off the now-deprecated `gen_ai.content.prompt` /
> `gen_ai.content.completion` events to the current `gen_ai.input.messages` /
> `gen_ai.output.messages` / `gen_ai.system_instructions` attributes; offload markers
> (`storage_uri`, `media_uris`, `truncated`) are marked explicitly as llm-d extensions.
> The standalone `otel-collector-with-payload-capture.yaml` has been **removed** and its
> payload pipeline folded into the existing `otel-collector.yaml` recipe per review.

---

### Design decisions addressed from #1108

**1. Which component emits the events?**
Three-layer answer with vLLM as the preferred attachment point (direct access to decoded
Python objects and the canonical `vllm:llm_request` span). Gateway (EPP) captures prompts
as fallback when vLLM tracing is off; P/D proxy captures completions from reassembled SSE.
`emitIfUpstreamTraced: false` prevents double-emission.

**2. Storage backend abstraction?**
`PayloadStore` interface with five backends: `noop` (default), `inline`, `gcs`, `s3`
(covers MinIO/Ceph), `filesystem`. The Go interface and backends are owned by the
**gateway / EPP**; the vLLM integration is a Python mirror of the same interface. Text
payloads above `inlineSizeThresholdBytes` (default 4 KB) auto-offload to the configured
`offloadBackend`. GCS uses Workload Identity; S3 uses IRSA.

**3. Non-text / multimodal payloads?**
Binary content (images, audio) cannot be carried as OTel attributes — the attribute type
system permits strings/numbers/bools/arrays only, on neither spans nor log records.
Non-text parts are therefore **always offloaded** to the configured `PayloadStore`,
mapped onto the upstream input-messages `parts[]` schema, replaced in the text skeleton
with a `$ref` marker, and surfaced via the `gen_ai.content.media_uris` attribute (llm-d
extension).

**4. PII / redaction policy?**
Opt-in via `payloadCapture.redaction.enabled`. Built-in regex patterns (SSN, credit
cards, Bearer tokens). Custom `{pattern, replacement}` list. Optional external DLP
`webhookURL`. Runs before any backend sees the payload. Redaction operates on the text
skeleton only; operators relying on redaction for compliance on multimodal workloads
should disable non-text capture or apply DLP at the object-store layer.

**5. Payload size threshold?**
4 KB inline for text; `maxCompletionBufferBytes: 256 KB` buffer for SSE completions,
truncated with `gen_ai.content.truncated: true` on overflow. `maxNonTextPartBytes: 8 MiB`
per non-text part by default; tunable per deployment.

---

`payloadCapture.enabled: false` by default — zero overhead on the hot path until
explicitly opted in. `backend: noop` is a secondary kill switch that records no
payload attributes even when the master switch is enabled.

## Test plan

- [ ] Proposal reviewed against `docs/proposals/PROPOSAL_TEMPLATE.md`
- [ ] OTel Collector recipe validates cleanly
- [ ] Reviewed by impacted component maintainers (gateway, P/D sidecar, vLLM integration)
- [ ] Multimodal/non-text handling reviewed against current OpenAI / Anthropic / Gemini request shapes

---

Proposal-only PR — no implementation code per CONTRIBUTING.md. Implementation will land in
follow-up PRs once the design is approved.

### DIFF
--- a/docs/proposals/genai-payload-events.md
+++ b/docs/proposals/genai-payload-events.md
@@ -1,0 +1,313 @@
+# Capture GenAI Prompts and Completions as OTel Events
+
+## Summary
+
+Extend llm-d's existing OpenTelemetry distributed tracing to optionally capture full
+LLM prompts and completions as OTel span events, following the upstream
+[open-telemetry/semantic-conventions#2010](https://github.com/open-telemetry/semantic-conventions/issues/2010)
+specification. Payloads above a configurable size threshold are offloaded to an object
+storage backend (GCS, S3, or local filesystem) with only a reference URI stored on the
+span. The feature is disabled by default and requires explicit opt-in, preserving the
+existing metadata-only privacy posture for operators who do not need payload visibility.
+
+This proposal extends [distributed-tracing.md](distributed-tracing.md).
+
+## Motivation
+
+The existing tracing proposal deliberately excludes request/response payloads to protect
+privacy and keep span records small. However, several production use-cases require payload
+visibility that cannot be satisfied by token counts and latency metadata alone:
+
+- **Offline quality evaluation** — eval pipelines need exact prompts and completions to
+  run scoring models, compute BLEU/ROUGE, or detect prompt injection.
+- **Prompt debugging** — operators correlating unexpected outputs need the verbatim input,
+  not a summary.
+- **RAG pipeline validation** — confirming that retrieved context was injected correctly
+  requires inspecting the assembled prompt.
+- **Cost attribution** — understanding *what* drove high token usage requires the content,
+  not just the count.
+
+The upstream OTel semantic-conventions community chose an **event-based model** (not span
+attributes) specifically to decouple large payload data from the lightweight span record.
+Implementing against that spec makes llm-d payload data consumable by any OTel-native
+tooling without custom parsing.
+
+### Goals
+
+- Emit `gen_ai.content.prompt` and `gen_ai.content.completion` OTel events on the
+  relevant span following the upstream semantic convention.
+- Inline small payloads (≤ configurable threshold, default 4 KB) directly on the event
+  attribute; offload larger payloads to object storage and record only a reference URI.
+- Provide a pluggable storage interface with four built-in backends: `noop` (default),
+  `inline`, `gcs`, `s3`, and `filesystem`.
+- Expose an opt-in redaction pipeline (configurable regex patterns + optional external
+  DLP webhook) that runs before any payload reaches a backend.
+- Ensure zero overhead on the hot path when the feature is disabled.
+- Surface configuration through the deployment overlay (Kustomize, once the in-flight
+  migration of model server deployments away from Helm lands) and through standard
+  environment variables.
+
+### Non-Goals
+
+- This proposal does not modify or replace the latency, token-count, or routing
+  attributes defined in [distributed-tracing.md](distributed-tracing.md).
+- Real-time PII detection or alerting is out of scope; the redaction pipeline is
+  best-effort and advisory.
+- Streaming token-by-token event emission is out of scope for v1; completions are
+  buffered and emitted as a single event after the response is complete.
+- Guaranteed delivery of payload events to object storage (best-effort with
+  configurable timeout).
+- Breaking changes to the existing tracing configuration schema.
+
+## Proposal
+
+Operators who need payload visibility enable `payloadCapture.enabled: true` in their
+deployment overlay (or set `LLMD_PAYLOAD_CAPTURE_ENABLED=true`). All other operators
+see no change in behaviour or performance.
+
+When enabled, the relevant llm-d component:
+
+1. Serialises the request messages / prompt to JSON.
+2. Passes the JSON through the configured redaction pipeline.
+3. Compares the byte length against `inlineSizeThresholdBytes` (default 4 096).
+4. If within threshold → attaches the JSON as a `gen_ai.prompt` attribute on a
+   `gen_ai.content.prompt` OTel event on the active span.
+5. If above threshold → uploads to the configured `PayloadStore` backend and attaches
+   only the returned URI as `gen_ai.content.storage_uri`.
+6. Repeats steps 1–5 for the response completion after the full response is assembled.
+
+Success is measured by: (a) `gen_ai.content.prompt` / `gen_ai.content.completion` events
+appearing in Jaeger traces when enabled, (b) object-store objects appearing at the
+expected path, (c) no measurable latency increase on sampled requests when disabled.
+
+### User Stories
+
+#### Story 1 — Offline eval engineer
+
+As an eval engineer, I want to retrieve the exact prompt and completion for any sampled
+inference request so that I can run my scoring pipeline without maintaining a separate
+request-logging sidecar.
+
+#### Story 2 — Platform operator with PII constraints
+
+As a platform operator, I want to enable payload capture with built-in SSN and credit-card
+redaction patterns so that my audit team can inspect prompts without exposing raw PII,
+and without writing a custom logging proxy.
+
+#### Story 3 — RAG pipeline debugger
+
+As an ML engineer debugging a RAG pipeline, I want to open a Jaeger trace for a
+low-quality response and immediately see the assembled prompt (with retrieved context
+injected) so that I can confirm whether retrieval or generation was the failure point.
+
+## Design Details
+
+### OTel Event Specification
+
+Per [semconv#2010](https://github.com/open-telemetry/semantic-conventions/issues/2010),
+payload events are emitted on the **same span** that represents the LLM call.
+
+**Prompt event**
+
+| Field | Value |
+|---|---|
+| Event name | `gen_ai.content.prompt` |
+| `gen_ai.prompt` | Serialised messages / prompt JSON (omitted when offloaded) |
+| `gen_ai.content.storage_uri` | Object store URI (present only when offloaded) |
+
+**Completion event**
+
+| Field | Value |
+|---|---|
+| Event name | `gen_ai.content.completion` |
+| `gen_ai.completion` | Serialised choices JSON (omitted when offloaded) |
+| `gen_ai.content.storage_uri` | Object store URI (present only when offloaded) |
+
+When the payload is truncated due to `maxCompletionBufferBytes` being exceeded, the event
+additionally carries `gen_ai.content.truncated: true`.
+
+### Capture Layers
+
+Payload events are attached to spans at the following layers, with vLLM preferred when
+available:
+
+| Layer | Prompt capture | Completion capture | Span attached to |
+|---|---|---|---|
+| **vLLM** (preferred) | Decoded Python request object | `RequestOutput` choices | `vllm:llm_request` |
+| **Gateway (EPP)** | Raw HTTP request body (JSON parse) | Not available | `gateway.request` |
+| **P/D proxy sidecar** | Not available | Reassembled SSE stream | `llm_d.pd_proxy.request` |
+
+To avoid double-emission when vLLM tracing is active, the gateway and sidecar skip
+prompt/completion events when `payloadCapture.emitIfUpstreamTraced` is `false` (default).
+
+### PayloadStore Interface
+
+```go
+// PayloadStore persists a payload and returns a retrieval URI.
+// Implementations: NoopStore, InlineStore, GCSStore, S3Store, FilesystemStore.
+type PayloadStore interface {
+    Store(ctx context.Context, traceID, spanID string, kind PayloadKind, data []byte) (uri string, err error)
+}
+
+type PayloadKind string
+const (
+    KindPrompt     PayloadKind = "prompt"
+    KindCompletion PayloadKind = "completion"
+)
+```
+
+Backend selection via `payloadCapture.backend`:
+
+| Value | Behaviour |
+|---|---|
+| `noop` (default) | Drop silently; no event emitted |
+| `inline` | Attach to OTel event attribute; auto-offload if above threshold |
+| `gcs` | Upload to GCS, emit `gs://bucket/path` URI |
+| `s3` | Upload to S3-compatible store, emit `s3://bucket/path` URI |
+| `filesystem` | Write to mounted volume, emit `file:///path` URI |
+
+Object store path convention:
+```
+{prefix}/{YYYY}/{MM}/{DD}/{traceID}/{spanID}-{kind}.json
+```
+
+### Redaction Pipeline
+
+Runs before any backend sees the payload:
+
+```
+raw JSON → regex replacements → [optional webhook POST] → PayloadStore
+```
+
+Built-in patterns (active when `redaction.builtinPatterns: true`): US SSN, credit card
+numbers (Luhn-format), Bearer tokens in Authorization values. Custom patterns are a list
+of `{pattern, replacement}` objects. The webhook client enforces a configurable timeout;
+on error the payload is truncated rather than forwarded unredacted.
+
+### Streaming Completion Buffering
+
+SSE responses are buffered in memory up to `maxCompletionBufferBytes` (default 256 KB)
+before the completion event is emitted. If the buffer limit is reached, the event is
+emitted with the buffered content and `gen_ai.content.truncated: true`; the response
+stream continues unaffected.
+
+### Configuration
+
+The configuration schema below is the canonical shape; the concrete operator surface
+(currently Helm `values.yaml`, transitioning to Kustomize patches as the model server
+deployments are migrated) will adopt the same field names and defaults. The example
+shows the GAIE / EPP component:
+
+```yaml
+inferenceExtension:
+  payloadCapture:
+    enabled: false                      # master switch; opt-in
+    backend: noop                       # noop | inline | gcs | s3 | filesystem
+    inlineSizeThresholdBytes: 4096
+    offloadBackend: ""                  # fallback when inline exceeds threshold
+    emitIfUpstreamTraced: false         # skip if vLLM tracing is active
+    maxCompletionBufferBytes: 262144
+    gcs:
+      bucket: ""
+      prefix: "llm-d/payloads"
+    s3:
+      bucket: ""
+      endpoint: ""
+      prefix: "llm-d/payloads"
+      region: "us-east-1"
+    filesystem:
+      mountPath: "/var/llm-d/payloads"
+    redaction:
+      enabled: false
+      builtinPatterns: true
+      patterns: []
+      webhookURL: ""
+```
+
+Equivalent environment variables: `LLMD_PAYLOAD_CAPTURE_ENABLED`, `LLMD_PAYLOAD_BACKEND`,
+`LLMD_PAYLOAD_INLINE_THRESHOLD`, `LLMD_PAYLOAD_GCS_BUCKET`, `LLMD_PAYLOAD_S3_BUCKET`,
+`LLMD_PAYLOAD_FS_PATH`, `LLMD_PAYLOAD_REDACTION_ENABLED`.
+
+### OTel Collector Pipeline
+
+A new `traces/payloads` pipeline filters for `gen_ai.content.*` events and routes them
+to a dedicated exporter (filesystem or OTLP bridge for GCS/S3). The existing
+`traces` pipeline is unchanged. The redaction processor can additionally be placed in
+the collector as a defence-in-depth layer.
+
+### Security
+
+Payload capture significantly raises the sensitivity of trace data. Operators must:
+
+1. Enable TLS on all OTLP gRPC endpoints.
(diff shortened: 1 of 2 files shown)


robertgshaw2-redhat:
cc @ahg-g - FYI

robertgshaw2-redhat:
just an FYI before you go to far with the helm. We are shifting the helm charts to use kustomize for the model server deployments

nickaggarwal:
Thanks for the heads-up @robertgshaw2-redhat — really good catch before this got further along.

I've reworked the PR to take the Helm churn out:

- **Dropped all 10 `guides/**/values.yaml` stanzas** (the original 212 lines of `payloadCapture` Helm config). They'd just have to be redone for Kustomize anyway.
- **Genericised the proposal text** in `docs/proposals/genai-payload-events.md` — the configuration block is now framed as the canonical schema (field names + defaults), with an explicit note that the concrete operator surface will be Kustomize patches/overlays as the model server deployments migrate. The same field names map cleanly either way.
- **Added a Phase-1 note** that the operator opt-in surface lands alongside the Kustomize migration + the Phase-1 implementation PR, not in this proposal.
- **Updated the OTel Collector config comment** to match.

 Happy to follow up on Kustomisation once it's further along.

docs/proposals/genai-payload-events.md
```diff
@@ -0,0 +1,313 @@
...
+
+- **Offline quality evaluation** — eval pipelines need exact prompts and completions to
+  run scoring models, compute BLEU/ROUGE, or detect prompt injection.
+- **Prompt debugging** — operators correlating unexpected outputs need the verbatim input,
+  not a summary.
+- **RAG pipeline validation** — confirming that retrieved context was injected correctly
+  requires inspecting the assembled prompt.
+- **Cost attribution** — understanding *what* drove high token usage requires the content,
+  not just the count.
+
```
JeffLuoo, line 42: and also non-text payload can't be placed in trace span or logs

JeffLuoo:
Thank you for working on the proposal. When do you think the proposal is ready for review? I see it's still in draft.

nickaggarwal:
Will get it ready for 5/15

docs/monitoring/tracing/otel-collector-with-payload-capture.yaml
```diff
@@ -0,0 +1,176 @@
+# OTel Collector configuration with GenAI payload capture support.
+#
+# This variant extends otel-collector.yaml with:
+#   - A tail-sampling processor that keeps traces containing payload events
+#   - An attributes processor that redacts configured PII patterns
```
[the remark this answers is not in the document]
nickaggarwal: Resolved

docs/proposals/genai-payload-events.md
```diff
@@ -0,0 +1,432 @@
...
+before the completion event is emitted. If the buffer limit is reached, the event is
+emitted with the buffered content and `gen_ai.content.truncated: true`; the response
+stream continues unaffected.
+
+### Configuration
+
+The configuration schema below is the canonical shape; the concrete operator surface
+(currently Helm `values.yaml`, transitioning to Kustomize patches as the model server
+deployments are migrated) will adopt the same field names and defaults. The example
+shows the GAIE / EPP component:
```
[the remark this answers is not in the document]
nickaggarwal: Updated the body for PR

nickaggarwal:
@JeffLuoo Its ready for review

nickaggarwal:
@JeffLuoo @robertgshaw2-redhat 

If the proposal looks good, does this make sense to start implementation on this ?

JeffLuoo:
Overall it LGTM, will let @gyliu513 and @nicolexin to take another closer look.

ahg-g:
Thanks @nickaggarwal, I have couple of questions:
- What are the concrete next steps in terms of execution? 
- Which component is expected to implement the `PayloadStore` interface?

docs/monitoring/tracing/otel-collector-with-payload-capture.yaml
```diff
@@ -0,0 +1,233 @@
+# OTel Collector configuration with GenAI payload capture support.
```
gyliu513: I think we already have the otel collector config at https://github.com/llm-d/llm-d/blob/main/guides/recipes/observability/tracing/otel-collector.yaml, can you update it?

docs/proposals/genai-payload-events.md
```diff
@@ -0,0 +1,438 @@
...
+attributes) specifically to decouple large payload data from the lightweight span record.
+Even so, event attributes are typed as strings/numbers/bools/arrays and cannot carry
+opaque bytes; non-text payloads must therefore be offloaded to object storage and
+referenced by URI, never inlined in a span or a log record. Implementing against the
+upstream spec makes llm-d payload data consumable by any OTel-native tooling without
+custom parsing.
+
+### Goals
+
+- Emit `gen_ai.content.prompt` and `gen_ai.content.completion` OTel events on the
```
gyliu513: I think the latest otel semantic convention is 

- gen_ai.input.messages
- gen_ai.output.messages
Check https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/ for detail.

docs/proposals/genai-payload-events.md
```diff
@@ -0,0 +1,438 @@
...
+deployment overlay (or set `LLMD_PAYLOAD_CAPTURE_ENABLED=true`). All other operators
+see no change in behaviour or performance.
+
+When enabled, the relevant llm-d component:
+
+1. Serialises the request messages / prompt to JSON.
+2. Passes the JSON through the configured redaction pipeline.
+3. Compares the byte length against `inlineSizeThresholdBytes` (default 4 096).
+4. If within threshold → attaches the JSON as a `gen_ai.prompt` attribute on a
+   `gen_ai.content.prompt` OTel event on the active span.
```
gyliu513: as above comment, there is no `gen_ai.content.prompt` event now

docs/proposals/genai-payload-events.md
```diff
@@ -0,0 +1,438 @@
...
+5. If above threshold → uploads to the configured `PayloadStore` backend and attaches
+   only the returned URI as `gen_ai.content.storage_uri`.
```
gyliu513: does otel has this semantic convention `gen_ai.content.storage_uri`?

docs/proposals/genai-payload-events.md
```diff
@@ -0,0 +1,438 @@
...
+(now closed; GenAI conventions have moved to the dedicated
+[semantic-conventions-genai](https://github.com/open-telemetry/semantic-conventions-genai)
+repository), payload events are emitted on the **same span** that represents the LLM
+call. Attribute names prefixed `gen_ai.content.*` that are **not yet defined** in the
+upstream GenAI semantic conventions are marked below as *(llm-d extension)*; we will
+track upstream as the GenAI conventions stabilise and rename these attributes to match
+once standard names exist. This includes `gen_ai.content.media_uris`,
+`gen_ai.content.storage_uri`, and `gen_ai.content.truncated`. The base event names
+(`gen_ai.content.prompt`, `gen_ai.content.completion`) and the inlined-text payload
+attributes (`gen_ai.prompt`, `gen_ai.completion`) follow the upstream spec.
```
gyliu513: I think the latest otel do not include those semantic conventions, yes, you can still have llm-d extension

docs/proposals/genai-payload-events.md
```diff
@@ -0,0 +1,438 @@
...
+| `gen_ai.content.media_uris` *(llm-d extension)* | Array of object store URIs for non-text content parts (present when the completion contains any non-text media) |
+
+When the payload is truncated due to `maxCompletionBufferBytes` being exceeded, or when
+a non-text part is dropped because the configured offload target cannot return a
+resolvable URI (see [Non-Text and Multimodal Payloads](#non-text-and-multimodal-payloads)
+for the cases that produce this), the event additionally carries
+`gen_ai.content.truncated: true`. `backend: noop` is distinct: it is the
+no-event-at-all kill switch and emits no `gen_ai.content.*` events of any kind.
+
+### Non-Text and Multimodal Payloads
```
gyliu513, line 198: For multimodal support, check https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-input-messages.json to get some detail

gyliu513 (commented):
@nickaggarwal I think you may want to check the latest semantic convention for genai, the semantic convention that you are using have been deprecated, you may want to update it to use latest semantic convention.

docs/proposals/genai-payload-events.md
nickaggarwal: Thanks @gyliu513 . I've updated the whole proposal changed the deprecated gen_ai.content.prompt  gen_ai.content.completion events to the current GenAI conventions (gen_ai.input.messages, gen_ai.output.messages, gen_ai.system_instructions)

docs/proposals/genai-payload-events.md
nickaggarwal: Updated. The proposal now records gen_ai.input.messages / gen_ai.output.messages (+ gen_ai.system_instructions) as structured attributes

docs/proposals/genai-payload-events.md
nickaggarwal: I moved it (and the other extensions) out of the reserved gen_ai.*  namespace into llm_d.payload.* to avoid squatting

nickaggarwal:
Hi @ahg-g 

> Thanks @nickaggarwal, I have couple of questions:
> 
> * What are the concrete next steps in terms of execution?
> * Which component is expected to implement the `PayloadStore` interface?

Ownership: the Go PayloadStore interface and the noop/inline/gcs/s3/filesystem backends are owned by the gateway / EPP  (inferenceExtension), where the request body is already parsed and the active span is in scope

For Implementation 
Phase 1 as the first implementation PR (interface + NoopStore/InlineStore + enabled wiring in the gateway + Kustomize  opt-in surface + unit tests) Next, Phases 2–4 in order. Each phase is independently shippable behind the default-off payloadCapture.enabled flag.

docs/proposals/genai-payload-events.md
nickaggarwal: Agreed. Those three are kept as explicit llm-d extensions, now under llm_d.payload.*

docs/monitoring/tracing/otel-collector-with-payload-capture.yaml
nickaggarwal: Updated the Proposal

docs/proposals/genai-payload-events.md
nickaggarwal: Right, made this explicit. The Non-Text sections now state that blob content can be carried neither on a span nor in a log record (the attribute type system permits strings/numbers/bools/arrays only), which is exactly why blob parts are always offloaded to object storage and referenced by URI.

docs/proposals/genai-payload-events.md
nickaggarwal: Done. The proposal now maps multimodal content onto the schema's part types : blob (inline bytes), file (provider file_id), uri (external) each with mime_type + modality

docs/proposals/genai-payload-events.md
```diff
@@ -0,0 +1,482 @@
...
+deployment overlay (or set `LLMD_PAYLOAD_CAPTURE_ENABLED=true`). All other operators
+see no change in behaviour or performance.
+
+When enabled, the relevant llm-d component:
+
+1. Serialises the request messages into the upstream `gen_ai.input.messages` structure
+   (an array of role/parts messages; system content is split out to
+   `gen_ai.system_instructions`).
+2. Passes the JSON through the configured redaction pipeline.
+3. Compares the byte length against `inlineSizeThresholdBytes` (default 4 096).
```
gyliu513: s/4 096/4096

guides/recipes/observability/tracing/otel-collector.yaml
```diff
@@ -27,17 +27,55 @@ data:
...
+      #   summary: debug
+      # Keep only span events that carry gen_ai payload attributes. NOTE: this is
+      # an event-level filter — it drops non-payload span *events*, not whole
+      # spans or traces. Parent spans still flow through the `traces` pipeline
+      # above unchanged.
+      # filter/payload-events-only:
+      #   error_mode: ignore
+      #   traces:
+      #     spanevent:
+      #       - 'attributes["gen_ai.input.messages"] == nil and attributes["gen_ai.output.messages"] == nil'
```
gyliu513: This filter condition drops the very events the pipeline is meant to capture in the offload case. Per the proposal's attribute table, `gen_ai.input.messages` is **omitted when the whole payload is offloaded**, the event then carries only `llm_d.payload.storage_uri`. Such an event matches this condition (both attributes are nil) and gets dropped, so any payload above `inlineSizeThresholdBytes` (4 KB) never reaches the payload sink, and consumers lose the only pointer (`storage_uri`) to the stored object.

The condition should also keep events carrying the llm-d extension attributes:

```suggestion
      #       - 'attributes["gen_ai.input.messages"] == nil and attributes["gen_ai.output.messages"] == nil and attributes["llm_d.payload.storage_uri"] == nil and attributes["llm_d.payload.media_uris"] == nil'
```

docs/proposals/genai-payload-events.md
```diff
@@ -0,0 +1,482 @@
...
+
+Payload content is recorded using the current upstream
+[GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) —
+`gen_ai.input.messages`, `gen_ai.output.messages`, and `gen_ai.system_instructions`
+([attribute registry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/)).
+These attributes are `Opt-In` and at `Development` stability upstream; they may still
+change, so this proposal tracks them rather than pinning a frozen copy. Per the
+convention, message content is recorded **in structured form on a span event or log
+record** attached to the LLM span; when recorded on the span itself it is serialised to a
+JSON string. Each message is `{role, parts: [...]}`, where each part has a `type`
```
gyliu513: The framing here doesn't quite match upstream. The GenAI semconv records message content in one of two ways: (a) as Opt-In **span attributes** (`gen_ai.input.messages` etc.), serialized to a JSON string, or (b) on the log-based event `gen_ai.client.inference.operation.details`, "event" in the semconv means an OTel Log-based Event, not a Span Event. Note also that span event attributes have exactly the same type restrictions as span attributes, so "recorded in structured form on a span event" is self-contradictory, on a span event the content can only be a JSON string, same as on the span.

Using Span Events is a reasonable llm-d implementation choice (and the `spanevent` OTTL filter in the collector recipe is consistent with it), but the proposal should state explicitly that this is an llm-d decision that deviates from the upstream convention, rather than implying the convention prescribes span events, especially since the OTel community is moving away from span events toward logs.

gyliu513 (commented):
thanks for the update @nickaggarwal

robertgshaw2-redhat:
is this still being pushed on? would be great to get this completed

nickaggarwal:
Hi @robertgshaw2-redhat 

will update the last set of feedback by tomorrow and go ahead with the implementation in the following week

docs/proposals/genai-payload-events.md
nickaggarwal: Fixed in [`b60f926`](https://github.com/nickaggarwal/llm-d/commit/b60f9268d1c8609e611b2d67da2ffaa34d06faf5) — `4 096` → `4096`.

guides/recipes/observability/tracing/otel-collector.yaml
nickaggarwal: Good catch — this would have silently dropped every offloaded payload event, exactly the case the pipeline exists to route. Fixed in [`b60f926`](https://github.com/nickaggarwal/llm-d/commit/b60f9268d1c8609e611b2d67da2ffaa34d06faf5) by extending the drop condition to also require `llm_d.payload.storage_uri` and `llm_d.payload.media_uris` to be nil, so events that omit the inline messages but still carry an offload pointer are kept. Added inline comments explaining the offload case so this doesn't regress on the next collector update.

docs/proposals/genai-payload-events.md
nickaggarwal: You're right that the framing was wrong — the GenAI semconv actually specifies two carriers (Opt-In span attributes as JSON string, or the log-based event `gen_ai.client.inference.operation.details`), and span-event attributes share the span-attribute type system so "in structured form on a span event" was self-contradictory.

Fixed in [`b60f926`](https://github.com/nickaggarwal/llm-d/commit/b60f9268d1c8609e611b2d67da2ffaa34d06faf5):

- Restated the upstream carriers accurately in **Summary**, **Motivation**, and a new **Upstream carriers** paragraph in the OTel Attribute Specification section.
- Added an **llm-d carrier — design decision** callout that explicitly states span events are an llm-d choice (not what the convention prescribes), gives the reasons (event-scoped Collector routing surface, no logs pipeline required, keeps the parent LLM span light for consumers that don't opt in), and acknowledges the OTel community's move toward Log-based Events.
- Documented the migration plan: if the upstream convention firms up around log records or Opt-In span attributes, a future revision migrates the carrier without changing the attribute names.
- Updated **Goals** and **Proposal step 4** to match the same framing (JSON-string encoding, span-event attributes typed identically to span attributes).

nickaggarwal:
@gyliu513 — pushed [`b60f926`](https://github.com/nickaggarwal/llm-d/commit/b60f9268d1c8609e611b2d67da2ffaa34d06faf5) addressing the three unresolved review comments. Individual replies inline above; short summary:

1. **`s/4 096/4096`** on `inlineSizeThresholdBytes` default.
2. **Collector `filter/payload-events-only`** — extended the drop condition to keep events that carry only `llm_d.payload.storage_uri` / `llm_d.payload.media_uris` (i.e. the offloaded case). Without this, any payload above 4 KB was silently discarded before reaching the payload sink.
3. **Span-event vs log-event framing** — the proposal previously implied the upstream convention prescribes span events, which it doesn't. Now correctly describes the two upstream carriers (Opt-In span attributes serialised to JSON, or the log-based event `gen_ai.client.inference.operation.details`) and explicitly declares llm-d's span-event carrier as a design decision with rationale, deviation note, and migration plan.

Sorry for the delay @robertgshaw2-redhat. Ready to start the Phase-1 implementation PR once these are resolved.

nickaggarwal:
With the three unresolved review comments now addressed in [`b60f926`](https://github.com/nickaggarwal/llm-d/commit/b60f9268d1c8609e611b2d67da2ffaa34d06faf5), I'll proceed with the Phase-1 implementation ?


**Phase 1 scope (from the proposal):**
- `PayloadStore` interface in the gateway / EPP (`inferenceExtension`) — the layer that already parses the request body and holds the active LLM span.
- `NoopStore` (default, zero-overhead) and `InlineStore` implementations.
- Wire into gateway `server.go` `Process()` behind the default-off `payloadCapture.enabled` flag; no code path runs when disabled.
- Kustomize-aligned opt-in surface (per @robertgshaw2-redhat's earlier steer to skip Helm churn), plus equivalent env-var aliases.
- Unit tests for the interface and both backends; integration coverage that the `payloadCapture.enabled: false` path is a no-op.

Phases 2–4 (object-store backends, redaction pipeline, vLLM native integration) will land as separate PRs, each independently shippable behind the same default-off flag.

@JeffLuoo @gyliu513 @ahg-g @robertgshaw2-redhat @nicolexin — flagging in case anyone wants a final look at the proposal before I open the implementation PR.

PR #1230, opened 2026-04-22
Support tiered prefix cache on Intel XPU with offloading connector

XinyuYe-Intel:
Support tiered prefix cache on Intel XPU with offloading connector.


ahg-g:
@XinyuYe-Intel can you please address copilot comments (if addressed please resolve them).

Also, can Intel provide capacity for CI test coverage for this guide?

XinyuYe-Intel:
> @XinyuYe-Intel can you please address copilot comments (if addressed please resolve them).
> 
> Also, can Intel provide capacity for CI test coverage for this guide?

Hi @ahg-g , copilot comments are all resolved now.

Yes, we can provide capacity for CI test coverage for this guide.

XinyuYe-Intel:
Added benchmarks info and data of Intel B60.

XinyuYe-Intel:
Hi @ahg-g , could you please review this pr, it's ready now.

yuanwu2017 (dismissed):

guides/tiered-prefix-cache/modelserver/xpu/vllm/lmcache-connector/cpu/base/kustomization.yaml
```diff
@@ -3,11 +3,25 @@ kind: Kustomization
 resources:
   - ../../../base
 
+images:
+  - name: ghcr.io/llm-d/llm-d-xpu-dev
+    newTag: pr-1731
```
yuanwu2017: If it inherits from the base, can we remove the images?

guides/tiered-prefix-cache/modelserver/xpu/vllm/lmcache-connector/cpu/base/patch-vllm.yaml
```diff
@@ -7,16 +7,19 @@ spec:
...
+                --dtype bfloat16 \
+                --block-size 64 \
+                --tensor-parallel-size=1 \
+                --max-num-seq 128 \
+                --max-model-len 20480 \
+                --uvicorn-log-level warning \
+                --kv-transfer-config \
+                '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'
           env:
             - name: HF_TOKEN
```
yuanwu2017, line 24: Same.

guides/tiered-prefix-cache/modelserver/xpu/vllm/native/cpu/base/patch-vllm.yaml
```diff
@@ -0,0 +1,37 @@
...
+                --dtype bfloat16 \
+                --block-size 64 \
+                --tensor-parallel-size=4 \
+                --gpu-memory-utilization 0.85 \
+                --kv-transfer-config \
+                '{"kv_connector":"OffloadingConnector","kv_role":"kv_both","kv_connector_extra_config":{"cpu_bytes_to_use":107374182400}}' \
+                --max-num-seq 128 \
+                --uvicorn-log-level warning
+          env:
+            - name: HF_TOKEN
```
yuanwu2017, line 24: Same

yuanwu2017 (changes requested):

guides/tiered-prefix-cache/modelserver/xpu/vllm/lmcache-connector/cpu/base/kustomization.yaml
XinyuYe-Intel: Yes,  you are right, I'll remove the image.

guides/tiered-prefix-cache/modelserver/xpu/vllm/lmcache-connector/cpu/base/patch-vllm.yaml
XinyuYe-Intel: This one can't just use base as it differ with base in vllm arguments and other settings.

guides/tiered-prefix-cache/modelserver/xpu/vllm/native/cpu/base/patch-vllm.yaml
XinyuYe-Intel: This one can't just use base as it differ with base in vllm arguments and other settings.

yuanwu2017 (dismissed):

yuanwu2017 (dismissed):

yuanwu2017:
Heads-up on the CUDA build failures showing up here — they are not caused by anything in this PR, and there is nothing to change on this branch.

Editing `docker/common-versions` is what triggered them. `detect-changes` in `build-image.yaml` lists that file as a path trigger for *every* platform (cuda, cuda-debug, gb200, aws, cpu, xpu), so before this change the CUDA jobs were being skipped entirely and afterwards they all started running. They then hit a pre-existing bug in `docker/scripts/common/setup-sccache.sh` that had been latent because no recent PR had touched the shared version file.

Filed and fixed separately:

- #2223 — the bug
- #2224 — the fix (three commits; the failure surfaced in three successive layers)
- #2225 — related, not blocking: the sccache S3 credentials are invalid, so the fallback path runs on every build in the first place

Merge order matters here: the CUDA checks on this PR will stay red until #2224 lands, after which this branch needs a rebase or a re-run. The XPU job itself is green — `xpu-build-llm-d` on [31362149292](https://github.com/llm-d/llm-d/actions/runs/31362149292) built and pushed successfully with the new base image.

Separately, on the XPU change itself: I verified the built image end to end on a 16× Max 1550 host, with `BASE_IMAGE` sourced from `docker/common-versions` exactly as CI does:

```
torch 2.12.0+xpu | xpu avail: True | count: 16
LMCache INFO: Using backend: lmcache.xpu_ops
vllm: 0.26.0+xpu
icpx: /opt/intel/oneapi/compiler/latest/bin/icpx
```

Worth being aware of for the future: the `ENV` block is a workaround for v0.26.0 shipping oneAPI without exporting its environment. The nightly images have already moved oneAPI into pip wheels under `/opt/venv`, which fixes the environment problem but removes `icpx` altogether — I checked, and there is no pip-installable DPC++ compiler (`dpcpp-cpp-compiler` and `intel-dpcpp-cpp-compiler` both 404 on PyPI). So when the base image is next bumped, the `ENV` block can go away but `BUILD_WITH_SYCL=1` will need a different arrangement, probably a multi-stage build or Intel's own package index. Not a problem for this PR, just something that will bite at the next upgrade.

yuanwu2017 (approved):

PR #1231, opened 2026-04-22
update: clean deprecated env variable for vllm to build AVX

zdtsw:
# Description

- since vllm 0.17+, this env is not in use
- build for both and runtime check node to use the correct one

ref https://github.com/vllm-project/vllm/pull/35466


zdtsw:
cc @diegocastanibm

diegocastanibm (approved):
LGTM

PR #1244, opened 2026-04-23
Test nightly with a llm-d-infra branch

diegocastanibm:
**Context:**

 When making changes that span both llm-d (guides, workflow files) and llm-d-infra (reusable
 workflows, infra scripts), there's no way to test them together before merging. The nightly
 workflows in llm-d hardcode @main when calling reusable workflows from llm-d-infra:

 uses: llm-d/llm-d-infra/.github/workflows/reusable-nightly-e2e-openshift-helmfile.yaml@main

 GitHub Actions does not support expressions in uses: — it must be a static string. The
 workaround is to rewrite the YAML on a temp branch before dispatching, replacing @main with
 the desired infra branch ref.

**New syntax:**

 `/test-nightly <workflow-name-or-glob> [infra:<branch>]`

 Examples:
```
 /test-nightly pd-disaggregation-ocp infra:my-feature-branch
 /test-nightly *-ocp infra:fix-helmfile
 /test-nightly simulated-accelerators              # uses @main (unchanged behavior)
```
 The infra: prefix makes the argument self-documenting and avoids ambiguity with potential
 future positional args.

@llm-d/release-crew

### DIFF
--- a/.github/workflows/slash-test-nightly.yaml
+++ b/.github/workflows/slash-test-nightly.yaml
@@ -1,6 +1,6 @@
 name: /test-nightly Slash Command
 
-# Allows running any nightly E2E workflow against a PR's code.
+# Allows running any nightly E2E workflow against a fork PR's code.
 #
 # Usage: Comment on a PR with:
 #   /test-nightly <nightly-name>       # exact match
--- a/.github/workflows/slash-test-nightly.yaml
+++ b/.github/workflows/slash-test-nightly.yaml
@@ -1,16 +1,25 @@
 name: /test-nightly Slash Command
 
-# Allows running any nightly E2E workflow against a fork PR's code.
+# Allows running any nightly E2E workflow against a PR's code.
 #
 # Usage: Comment on a PR with:
 #   /test-nightly <nightly-name>       # exact match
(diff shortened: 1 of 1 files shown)


diegocastanibm:
This PR always runs the workflow from the default branch. So the `workflows: write` permission we added won't take effect until it's merged to main. Therefore I need to merge just the `workflows: write` permission to main first to avoid a chicken-and-egg problem. The PR to do so is: https://github.com/llm-d/llm-d/pull/1245

diegocastanibm:
I was wrong. `workflows` is not a valid permission for the permissions: block in GitHub Actions. The `GITHUB_TOKEN` cannot be granted the workflows scope. That's only available on Personal Access Tokens (PATs).
This means `GITHUB_TOKEN` can never push a branch that contains modified workflow files. And since our `infra:` feature proposed in this PR always modifies `nightly-e2e-*.yaml` files via `sed`, the push will always fail when `infra:` is used.

I'll discuss the best solution for this issue with the @llm-d/release-crew team

diegocastanibm:
Closing based on conversations with the @llm-d/release-crew

PR #1277, opened 2026-04-28
Added API and HTTP Headers References and updated the inferencepool architectural doc

ahg-g:


chcost (commented):
Thanks for this PR! A few minor suggestions below.

docs/wip-docs-new/architecture/core/inferencepool.md
```diff
@@ -1,170 +1,38 @@
...
+The following diagram visualizes how the `InferencePool` resource is involved in the control path of both the EPP and Gateway Controller:
 
-### InferencePool Spec
+![InferencePool](../../../assets/inferencepool.svg)
 
-The InferencePool custom resource has three core fields:
+### 1. Endpoint Discovery (EPP Perspective)
 
-#### `selector`
+The EPP uses the `InferencePool` to disocver which pods it can pick from.
```
chcost, line 21: Typo: "disocver" → "discover"

Also, consider adding a link to the EPP documentation:

```suggestion
-The EPP uses the `InferencePool` to disocver which pods it can pick from.
+The [EPP](epp/index.md) uses the `InferencePool` to discover which pods it can pick from.
```

docs/wip-docs-new/api-reference/http-headers.md
```diff
@@ -0,0 +1,29 @@
...
+| Header | Description |
+| --- | --- |
+| `x-slo-ttft-ms` | Specifies the target Time To First Token (TTFT) in milliseconds. Used by plugins to determine if a request can be admitted while meeting the latency goal. |
+| `x-slo-tpot-ms` | Specifies the target Time Per Output Token (TPOT) in milliseconds. Used for admission control based on predicted or observed token generation latency. |
+
+---
+
+## Implementation Notes
+
+- **Case Sensitivity:** All header lookups are case-insensitive.
```
chcost, line 28: Consider clarifying the case-sensitivity note to reference the HTTP standard:

```suggestion
-- **Case Sensitivity:** All header lookups are case-insensitive.
+- **Case Sensitivity:** HTTP header names are case-insensitive per RFC 7230. The EPP normalizes all header names for lookup.
```

This clarification is helpful because HTTP/2 headers are always lowercase, so the current wording might be confusing.

chcost:
The API reference README.md should be updated to include navigation links to the new pages:
- InferencePool API Reference
- InferenceObjective API Reference
- InferenceModelRewrite API Reference
- HTTP Headers Reference

This will improve discoverability of the new documentation.

robertgshaw2-redhat:
do you think we should put the EPPConfig here instead of in epp/configuration?

Can be a separate PR

chcost:
i'd suggest we add cross-references between the two. In InferencePool API reference: mention that EPP consumes these resources, link to EPP docs. In EPP configuration: reference InferencePool API when explaining pool discovery. This follows the Kubernetes docs pattern where API references are kept separate from component configuration guides.

ahg-g:
I added the readme.md file, I also added the config api reference and made it clear that it is not a k8s object api, it is a config api.

docs/wip-docs-new/api-reference/endpointpickerconfig.md
```diff
@@ -0,0 +1,93 @@
+# EndpointPickerConfig (EPP Configuration)
```
robertgshaw2-redhat, line 1: do we need to remove `architecture/core/epp/configuration.md`?

ahg-g: No, this is just an api reference for the api fields themselves, just like we have api-reference for inferencepool and a doc on inferencepool in the architecture.

architecture/core/epp/configuration.md is the doc on how to configure the various EPP components in detail with examples.

robertgshaw2-redhat (approved):

PR #1282, opened 2026-04-28
Add details on gke patch for optimized baseline

liu-cong:


liu-cong:
The CI is passing: https://github.com/llm-d/llm-d/actions/runs/25076824679/job/73471563439

PR #1184, opened 2026-04-16
docs: llm-d-inference-payload-processor proposal

nilig:
Proposal to create the `llm-d/llm-d-inference-payload-processor` repository and migrate code from BBR.

Background and naming discussion: #1154

cc @nirrozenbaum @ahg-g

Closes #1154

PR #1294, opened 2026-04-29
docs: inference-payload-processor-rs incubation repo proposal

nilig:
Proposal to create `llm-d-incubation/inference-payload-processor-rs` to experiment with a Rust-based implementation of the inference payload processor.

Related: #1184

cc @nirrozenbaum @chcost


chcost (approved):
/approve

PR #1296, opened 2026-04-29
Ensure `precise-prefix-cache-aware` passes both CKS and OCP

maugustosilva:
After a fix directly on `llm-d-infra`, only OCP required changes


Gregory-Pereira (approved):
/lgtm

PR #1258, opened 2026-04-24
precise-prefix-cache-scheduling guide helm -> kustomize

vMaroon:
## Summary

Refactors `guides/precise-prefix-cache-aware/` from helmfile to kustomize, aligned with the `optimized-baseline` guide layout, and adopts the data-layer config shape from [llm-d-inference-scheduler#862](https://github.com/llm-d/llm-d-inference-scheduler/pull/862).

## Changes

**Install shape** — single `standalone` chart helm install (no separate `llm-d-infra` / `inferencepool` releases). UDS tokenizer sidecar is appended to the chart's pod via a helm post-renderer at `scheduler/patches/uds-tokenizer/` (v4 `postrenderer/v1` plugin, also works as a v3 path-based `--post-renderer`).

**Layout** (mirrors `liu-cong/llm-d` optimized-baseline):
- `modelserver/` — 7 accelerator overlays (`gpu`, `amd`, `cpu`, `hpu`, `tpu-v6`, `tpu-v7`, `xpu`), all vLLM.
- `modelserver/components/active-active/` — kustomize Component that switches vLLM from central ZMQ to per-pod publishing (`tcp://*:5556` + container port `5556`).
- `scheduler/precise-prefix-cache-aware.values.yaml` — central mode (default).
- `scheduler/features/active-active.values.yaml` — layered overlay: `replicas: 2`, data-layer `endpoint-notification-source` → scorer `ExtractEndpoint`, and `--ha-enable-leader-election=false` override (chart would otherwise auto-add the bare flag and collapse to active-passive).

**Config updates** — vLLM patches add `--block-size=64`, `--kv-events-config` (with `$(KV_EVENTS_ENDPOINT)` / `$(POD_IP)` / `$(POD_PORT)` env-var expansion). Scheduler uses `v0.8.0-rc.1`.

**Sub-guide** — pod-discovery renamed to active-active; `active-active.md` documents the 2-gateway + 2-scheduler topology, the leader-election workaround, and verification steps.

## Testing

End-to-end on an OpenShift cluster with 2× `Qwen/Qwen3-32B` replicas × TP=2.

**Central mode (default install):**
- Scheduler pod `3/3 Running` (`tokenizer-uds` + `epp` + `envoy-sidecar`).
- Repeat-prompt test: request 1 → `scores: {}`; request 2 (same prompt) → `scores: {"<pod-ip>:8000": 2}` on the cache-warm pod. Precise prefix cache scorer routed traffic based on real KV-cache state.

[...]

PR #1251, opened 2026-04-24
fix: migrate e2e-optimized-baseline-xpu workflow from helmfile to kus…

yuanwu2017:
…tomize+helm

The optimized-baseline guide was migrated from helmfile to kustomize+helm in commit 645891a (#1131), but the XPU e2e workflow was not updated accordingly. This caused the CI to fail because:
- helmfile.yaml.gotmpl no longer exists in guides/optimized-baseline/
- httproute.yaml no longer exists in guides/optimized-baseline/
- helmfile destroy in cleanup also fails

This commit updates the workflow to:
- Deploy model server via kustomize build (guides/optimized-baseline/modelserver/xpu/vllm/)
- Deploy scheduler via helm install (inferencepool chart v1.4.0)
- Deploy gateway+HTTPRoute via kustomize (guides/recipes/gateway/istio)
- Use kustomize edit set image for custom image overrides
- Clean up with helm uninstall + kubectl delete instead of helmfile destroy

PR #1297, opened 2026-04-29
ci: migrate e2e-prefix-cache-xpu from helmfile to kustomize

xiaojun-zhang:
## Summary

Migrates `.github/workflows/e2e-prefix-cache-xpu.yaml` from the old helmfile-based deploy to the kustomize + helm-standalone layout introduced by #1258. The nightly XPU workflow has been failing at the deploy step since #1258 merged because it still called `helmfile apply -e xpu` against files that no longer exist (`helmfile.yaml.gotmpl`, `ms-kv-events/values_xpu*.yaml`, `gaie-kv-events/values*.yaml`, `httproute.yaml`).

Structurally mirrors the optimized-baseline XPU migration in #1251, with two precise-prefix-specific additions:

1. `--post-renderer ./guides/precise-prefix-cache-aware/scheduler/patches/uds-tokenizer/post-renderer.sh` on the helm install — grafts the UDS tokenizer sidecar onto the scheduler pod (the standalone chart's `sidecar.*` slot is owned by Envoy).

2. Pre-install `sed` rewrite on the shared scheduler values file to pin `modelName` to `Qwen/Qwen3-0.6B`:

   ```
   sed -i 's|modelName: Qwen/Qwen3-32B|modelName: Qwen/Qwen3-0.6B|g' \
     guides/precise-prefix-cache-aware/scheduler/precise-prefix-cache-aware.values.yaml
   ```

   This is the CI automation of the existing operator-facing NOTE in [`guides/precise-prefix-cache-aware/README.md`](https://github.com/llm-d/llm-d/blob/main/guides/precise-prefix-cache-aware/README.md?plain=1#L42) — not a new concept. The shared scheduler values file ships the Qwen/Qwen3-32B default (nvidia-gpu) while the XPU overlay deploys Qwen/Qwen3-0.6B; without this rewrite the scorer's in-EPP tokenizer hashes against the 32B vocab and misses every KV lookup, silently dropping precise-prefix-cache scores to 0. On today's main, the string `Qwen/Qwen3-32B` appears only in the two `modelName:` keys of that file (verified with grep), so the sed is safe.

Other structural changes (lifted from #1251):

- Drop the helmfile-era HTTPRoute step — standalone chart runs Envoy as a scheduler sidecar, no `Gateway` resource.
- Install Gateway API Inference Extension CRDs via `kubectl apply -k .../config/crd?ref=v1.4.0` instead of the gateway-provider helmfile.
- `GATEWAY_HOST=precise-prefix-cache-aware-epp.<ns>.svc.cluster.local` for `e2e-validate.sh` (standalone mode; Service port 80 → targetPort 8081 Envoy sidecar).
- `custom_image_tag` input supports bare tags (expanded to `ghcr.io/<repo>-xpu-dev:<tag>`) and full image references. Applied via `kustomize edit set image REPLACE_MODEL_SERVER_IMAGE=...` before the `kubectl apply`.
- Artifact bundle gains a per-container EPP logs bundle and a snapshot of the rendered plugins `ConfigMap` for post-mortem debugging of scorer config coupling.
- Cron shifted to 10:30 UTC so it doesn't pile on top of `e2e-optimized-baseline-xpu`'s 10:00 UTC slot on the shared self-hosted runner.

### Back-to-back workflow runs on a fork-registered self-hosted XPU runner

> Disclosure: I do not have permission to trigger workflows against the upstream `llm-d/llm-d` runners, so the runs below were executed on **my own fork** (`xiaojun-zhang/llm-d`) using a self-hosted runner I registered on the same physical XPU host as the upstream `xpu` runner (`bmg817225`, 8× Intel Arc Pro B60, Ubuntu 24.04, kind cluster with DRA). They are labelled `xpu`, run the same workflow file in this PR, and exercise identical hardware — the only difference from an upstream run is the `github.repository` guard at the top of the job. The upstream nightly on the `llm-d/llm-d` runner is still required to consider this PR truly validated in-tree.

Both runs dispatched via `workflow_dispatch` on a `xpu-precise-prefix-cache-refactor-testing` branch that relaxes the `github.repository == 'llm-d/llm-d'` guard (testing branch only — not included in this PR):

| Run | Conclusion | Duration | URL |
|-----|------------|----------|-----|
| 1 | success | 3m55s | https://github.com/xiaojun-zhang/llm-d/actions/runs/25134955902 |
| 2 | success | 3m52s | https://github.com/xiaojun-zhang/llm-d/actions/runs/25135248578 |

All 21 steps green on both runs, including the sed-patch, helm install (standalone + UDS post-renderer), kustomize apply, pod ready wait, `e2e-validate.sh` (10 chat + 10 completion iterations), artifact upload, and full namespace teardown. The workflow is self-contained and idempotent across runs.

### Additional manual replay (pre-CI)

Separately verified end-to-end on the same host before registering the runner:

- `helm install` rendered scheduler Deployment **3/3 Ready** (epp + envoy-sidecar + tokenizer-uds).
- Scheduler ConfigMap carried `Qwen/Qwen3-0.6B` in both `modelName` fields + `blockSize: 64` after the sed pre-step.
- `kubectl apply -k modelserver/xpu/vllm`: 2 decode pods bound to clean XPUs via DRA, **1/1 Ready in ~3 min**.
- Repeat-prompt smoke test: identical long-prompt pair R1 (`81e0b6b5`) and R2 (`0093fa5a`) both routed to endpoint `10.244.0.22:8000` — precise-prefix-cache-scorer end-to-end working (second request landed on the pod whose KV cache was warmed by the first).

## Commit history

Six DCO + SSH-signed commits telling the full deliberation:

- Initial option-A work (`scheduler/plugins/xpu.yaml` + `--set-file`): `4f45409`, `831808d`, `fce2cb0`.
- Two reverts switching away from option A: `06d4a8f`, `80451dc`.
- Final option-B workflow (sed-based rewrite): `25de3db`.

Net diff against main is a single file: `.github/workflows/e2e-prefix-cache-xpu.yaml` (+92 / −87 from the helmfile era). Happy to squash on request.

## Test plan

- [x] `kustomize build guides/precise-prefix-cache-aware/modelserver/xpu/vllm` renders cleanly (RCT namePrefix rewrite via `namereference.yaml` works).
- [x] `helm template ... + post-renderer` produces 3-container EPP Deployment (epp + envoy-sidecar + tokenizer-uds).
- [x] After sed, ConfigMap `precise-prefix-cache-aware-epp` carries `Qwen/Qwen3-0.6B` in both `modelName` keys.
- [x] Local e2e (manual `helm`/`kubectl`) on self-hosted XPU: 2 decode pods bound to idle cards via DRA, 1/1 Ready.
- [x] Repeat-prompt smoke: R1 and R2 routed to same endpoint after KV-events propagate.
- [x] **Workflow runs green end-to-end on a GitHub Actions self-hosted XPU runner I registered on my fork `xiaojun-zhang/llm-d` (same physical host as upstream `xpu`; 2× back-to-back, ~4 min each, idempotent teardown).** See "Evidence" section above for links + caveats.
- [ ] **Upstream-runner validation**: once a maintainer triggers this workflow on the `llm-d/llm-d` self-hosted `xpu` runner (or merges and the nightly fires), confirm it reaches the same green state.

cc @vMaroon (author of #1258)


xiaojun-zhang:
@liu-cong @Gregory-Pereira @vMaroon @yuanwu2017 can you please take a look? I'm trying to make this for 0.7.0 release. 
I have tested the workflow with a local github runner.

liu-cong (approved):

PR #1308, opened 2026-04-30
update the llm-d-fs-connector kustomizations for v0.8

effi-ofer:
1. Updated the wheel release
2. pip install fails because it doesn't have access to the global install dir, and the local directory is not in the python path.  There are multiple ways of fixing this, I picked installing into /tmp and updating the path to reference it.
3. the pvc is different than the one created in guides/tiered-prefix-cache/storage/manifests/pvc.yaml which is used by the lmcache kustomizations.

### DIFF
--- a/guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
+++ b/guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
@@ -9,7 +9,7 @@ patches:
     patch: |-
       - op: replace
         path: /spec/template/spec/containers/0/image
-        value: vllm/vllm-openai:v0.15.1
+        value: vllm/vllm-openai:v0.19.1
 
   # Add Offloading Connector with llm-d FS Backend
   - target:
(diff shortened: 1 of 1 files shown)


effi-ofer:
@kfirtoledo please review

guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
```diff
@@ -61,7 +66,7 @@ patches:
         value:
           name: files-storage
           persistentVolumeClaim:
-            claimName: pvc-storage-files
+            claimName: llm-d-kv-cache-storage
```
kfirtoledo, line 70: Is this consistent with the PVC yaml name?

guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
```diff
@@ -9,33 +9,38 @@ patches:
...
-                "spec_module_path": "llmd_fs_backend.spec"
-              }
-            }'
+        path: /spec/template/spec/containers/0/command
+        value: ["/bin/sh", "-c"]
+      - op: replace
+        path: /spec/template/spec/containers/0/args
+        value:
+          - |-
+            pip install --target /tmp/llmd-packages https://github.com/llm-d/llm-d-kv-cache/releases/download/v0.8.0-rc.1/llmd_fs_connector-0.19-cp312-cp312-linux_x86_64.whl
```
kfirtoledo: Can you install it like the readme
```
pip install 'llmd-fs-connector==0.19' \
  --extra-index-url https://llm-d.github.io/llm-d-kv-cache/simple/
```

kfirtoledo (changes requested):

guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
effi-ofer: Yes

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: llm-d-kv-cache-storage # If you change this, update the ./manifests/vllm patches accordingly.
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 18000Gi
  storageClassName: $STORAGE_CLASS # Specify the storage class here, or remove this line to use the default storage class.
```

guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
```diff
@@ -9,33 +9,39 @@ patches:
...
-                "spec_module_path": "llmd_fs_backend.spec"
-              }
-            }'
+        path: /spec/template/spec/containers/0/command
+        value: ["/bin/sh", "-c"]
+      - op: replace
+        path: /spec/template/spec/containers/0/args
+        value:
+          - |-
+            pip install --target /tmp/llmd-packages 'llmd-fs-connector==0.19' \
```
Sneha-at, line 26: Is this tied to a version? Can we use this guide as-is for other llm-d- s connector versions?

vMaroon: https://llm-d.ai/llm-d-kv-cache/simple/llmd-fs-connector/ works for all in here

effi-ofer: I think it is safer than using latest and possibly not being compatible.

maugustosilva (approved):

PR #1313, opened 2026-04-30
[Docs]: Update docs with latest data producer plugin changes

rahulgurnani:
Recent changes in data producer plugins in https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2846
make configuring data producer explicitly in order for parameters to pass. 
This approach of explicitly specifying can still be used in v1.5 release of EPP. So, we can update the docs so that future version don't break as in future versions, specifiying these parameters explicitly with data producers is necessary.


ahg-g:
pls wait until https://github.com/llm-d/llm-d/pull/1305 merges, which will move the `epp` directory to be under `router`

PR #1318, opened 2026-04-30
[Guides Storage]: Align tiered-prefix-cache storage guide with optimized-baseline

kfirtoledo:
## Summary

The storage kustomization referenced \`recipes/vllm/standard\` (removed in #1131), so \`kubectl kustomize\` errored out before any apply. This PR aligns the guide with the [optimized-baseline](https://github.com/llm-d/llm-d/tree/main/guides/optimized-baseline) pattern that's already CI-tested.

- Switch base to \`recipes/modelserver/base/single-host/default\`, use kustomize \`namePrefix\`/\`images\`, and replace JSON patches with a strategic-merge \`deployment-patch.yaml\`.
- Bump vLLM to v0.19.1 and install \`llmd-fs-connector==0.19\` from the pip index (matching the connector's [README](https://github.com/llm-d/llm-d-kv-cache/blob/main/kv_connectors/llmd_fs_backend/README.md)).
- Add the volume mounts the model server needs under restricted SCC (\`/.cache\`, \`/.triton\`, \`/.local\`, \`/dev/shm\`) and a longer \`startupProbe\` so Qwen3-32B has time to load before kubelet kills the pod.
- README: fix stale resource names users would see when verifying (\`llm-d-route\` → \`llm-d-infpool\`, \`llm-d-model-server-*\` → \`llm-d-decode-*\`, PVC and storage class shown as placeholders) and add a port-forward curl example to test inference end-to-end through the Gateway.

## Test plan

- [x] Deployed end-to-end on OpenShift (IBM Spectrum Scale RWX PVC) 
- [x] Verified KV cache write to shared PVC (\`/mnt/files-storage/kv-cache/Qwen/Qwen3-32B/...\`).
- [x] Verified cross-pod prefix cache hit (pod 2 reads cached blocks pod 1 wrote — \`External prefix cache hit rate: 99.9%\`).

### DIFF
--- a/guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
+++ b/guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
@@ -1,92 +1,14 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - ../../../../../recipes/vllm/standard
+  - ../../../../../recipes/modelserver/base/single-host/default
+
+namePrefix: llm-d-
+
+images:
+  - name: REPLACE_MODEL_SERVER_IMAGE
+    newName: vllm/vllm-openai
+    newTag: v0.19.1
+
 patches:
-  - target:
-      kind: Deployment
-      name: llm-d-model-server
-    patch: |-
-      - op: replace
-        path: /spec/template/spec/containers/0/image
-        value: vllm/vllm-openai:v0.19.1
-
-  # Add Offloading Connector with llm-d FS Backend
-  - target:
-      kind: Deployment
-      name: llm-d-model-server
-    patch: |-
-      - op: replace
-        path: /spec/template/spec/containers/0/command
-        value: ["/bin/sh", "-c"]
-      - op: replace
-        path: /spec/template/spec/containers/0/args
-        value:
-          - |-
-            pip install --target /tmp/llmd-packages 'llmd-fs-connector==0.19' \
-              --extra-index-url https://llm-d.github.io/llm-d-kv-cache/simple/
-            export PYTHONPATH="/tmp/llmd-packages:${PYTHONPATH:-}"
-            exec vllm serve \
-              Qwen/Qwen3-32B \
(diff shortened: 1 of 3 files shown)


kfirtoledo:
@vMaroon @liu-cong , @Sneha-at   PTAL

guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
```diff
@@ -1,92 +1,15 @@
...
-    patch: |-
-      - op: add
-        path: /spec/template/spec/containers/0/resources
-        value:
-          limits:
-            nvidia.com/gpu: 2
-          requests:
-            nvidia.com/gpu: 2
-            memory: 400G
+namePrefix: llm-d-
```
liu-cong, line 6: For consistency with other guides, use `tiered-prefix-cache-storage-nvidia-gpu-vllm-`

guides/tiered-prefix-cache/storage/manifests/vllm/llm-d-fs-connector/kustomization.yaml
```diff
@@ -1,92 +1,15 @@
...
-      - op: add
-        path: /spec/template/spec/volumes/-
-        value:
-          name: files-storage
-          persistentVolumeClaim:
-            claimName: llm-d-kv-cache-storage
+images:
+  - name: REPLACE_MODEL_SERVER_IMAGE
+    newName: vllm/vllm-openai
+    # Keep in sync with deployment-patch.yaml: llmd-fs-connector==0.X must match vLLM 0.X.x.
```
liu-cong, line 11: since this needs to match the fs connector version, let's just add the image directly in the patch yaml so it's easier to discover

liu-cong (commented):
We still need to change the guide style for consistency but I can do that as a follow up.

Thanks for fixing the manfiest!

robertgshaw2-redhat:
🐐

vMaroon:
@liu-cong can you follow-up with the requested changes directly? Availability on Friday is limited, better not hold it. Will merge this as-is.

vMaroon (approved):

PR #1321, opened 2026-04-30
Fix precise-prefix-cache guide: remove tokenizer plugin, fix speculativeIndexing placement

bongwoobak:
## Summary
1. Remove `type: tokenizer` plugin and `pluginRef: tokenizer` from scheduling profiles in both central and active-active configs. In v0.8.0 the tokenizer plugin is a Scorer — without a `pluginRef` in the scheduling profile it never executes. The `precise-prefix-cache-scorer` has its own tokenizer pool (`tokenizersPoolConfig.uds`) and does not need the external plugin.

2. Fix `speculativeIndexing` placement — move from inside `indexerConfig` (where it was silently ignored as an unknown field on `kvcache.Config`) to the correct position under `parameters` (`PluginConfig` level).

## Files changed
- `scheduler/precise-prefix-cache-aware.values.yaml`
- `scheduler/features/active-active.values.yaml`

### DIFF
--- a/guides/precise-prefix-cache-aware/scheduler/precise-prefix-cache-aware.values.yaml
+++ b/guides/precise-prefix-cache-aware/scheduler/precise-prefix-cache-aware.values.yaml
@@ -79,8 +79,8 @@ inferenceExtension:
           parameters:
             tokenProcessorConfig:
               blockSize: 64           # must match vLLM --block-size
+            speculativeIndexing: true
             indexerConfig:
-              speculativeIndexing: true
               tokenizersPoolConfig:
                 modelName: Qwen/Qwen3-32B
                 uds:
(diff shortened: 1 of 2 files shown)


guides/precise-prefix-cache-aware/scheduler/precise-prefix-cache-aware.values.yaml
```diff
@@ -68,19 +68,14 @@ inferenceExtension:
       apiVersion: inference.networking.x-k8s.io/v1alpha1
       kind: EndpointPickerConfig
       plugins:
-        - type: tokenizer
```
vMaroon, line 71: Keep the plugin entry because the next release will fully utilize, in this one, removing the entry in plugins is sufficient.

vMaroon (approved):

PR #1324, opened 2026-05-01
[Guides] Bump WideEP

robertgshaw2-redhat:
SUMMARY
- [x] migrate guides to vllm 0.19.1 images
- [x] update benchmarks
- [x] validate on coreweave benchmarking
- [x] validate cks ci (https://github.com/llm-d/llm-d/actions/runs/25230809312/job/73985557568) - passing
- [ ] validate gke ci 

@liu-cong and I discovered the v0.19 images are not working on wideep for gke. so we will use the llm-d images for this release and resolve the issue later.


robertgshaw2-redhat:
throughput numbers look reasonable

image

guides/recipes/scheduler/base.values.yaml
```diff
@@ -10,6 +10,6 @@ inferenceExtension:
     # Image override for helm chart version 1.5 and above
     registry: ghcr.io
     repository: llm-d/llm-d-inference-scheduler
-    tag: v0.8.0
```
robertgshaw2-redhat, line 13: this is failing due to `prepareDataPlugins`, need to fix before merge

guides/wide-ep-lws/modelserver/gpu/vllm/base/decode.yaml
```diff
@@ -92,11 +92,15 @@ spec:
...
                   --port 8200 \
+                  --gpu-memory-utilization 0.9 \
                   --trust-remote-code \
-                  --disable-uvicorn-access-log \
+                  --api-server-count 1 \
+                  --disable-access-log-for-endpoints=/health,/metrics \
                   --data-parallel-hybrid-lb \
                   --enable-expert-parallel \
                   --tensor-parallel-size $TP_SIZE \
+                  --max-num-batched-tokens ${MAX_TOKENS} \
```
chcost, line 106: `MAX_TOKENS` is used for both `--max-num-batched-tokens` and `--max-num-seqs`, which are semantically different (tokens vs sequences). They happen to be equal at 256 but this is a footgun for anyone tuning. Consider splitting into two env vars or adding a comment explaining they are intentionally coupled.

guides/wide-ep-lws/modelserver/gpu/vllm/base/prefill.yaml
```diff
@@ -78,16 +80,15 @@ spec:
...
-                  --async-scheduling \
                   --enable-dbo \
                   --dbo-prefill-token-threshold 32 \
                   --enable-eplb \
                   --eplb-config '{"window_size":"1000",
                                   "step_interval":"3000",
                                   "num_redundant_experts":"32",
                                   "log_balancedness":"False"}' \
+                  --moe-backend deep_gemm \
                   --all2all-backend deepep_high_throughput \
```
chcost, line 95: Now that `--gpu-memory-utilization 0.75` was removed, `--all2all-backend deepep_high_throughput` is the last real flag but still has a trailing `\`. The shell handles it fine (the `#` comment on the next line is still treated as a comment after line continuation), but it's unnecessary and a bit misleading — looks like there should be more flags after it. Consider removing the trailing `\`.

.github/workflows/nightly-e2e-wide-ep-lws-cks.yaml
```diff
@@ -280,7 +68,7 @@ jobs:
           --version v1.5.0
 
         echo "Deployment complete"
-      image_override: 'ghcr.io/llm-d/llm-d-cuda-dev:latest'
+      # Wide-EP-specific validation: check prefill + decode pods
+      e2e_validate_args: '-v'
       allow_gpu_preemption: true
       skip_cleanup: ${{ github.event.inputs.skip_cleanup == 'true' }}
```
chcost, line 82: `secrets: inherit` was removed below this line. If the reusable workflow needs secrets (HF tokens, cloud credentials), this will break CI. Was this intentional?

chcost (commented):
A few items — see inline comments.

chcost (commented):
a few issues below, including a CI failure and what looks like an image mismatch for the coreweave overlay

.github/workflows/nightly-e2e-wide-ep-lws-cks.yaml
```diff
@@ -64,223 +49,26 @@ jobs:
...
+        # 2. Deploy Standalone Inference Scheduler via helm
+        echo "Deploying Standalone Inference Scheduler..."
+        helm install wide-ep-lws \
+          oci://registry.k8s.io/gateway-api-inference-extension/charts/standalone \
+          -f guides/recipes/scheduler/base.values.yaml \
+          -f guides/recipes/scheduler/pd.values.yaml \
+          -f guides/wide-ep-lws/scheduler/wide-ep-lws.values.yaml \
+          -n "$NAMESPACE" --version v1.4.0
 
         echo "Deployment complete"
```
chcost, line 78: `secrets: inherit` was removed from the end of this job. The reusable workflow likely needs secrets for cluster access, image pulls, etc. Was this intentional? If the reusable workflow requires any secrets, this will silently fail.

guides/wide-ep-lws/modelserver/gpu/vllm/base/prefill.yaml
```diff
@@ -31,7 +31,7 @@ spec:
             emptyDir: {}
         containers:
           - name: vllm
-            image: ghcr.io/llm-d/llm-d-cuda:v0.7.0
+            image: ghcr.io/llm-d/llm-d-cuda-gb200:v0.7.0
```
chcost: The base prefill image changed from `ghcr.io/llm-d/llm-d-cuda:v0.7.0` to `ghcr.io/llm-d/llm-d-cuda-gb200:v0.7.0`, while decode stays as `ghcr.io/llm-d/llm-d-cuda:v0.7.0-rc.2`.

The coreweave kustomization overrides image name `ghcr.io/llm-d/llm-d-cuda` → `docker.io/vllm/vllm-openai:v0.19.1`, but that pattern won't match `ghcr.io/llm-d/llm-d-cuda-gb200`. So on coreweave, decode gets the upstream image but prefill stays on the llm-d gb200 image. Is that intentional? If both roles should use the same image on coreweave, the kustomization needs a second image entry for `llm-d-cuda-gb200`.

guides/wide-ep-lws/modelserver/gpu/vllm/base/decode.yaml
```diff
@@ -61,7 +61,7 @@ spec:
             emptyDir: {}
         containers:
           - name: vllm
-            image: ghcr.io/llm-d/llm-d-cuda:v0.7.0
+            image: ghcr.io/llm-d/llm-d-cuda:v0.7.0-rc.2
```
chcost: Minor: decode base uses `v0.7.0-rc.2` while prefill base uses `v0.7.0` (GA). Is this intentional? Release candidates and GA images can have different behavior. Usually both roles should be on the same version.

guides/wide-ep-lws/modelserver/gpu/vllm/base/prefill.yaml
```diff
@@ -78,16 +80,15 @@ spec:
...
                                           "kv_load_failure_policy":"fail"}' \
-                  --async-scheduling \
                   --enable-dbo \
                   --dbo-prefill-token-threshold 32 \
                   --enable-eplb \
                   --eplb-config '{"window_size":"1000",
                                   "step_interval":"3000",
                                   "num_redundant_experts":"32",
                                   "log_balancedness":"False"}' \
+                  --moe-backend deep_gemm \
```
chcost, line 94: The `\` on `--all2all-backend deepep_high_throughput \` makes this a line continuation into the comment below (`# To enable tracing`). It works because bash treats `#` as a comment even after line continuation, but it's fragile — if someone removes or edits the comment line, the `\` would join to whatever comes next.

Since `--gpu-memory-utilization` moved earlier and this is now the last real argument, the trailing `\` should be removed.

guides/wide-ep-lws/modelserver/gpu/vllm/base/decode.yaml
```diff
@@ -92,11 +92,15 @@ spec:
...
+                  --gpu-memory-utilization 0.9 \
                   --trust-remote-code \
-                  --disable-uvicorn-access-log \
+                  --api-server-count 1 \
+                  --disable-access-log-for-endpoints=/health,/metrics \
                   --data-parallel-hybrid-lb \
                   --enable-expert-parallel \
                   --tensor-parallel-size $TP_SIZE \
+                  --max-num-batched-tokens ${MAX_TOKENS} \
+                  --max-num-seqs ${MAX_TOKENS} \
```
chcost, line 107: +1 to Copilot's point: `MAX_TOKENS` drives both `--max-num-batched-tokens` (token budget per iteration) and `--max-num-seqs` (max concurrent sequences). These are semantically different limits. At `MAX_TOKENS=256`, you're capping both to 256 — intentionally tying them to the same value. If that's a deliberate design choice for decode (each sequence gets ~1 token per iteration), a comment explaining the rationale would help. Otherwise, separate env vars would prevent accidental mis-tuning.

guides/wide-ep-lws/README.md
```diff
@@ -108,7 +108,7 @@ helm install ${GUIDE_NAME} \
 Apply the Kustomize overlays for your specific backend:
 
 ```bash
-export INFRA_PROVIDER=gke # options: gke (H200), gke-a4 (B200), coreweave
+export INFRA_PROVIDER=gke # options: gke (H200), gke-a4 (B200), coreweave, dgx-cloud-gb200
```
chcost: `dgx-cloud-gb200` is added as a new option here but there's no corresponding `dgx-cloud-gb200/kustomization.yaml` in this diff. Is it being added in a separate PR, or is it already on main?

maugustosilva:
Will merge "as is", and then open a shared PR where we (@maugustosilva and @Gregory-Pereira) can work together.

maugustosilva (approved):

PR #1326, opened 2026-05-01
[CI] build-image.yml use wrong github.event_name

weizhoublue:
.github/workflows/ci-pr-test.yaml  states that all PRs use pull_request_target, and the same file invokes build-image.yml under that model. Within build-image.yml itself, the equivalent latest tag guards for other image jobs already use github.event_name != 'pull_request_target'. 
 The GB200 condition was the only remaining outlier, so this change simply aligns it with both the external trigger model and the existing in-file pattern.


maugustosilva (dismissed):

weizhoublue:
fixed the main conflict and update the pr

PR #1333, opened 2026-05-01
[Docs] Polish Getting Started and Glossary

ahg-g:


robertgshaw2-redhat:
we need to add some of the flow control / batch gateway stuff to the introduction

robertgshaw2-redhat:
do you think we should do this now, or in a follow up

ahg-g:
I thought I added batch! let me do both in the PR.

ahg-g:
> do you think we should do this now, or in a follow up

done, please take a look

ahg-g:
actually, wait, I need to stress that those are all composable features and one doesn't need to subscribe to the full stack to get one of them.

ahg-g:
now it is done.

robertgshaw2-redhat (approved):

robertgshaw2-redhat:
great work

PR #1356, opened 2026-05-01
Fix AWS image

diegocastanibm:
## Summary

- Simplify `release-aws-llm-d` to push directly with tags, removing the broken digest-passing mechanism
- Remove `release-aws-success-check` and `release-aws-manifest` jobs
- Update `verify-release-images` dependency from `release-aws-manifest` to `release-aws-llm-d`

## Problem

The `release-aws-manifest` job was silently skipped during the `v0.7.0-rc.1` release, resulting in the AWS image not being tagged.

**Root cause:** The digest-passing mechanism between jobs broke after `actions/upload-artifact` was bumped from v6 to v7 (#1149) and `actions/download-artifact` from v4 to v8 (#1219). With `upload-artifact@v7` (which defaults to `archive: true`) and `download-artifact@v8`, the artifact is downloaded successfully but the files are not extracted into the expected subdirectory structure. The `release-aws-success-check` job's reorganize step (`cp -v "$artifact_dir"/* ...`) finds nothing to copy, reports `any_succeeded=false`, and `release-aws-manifest` is skipped entirely.

This went unnoticed for CUDA because with 4 matrix variants the extraction behaves differently — AWS only has a single artifact (`digests-aws-rhel-amd64`), which triggers the incompatibility.

## Fix

Simplify `release-aws-llm-d` to push directly with tags (like `release-cpu-llm-d`, `release-rocm-llm-d`, and `release-hpu-llm-d` already do), removing the unnecessary `release-aws-success-check` and `release-aws-manifest` jobs. The multi-step digest→manifest pattern only benefits multi-arch builds (like CUDA with 4 platform variants). AWS is single-platform (`linux/amd64`, RHEL-only) so the extra indirection adds complexity with no benefit.

## Test plan

- [ ] Push a new RC tag and verify the `release-aws-llm-d` job pushes the image with the correct tag
- [ ] Verify `ghcr.io/llm-d/llm-d-aws:<tag>` is accessible after the workflow completes
- [ ] Verify `verify-release-images` job passes for the AWS image

### DIFF
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -374,23 +374,7 @@ jobs:
           fi
 
   release-aws-llm-d:
-    strategy:
-      fail-fast: false
-      # AWS/EFA release matrix: RHEL-only, amd64-only (for now)
-      # Ubuntu and ARM64 variants commented out - can be enabled if needed
-      matrix:
-        include:
-          - platform: linux/amd64
-            runner: vllm-runner
-            os: rhel
-            base_image_suffix: ubi9
-            image_suffix: ""
-            # - platform: linux/arm64
-            #   runner: vllm-runner-arm
-            #   os: rhel
-            #   base_image_suffix: ubi9
-            #   image_suffix: ""
-    runs-on: ${{ matrix.runner }}
+    runs-on: vllm-runner
     steps:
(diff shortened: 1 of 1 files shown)


Gregory-Pereira (approved):
/lgtm

PR #1362, opened 2026-05-02
[Build] Use Proper Wheel Variant

robertgshaw2-redhat:
SUMMARY:
- in 0.7.0-rc2 vllm startup fails, looking for cu12 runtime
- the images are cu13 now
- we are installing the prebuild vllm binary, which defaults to cu129 for 0.19.1
- this pr updates the precompiled wheel variant to match


Gregory-Pereira (approved):
/lgtm

PR #1367, opened 2026-05-03
[Build] DeepGEMM Version

robertgshaw2-redhat:
SUMMARY:
* deepgemm version diverges from vllm (we are well behind)
* additionally, starting in 0.20, vLLM's wheel has deepgemm built in


docker/Dockerfile.cuda
```diff
@@ -61,7 +61,7 @@ ARG DEEPEP_VERSION="38d21b7f9bb6f3b102b1819d09439686eaa87ce8"
 ARG DEEPEP_GB200_REPO="https://github.com/tlrmchlsmth/DeepEP"
 ARG DEEPEP_GB200_VERSION="sgl-gb200-blog-pt2"
 ARG DEEPGEMM_REPO="https://github.com/deepseek-ai/DeepGEMM"
-ARG DEEPGEMM_VERSION="v2.1.1.post3"
+ARG DEEPGEMM_VERSION="477618cd51baffca09c4b0b87e97c03fe827ef03"
```
ahg-g, line 64: can we add a comment explaining why this is pointing to a specific hash instead of a released version

Gregory-Pereira: This is what vLLM pins to, so basically just that its a known version

Gregory-Pereira (approved):
/lgtm

Pin to what vLLM uses for now

PR #1378, opened 2026-05-04
add back hpu optimized baseline

Gregory-Pereira:
cc @praveingk @llm-d/release-crew


Gregory-Pereira:
Forcing this through to unblock builds

diegocastanibm (commented):
yes, we missed this one.

PR #1379, opened 2026-05-04
Fixes for missing tags which break docusaurus website builds

petecheslock:
Some of the new docs were missing `<details>` tags which looks fine on github but breaks the build on docusaurus, fixed those issues and added a pre-commit test to check for these in the future.


petecheslock:
This got resolved in the other PR so to save complexity i'm going to close this unmerged and we can address this after we get the docs in a solid place.

PR #1386, opened 2026-05-04
[bugfix] Fix DeepGEMM JIT in llm-d builds

robertgshaw2-redhat:
SUMMARY:
* deepgemm need cuobjdump in the runtime layer of the image
* this is not installed
* nvcc works, but then loading the cubin fails for this reason


Gregory-Pereira (approved):
/lgtm

PR #1389, opened 2026-05-05
[Build] bump nvshmem version to one that works with RoCE

robertgshaw2-redhat:
SUMMARY:
* After https://github.com/llm-d/llm-d/pull/1134, we no longer build NVSHMEM from source and instead install via pypi from nvidia-nvshmem-cu13
* We are ending up with nvsheme 3.4.5

```
Name: nvidia-nvshmem-cu13
Version: 3.4.5
Location: /opt/vllm/lib/python3.12/site-packages
Requires:
Required-by: torch
```

* this version does not support RoCE (https://llm-d.ai/docs/guide/InfraProviders/gke#nvshmem-reports-unable-to-create-ah-on-initialization-for-deepep), so NVSHMEM is failing to get AH. PReviously we had a patch for this (https://github.com/llm-d/llm-d/pull/407/changes), but since we no longer build NVSHMEM, we cannot have this
* update to a newer version


tlrmchlsmth (dismissed):

robertgshaw2-redhat:
this didnt end up installing the right version, because torch overrode it

PR #1400, opened 2026-05-05
fix: precommit lints

zdtsw:
ref: https://github.com/llm-d/llm-d/actions/runs/25364825218/job/74372910926?pr=1399 from https://github.com/llm-d/llm-d/pull/1399


zdtsw:
cc @petecheslock

PR #1415, opened 2026-05-05
Make active-active HA the default for precise-prefix-cache-aware

vMaroon:
Make active-active HA the default for precise-prefix-cache-aware

Pod-discovery + 2 scheduler replicas (active-active) become the default install path; central ZMQ mode is removed (still supported though). Single-replica setups are supported via `--set inferenceExtension.replicas=1`.

Over the past week:
1. This deployment was ran in an IBM internal production LLM environment continuously for days
2. Several benchmarks were ran


guides/precise-prefix-cache-aware/modelserver/cpu/vllm/patch-vllm.yaml
```diff
@@ -25,6 +25,12 @@ spec:
             # Uncomment for distributed tracing
             # - "--otlp-traces-endpoint=http://otel-collector:4317"
             # - "--collect-detailed-traces=all"
+          ports:
```
liu-cong, line 28: Consider adding this to a `base` that all accelerator variants can derive from

guides/precise-prefix-cache-aware/scheduler/precise-prefix-cache-aware.values.yaml
```diff
@@ -73,13 +49,15 @@ inferenceExtension:
...
+        - type: endpoint-notification-source
+        - type: metrics-data-source
+        - type: core-metrics-extractor
         - type: single-profile-handler
         - type: decode-filter
         - type: precise-prefix-cache-scorer
           parameters:
             tokenProcessorConfig:
               blockSize: 64           # must match vLLM --block-size
-            speculativeIndexing: true
```
liu-cong, line 82: why removing disabling this?

Gregory-Pereira:
successful run: https://github.com/llm-d/llm-d/actions/runs/25405438757/job/74515128463

Gregory-Pereira (dismissed):
/lgtm but I think Cong had some comments

liu-cong (dismissed):
my comments are not blocking

guides/precise-prefix-cache-aware/scheduler/precise-prefix-cache-aware.values.yaml
vMaroon: Lowkey forgot we introduced this in v0.6.0, recovered. With external tokenization it currently causes double tokenization, which kind of beats the point, but we can live with that for a release.

guides/precise-prefix-cache-aware/scheduler/patches/uds-tokenizer/patch-uds-sidecar.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
         - name: tokenizer-uds
-          image: ghcr.io/llm-d/llm-d-uds-tokenizer:v0.7.1
+          image: ghcr.io/llm-d/llm-d-uds-tokenizer:vllm-v0.19.1
```
zdtsw, line 17: we do not want use release tag v0.8.0? even they are from the same source

vMaroon: We can do v0.8.0, although this tag is more informative. Is there a preference?

Gregory-Pereira (approved):
/lgtm

PR #1416, opened 2026-05-05
Fix InferencePool diagram reference

chcost:
The InferencePool documentation referenced `gateway-design.svg` but should reference `inferencepool.svg` to match the actual diagram for this component.

Both SVG files exist in `docs/assets/`, but the markdown had the wrong filename.


ahg-g:
Thanks Carlos, this is intentional, we can remove inferencepool.svg; it is outdated and we can just reuse gateway-design.svg in place it

ahg-g:
We can remove inferencepool.svg for clarity

PR #1426, opened 2026-05-06
attempt fix(ci): align WVA nightlies with nightly namespace and CKS Prom settings

mamy-CS:
Note: I dont have access to the cks cluster, hence having difficulties patching the fix. This is an almost blind attempt.


ahg-g:
is this PR still relevant?

PR #1431, opened 2026-05-07
Unified drawings color theme, sizes and placement

ahg-g:


chcost:
looks good to me. Thanks @ahg-g. The light/dark switch was aspirational (assuming some automation which would invert SVG colors automatically at doc sync). But it didn't work well, and this was left over. Agree to keep it single color theme now is better.

PR #1434, opened 2026-05-07
Fix nightly PD OCP workflow for kustomize deployment

### LINKED ISSUE #1433: [CI/CD] Add Nightly P/D test for OCP
Currently it is failing because it was not updated to use Kustomize.

sudoalok:
This updates the OCP nightly PD workflow after the guide moved to kustomize.

The old workflow was still using the helmfile/ms-pd flow, which no longer matches the current guide setup.

Changes:
- switched to the newer reusable OpenShift workflow
- deploy PD manifests with kubectl apply -k
- install the scheduler from custom_deploy_script
- update the slim test transforms to modify:
  - patch-decode.yaml
  - patch-prefill.yaml

Checked locally by:
- validating the workflow YAML
- rendering the kustomize manifests successfully
- verifying the yq patch updates

Fixes #1433

### DIFF
--- a/.github/workflows/nightly-e2e-pd-disaggregation-ocp.yaml
+++ b/.github/workflows/nightly-e2e-pd-disaggregation-ocp.yaml
@@ -1,8 +1,5 @@
 name: Nightly - PD Disaggregation E2E (OpenShift)
 
-# Nightly regression test for the pd-disaggregation guide on OpenShift.
-# Deploys the guide and validates with e2e-validate.sh.
-
 on:
   schedule:
     - cron: '30 0 * * *'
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-e2e-pd-disaggregation-ocp.yaml
```diff
@@ -27,53 +17,79 @@ permissions:
...
-        yq e 'del(.prefill.containers[0].resources.limits."rdma/ib")' -i ms-pd/values.yaml
-        yq e 'del(.prefill.containers[0].resources.requests."rdma/ib")' -i ms-pd/values.yaml
-        yq e '.decode.containers[0].resources.limits["nvidia.com/gpu"] = "1"' -i ms-pd/values.yaml
-        yq e '.decode.containers[0].resources.requests["nvidia.com/gpu"] = "1"' -i ms-pd/values.yaml
-        yq e '.decode.parallelism.tensor = 1' -i ms-pd/values.yaml
-        yq e '.prefill.replicas = 1' -i ms-pd/values.yaml
-        yq e '.decode.volumes[1].emptyDir.sizeLimit = "2Gi"' -i ms-pd/values.yaml
-        yq e '.prefill.volumes[1].emptyDir.sizeLimit = "2Gi"' -i ms-pd/values.yaml
-        echo "Slim transform applied — model: Qwen3-0.6B, 1 GPU decode, 1 prefill replica"
+
```
robertgshaw2-redhat: please remove this, it should run gpt-oss

sudoalok:
Removed the slim-model overrides so the workflow now deploys the default gpt-oss PD configuration via kustomize.

robertgshaw2-redhat:
we aren't ready to add this, because the OCP example uses RoCEv2 and we do not have proper performance yet for exposing the NIC up to the pod in the P/D setup

sudoalok:
Got it, makes sense then.

the workflow side is updated for the current kustomize PD flow, but yeah if the OCP P/D setup itself still has perf issues around RoCEv2/NIC exposure then probably better not to enable the nightly yet.

i’ll leave this here for now and can rework/rebase it later once that setup is in a better state.

robertgshaw2-redhat:
> Got it, makes sense then.
> 
> the workflow side is updated for the current kustomize PD flow, but yeah if the OCP P/D setup itself still has perf issues around RoCEv2/NIC exposure then probably better not to enable the nightly yet.
> 
> i’ll leave this here for now and can rework/rebase it later once that setup is in a better state.

Thank you!

PR #1444, opened 2026-05-07
Add a nightly P/D on GKE badge to the main matrix

maugustosilva:
Attempted fix for tiered-prefix-cache on OCP


guides/README.md
```diff
@@ -46,7 +46,7 @@ Auto-updated from GitHub Actions:
 |-------|-----|-----|-----|
```
ahg-g, line 46: can we pls add the predicted latency ones as well, we have coverage for gke and cks

maugustosilva: Absolutely!

Gregory-Pereira (approved):
/lgtm

PR #1448, opened 2026-05-08
[Bugfix] Fix WideEP Build + IMA

robertgshaw2-redhat:
SUMMARY:
* wideep is failing because nvshmem 3.4.5 does not work with RoCE (failed to create ah)
* we cannot simply update to 3.5.19 (which has the ah failure fix), because this is causing an IMA in deepep HT
* so, we need to bring back the nvshmem patch for 3.4.5 RoCE --> revert to the "build nvshmem from source" method
* additionally, after this fix, there is a bug in vllm ep scatter. we cherry pick matej's commit:https://github.com/vllm-project/vllm/pull/39213)

test image (only for b200):
- `ghcr.io/llm-d/llm-d-cuda-dev:sha-81f2333`


Gregory-Pereira (dismissed):
/lgtm

Gregory-Pereira (approved):
/lgtm

PR #1455, opened 2026-05-08
upgrade to CNCF runners

Gregory-Pereira:
Should enable our jobs to consume new CNCF runners


Gregory-Pereira:
Will fix if it breaks something but forcing this through to unblock us

PR #1467, opened 2026-05-10
Various fixes to the guides

ahg-g:


guides/recipes/scheduler/base.values.yaml
```diff
@@ -11,5 +11,54 @@ inferenceExtension:
...
+  sidecar:
+    # Resource requests and limits for the Envoy sidecar.
+    resources:
+      requests:
+        cpu: "4"
+        memory: 8Gi
+      limits:
+        memory: 16Gi
+      
+  tracing:
```
ahg-g, line 53: this and the below settings are applying default values that exist in the infpool chart but not in the standalone one.

vMaroon (approved):

PR #1476, opened 2026-05-11
removed redundant nightly tests badges

ahg-g:
Each guide already have their test status, this table is redundant and can go out of sync.


maugustosilva:
I propose we keep this table as an aggregator for all the guides, at least for the time being. There is value, especially during the release process - which is supposed to happen with a monthly periodicity - to have a (admittedly, poor man's) "single glass pane" view of the stability and testing coverage.

ahg-g:
I strongly prefer to not have redundant info on the site, they go out of sync quickly, it is already out of sync now.

ahg-g:
Unless there is an automated process to add them to the table.

maugustosilva:
I was about to suggest this. How about we keep it, and add a pre-commit hook to ensure the information is synchronized?

ahg-g:
I recommend to separate user facing docs from the ones for our operations. We can have that, but in some release/README.md file/pkg.

maugustosilva (approved):

PR #1484, opened 2026-05-12
Fix: PD disaggregation CKS nightly values

weizhoublue:
When the PD disaggregation CKS nightly job runs, the scheduler deployment attempts to load a missing Helm values input. This causes deployment to fail before validation starts, blocking that nightly run. The expected behavior is that the job deploys using only existing base and PD-specific scheduler configuration.

Root cause: 
The CKS nightly path still referenced an obsolete shared PD scheduler values input after the PD-specific scheduler configuration had already been moved into the guide-specific configuration.


clubanderson:
### Nightly E2E Result: `pd-disaggregation` (CKS)

| | |
|---|---|
| Guide | `pd-disaggregation` |
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/25735443445) |
| Ref | `b5e7036` |

maugustosilva (approved):

PR #1503, opened 2026-05-13
ci: add xpu-sglang image build job

xiaojun-zhang:
## Summary

Adds a new `xpu-sglang-build-llm-d` job to `.github/workflows/build-image.yaml` to build sglang's XPU
Dockerfile for llm-d. The docker image will be pushed as "llm-d-xpu-sglang". 

Later when SGLang upstream has a docker image for XPU, we will switch to use upstream image instead of building within llm-d.

## Dependent PR
There is a follow-up PR #1508 for Optimized Baseline guide that depends on this PR.

## Test plan

- [x] `workflow_dispatch` against a fork + self-hosted runner, image lands in the expected registry path: https://github.com/users/xiaojun-zhang/packages/container/package/llm-d-xpu-sglang
- [x] Trivy step runs and uploads SARIF without colliding with the vLLM-XPU job's upload
- [ ] Nightly / path-filter trigger on merge — covered by `paths-filter` on `docker/common-versions`

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -17,13 +17,16 @@ on:
       xpu:
         description: 'Whether XPU image was built'
         value: ${{ jobs.detect-changes.outputs.xpu }}
+      sglang-xpu:
+        description: 'Whether sglang-XPU image was built'
+        value: ${{ jobs.detect-changes.outputs.sglang-xpu }}
       hpu:
         description: 'Whether HPU image was built'
         value: ${{ jobs.detect-changes.outputs.hpu }}
   workflow_dispatch:
     inputs:
       platforms:
-        description: "Platforms to build (comma-separated: cuda,aws,gb200,xpu,cpu,hpu)"
+        description: "Platforms to build (comma-separated: cuda,aws,gb200,xpu,sglang-xpu,cpu,hpu)"
         required: false
         default: "cuda"
         type: string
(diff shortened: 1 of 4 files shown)


xiaojun-zhang:
@liu-cong  @diegocastanibm does this approach make sense to you?

diegocastanibm (commented):
I'm missing the intention of this PR. If your intention is to have `sglang-xpu` in the release process, then you need to add a release job to `ci-release.yaml` (mirroring `release-xpu-llm-d`) and add it also to the verification array. If it's intentionally nightly-only, please document this explicitly in the PR.

.github/workflows/build-image.yaml
```diff
@@ -1376,6 +1393,110 @@ jobs:
...
+          echo "tag=${TAG}" >> $GITHUB_OUTPUT
+          PR_TAG="${{ github.event.pull_request.number }}"
+          if [ -n "${PR_TAG}" ]; then
+            echo "pr_tag=pr-${PR_TAG}" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Calculate Cache Bust Timestamp
+        id: cache_buster_timestamp
+        run: |
+          echo "cache_buster_timestamp=$(date)" >> $GITHUB_OUTPUT
```
diegocastanibm: Suggest date +%s

xiaojun-zhang: resolved.

xiaojun-zhang:
> I'm missing the intention of this PR. If your intention is to have `sglang-xpu` in the release process, then you need to add a release job to `ci-release.yaml` (mirroring `release-xpu-llm-d`) and add it also to the verification array. If it's intentionally nightly-only, please document this explicitly in the PR.

@diegocastanibm , thanks for pointing out! Yes, the intention is to have sglang-xpu in the release process. Added the release job and verification array. Also tested on a self hosted github runner.

PR #1510, opened 2026-05-15
Add llm-d on a kind cluster (Mac) getting-started series

arun-gupta:
## Summary

- Adds a six-part tutorial series under `docs/getting-started/kind-mac/` for running the full llm-d stack locally on macOS using a kind cluster, no GPU required
- All content is normative in the repo (no external gist links)
- Series covers: vLLM setup → llm-d Gateway + EPP → load distribution → model aliasing → fault tolerance → EPP observability
- Linked from the getting-started README alongside the existing Quickstart guide

## What's included

| File | Content |
|---|---|
| `kind-mac/README.md` | Series overview, architecture diagram, guide table, prerequisites, cleanup guidance |
| `kind-mac/01-vllm.md` | Guide 1: Deploy vLLM in CPU mode on a local kind cluster |
| `kind-mac/02-llmd.md` | Guide 2: Add the llm-d Gateway and EPP scheduling layer |
| `kind-mac/03-load-distribution.md` | Guide 3: Scale to 3 replicas, observe EPP load spreading |
| `kind-mac/04-model-aliasing.md` | Guide 4: Model aliasing with `InferenceModelRewrite` |
| `kind-mac/05-fault-tolerance.md` | Guide 5: Pod failure and automatic EPP recovery |
| `kind-mac/06-observability.md` | Guide 6: EPP Prometheus metrics and pool size changes |

## Test plan

- [ ] Follow guides 1–6 in order on macOS with Docker Desktop and verify each step produces the expected output
- [ ] Verify all internal links between guides resolve correctly
- [ ] Verify the architecture diagrams render in GitHub markdown


robertgshaw2-redhat:
Thank you for your contribution! I have a couple thoughts:

- Infra provider specific content should be under `docs/infrastructure-providers`, which is where infra-specific knowledge is places

- This content in this is interesting and useful, but it reads more like an end-to-end tutorial / blog vs documentation of how to run llm-d with kind. The only "kind-specific" content is --- deploying cpu and setting up a kind cluster. The content is useful though. But it should be slotted into appropriate feature specific things. For instance, we could have a section in docs called operations to cover things like fault tolerance. We already have a section on observability (which needs improving). I think that we should integrate these pieces into the existing documentation structure rather than having another end-to-end explanation of each piece

Also, its Endpoint Picker not Endpoint Policy Processor

arun-gupta:
Thanks for the feedback, and good call on the location and terminology - both fixed in the latest push.

On the tutorial structure: I see real value in keeping an end-to-end format for this particular guide. A newcomer to llm-d who wants to see the full system working (Gateway, EPP, routing, observability) benefits from a single narrative that builds up state step by step. Individual feature docs answer "how does X work?" but don't answer "how do I get from zero to a running llm-d stack?" That on-ramp is what this series is trying to provide.

That said, I'm open to integrating more tightly with the existing docs structure rather than having it be a standalone silo. A few directions that could work - would appreciate your thoughts on which fits best:

1. **Cross-link from feature docs** - keep the tutorial as-is under `infra-providers/kind/`, but add "Try this on a local kind cluster" links from the relevant sections in architecture/observability/etc. docs
2. **Extract the kind-specific parts** - slim the series down to just what's truly kind-specific (cluster setup, CPU mode), and point to feature docs for everything else (fault tolerance, observability, model aliasing)
3. **Something else** - if there's a specific structure you have in mind for how these pieces should slot in, I'm happy to reshape around that

robertgshaw2-redhat:
> Thanks for the feedback, and good call on the location and terminology - both fixed in the latest push.
> 
> On the tutorial structure: I see real value in keeping an end-to-end format for this particular guide. A newcomer to llm-d who wants to see the full system working (Gateway, EPP, routing, observability) benefits from a single narrative that builds up state step by step. Individual feature docs answer "how does X work?" but don't answer "how do I get from zero to a running llm-d stack?" That on-ramp is what this series is trying to provide.
> 
> That said, I'm open to integrating more tightly with the existing docs structure rather than having it be a standalone silo. A few directions that could work - would appreciate your thoughts on which fits best:
> 
> 1. **Cross-link from feature docs** - keep the tutorial as-is under `infra-providers/kind/`, but add "Try this on a local kind cluster" links from the relevant sections in architecture/observability/etc. docs
> 2. **Extract the kind-specific parts** - slim the series down to just what's truly kind-specific (cluster setup, CPU mode), and point to feature docs for everything else (fault tolerance, observability, model aliasing)
> 3. **Something else** - if there's a specific structure you have in mind for how these pieces should slot in, I'm happy to reshape around that

We don't have the capacity or desire to maintain a per-provider end to end tutorial in the project since any changes we make to the project will have to be propagated to each per-provider tutorial.

If you want to include some information about `kind`, it should be limited to how to setup `kind` such that its ready to deploy llm-d.

The rest of the content that you have is very very good and very valuable. Currently, our section on "operations" is weak. It would be great if you could include the content you have here inside this section (since most of the tutorial is actually about showing off certain operational flows (e.g. what happens if a replica dies) https://llm-d.ai/docs/resources/operations/router

PR #1513, opened 2026-05-15
fix(xpu): use VLLM_XPU_COMMIT_SHA as single source of truth

yuanwu2017:


yuanwu2017:
@Gregory-Pereira Please help to review.

docker/common-versions
```diff
@@ -22,7 +22,10 @@ VLLM_PRECOMPILED_WHEEL_COMMIT=b1388b1fbf5aaef47937fabe98931211684666a6 # maps to
 
 
 
-# Used by XPU, commit from vllm, maps to v0.19.1 tag
+# Used by XPU. The Makefile's xpu-prepare target clones VLLM_XPU_REPO and
+# checks out VLLM_XPU_COMMIT_SHA.
+# Currently maps to vllm-project/vllm v0.19.1.
+VLLM_XPU_REPO=https://github.com/vllm-project/vllm.git
```
diegocastanibm, line 28: Why you have to point to a different `vLLM` repo? Is there a reason for not using `VLLM_REPO=https://github.com/neuralmagic/vllm.git`

docker/common-versions
```diff
@@ -22,7 +22,10 @@ VLLM_PRECOMPILED_WHEEL_COMMIT=b1388b1fbf5aaef47937fabe98931211684666a6 # maps to
...
 VLLM_XPU_COMMIT_SHA=b1388b1fbf5aaef47937fabe98931211684666a6
```
diegocastanibm, line 29: This SHA is exactly the same one that we have in `VLLM_PRECOMPILED_WHEEL_COMMIT`. Two separate SHAs for the same concept can cause a silent drift

docker/common-versions
yuanwu2017: I don't know why use the https://github.com/neuralmagic/vllm.git. The official vllm repo should be https://github.com/vllm-project/vllm. Intel submits the XPU PRs into the official vllm repo. I'm not sure if these two repos are completely synchronized. If you are sure, then I can use neuralmagic's.

docker/common-versions
yuanwu2017: Intel hasn't official docker image release from https://github.com/vllm-project/vllm, and the PVC cards are not support. Currently the llmd CICD server use the PVC cards. Not every release tag will work. So sometimes XPU uses different variable name. We are preparing the new servers with BMG for llmd CICD. The official image release may occur in Q3.

docker/common-versions
diegocastanibm: I'm not sure either. @robertgshaw2-redhat should know it. 
I just think that having two separate `vLLM` repos on the same project is not something I like to see, but it could be a hidden reason for that that I do not know

PR #1264, opened 2026-04-25
Add AWS EFS backend guide for tiered prefix cache storage

sudoalok:
## Summary

Adds documentation and configuration for using AWS EFS as a shared storage backend for llm-d tiered prefix cache.

This follows the existing GCP Lustre guide structure and provides:

- EFS CSI driver setup guidance
- StorageClass configuration for RWX access
- PVC example for shared KV cache storage

## Changes

- Added `aws/README.md` with provisioning and setup instructions
- Added `aws/storage_class.yaml` for EFS CSI-based dynamic provisioning

## Notes

- This mirrors the structure and style of the existing Lustre backend guide
- Uses RWX PVC to enable shared KV cache across multiple vLLM pods

## Validation

- Documentation and configuration prepared based on AWS EFS CSI best practices
- Further validation with multi-pod KV cache reuse and benchmarking can be added in follow-up

Issue #1252, opened 2026-04-24
Validate llm-d with shared storage across multiple vendors


kfirtoledo:
## Summary

Validate and document llm-d running with shared storage across multiple vendors. The [tiered prefix cache storage guide](https://github.com/llm-d/llm-d/blob/main/guides/tiered-prefix-cache/storage/README.md) shows how to connect a storage connector (llm-d FS / LMCache) to a shared POSIX volume, but today only covers **GCP Lustre** as a concrete backend. We want to guide how llm-d works on other POSIX-backed storage systems, surface any implementation gaps, and extend the guide with a path per backend. Beyond functional validation, each backend should be configured for high-bandwidth access so the storage can sustain the throughput required for KV-cache reuse 

## Backends

- [ ] GCP Lustre - [existing guide](https://github.com/llm-d/llm-d/blob/main/guides/tiered-prefix-cache/storage/manifests/backends/lustre/README.md)
- [ ] IBM Storage Scale
- [ ] Azure
- [ ] AWS 
- [ ] VAST
- [ ] CephFS

## Scope per backend

- Provisioning (StorageClass / CSI) + RWX PVC.
- High-performance configuration for storage to sustain KV-cache reuse bandwidth.
- Smoke test: KV blocks written and reused across vLLM pods.
- Short TTFT + throughput benchmark.
- File any llm-d implementation issues uncovered during bring-up.

alok7058:
Hi @kfirtoledo , I’d like to take this up.
I’m planning to validate llm-d on AWS (EFS) first and document:
RWX PVC setup
vLLM multi-pod KV cache reuse
TTFT + throughput benchmarks
Let me know if that aligns with expectations or if you prefer another backend first.

kfirtoledo:
@alok7058 Sounds great, that's exactly the plan: documenting AWS EFS installation and adding performance recommendations like the GCP Lustre guide.

PR #1549, opened 2026-05-19
Add Azure Managed Lustre backend guide for tiered prefix cache storage

chokevin:
## Summary

Adds an Azure Managed Lustre Filesystem (AMLFS) backend guide and `StorageClass` to `guides/tiered-prefix-cache/storage/manifests/backends/azure/`, matching the existing per-backend layout under `aws/` (EFS) and `lustre/` (GCP Lustre).

Concretely:

- `manifests/backends/azure/storage_class.yaml` — `StorageClass` using the upstream [`azurelustre.csi.azure.com`](https://learn.microsoft.com/en-us/azure/azure-managed-lustre/use-csi-driver-kubernetes) driver. Dynamic provisioning (AKS creates the AMLFS cluster on PVC bind), `mountOptions: [noatime]`, `AMLFS-Durable-Premium-250` as a reasonable default SKU.
- `manifests/backends/azure/README.md` — Prerequisites / cluster setup / provisioning / PVC verification / performance recommendations / cleanup. Mirrors the structure of `backends/aws/README.md`.
- Updates `guides/tiered-prefix-cache/storage/README.md` to link the new Azure backend guide alongside the existing GCP Lustre and AWS EFS links, and adds `azurelustre-class` to the `STORAGE_CLASS` env-var hint.

## Validation

The provisioning instructions in `backends/azure/README.md` were exercised on an AKS H200 cluster against an existing AMLFS (16 TiB, `mgs-ip-address` + `fs-name` for static, plus the dynamic provisioning path) with the official `azurelustre.csi.azure.com` driver. A two-pod `gpu-vllm-decode` deployment ran the tiered-prefix-cache storage guide end-to-end and produced observable cross-pod KV reuse via the shared Lustre PVC (`External prefix cache hit rate: 69.9%` on a reader pod that never received the writer pod's traffic, with content-addressed block files visible on the AMLFS mount). Full validation evidence will be posted under #1252 separately; that benchmark is intentionally **not** committed here to keep this PR scoped to the per-backend provisioning guide.

## Related

- Per-backend guide pattern established by #1264 (AWS EFS).
- Validates the Azure backend dimension of #1252 ("Validate llm-d with shared storage across multiple vendors").


natoscott:
@chokevin I'm curious about the flock(2) requirement - I don't see any such call in the llm-d fs connector code and this is the first time I've heard that mentioned.  Can you expand on that?  Thanks!

chokevin:
> @chokevin I'm curious about the flock(2) requirement - I don't see any such call in the llm-d fs connector code and this is the first time I've heard that mentioned. Can you expand on that? Thanks!

Yeah honestly I have no idea where that came from. I think it was hallucinated during my testing since it took a bit of tinkering to get the luster setup to work on my AKS cluster. I removed it. Thanks!

chokevin:
Update the description to avoid the reference to flock. Retested everything end to end to make sure it was running appropriately on my AKS cluster with managed lustre set up. LMK if theres anything else we want to benchmark!

natoscott:
> Yeah honestly I have no idea where that came from. I think it was hallucinated during my testing since it took a bit of tinkering to get the luster setup to work on my AKS cluster. I removed it. Thanks!

No problem, thanks for looking into it.  I see your table stops at NFS v4.1 - I'm curious if you had any experiences to share with NFS v4.2 and pNFS?  It seems like it might be another possible option (simpler than the clustered filesystems) - with O_DIRECT, RDMA - could be well-suited for kv cache offload.

chokevin:
I personally don't. I just offer the setup/benchmark upstream as I have access to managed lustre + my aks cluster set up. 

We did originally have a contact at VAST but I don't personally have that connection and not sure which teams are currently enabled on that so wouldn't be able to offer that data up

maugustosilva:
@chokevin Please fix the conflict and we will merge... thanks!

chokevin:
@maugustosilva done. Thanks!

chokevin:
this should be good to merge whenever @maugustosilva

Issue #1567, opened 2026-05-20
[CI/CD] nightly-e2e-predicted-latency-ocp references non-existent reusable workflow


hexfusion:
`nightly-e2e-predicted-latency-ocp` `uses:` `reusable-nightly-e2e-ocp.yaml@main` in `llm-d-infra`, which doesn't exist. It also passes GKE-only inputs. 7/7 scheduled runs fail.

Tracked in #1565.

soumyajyotiii:
Took a pass through the current state of this one. The `nightly-e2e-predicted-latency-ocp.yaml` caller referenced here doesn't exist anymore — it was removed as part of the workflow-name normalization in #1721, and the dead `reusable-nightly-e2e-ocp.yaml@main` reference is gone repo-wide (no hits under `.github/` today). For context, llm-d-infra never shipped a `reusable-nightly-e2e-ocp.yaml`; the OpenShift reusables it exposes are `reusable-nightly-e2e-openshift.yaml` and `-openshift-helmfile.yaml`.

The predicted-latency nightlies that exist now (`cks`/`gke`/`ibm`) all reference reusable workflows that are actually present upstream (`reusable-ci-nightly-benchmark.yaml`, `reusable-update-badge.yaml`), and the OCP path is covered by the `ibm` variant — its consolidate-status is titled "Predicted Latency Routing E2E (OCP GPU)". OCP predicted-latency was also explicitly disabled back in #1452.

So rather than a fix, my plan is to confirm there's no lingering OCP coverage gap the tracker (#1565) still cares about, and if not, this looks safe to close as resolved by #1721. Happy to re-add a correctly-wired OCP caller instead if the intent is to bring that lane back — just let me know which direction the CI folks want.

PR #1574, opened 2026-05-20
fix: e2e nightly on OCP

zdtsw:
# Changes
Try fix https://github.com/llm-d/llm-d/tree/main/release for nightly result on wideep, slo-aware , p/d
- wrong reuseable template name
- remove gke related input for ocp
- switch to use kustomize template than helmfile
ref #1567


maugustosilva:
@ahg-g at this moment, `pd-disaggregation` (with RoCE/GDR) will most likely not work on `OCP` (a new layer will have to added `modelserver/gpu/vllm` to that particular guide) and `wide-ep-lws` will definitely 100% not work :-)

I am ok merging the PR as is (after #1581 was merged), but more work has to be done for OCP.

PR #1581, opened 2026-05-21
chore: add missing nightly test into slash command

zdtsw:
# Notes
- predicted-latency-* group is missing
- precise-prefix-cache-cks and gks are missing too
- format and fix lint
- remove build-image and benchmark-cks

cc @diegocastanibm @maugustosilva


Gregory-Pereira (commented):
generally speaking /lgtm but not sure if @maugustosilva left these out for intentional reasons

maugustosilva (approved):

PR #1583, opened 2026-05-21
Revert "fix: add timeout 90mins for nightly GithubAction"

maugustosilva:
Reverts llm-d/llm-d#1575


maugustosilva:
Hey @zdtsw, it looks like this is a problem, which none of our automated tests caught! 

```
when a reusable workflow is called with "uses", "timeout-minutes" is not available. only following keys are allowed: "name", "uses", "with", "secrets", "needs", "if", and "permissions" in job "nightly"
```

zdtsw:
sorry, looks like this need to set  in the reusable ones, than directly here.

diegocastanibm (approved):

PR #1592, opened 2026-05-22
docs nitpick: remove RDMA from "RoCE RDMA"

VadimEisenberg:
RoCE is RDMA over Converged Ethernet. so "RoCE RDMA" is a tautology. 

https://en.wikipedia.org/wiki/RDMA_over_Converged_Ethernet

InfiniBand also provides RDMA, so RDMA does not make RoCE special in this aspect and there is no point to mention it. If RDMA should be mentioned, it can be put this way:

high-speed, RDMA interconnects (InfiniBand, RoCE)


robertgshaw2-redhat (approved):
thanks for the fix

Issue #1599, opened 2026-05-25
DigitalOcean and openshift infr-provider guides are out of sync


ahg-g:
This is referring to the guides at:
- https://github.com/llm-d/llm-d/tree/main/docs/infra-providers/openshift
- https://github.com/llm-d/llm-d/tree/main/docs/infra-providers/digitalocean

Both still assume helm and completely out of sync with the canonical guides for the well-lit paths. They also present a set of instructions that is also not aligned with what all infra-provider guides are offering, which is provider specific instructions. Both guides needs a full re-work to remove anything that is already detailed at the individual well-lit path guides at https://github.com/llm-d/llm-d/tree/main/guides

PR #1600, opened 2026-05-25
Cleaned up the guides/prereq/gateway-provider directory

ahg-g:
Cleaned up the guides/prereq/gateway-provider directory. All guides reference guides/prereq/gateways.

Note that both openshift and digitalocean infra-provider guides are completely out of sync with the reset of the guides,  https://github.com/llm-d/llm-d/issues/1599 is to track fixing that.


maugustosilva:
@ahg-g can you please rebase? I have merged #1612 and should be possible to fully test (`dry-run`) this PR now.

ahg-g:
@maugustosilva rebased, I kept the gateway and gaie script for now, we an probably also remove it.

The istio and agentgateway installations are now not backed by any scripts or helm values in the guides, we just have explicit steps to be followed, see istio as an example: https://github.com/llm-d/llm-d/blob/main/guides/prereq/gateways/istio.md#step-2-install-istio

This is now possible because all nightlies moved to use the standalone mode, which doesn't require setting up any of the gateways.

maugustosilva (approved):

PR #1602, opened 2026-05-26
fix(nightly):   e2e for OCP on P/D + WideEP + WVA

zdtsw:
# Notes
- p/d use ksutomize patches than values.yaml from ms-pd + mount .triton for write on ocp because of scc
- wva: create namespace before it gets used
- wideep wrong path to servieraccount file, use GIE standalone chart, similar to gke and cks


clubanderson:
### Nightly E2E Result: `pd-disaggregation` (OpenShift)

| | |
|---|---|
| Result | :x: **Failed** |
| Guide | `pd-disaggregation` |
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/26462704906) |
| Ref | `021d6c2` |

maugustosilva (approved):

PR #1603, opened 2026-05-26
removing helmfile and helmdiff plugin from deps

Gregory-Pereira:
we no longer need the `helm diff` plugin or `helmfile`. Sad to see them go

cc @zdtsw


maugustosilva:
This is breaking `dry-run` for `pd-disaggregation`

PR #1608, opened 2026-05-26
feat(guides): add ModelExpress P2P weight-loading guide

wseaton:
Add a guide for integrating model express with llm-d for P2P model weight loading.


guides/modelexpress-p2p/README.md
```diff
@@ -0,0 +1,633 @@
...
+* **Fixed metadata/worker ports**: `MX_METADATA_PORT` (default `5555`) and `MX_WORKER_GRPC_PORT` are _base_ ports; each TP rank uses `base + device_id`. With TP=2 each pod consumes `5555..5556` and `6555..6556`. If you change TP size, make sure the port range stays free and the values match across pods.
+* **Mixed-version fleets**: the kubernetes backend indexes by `mx_source_id`, which is content-addressed via `SourceIdentity.revision`. Multiple revisions of the model can coexist in the same namespace; pods only consume from sources that match their identity.
+* **No shared storage (P2P arm)**: the P2P arm deliberately uses per-pod `emptyDir` model caches. The seed pod fills its local cache, but receivers never touch theirs - weights land straight in HBM via RDMA. (The fastsafetensors baseline arms are the opposite: they _want_ a shared/warm cache, which is the whole point of that comparison.)
+* **`MX_P2P_METADATA=1`**: tells source workers to serve NIXL agent and tensor manifest data directly to targets instead of routing it through the central server.
```
nicolasnoble: Under the `kubernetes` backend, `MX_P2P_METADATA` is a no-op: `MxK8sServiceClient` forces the P2P path on via `REQUIRES_P2P_METADATA = True`, and setting it to `0` just logs a warning and gets overridden. The Notes & Trade-offs bullet presents it as the switch that enables P2P, which will mislead people since this backend has no central metadata server to route through. Worth dropping it from `patch-vllm.yaml` or noting it's redundant under kubernetes.

wseaton: sounds good, thanks for the catch!

wseaton:
Before merge, need to re-run benchmark numbers with v0.5 bits and get a time estimate for the cache artifact improvements. Will do this tomorrow.

guides/modelexpress-p2p/README.md
```diff
@@ -0,0 +1,709 @@
...
+
+# `--no-deps` keeps the client from shadowing the image's pinned
+# vllm / torch / nixl, so its remaining deps must be listed explicitly:
+# google-crc32c (artifact chunk checksums) is not in the vLLM base image.
+# The 0.5.x wheels ship a prebuilt VMM allocator extension; no build step.
+ARG MODELEXPRESS_VERSION=0.5.0
+RUN python3 -m pip install --target=/opt/modelexpress --no-deps --no-cache-dir \
+        "modelexpress==${MODELEXPRESS_VERSION}" google-crc32c
+ENV PYTHONPATH=/opt/modelexpress
+```
```
omerap12, line 169: Does it make sense to pin the original Dockerfile here? Otherwise, if Modelexpress changes their Dockerfile, we'll have to update this one as well.

guides/modelexpress-p2p/README.md
```diff
@@ -0,0 +1,709 @@
...
+
+## Installation Instructions
+
+### 1. Deploy the ModelExpress Server
+
+This step provisions the `kubernetes`-backend ModelExpress server, RBAC for the `ModelMetadata` and `ModelCacheEntry` CRDs, and a `modelexpress-server` Service. The inference pods find this Service through DNS.
+
+<!-- guide:deploy.mx_server start -->
+```bash
+kubectl apply -n ${NAMESPACE} -f ${REPO_ROOT}/guides/${GUIDE_NAME}/modelexpress/modelexpress-server.yaml
```
omerap12, line 206: Why not using the Modelexpress chart?

guides/modelexpress-p2p/README.md
```diff
@@ -0,0 +1,709 @@
...
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+
+# `--no-deps` keeps the client from shadowing the image's pinned
+# vllm / torch / nixl, so its remaining deps must be listed explicitly:
+# google-crc32c (artifact chunk checksums) is not in the vLLM base image.
+# The 0.5.x wheels ship a prebuilt VMM allocator extension; no build step.
+ARG MODELEXPRESS_VERSION=0.5.0
+RUN python3 -m pip install --target=/opt/modelexpress --no-deps --no-cache-dir \
+        "modelexpress==${MODELEXPRESS_VERSION}" google-crc32c
```
omerap12: I am not sure what is the purpose of `google-crc32c` (I lack the knowledge) but should we pin that as well?

wseaton: it was added as an mx python client dep for checksumming IIRC. we can pin it for the guide yes

guides/modelexpress-p2p/README.md
wseaton: The upstream chart is not istio compatible. I can open a PR to fix and we can swap to it in a new version of the guide.

guides/modelexpress-p2p/README.md
```diff
@@ -0,0 +1,709 @@
+# ModelExpress P2P Weight Transfer
+
+## Overview
```
vMaroon, line 3: I think the guide's current state can be split to 2 or more docs that are referenced within a straightforward, focused main guide path that follows the example of the other guides.

It should include a high level overview, concrete guide deployment steps, verification and **benchmarking/numbers that show measurable value**. Additional side-quests can be pointed to from the main flow where sensible.

vMaroon (approved):
Looks good, considering that this is not yet exposed as well-lit path (the latter should aggregate the fast-model loading experience better) - merging this has low risks of positioning misalignments.

Following the discussion in the leads call and the leadership's OK, merging to avoid missing the release bus.

Issue #1532, opened 2026-05-16
[Install] Migrate WideEP CUDA Images to Upstream vLLM


robertgshaw2-redhat:
SUMMARY:
- we currently build a CUDA image for WideEP, due to a bug in NVSHMEM on RoCE (we have a patch) and various patches we are maintaining
- we should upstream these changes and switch over to vllm official releases

wseaton:
starting to investigate build bumps

ahg-g:
What is the status of this issue? is this done or should we move it to 0.9.0 milestone?

robertgshaw2-redhat:
move to 0.9.0

PR #1610, opened 2026-05-26
Migrate WideEP CUDA image to upstream vLLM + bump NVSHMEM

wseaton:
Refs #1532.

- Pin vLLM to upstream `v0.21.0` (drops the `neuralmagic/vllm` fork; all three cherry-picks are now in upstream).
- Bump NVSHMEM to `v3.5.21-0` and delete the carried `ah_attr` / CKS patches (both fixes are upstream).


Gregory-Pereira:
Are we sure we want to bump NVSHMEM? I thought our new approach was to be stable with versions, particularly staying on v3.4.5 if we can for support of GB200 builds.

PR #1616, opened 2026-05-26
chore: ensure all dry-run steps can be executed with different helm versions

maugustosilva:


maugustosilva:
Talked to @ahg-g and this should no longer be needed.

PR #1622, opened 2026-05-27
feat(recipes): add gke autopilot node selection component

xzhaosg:
This component provides a Kustomize patch to inject GKE Autopilot-specific node selection labels ('cloud.google.com/gke-accelerator: nvidia-l4' and 'cloud.google.com/gke-spot: "true"') into Deployment specifications. This decouples cluster-specific GPU hardware matching from core engine-agnostic deployments.


ahg-g:
@liu-cong should we group gke specific patches under one directory?

guides/recipes/modelserver/components/gke/gke-autopilot/kustomization.yaml
```diff
@@ -0,0 +1,20 @@
...
+        name: all-deployments
+      spec:
+        template:
+          spec:
+            nodeSelector:
+              cloud.google.com/gke-accelerator: nvidia-l4
+              # GKE Autopilot Spot Instance selector.
+              # Set to "true" to use Spot instances (highly cost-effective for dev/test).
+              # Set to "false" or remove this line to use On-Demand instances.
+              cloud.google.com/gke-spot: "true"
```
liu-cong, line 20: I think we should default to false.

liu-cong:
This adds an example patch for gke AP but it's not directly usable in guides (e.g., need to tweak the gpu type). I think we should:
1. Still keep the recipe as as example so users can adapt. 
2. Group GKE patches in one folder as @ahg-g mentioned. Currently we have a `disable-gke-nccl-tuner-patch` component, we can move both under a `gke` folder.
3. In the GKE overlay of each guide,  add an autopilot variant, so it will look like (using optimized baseline as example)
```
guides/optimized-baseline/modelserver/gpu/vllm/
|__gke
|____kustomization.yaml
|____standard
|________kustomization.yaml
|____autopilot
|________kustomization.yaml # here we should apply the correct gpu type that matches the guide
```

xzhaosg:
make sense. let me update that.

guides/optimized-baseline/modelserver/gpu/sglang/gke/autopilot/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../standard
+components:
+  - ../../../../../../recipes/modelserver/components/gke/gke-autopilot
```
liu-cong, line 6: we can't directly apply the component here - we need to pick the correct gpu type for this guide (which is H100). And different guides have different default accelerator types.

guides/optimized-baseline/modelserver/gpu/sglang/gke/standard/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base
+components:
+  - ../../../../../../recipes/modelserver/components/gke/disable-gke-nccl-tuner-patch
```
liu-cong, line 6: this will be needed for both standard and autopilot, so we should add this to the root `gke/kustomization.yaml` or `gke/base/kustomization.yaml`

guides/optimized-baseline/modelserver/gpu/sglang/gke/standard/kustomization.yaml
liu-cong, line 1: since the file paths are changed, we also need to update the README.md of each guide where we set the `INFRA_PROVIDER` env var, e.g., change to `export INFRA_PROVIDER=base # base | gke/standard | gke/autopilot`

PR #1624, opened 2026-05-28
[hardware-support] Build and publish flagos images for cuda and NPU ascend

Gregory-Pereira:
vllm-plugin-FL is a plugin for the [vLLM](https://github.com/vllm-project/vllm) inference/serving framework, built on FlagOS's unified multi-chip backend — including the unified operator library [FlagGems](https://github.com/flagos-ai/FlagGems) and the unified communication library [FlagCX](https://github.com/flagos-ai/FlagCX). It extends vLLM's capabilities and performance across diverse hardware environments. Without changing vLLM's original interfaces or usage patterns, the same command can run model inference/serving on different chips.

For more information see: https://github.com/flagos-ai/vllm-plugin-FL

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -23,7 +23,7 @@ on:
   workflow_dispatch:
     inputs:
       platforms:
-        description: "Platforms to build (comma-separated: cuda,aws,gb200,xpu,cpu,hpu)"
+        description: "Platforms to build (comma-separated: cuda,aws,gb200,xpu,cpu,hpu,flagos-cuda,flagos-ascend)"
         required: false
         default: "cuda"
         type: string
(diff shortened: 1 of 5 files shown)


ahg-g:
what is the intended deployment environment? which guides will be configured to use with these images?

Gregory-Pereira:
> what is the intended deployment environment? which guides will be configured to use with these images?

I was thinking starting with optimized baseline on Cuda since thats what we have access to. Once we also get NPU devices in CI we can integrate those. PD would be a pretty safe bet extending there after this if we so choose

avinashsingh77:
@Gregory-Pereira Created https://github.com/llm-d/llm-d/pull/2097 for adding deployment guides. Once we have the image ready in this PR, we can test deployment on cluster.

avinashsingh77:
Hey @Gregory-Pereira  While trying to build the cuda image,  found two minor fixes:

  1. uv pip install can't find the virtual environment (line 89)

  uv pip install --no-cache -U wheel fails with No virtual environment found because ENV VIRTUAL_ENV=/opt/vllm (line 91) only takes effect in subsequent RUN instructions, not the current one.

  Fix: Add export VIRTUAL_ENV=/opt/vllm inline before uv pip install:
      
```
ln -s /opt/vllm/bin/pip /usr/bin/pip && \
      export VIRTUAL_ENV=/opt/vllm && \
      uv pip install --no-cache -U wheel
```

  2. PyTorch not installed before --no-build-isolation builds

  The vLLM source build (uv pip install --no-build-isolation -e .) expects PyTorch to already be present in the environment, but nothing installs it. The build fails during CMake configuration.

  Fix: Add a PyTorch install step after the build dependencies and before the Mooncake/vLLM steps:
 
 ```
# ---- PyTorch (must precede vLLM and all --no-build-isolation installs) ----
  RUN --mount=type=cache,target=/root/.cache/uv \
      uv pip install \
          --extra-index-url https://download.pytorch.org/whl/cu130 \
          numpy \
          "torch==2.11.0" \
          pyyaml \
          types-PyYAML \
          "patchelf>=0.11.0"
```

  My branch with both fixes: https://github.com/avinashsingh77/llm-d/tree/flagos-dockerfile-fix

  I've successfully built the image (via Kaniko) and validated it deploys and serves inference on an NVIDIA GPU cluster with Qwen3-32B TP=2.

Gregory-Pereira:
Since we've gotten no word back from the flagos team im going to close this out.

PR #1627, opened 2026-05-28
refactor(nightly/wva): extract custom_deploy_script into standalone shell scripts

lionelvillard:
## Summary

- Extracts the ~35-line inline bash body from `nightly-e2e-wva-ocp.yaml` and `nightly-e2e-wva-cks.yaml` into `guides/workload-autoscaling/scripts/nightly-deploy-ocp.sh` and `nightly-deploy-cks.sh`
- Hoists the `${{ needs.build-wva-image.outputs.image_tag || github.event.inputs.image_tag }}` GHA expression to the top of `custom_deploy_script` as an `export WVA_TAG=` assignment; the script files use plain `${WVA_TAG:-}` with no GHA syntax
- No logic changes — identical deploy steps, just reorganized

### DIFF
--- a/guides/workload-autoscaling/scripts/nightly-deploy-cks.sh
+++ b/guides/workload-autoscaling/scripts/nightly-deploy-cks.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# WVA kustomize builds target namespace llm-d-autoscaler (do not pass a conflicting kubectl -n).
+WVA_DEPLOY_NS=llm-d-autoscaler
+yq '.spec.template.spec.priorityClassName="nightly-gpu-critical"' -i guides/optimized-baseline/modelserver/gpu/vllm/base/patch-vllm.yaml
+yq '.spec.template.spec.volumes += {"name": "triton-cache", "emptyDir": {}}' -i guides/optimized-baseline/modelserver/gpu/vllm/base/patch-vllm.yaml
+yq '.spec.template.spec.containers[0].volumeMounts += {"mountPath": "/.triton", "name": "triton-cache"}' -i guides/optimized-baseline/modelserver/gpu/vllm/base/patch-vllm.yaml
+yq '.spec.replicas=2' -i guides/optimized-baseline/modelserver/gpu/vllm/base/patch-vllm.yaml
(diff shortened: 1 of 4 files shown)


lionelvillard:
@mamy-CS @maugustosilva @Gregory-Pereira

guides/workload-autoscaling/scripts/nightly-deploy-cks.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# WVA kustomize builds target namespace llm-d-autoscaler (do not pass a conflicting kubectl -n).
```
ahg-g, line 4: why do we need a special script per provider? I thought wva is agnostic.

lionelvillard: That's a good point, probably not. I first want to move these scripts out of gh workflow files.

lionelvillard: There is actually differences in the way applications (including WVA) connects to prometheus.

mamy-CS (approved):
/lgtm as this is a pure refactor as a correct extract with no logic changes. For existing issues, other prs will follow.

guides/workload-autoscaling/scripts/nightly-deploy-cks.sh
mamy-CS: ocp has tls cert extraction from thanos-querier-tls that cks doesn't need

lionelvillard: In OCP there is no need for cert extraction, it is injected by the service CA operator.

lionelvillard:
@mamy-CS @maugustosilva @Gregory-Pereira @ahg-g Any objections in merging this PR as it is blocking the subsequent one(s)? Our nightly has been failing for a week now and I'd like to fix it ASAP.

guides/workload-autoscaling/scripts/nightly-deploy-cks.sh
ahg-g: we can parameterize it, no?

lionelvillard: obviously :-)

PR #1651, opened 2026-06-01
Optimized Baseline Architectural Shift & Updated Benchmarks

kaushikmitr:
## Summary
This PR transitions the **Optimized Baseline** configuration to a new scheduling strategy that combines the **Token Load Scorer** with **Prefix Cache Affinity Filtering**. This approach optimizes for both backend saturation (via tokens-in-flight) and prefix reuse, resulting in significantly improved throughput and latency profiles compared to standard Kubernetes round-robin. 

While the previous optimized-baseline configuration (based on queue-depth and kv-cache utilization) was hard to generalize across diverse workloads (detailed in [this design doc](https://docs.google.com/document/d/1AdGHoDksnXYI3kA71I5MkH66zPrDsvBzq9aomRXZlpE/edit?tab=t.0#heading=h.p2dtnj3m44at)), this new setup provides a robust "performance floor" that remains vastly superior to stock K8s (+67.8% throughput, -99.5% TTFT). Although we observe a slight decrease in aggregate output tokens per second in this specific benchmark compared to the previous version, this new configuration has proven to work well across a wider range of production workloads.

## Key Architectural Changes
*   **Token-Load Aware Scheduling:** Integrated the `token-load-scorer` as the primary scoring mechanism to balance requests based on real-time token processing load.
*   **Prefix Affinity Overrides:** Implemented the `prefix-cache-affinity-filter` with an automated saturation valve ($\tau$). This ensures sticky routing for prefix cache benefits until a pod reaches a measured load threshold.

## Benchmarking Report Refresh
The `Benchmarking Report` has been updated with final results using this new architecture. The data shows that balancing by token load while respecting prefix affinity provides a superior performance floor compared to standard round-robin.


guides/optimized-baseline/README.md
```diff
@@ -268,11 +345,11 @@ Summary across the full ladder (rates 3 → 60):
 
 | Metric              | k8s service (RR) | llm-d Optimized | Δ% vs k8s |
 | :------------------ | :--------------- | :-------------- | :-------- |
-| Output tokens/sec   | 5,722            | 13,163          | +130.0%   |
-| Requests/sec        | 35.87            | 36.38           | +1.4%     |
-| TTFT mean (s)       | 58.10            | 0.156           | −99.73%   |
-| TTFT p90 (s)        | 107.43           | 0.206           | −99.81%   |
-| ITL mean (ms)       | 44.0             | 47.0            | +6.8%     |
+| Output tokens/sec   | 5,511            | 9,247           | +67.8%    |
```
ahg-g: there is a significant drop here for llm-d, why?

guides/optimized-baseline/README.md
```diff
@@ -268,11 +345,11 @@ Summary across the full ladder (rates 3 → 60):
...
+| Requests/sec        | 5.12             | 9.42            | +84.1%    |
```
ahg-g: this is completely different from the previous results, is this running a different set up compared to before?

guides/optimized-baseline/README.md
```diff
@@ -157,6 +159,81 @@ kubectl apply -n ${NAMESPACE} -k guides/${GUIDE_NAME}/modelserver/cpu/vllm/
 kubectl apply -n ${NAMESPACE} -k guides/recipes/modelserver/components/monitoring
 ```
 
+### 4. (Optional) Calibrate prefix-cache-affinity-filter
+
+This step adds the `prefix-cache-affinity-filter` plugin to the EPP scheduling pipeline, with its saturation-valve parameter `maxTokensInFlightPenalty` (τ) measured against your live deployment. With τ correctly tuned, sticky prefix-affinity routing keeps its TTFT benefits without creating hot-pod imbalance under high load.
```
ahg-g:
```suggestion
This step adds the `prefix-cache-affinity-filter` plugin to the EPP scheduling pipeline, with its saturation-valve parameter `maxTokensInFlightPenalty` (τ) measured against your live deployment. With τ correctly tuned, sticky prefix-affinity routing keeps its TTFT benefits without creating hot-endpoint imbalance under high load.
```

guides/optimized-baseline/README.md
```diff
@@ -12,6 +12,8 @@ The optimized-baseline defaults to two main routing criteria:
 
 - **Load-aware** using both the [kv-cache utilization](https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/scheduling/scorer/kvcacheutilization) and the [queue size](https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/scheduling/scorer/queuedepth) scorers.
 
+An optional [calibration step](#4-optional-calibrate-prefix-cache-affinity-filter) is also available to add a saturation-aware override gate (the `prefix-cache-affinity-filter`) on top of the default plugins, with a per-deployment τ value measured against the live cluster. This is useful when sticky prefix-affinity routing under heavy load could create hot-pod imbalance.
```
ahg-g:
```suggestion
An optional [calibration step](#4-optional-calibrate-prefix-cache-affinity-filter) is also available to add a saturation-aware override gate (the `prefix-cache-affinity-filter`) on top of the default plugins, with a per-deployment τ value measured against the live cluster. This is useful when sticky prefix-affinity routing under heavy load could create hot-spotting.
```

guides/optimized-baseline/router/optimized-baseline.values.yaml
```diff
@@ -15,21 +15,20 @@ router:
...
-        - type: no-hit-lru-scorer
+        - type: token-load-scorer
+        - type: approx-prefix-cache-producer
+          parameters:
+              maxPrefixTokensToMatch: 40960 # Must match vLLM --max-model-len (can be found via /v1/models endpoint)
+        - type: prefix-cache-affinity-filter
+          parameters:
+            affinityThreshold: 0.80
+            maxTokensInFlightPenalty: 286720
+            maxTTFTPenaltyMs: 0
```
ahg-g: this defaults to 0, so we can drop it, right?

guides/optimized-baseline/README.md
```diff
@@ -157,6 +159,81 @@ kubectl apply -n ${NAMESPACE} -k guides/${GUIDE_NAME}/modelserver/cpu/vllm/
 kubectl apply -n ${NAMESPACE} -k guides/recipes/modelserver/components/monitoring
 ```
 
+### 4. (Optional) Calibrate prefix-cache-affinity-filter
```
ahg-g: A couple of comments:
- This adds a fair amount of complexity, which may well be needed, but what is the cost of not tuning this correctly?
- I think this needs to be a little less automated, the tool suggests a value and perhaps produces a helm values file, but lets not automatically apply that.

guides/optimized-baseline/README.md
kaushikmitr: The k8 runs are almost identical and any difference is due to run to run varaince. 
For llm-d, here is a comparison between old and new, we can probably further tune the threshold (TTFTPenalty) to match it but the idea is to keep it general enough that works for different scenarios. wdyt? : 
image
image
image
One additional point, the data collected [here](https://docs.google.com/document/d/1KonmVn5d4nJjg72dhWzcvpYYd8hFZO6G0oCRyCVjKLk/edit?tab=t.0#heading=h.p2dtnj3m44at) is with and older deployemnt that did not have this [line](https://github.com/llm-d/llm-d/blob/e3ed09a5918b9cc418723f87af18b09bb14efb28/guides/optimized-baseline/modelserver/gpu/vllm/base/patch-vllm.yaml#L17). Setting gpu max to 95% improved the older baseline little bit

guides/optimized-baseline/README.md
kaushikmitr: I think the Requests/sec summary was wrong before (it was a bug). @vMaroon i guess requests/sec in the summary table had a typo? I see the numbers much lower (I get them from the overall summary json file generated by inference perf)

guides/optimized-baseline/README.md
kaushikmitr: * The cost is that if we use the same config for a bigger model 128b with the same hardware, or same model with a lower hardware like L4, it can lead to hot spot. Similarly if we use the same config for a smaller model (llama3) but with same hardware, or with same model but B200 the prefix caching will be suboptimal. 
* Yes I can make that change

guides/optimized-baseline/router/optimized-baseline.values.yaml
kaushikmitr: actually default is not 0 but i think we should make it 0. I will create a small PR on this.

guides/optimized-baseline/README.md
kaushikmitr: done

guides/optimized-baseline/router/optimized-baseline.values.yaml
kaushikmitr: if i update maxTTFTPenaltyMs to 0 also need to change the latency predictor values.yaml. Will chnge the defaults everywhere (including here) with a follow up PR

guides/optimized-baseline/README.md
ahg-g: From this benchmarks it is not clear that adding this complexity is warranted, how can we show that it is worth it?

kaushikmitr: agreed, i have not shown any evidence that this calibraiton is needed and will work for other models/hardware. Need to run benchmarks on a different hardware or model

kaushikmitr: evaluated this workload on Llama-3-70B and same hardware (tp=2, 8 pods on H100). Because of the model's size, I adjusted the QPS range downwards to match our hardware capacity.  (pasting below for record). As you see the current default of $286,720$ shows higher TTFT compared to what the calibrated run of $140,394$ shows. This might not rigorously prove that the calibration value is perfectly optimal, but I think it signals that the tool is heading in the right direction.

image

image

image


      stages:
        # Warmup — seat residents (~1×S)
        - rate: 3
          duration: 250            # 15*50  = 750

        # Main ladder
        - rate: 1
          duration: 20
        - rate: 2
          duration: 20
        - rate: 3
          duration: 20
        - rate: 4
          duration: 38
        - rate: 5
          duration: 34
        - rate: 6
          duration: 30
        - rate: 7
          duration: 25
        - rate: 8
          duration: 21
        - rate: 9
          duration: 38
        - rate: 10
          duration: 36
        - rate: 11
          duration: 33
        - rate: 12
          duration: 30
        - rate: 13
          duration: 29
        - rate: 14
          duration: 27
        - rate: 15
          duration: 26
        - rate: 16
          duration: 25

ahg-g: I thought the agentic workload did show benefits; also, lets not use "legacy" models, try with gpt 120B

guides/optimized-baseline/router/optimized-baseline.values.yaml
```diff
@@ -15,21 +15,20 @@ router:
...
         kind: EndpointPickerConfig
         plugins:
-        - type: queue-scorer
-        - type: kv-cache-utilization-scorer
-        - type: prefix-cache-scorer
-        - type: no-hit-lru-scorer
+        - type: token-load-scorer
+        - type: approx-prefix-cache-producer
+          parameters:
+              maxPrefixTokensToMatch: 40960 # Must match vLLM --max-model-len (can be found via /v1/models endpoint)
```
vMaroon: what if the default of this is very high? e.g., 1m? or technically better: if not set, unbounded. Then all models would be covered. If we see this as the new default deployment.

guides/optimized-baseline/router/optimized-baseline.values.yaml
```diff
@@ -15,21 +15,20 @@ router:
...
+        - type: prefix-cache-affinity-filter
+          parameters:
+            affinityThreshold: 0.80
+            maxTokensInFlightPenalty: 286720
```
vMaroon: Can these be transformed to intuitive params that can then translate back to these magic numbers? It would also assist in algorithm explainability.

guides/optimized-baseline/README.md
```diff
@@ -148,15 +150,100 @@ kubectl apply -n ${NAMESPACE} -k guides/${GUIDE_NAME}/modelserver/cpu/vllm/
...
+- Install the [Monitoring stack](../../docs/monitoring/README.md).
 - Deploy the monitoring resources for this guide.
 
 ```bash
 kubectl apply -n ${NAMESPACE} -k guides/recipes/modelserver/components/monitoring
 ```
 
+### 4. (Optional) Calibrate `prefix-cache-affinity-filter`
+
+The guide ships with a static default for `maxTokensInFlightPenalty` (τ = 286720), calibrated for Qwen3-32B on H100 TP=2 at T_max=17s. This default works well for that reference setup but is not optimal for other hardware/model combinations:
```
vMaroon: See this effort: https://github.com/vllm-project/vllm/issues/38147

What information can be exposed there, and then attached to the endpoint identity to fully automate calibration? vLLM already runs profiling on startup, we can expand it perhaps if info is missing.

guides/optimized-baseline/router/optimized-baseline.values.yaml
kaushikmitr: yes i am thinking: PeakPrefillThrougput (that can be calibrated) and MaxTTFTPenaltyInSecs that is already there. Together we can calcuate maxTokensInFlightPenalty

guides/optimized-baseline/router/optimized-baseline.values.yaml
kaushikmitr: yes. I think 0 will make it unbounded.

guides/optimized-baseline/README.md
kaushikmitr: for 120B we will need more GPUs  etc. which is okay but that may not be desirable for optimized-baseline setup for easy reproduction?

vMaroon: `gpt-oss-120b` can run on a single H100 with 150k tokens KV-cache. With 2 of them it can get to 1m.

kaushikmitr: oh i see, i am surprised since its ~4 times bigger.  can you share reccommened deployment yaml of GPT OSS 120 on H100 if you have in handy i will try this there

kaushikmitr: i tried gpt-0ss-120b with tp = 2, 8 H100 with the code-geenration workload and the results are infact consistent with what we saw with qwen-32. Current default is referred to as load+prefix scorer

image

kaushikmitr: For the gpt deployment a couple of tweaks i had to make to prevent OOM errors: 
           - "openai/gpt-oss-120b"
            - "--tensor-parallel-size=2"
            # Cap concurrency: the sampler/activation warmup sizes itself to
            # max-num-seqs (default 1024), which OOMs on this memory-tight 120B
            # model. 256 frees that headroom without shrinking the KV cache.
            - "--max-num-seqs=256"
            # 0.90 (not 0.95): at 0.95 vLLM filled the budget with ~1.12M tokens of
            # KV cache, leaving only ~166Mi free — too little for the sampler/CUDA-graph
            # warmup that runs after (OOM by ~32Mi). 0.90 frees ~4Gi headroom while
            # still giving ~820k tokens of KV cache.
            - "--gpu-memory-utilization=0.90"

guides/optimized-baseline/README.md
kaushikmitr: The cleanest will be: 
Endpoint: add peak_prefill_tokens_per_sec (+ the batch/seq-len assumptions it was measured under) to /v1/server/kv-cache or a runtime-data endpoint.
Measurement: hook a short timed prefill after warmup; gate behind a flag like --profile-prefill-throughput so default startup latency is unchanged.

however i am not sure if this belongs to core vLLM. Although it does warmup profile run, its pre cuda graphs  so is pessimistic. So i think any profiling should be done outside vLLM. We can estimate theoritical peak based on model sizes and hardware specs but that again is not something vLLM owns.

guides/optimized-baseline/router/optimized-baseline.values.yaml
ahg-g: does that mean a user needs to provide two parameters instead of one?

kaushikmitr: MaxTTFTPenaltyInSecs is something that will mostly be default and not needed to update (atleast that is what th ebenchmark indiciates)

guides/optimized-baseline/router/optimized-baseline.values.yaml
kaushikmitr: did not make it 0 as discussed and made it 130172: https://github.com/llm-d/llm-d-router/pull/1548 As making it unbounded might have adverse affect on CPU for very long prompt in millions while may not be needed. However it should be tuned based on use cases.

guides/optimized-baseline/router/optimized-baseline.values.yaml
kaushikmitr: This should make it more intuitive: https://github.com/llm-d/llm-d-router/pull/1547

guides/optimized-baseline/README.md
ahg-g: this is interesting, so roughly 135 concurrency for your benchmark is the pool saturation

kaushikmitr: yes though the deployment yaml says 256 the tput kind of saturated at 135.

vMaroon:
If we go this path, can we build a configuration matrix of (model, accelerator) for every combination we have in guides/we want to support?

ahg-g:
I think we should delay this to the next release to come up with a structure for the guide to show case multiple models and accelerators.

rahulgurnani:
I observed good performance with this scheduling profile with SGLang as well, plots: 
image

This was even before bumping up the kv cache memory for sglang to 0.9 in the change https://github.com/llm-d/llm-d/pull/1840

kaushikmitr:
> I think we should delay this to the next release to come up with a structure for the guide to show case multiple models and accelerators.

Makes sense, I just created this: https://github.com/kaushikmitr/llm-d/blob/d3bfa5d20c9f2430e2ecbbf9bc597d0142fe10b3/guides/recipes/router/calibration/configuration-matrix.md

@rahulgurnani  helped with getting the sglang/vllm numbers. For amd/xpu/hpu i dont have h/w to measure so using the  equivalent nvidia or cpu numbers

kaushikmitr:
created a grid: 
https://github.com/kaushikmitr/llm-d/blob/updated-optimized-benchmark/guides/recipes/router/calibration/configuration-matrix.md
@vMaroon @ahg-g

kaushikmitr:
What are the next steps here? https://github.com/llm-d/llm-d/pull/1651 I was thinking of just updating the benchmark results with the new router. Any other concern?

kaushikmitr:
I updated the benchmark results folder with the results from the new router. 
Note that the optimized-baseline workload replays one fixed prompt set at increasing concurrency, after a seed round, 150 prefix groups × 5 questions drawn from the same corpus at every rate, with a warmup stage that seats those prefixes in cache. 

I think the current router is better by design in exactly that regime, especially no-hit-lru-scorer, whose job is spreading cache-missing requests. When the corpus is pre-seeded and never grows, misses are rare and predictable, so LRU-style spreading should be close to optimal. 

But I don't think that advantage generalizes. It seems to depend on prompts being reused and pre-seeded. Where each concurrency level brings an independent, unseeded set of prompts, agentic workloads, say, the cache is continuously cold on new work, and we see the current set of scorers perform worse than k8.

That's why I lean toward the token-based routing as the more principled default. It reasons about actual prefill token load rather than mix of different weights, and the exposed configurations feels right to me:

**peakPrefillThroughput**: configured, not tuned. It looks like a genuinely measurable property of a (model, accelerator, TP, chunk-size) deployment, obtained deterministically from the calibration recipe. (gpt-oss's matrix value of 39,065 reproduced to within 1% for me — measured 38,980 / 39,364.)
**maxTTFTMs**: the one knob I'd actually tune. I kept it at its default across every benchmark here, so none of these results depend on tuning it.

So I'd read the numbers as a clear win over no routing at all, reached with one measured parameter and one policy parameter, rather than as a claim that it beats the incumbent scorers on a benchmark that's arguably built around the property those scorers exploit. Let me know what you think.

ahg-g:
/approve

Thanks!

PR #1640, opened 2026-05-30
fix(flow-control): migrate InferenceObjective to llm-d.ai/v1alpha2 API group

gyliu513:
- `guides/flow-control/objectives.yaml`: update `apiVersion` for all three `InferenceObjective` objects from `inference.networking.x-k8s.io/v1alpha2` to `llm-d.ai/v1alpha2`
- `guides/flow-control/README.md`: update the CRD install command to use `llm-d-router/config/crd` instead of the upstream GAIE repo

PR #1656, opened 2026-06-02
docs(guides): install only the InferencePool CRD via v1-manifests.yaml

gyliu513:
This is a follow up for #1640: Updates guides and getting-started docs to install the GAIE **InferencePool** CRD via the released `v1-manifests.yaml` manifest.

/cc @ahg-g


gyliu513:
CI is being fixed at https://github.com/llm-d/llm-d/pull/1657

PR #1507, opened 2026-05-14
update: add guide for precies-prefix to use new vllm launch render

zdtsw:
# Notes

- keep old UDS example but mark as legancy support:
- set to use 0.19.1 vllm image for render
- remove tokenizerPoolConfig from uds values.yaml after https://github.com/llm-d/llm-d-router/pull/1112

# Dependency
require: https://github.com/llm-d/llm-d-router/pull/1079
depends on ~~https://github.com/llm-d/llm-d-router/pull/1371~~  https://github.com/llm-d/llm-d-router/pull/1428

PR #1657, opened 2026-06-02
fix(ci): satisfy chart v0 tokenizer modelName validation

gyliu513:
Fixes `dry-run-precise-prefix-cache-routing` CI failure where `helm template` against `llm-d-router-standalone-dev:v0` exits with:

```text
.Values.router.tokenizer.modelName is required when the tokenizer is enabled.
```

/cc @zdtsw @ahg-g

### DIFF
--- a/guides/precise-prefix-cache-routing/router/precise-prefix-cache-routing.values.yaml
+++ b/guides/precise-prefix-cache-routing/router/precise-prefix-cache-routing.values.yaml
@@ -8,6 +8,8 @@
 router:
   tokenizer:
     enabled: true
+    # Required by llm-d-router chart v0 when tokenizer is enabled; must match model server + plugins below.
+    modelName: Qwen/Qwen3-32B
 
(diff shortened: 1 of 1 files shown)


gyliu513:
BTW: This is related to https://github.com/llm-d/llm-d-router/pull/1428, this PR introduced Tokenizer validations: require modelName for the render sidecar's command args.

zdtsw:
> BTW: This is related to [llm-d/llm-d-router#1428](https://github.com/llm-d/llm-d-router/pull/1428), this PR introduced Tokenizer validations: require modelName for the render sidecar's command args.

something was planned to get out in https://github.com/llm-d/llm-d/pull/1507 along with the llm-d-route chart changes i believe.

gyliu513:
@zdtsw I think we may need to fix the CI error as soon https://github.com/llm-d/llm-d/pull/1656.

PR #1675, opened 2026-06-03
Update Optimized baseline guide benchmarks

oshfeder:
Added gpt-oss-120b benchmark configuration and results for Optimized baseline guide as optional deployment

Model tested *only* on NVIDIA H100 GPU (not tested on other hardware listed as supported)
Benchmark with vllm image vllm/vllm-openai:v0.22.0

Requires model-pvc

The benchmark run by [compare-llm-d-configurations skill](https://github.com/llm-d-incubation/llm-d-skills/tree/main/skills/compare-llm-d-configurations)


oshfeder:
@vMaroon

clubanderson:
### Nightly E2E Result: `tiered-prefix-cache` (GKE)

| | |
|---|---|
| Guide | `tiered-prefix-cache` |
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/27336686354) |
| Ref | `65a1c38` |

clubanderson:
### Nightly E2E Result: `tiered-prefix-cache` (OpenShift)

| | |
|---|---|
| Guide | `tiered-prefix-cache` |
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/27336687712) |
| Ref | `65a1c38` |

oshfeder:
rebased

guides/optimized-baseline/modelserver/gpu/vllm/gpt-oss/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
ahg-g, line 1: @jjk-g what is your recommendation for a kustomize directory structure if we are to support multiple models? Do you recommend a different approach than what this PR proposes?

ahg-g (commented):
I think before adding a new model, we need to come up with a structure to support multiple models, specifically:

1) The kustomize directory structure
2) The guide (README.md) structure to easily select and apply one of the models
3) The benchmarking results (ideally split into independent files that the guide points to to avoid cluttering it)

guides/optimized-baseline/modelserver/gpu/vllm/gpt-oss/kustomization.yaml
oshfeder: @ahg-g @jjk-g see aligned https://github.com/llm-d/llm-d/pull/1676, add benchmark results for kv cache cpu offload on gpt-oss-120b

ahg-g: It does not seem that PR addressed the above points: (1) it is not clear how to deploy the guide for gpt-oss, (2) the presentation of the benchmark results is not consistent (the gpt-oss in a separate file while the qwen are in the guide itself).

oshfeder: @ahg-g Are you referring to this PR? both model results are in the guide itself (README.md)

jjk-g: @oshfeder, @ahg-g is referencing splitting the benchmark results to a seperate md file https://llm-d.slack.com/archives/C08TSFYMSCQ/p1781211176020299. The qwen and gpt-oss results should be in a seperate md outside of the guide readme. 

For kustomize directory structure, borrowing from the [serving-catalog](https://github.com/kubernetes-sigs/wg-serving/tree/main/serving-catalog) convention, we should

1. Treat well-lit path as a base (similar to [deployment/lws](https://github.com/kubernetes-sigs/wg-serving/tree/main/serving-catalog/core/deployment) in the serving-catalog)
2. Treat each accelerator as [components](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/components.md). These are common patches used over the bases.
3. Specific models then become the final implementation that defines the composition of base + components as well as defining any model specific configuration as [overlays](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#bases-and-overlays). You can use a series of base/overlays to further specify model servers / providers etc.

2 & 3 are a slight divergence from the current structure. Following the suggested pattern enables generating via kustomize as something like:

```
kustomize build optmized-baseline/vllm/gpt-oss/gke | kubectl apply -f -
```

In this example the final provider directory contains a kustomize.yaml ([example](https://github.com/kubernetes-sigs/wg-serving/blob/main/serving-catalog/core/deployment/vllm/gemma-2b/gke/kustomization.yaml)) that builds on the overlays, references componnents, and specifies any further specific patches.

oshfeder: Updated structure per @jjk-g  and @ahg-g comments

guides/optimized-baseline/README.md
```diff
@@ -27,19 +27,33 @@ The optimized-baseline defaults to two main routing criteria:
 | GPUs per replica   | 2                                                       |
 | Total GPUs         | 16                                                      |
 
+## Additional Configuration
```
vMaroon: This should be combined with the above, as a new column. Can call it "example configurations", and add "(default)" next to Qwen.

The guide is also missing how to deploy the non-default config. It should be a comment/additional instruction on the deployment steps

oshfeder: Update configuration tables
Added how to deploy the non-default model config section

ahg-g: Are there any other tunings that were done to the router specific to gpt-oss or was it exactly the same as the one used for qwen?

ahg-g:
I will merge this and send a follow up PR to restructure the results directory a bit

guides/optimized-baseline/README.md
oshfeder: same router, modified question_len and output_len

Issue #1521, opened 2026-05-16
[P/D] Graduate DP-Aware WideEP Scheduling to Production


robertgshaw2-redhat:
SUMMARY:
- [x] land vllm-side support - https://github.com/vllm-project/vllm/pull/40841
- [x] update cks to use this feature, re-run benchmark
- [x] update gke to use this feature, re-run benchmark
- [x] update docs to cover this in more detail

tessapham:
I'd like to take this on.

robertgshaw2-redhat:
vllm side pr has landed: https://github.com/vllm-project/vllm/pull/40841

PR #1678, opened 2026-06-03
update deploy configs to use DP supervisor

tessapham:
Part of https://github.com/llm-d/llm-d/issues/1521.

### DIFF
--- a/guides/wide-ep-lws/experimental-dp-aware/manifests/modelserver/base/decode.yaml
+++ b/guides/wide-ep-lws/experimental-dp-aware/manifests/modelserver/base/decode.yaml
@@ -219,6 +219,9 @@ spec:
               - containerPort: 8207
                 name: rank7
                 protocol: TCP
+              - containerPort: 8208
+                name: supervisor
+                protocol: TCP
 
             startupProbe:
               httpGet:
--- a/guides/wide-ep-lws/experimental-dp-aware/manifests/modelserver/base/decode.yaml
+++ b/guides/wide-ep-lws/experimental-dp-aware/manifests/modelserver/base/decode.yaml
@@ -108,35 +108,32 @@ spec:
                 find /dev/shm -type f -delete
 
                 #################
-                # RUN vLLM decode workers
+                # RUN vLLM decode workers via DP Supervisor
+                # The supervisor automatically assigns CUDA_VISIBLE_DEVICES and
+                # --data-parallel-rank to each child process based on local_rank.
                 #################
                 START_RANK=$(( ${LWS_WORKER_INDEX:-0} * DP_SIZE_LOCAL ))
 
-                COMMON_VLLM_ARGS=(
-                  --disable-uvicorn-access-log
-                  --data-parallel-size $((LWS_GROUP_SIZE * DP_SIZE_LOCAL))
-                  --data-parallel-address="${LWS_LEADER_ADDRESS}"
-                  --data-parallel-rpc-port=5555
-                  --enable-expert-parallel
-                  --kv_transfer_config '{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_load_failure_policy":"fail"}'
-                  --all2all-backend deepep_low_latency
-                  --max-model-len 65536
+                exec vllm serve \
+                  deepseek-ai/DeepSeek-R1-0528 \
+                  --host=0.0.0.0 \
+                  --port=8200 \
+                  --disable-uvicorn-access-log \
+                  --data-parallel-multi-port-external-lb \
+                  --data-parallel-supervisor-port=8208 \
+                  --data-parallel-size $((LWS_GROUP_SIZE * DP_SIZE_LOCAL)) \
+                  --data-parallel-size-local $DP_SIZE_LOCAL \
+                  --data-parallel-start-rank $START_RANK \
+                  --data-parallel-address="${LWS_LEADER_ADDRESS}" \
+                  --data-parallel-rpc-port=5555 \
+                  --enable-expert-parallel \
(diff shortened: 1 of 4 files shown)


Gregory-Pereira:
cc @tlrmchlsmth @shmuelk @vMaroon

shmuelk:
This looks good.

I have one question. The supervisor is launched listening on the "base" port. It launches instances of vLLM to process each rank. Does the supervisor process also handle rank zer? In other words, if --data-parallel-size=4, are four processes launched or only three?

If four processes are launched, the port of the supervisor needs adjusting in all of the cases.

tessapham:
Good question! To clarify, the supervisor listens on port 8208 (in decode), which is not the "base" port. It also doesn't handle any rank itself. Looking at the [vLLM DP Supervisor code](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/dp_supervisor.py):

- With `--data-parallel-size-local=N`, the supervisor launches N child processes (one per rank).
- Children are assigned ports `base_port + 0` through `base_port + (N-1)` (e.g., 8200-8207 for decode).
- The supervisor itself only serves health/readiness endpoints (`/health`, `/readyz`) on the `--data-parallel-supervisor-port`.
- The supervisor [validates at startup](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/dp_supervisor.py#L105-L112) that the supervisor port does NOT overlap with the child port range.

So in our config: 8 children on ports 8200-8207, supervisor on 8208.

shmuelk:
Thanks for the explanation. I missed the fact that the supervisor port was either 8208 or 8008 as appropriate.

/lgtm

tessapham:
@robertgshaw2-redhat I used these configs to run benchmarks on CKS:

## CKS Benchmark Results

Ran `2048_concurrent_2k_isl_2k_osl` benchmark on CKS (4x H200 nodes, 32 GPUs, InfiniBand) with the DP Supervisor config.

  | Metric | Hybrid-LB (baseline) | DP Supervisor | Change |
  |---|---|---|---|
  | Output tokens/s | 22,125 | 27,853 | +26% |
  | Input tokens/s | 22,586 | 29,329 | +30% |
  | Total tokens/s | 44,711 | 57,183 | +28% |
  | Requests/s | 11.03 | 14.33 | +30% |

 Configuration:
- Model: DeepSeek-R1-0528
- vLLM image: `vllm/vllm-openai:v0.22.1-x86_64-cu129-ubuntu2404`
- Prefill: 2 pods × 8 DP ranks (EP=16)
- Decode: 2 pods × 8 DP ranks (EP=16)
- Benchmark: 2048 concurrent requests, 2K ISL, 2K OSL

If this looks good the deploy configs are good to go. I've published the results in https://github.com/llm-d/llm-d/pull/1707.

robertgshaw2-redhat:
this is great! we should remove experimental-dp-aware repo and make this the main

tessapham:
@robertgshaw2-redhat Makes sense. I'll do that in a separate PR to make it easier to review config changes here. Can we get this in first?

guides/wide-ep-lws/experimental-dp-aware/manifests/modelserver/base/decode.yaml
Gregory-Pereira, line 92: When did this work land in vLLM? Do we need to use a newer image for this to work?

guides/wide-ep-lws/experimental-dp-aware/manifests/modelserver/base/decode.yaml
```diff
@@ -108,35 +108,32 @@ spec:
...
-                  --data-parallel-rpc-port=5555
-                  --enable-expert-parallel
-                  --kv_transfer_config '{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_load_failure_policy":"fail"}'
-                  --all2all-backend deepep_low_latency
-                  --max-model-len 65536
+                exec vllm serve \
+                  deepseek-ai/DeepSeek-R1-0528 \
+                  --host=0.0.0.0 \
+                  --port=8200 \
+                  --disable-uvicorn-access-log \
```
Gregory-Pereira: Personally I prefer getting access logs just not for the non-helpful API routes:
```
--disable-access-log-for-endpoints=/health,/metrics,/v1/models
```
It does come down to personal preference though

Gregory-Pereira (commented):
Just some questions, all this /lgtm

guides/wide-ep-lws/experimental-dp-aware/manifests/modelserver/base/decode.yaml
tessapham: It landed in vllm v0.22.0. The `llm-d-cuda` image still uses vllm v0.19.1 (pinned [here](https://github.com/llm-d/llm-d/blob/main/docker/common-versions#L6)), but updating it would require a broader change. I've overridden the vllm image in the CKS overlay instead: https://github.com/llm-d/llm-d/pull/1678/changes/5ec6883dcd2f7830696bf07d152cf41be3015424.

Gregory-Pereira: Excellent

Gregory-Pereira (dismissed):
/lgtm

guides/wide-ep-lws/experimental-dp-aware/manifests/modelserver/base/decode.yaml
tessapham: Good point, this gives more visibility. Updated the flag and also added `/readyz` endpoint.

tessapham:
@Gregory-Pereira Oops could you re-stamp?

Gregory-Pereira (approved):
/lgtm

tessapham:
@Gregory-Pereira Could you give me write access to the repo so I can merge PRs moving forward? Thanks!

PR #1686, opened 2026-06-04
fix(pd): increase keep-alive timeout in remaining vLLM prefill manifests

weizhoublue:
Issue: 
Deploy P/D disaggregation guides. Decode routing sidecar forwards  requests to prefill server. Connection idle time stays between 5s and 90s.
 Intermittent prefill or inference request failures. Connection terminations via  TCP RST.
 
Root Cause: 
PR #1564 identified HTTP keep-alive timeout mismatch between sidecar  idle timeout (90s) and uvicorn default timeout (5s) but only applied fixes to HPU and select  GPU manifests. Mismatch resides in software layer (vLLM/uvicorn config) and affects all  hardware setups. Remaining hardware-specific manifests lack this uniform alignment.

 Remediation: 
Configure  VLLM_HTTP_TIMEOUT_KEEP_ALIVE: "120"  in patch-prefill.yaml,  patch-prefill.yaml, patch-prefill.yaml, and patch-prefill.yaml.


weizhoublue:
@ilmarkov Hi, I feel that your changes also apply to other YAML files, right? Please double-check. Thanks.

robertgshaw2-redhat:
shouldnt we just make the router's default better?

weizhoublue:
> shouldnt we just make the router's default better?

 Thanks for the suggestion! Before we proceed, want to make sure we're aligned on what "make the router's default better" means here.
 
 My reading: reduce the router's IdleConnTimeout from 90s to something below uvicorn's default keep-alive (5s), e.g. 4s — so it's safe out of the box without any manifest changes.
 
 Is that what you had in mind? Or were you thinking of a different approach, such as exposing IdleConnTimeout as a configurable flag with a safer default?
 
 For context: the tradeoff with a very short idle timeout (e.g. 4s) is more frequent TCP reconnects under bursty load. The manifest-side fix (setting VLLM_HTTP_TIMEOUT_KEEP_ALIVE=120) avoids that cost but requires touching every hardware config. Ideally both layers get fixed — router default is safe, manifests set a high keep-alive for performance

maugustosilva (approved):

PR #1542, opened 2026-05-18
[Docs] Streamline monitoring setup: consolidate docs site and move scripts to guides/recipes

sudoalok:
Closes #1537

## What changed

- added `docs/resources/monitoring/setup.md` — one place to get prom/grafana running, all guides now point here
- added `docs/resources/monitoring/promql.md` — PromQL reference that was sitting in docs/monitoring but never on the docs site
- cleaned up `metrics.md` — pulled out the prom install section since that's now setup.md's job, fixed script paths
- updated `tracing.md` — just script and manifest path updates
- moved everything runnable (scripts, dashboards, tracing yamls) from `docs/monitoring/` to `guides/recipes/observability/` — scripts don't belong in website content
- deleted `docs/monitoring/` entirely
- updated 10 guide READMEs and the relevant CI workflows to use the new paths

## Why

there were three separate locations with overlapping monitoring content and no clear entry point for a new user. this just consolidates it — docs site content lives in `docs/resources/monitoring/`, runnable stuff lives in `guides/recipes/observability/`, and every guide links to one setup page instead of wherever they happened to point before.

PR #1687, opened 2026-06-04
docs: add SGLang overview Grafana dashboard

### LINKED ISSUE #1665: [Docs] Add SGLang overview Grafana dashboard
### Feature Area

Observability / Monitoring

### Problem Statement

The observability docs (#1542) ship a vLLM overview Grafana dashboard (`llm-d-vllm-overview`) but there is no SGLang equivalent. SGLang metrics are already documented in `docs/resources/observability/metrics.md` (Key SGLang Metrics), so users running SGLang model servers can read the metrics but have no pre-built dashboard to visualize them.

### Proposed Solution

Author and test an SGLang overview Grafana dashboard (parity with the existing vLLM one) and add it to `guides/recipes/observability/grafana/dashboards/`, e.g. `llm-d-sglang-overview.json`. It should cover the documented SGLang metrics: running/waiting requests, token usage (KV cache), TTFT, ITL, and prompt/generation token throughput. Wire it into `load-llm-d-dashboards.sh` and the dashboard tables in the observability docs.

### Alternatives Considered

Hand-writing the dashboard JSON as part of #1542 was considered but rejected — an untested dashboard risks shipping broken "No data" panels. Tracking it separately lets it be authored and verified against a live SGLang deployment first.

### Willingness to Contribute

Yes, with guidance

### Additional Context

Deferred follow-up from #1542 (observability docs restructure). Requested by @gyliu513 during review.

sudoalok:
This adds a Grafana dashboard for SGLang, similar to the vLLM overview dashboard we already have. It's the follow-up from #1542. Closes #1665.

The dashboard shows:
- Running and waiting requests
- Token throughput (prompt + generation)
- Time to first token (TTFT)
- Inter-token latency (ITL)
- Token / KV cache usage

I kept the same templating as the vLLM dashboard (DS_PROMETHEUS, model_name, namespace), so it behaves the same way. The loader script picks it up automatically as the `llm-d-sglang-overview` ConfigMap, so no script change was needed. I also added it to the dashboard tables in metrics.md and setup.md.

## Testing

I don't have a GPU, so I couldn't test against a real SGLang server. Instead I ran a small fake exporter that emits the same sglang metrics on a local kind cluster with Prometheus + Grafana, loaded the dashboard, and checked that all the panels show data (including the percentile ones).

If someone has a real SGLang deployment, it'd be great if you could confirm the metric names and labels line up with a live server.


guides/recipes/observability/grafana/dashboards/llm-d-sglang-overview.json
```diff
@@ -0,0 +1,812 @@
...
+          "useBackend": false
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "disableTextWrap": false,
+          "editorMode": "code",
+          "expr": "sglang:num_waiting_reqs{model_name=~\"$model_name\",namespace=~\"$namespace\"}",
```
JustinTong0323: Use `sglang:num_queue_reqs` here. SGLang v0.5.12.post1 exports queued requests as `sglang:num_queue_reqs`, so `sglang:num_waiting_reqs` leaves the waiting-request series empty on a live server.

JustinTong0323 (commented):
Fix the SGLang queue metric before landing the dashboard.

gyliu513 (commented):
We should be fine once @JustinTong0323 's comments addressed.

PR #1694, opened 2026-06-04
fix: fixed typo

capri-xiyue:
I checked my previous benchmark config. I used image 4 instead of 3 and the maximum qps is 35 instead of 40

And in the approximate prefix cache routing, as it is estimate, we just ignore +2 token


capri-xiyue:
assign @liu-cong

maugustosilva (approved):

PR #1705, opened 2026-06-05
 Add benchmarking templates and report to the predicted-latency-routing guide

kaushikmitr:
## What this does

Adds a reusable benchmarking setup and a published results report to the **predicted-latency-routing** guide, plus a few correctness fixes to the router and model-server manifests.

## Changes

**Benchmarking templates** (`benchmark-templates/`)
- `run_benchmark.sh` — reusable orchestration script. Restarts the EPP, provisions a per-run PVC, submits one `inference-perf` Job per concurrency level, waits with early failure detection, then collects results from the PVC via a copy pod. Workload-agnostic via the `WORKLOAD` env var, with overridable `CONCURRENCY_LEVELS`, `MODEL_NAME`, and `EPP_DEPLOYMENT`.
- `bench-job.yaml` — Kubernetes Job template (pinned to `inference-perf:v0.5.0`).
- `workloads/code-generation/code-generation.yaml` — synthetic code-generation workload (conversation replay; long shared/dynamic system prompts; multi-turn; lognormal input/output lengths).

**Model server overlays** (`modelserver/gpu/vllm/{base,gke}`)
- Kustomize overlays that reuse the optimized-baseline model server with a predicted-latency patch (10 decode replicas, TP=2, 131072 context, YARN rope scaling).
- The `gke` overlay includes the `disable-gke-nccl-tuner-patch` component for parity with optimized-baseline.

**Router values** (`router/`)
- `predicted-latency.values.yaml` (routing-only): corrected the header to describe the routing-only setup (predictor trains on end-to-end latency, no SLO headers, no streaming required); fixed the `pluginsConfigFile` key so it matches the provided custom config and the pipeline is actually applied; fixed `prefix-cache-scorer` parameter indentation.
- `predicted-latency-slo.values.yaml` (SLO-aware): corrected `prefix-cache-scorer` indentation / `maxPrefixTokensToMatch` nesting and declared the missing `slo-headroom-tier-filter` plugin.

**README**
- New **Benchmarking** section (workload description, run steps, env-var override table, `<run-id>` clarification).
- New **Benchmarking Report** with the code-generation comparison plot and full per-concurrency data table (predicted-latency routing vs. plain k8s Service).
- Fixed the cleanup step to delete the predicted-latency-routing model-server overlay (was pointing at optimized-baseline).

## Benchmark results

Run on 10 vLLM decode pods (TP=2, 20 × H100), `Qwen/Qwen3-32B`. Predicted-latency routing vs. a plain Kubernetes Service:

| Metric | Δ vs k8s Service |
|---|---|
| Input throughput | **+38.3%** |
| Output throughput | **+40.2%** |
| TTFT p50 | **−13.0%** |
| TTFT p90 | **−46.8%** |

## Notes

- Reasoning workload is intentionally out of scope for this PR; to be added later.


guides/predicted-latency-routing/benchmark-templates/run_benchmark.sh
```diff
@@ -0,0 +1,175 @@
+#!/bin/bash
```
ahg-g: why do we need yet another benchmarking script? /cc @achandrasekar @liu-cong

ahg-g: if something missing in the common benchmarking scripts, then we should fix that there instead of having a custom script for each guide.

kaushikmitr: @achandrasekar is there a way to integrate this with inference-perf. We wanted a fresh job for every concurrent to allow separate num_conversation (= conc) and no overlap in system prompt: (as discussed here) https://docs.google.com/document/d/1fJLNYM9Kg4Cq-67I06Yqd0DIs96Tf34zN3_uaAKVXCU/edit?resourcekey=0-v0SEw2ra0vxKbUe9GTZHhg&disco=AAAB84nTgNc

kaushikmitr: actually only reason we needed the custom script is to sweep conc but we can instead adopt #1727
approach, just run it once for each conc. Thats cleaner for now. I remove the custom script and using the standard benchmarking harness

achandrasekar: +1 to what @ahg-g mentioned. For sweeping, let's address the issue in inference-perf directly. But for cases, where we need to do multiple clean runs, we can rely on llm-d-benchmark's sweep automation - https://github.com/llm-d/llm-d-benchmark/blob/main/experiments/max-concurrency-sweep.yaml. I believe this one does it for vllm benchmark. But we should be able to do something similar with inference-perf as well. cc @jjk-g @maugustosilva

kaushikmitr: @achandrasekar see the updated README, wdyt

guides/predicted-latency-routing/README.md
```diff
@@ -185,13 +185,90 @@ Once traffic is flowing, confirm three things in Prometheus (see the [architectu
...
+
+envsubst < guide.yaml > config.yaml
+./run_only.sh -c config.yaml -o ./results
+```
+
+To sweep concurrency, set a new `CONCURRENCY` (and `NUM_REQUESTS`) and re-run the block above for each level. Each run is saved under `./results/`.
+
+## Benchmarking Report
+
+The benchmark runs on 10 vLLM decode pods, each with tensor parallelism of 2 (20 × H100 GPUs total), using the `Qwen/Qwen3-32B` model. Results compare predicted-latency routing against a plain Kubernetes Service (round-robin, no EPP).
```
ahg-g, line 244: Why 10? can we align on one configuration for qwen so that benchmarks are comparable across guides?

ahg-g:
thanks @kaushikmitr , but why do you use 10 replicas instead of 8 like the optimized-baseline?

/approve

guides/predicted-latency-routing/benchmark-templates/run_benchmark.sh
jjk-g: afaik the mentioned bug fix in inference-perf is merged, @kaushikmitr is the only dependency to support this guide a new released inference-perf version?

+1 to clean interdependent runs being managed by llmdbenchmark experiments. @Vezio

guides/predicted-latency-routing/README.md
kaushikmitr: I did my runs with 10 replicas, i could have aligned with optimized baseline and rerun with 8 but we are anyway doing rope scaling to support longer prompts in qwen 32. So since we are anyway have our own model server patch_vllm so did not change it

guides/predicted-latency-routing/benchmark-templates/run_benchmark.sh
kaushikmitr: > afaik the mentioned bug fix in inference-perf is merged, @kaushikmitr is the only dependency to support this guide a new released inference-perf version?
> 
> +1 to clean interdependent runs being managed by llmdbenchmark experiments. @Vezio

The current approach actually works well. If llmdbenchark has a sweep automation, i can update to that in a follow up PR. 

To do it in inference-perf we need 2 things, we can modify the current benchmark to remove 2 but i think its a good feature to have: 
1. have llm-d-bencchmark support latest inference-perf (which will probably happen in next release). 
2. This feature request: https://github.com/kubernetes-sigs/inference-perf/issues/540

jjk-g: ack, thanks for filing!

Can you create a high level llm-d issue to track updating the guide and the dependent issues so we can track and align on timelines?

PR #1715, opened 2026-06-08
[Docs] Replace deprecated `kv_both` for NIXLConnector and with explicit P/D Roles

NickLucche:
We're deprecating `kv_both` in vLLM for the NixlConnector in favor of explicit roles https://github.com/vllm-project/vllm/issues/43807 so startup vllm serve commands for P and D will be asymmetric (they already are when configuring things like DeepEP but anyway..).
The TL;DR: this allows vLLM to have finer control at config time, that is prior to handshaking/pod discovery, to handle some important edge-cases more cleanly. 

This PR simply changes the docs to recommend setting roles.
Mind that at the time of writing `kv_both` isn't yet fully deprecated, so nothing is happening in the current release when setting kv_both.

### DIFF
--- a/docs/resources/rdma/README.md
+++ b/docs/resources/rdma/README.md
@@ -76,16 +76,21 @@ NIXL selects the backend based on what is available and the memory types involve
 Enable NIXL-based KV Cache transfer via the `--kv-transfer-config` flag:
 
 ```bash
+# Prefill pods
 vllm serve <model> \
-  --kv-transfer-config '{"kv_connector":"NixlConnector","kv_role":"kv_both"}'
+  --kv-transfer-config '{"kv_connector":"NixlConnector","kv_role":"kv_producer"}'
+
+# Decode pods
+vllm serve <model> \
+  --kv-transfer-config '{"kv_connector":"NixlConnector","kv_role":"kv_consumer"}'
 ```
 
-The `kv_role` is `kv_both` for both prefill and decode pods — each pod can both send and receive KV Cache.
+Set `kv_role` to `kv_producer` on prefill pods and `kv_consumer` on decode pods.
 
 For XPU and HPU devices where KV transfer happens via CPU memory, add:
 
 ```bash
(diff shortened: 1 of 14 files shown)


ahg-g:
/assign @robertgshaw2-redhat 

Rob, can you pls take a look

ahg-g:
/assign @robertgshaw2-redhat 

Not sure about the deprecation timeline, can we do this in the next release?

Gregory-Pereira:
I have some questions here, traditionally in llm-d decode can do both types of work, and so I would assume its default role would be kv_both. Also what about cases like using NIXL bidirecitonal transfer? I understand all the examples updated didn't have that in the KV connector config im just trying to nail down the mental model here. But I can work with Nick to make sure this is updated

junuxyz:
@Gregory-Pereira

> I have some questions here, traditionally in llm-d decode can do both types of work, and so I would assume its default role would be kv_both.

To my understanding, a decode worker performing local prefill or recomputation does not make it a NIXL producer (`kv_producer`). As described in llm-d’s [failure handling docs](https://github.com/llm-d/llm-d/blob/main/docs/operations/disaggregation/vllm.md#prefill-instance-failure), the decode worker runs the full request locally, including prefill. Since there is no remote KV handoff occurring here, this shouldn't affect its NIXL role.

> Also what about cases like using NIXL bidirectional transfer?

Think this was previously discussed in the [upstream vLLM RFC](https://github.com/vllm-project/vllm/issues/43807#issuecomment-4564937443). The conclusion was: bidirectional transfer is a secondary capability and does not change the primary P/D roles. Likewise, the [current vLLM documentation](https://github.com/vllm-project/vllm/blob/main/docs/features/nixl_connector_usage.md#configuration) configures bidirectional transfer with kv_producer on P and kv_consumer on D.

tlrmchlsmth (dismissed):

Gregory-Pereira (approved):
/lgtm

Issue #1057, opened 2026-04-02
[Feature]: Make Standalone mode the default for inference-scheduler


ahg-g:
### Feature Area

Inference Scheduling

### Problem Statement

Not all users are necessarily accustomed to everything k8s, while most know how to deploy simple things on k8s, advanced features like Gateway is a friction point. 

### Proposed Solution

We should make [the standalone deployment mode](https://gateway-api-inference-extension.sigs.k8s.io/guides/standalone/) the default for llm-d docs; the gateway-based ones (GKE, Istio etc.) are advanced options.

For production deployments, llm-d should recommend the gateway mode; the inference-scheduler offers this native integration with Gateway which differentiates it compared to other solutions like smg, sglang-router, vllm-router or even dynamo (though they are trying to address that). 

### Willingness to Contribute

Yes, I can submit a PR

robertgshaw2-redhat:
done

PR #1719, opened 2026-06-08
docs: surface agentgateway as a supported proxyType in standalone router guides

### LINKED ISSUE #1713: feat: add agentgateway as a supported proxyType for standalone router mode
## Summary

The `llm-d-router-standalone` chart (in `llm-d/llm-d-router`) already supports agentgateway as a `proxyType` via `router.proxy.proxyType=agentgateway` — confirmed by [@abdullahgh](https://github.com/abdullahgh). The chart has a full agentgateway preset including image, ports, readiness probe, and service configuration.

However, the llm-d **guides and documentation** do not expose or document this option. Every guide, values file, and doc in this repo describes standalone mode as Envoy-only — agentgateway as a `proxyType` is invisible to users following the llm-d guides.

With standalone mode becoming the default deployment path (ref: #1057), users should be able to choose agentgateway as the sidecar proxy without having to discover this from the llm-d-router chart directly.

## Current State (in llm-d repo)

- `guides/recipes/router/base.values.yaml` — only contains Envoy proxy config. No `proxyType` field, no agentgateway option documented
- `guides/recipes/router/README.md` — describes standalone as "deploys the llm-d Router with an **Envoy sidecar**" — no mention of agentgateway
- `docs/getting-started/artifacts.md` — explicitly says "standalone Envoy proxy as sidecar in EPP pod" — no mention of agentgateway option
- `guides/optimized-baseline/README.md` (section 1: Deploy the llm-d Router) — only shows standalone helm install without `proxyType` option
- `proxyType` field: **zero occurrences** across all llm-d guides and docs

## What Needs to Be Done (guides/docs only — chart is already complete)

1. **`guides/recipes/router/README.md`** — Add agentgateway as a documented `proxyType` option alongside Envoy, with a working helm install example using `--set router.proxy.proxyType=agentgateway`
2. **`guides/recipes/router/base.values.yaml`** — Add `proxyType` field and document agentgateway-specific settings so users know it exists
3. **`docs/getting-started/artifacts.md`** — Update standalone chart description from "Envoy proxy as sidecar" to "Envoy or agentgateway proxy as sidecar"
4. **`guides/optimized-baseline/README.md`** — Add agentgateway sidecar as an alternative to Envoy in section 1 (Deploy the llm-d Router)
5. **CI** — Add agentgateway standalone path to nightly E2E workflows (currently only Envoy standalone is exercised)

## Reference

- llm-d-router chart already supports this: [`config/charts/llm-d-router-standalone/values.yaml`](https://github.com/llm-d/llm-d-router/blob/main/config/charts/llm-d-router-standalone/values.yaml)
- llm-d-router chart docs: [`config/charts/README.md`](https://github.com/llm-d/llm-d-router/tree/main/config/charts)
- Initial GAIE implementation reference: [kubernetes-sigs/gateway-api-inference-extension#2905](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2905)
- Standalone as default: [llm-d/llm-d#1057](https://github.com/llm-d/llm-d/issues/1057)

---

Assigning to myself as part of GSoC 2026. Will start on the guides/docs PR once scope is confirmed.

abhay1999:
Closes #1713

## what this does

the llm-d-router-standalone chart already supports agentgateway as a sidecar proxy (`router.proxy.proxyType=agentgateway`) but none of the guides mentioned it, everything just assumed envoy. this PR adds the agentgateway docs and lines up the install commands with the chart's pseudo-service model (no service.name/service.ports needed anymore).

## changes

**guides/recipes/router/README.md**
- new "Standalone with agentgateway" section next to the existing envoy one, with a working helm install example
- NOTE callout for the required flags (`inferencePool.create=false`, `epp.flags.secure-serving=false`)
- added a Prerequisites section for the env vars used across these guides (REPO_ROOT, NAMESPACE, ROUTER_CHART_VERSION, via guides/env.sh)

**guides/recipes/router/base.values.yaml**
- kept the explicit envoy args here instead of switching to `args: []`. tried that first but the chart preset uses `--cpuset-threads` instead of `--concurrency 8`, and since this pod has no cpu limit that ends up spawning threads off the whole node instead of the tuned value from the v0.8.0 benchmark run. so args stayed as is, agentgateway gets its own args from the features overlay below

**guides/recipes/router/features/agentgateway-proxy.values.yaml** (new)
- sets proxyType: agentgateway + the required flags, swaps out the envoy args for agentgateway's

**guides/optimized-baseline/README.md**
- just points to the router recipes guide for agentgateway now instead of duplicating a whole snippet here, didn't want to add another dimension to an already complex guide

**docs/api-reference/artifacts.md**
- updated the standalone chart description to mention agentgateway as an option

## testing

ran `helm template` against both proxyType=envoy and proxyType=agentgateway for the affected guides.

also did a real e2e run on a fresh kind cluster, not just template. ran the exact helm install commands from `guides/recipes/router/README.md` for both paths:
- standalone with envoy: pod came up 2/2 running, sent a real `/v1/completions` request, got a real completion back
- standalone with agentgateway: same, pod 2/2 running, real completion back

only deviations from the literal guide commands were cpu/memory resource overrides (the guide's defaults are 4 cpu / 8gi per component, meant for real clusters, too big for a kind node) and the agentgateway image tag override, since the chart's built-in default tag doesn't exist in the registry. everything else ran exactly as documented.

## related

- closes #1713
- llm-d-router#1600 — agentgateway pseudo-service rework this depends on
- llm-d-router#1926 — companion chart fix so `--set proxyType` actually works
- #1057 — standalone as default


guides/optimized-baseline/README.md
```diff
@@ -99,6 +99,20 @@ helm install ${GUIDE_NAME} \
...
+To use **agentgateway** as the sidecar proxy instead of Envoy, add the following flags. See [router recipes](../recipes/router/README.md) for full details.
+
+```bash
+helm install ${GUIDE_NAME} \
+    oci://ghcr.io/llm-d/charts/llm-d-router-standalone-dev \
+    -f guides/recipes/router/base.values.yaml \
+    -f guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml \
+    --set router.proxy.proxyType=agentgateway \
+    --set router.inferencePool.create=false \
+    --set router.proxy.agentgateway.service.name=<your-model-service-name> \
```
danehans: This helm install slightly differs from the [GAIE standalone docs](https://gateway-api-inference-extension.sigs.k8s.io/guides/standalone/#deploy-endpoint-picker-extension-with-a-sidecar-proxy):
- Agentgateway creates a psuedo service so `--set router.proxy.agentgateway.service.name=<your-model-service-name> \` should not be needed.
- `--set-string inferenceExtension.flags.secure-serving=false \` is set so TLS is not used between EPP and the sidecar proxy (agentgateway).
- Is `--set inferenceExtension.endpointsServer.endpointSelector="app=vllm-qwen3-32b" \` no longer required?

guides/recipes/router/README.md
```diff
@@ -4,17 +4,41 @@ llm-d uses the **llm-d Router** to make intelligent request routing decisions fo
...
+```
+
+### Standalone with agentgateway
+
+agentgateway can be used as the sidecar proxy in place of Envoy. In this mode agentgateway runs alongside the EPP in the same pod and communicates with it over localhost via ext-proc — no Kubernetes Gateway API infrastructure required.
+
+> [!NOTE]
+> When using `proxyType=agentgateway`, set `router.inferencePool.create=false`
+> and provide the model service name via `router.proxy.agentgateway.service.name`.
+> A Kubernetes Service for the model workloads will be created automatically
```
danehans: See https://github.com/llm-d/llm-d/pull/1719/changes#r3376540200. Defining a service is not required since agtw auto-creates the required pseudo service.

guides/recipes/router/README.md
```diff
@@ -4,17 +4,41 @@ llm-d uses the **llm-d Router** to make intelligent request routing decisions fo
...
+
+```bash
+helm install <release-name> \
+  oci://ghcr.io/llm-d/charts/llm-d-router-standalone-dev \
+  -f ${REPO_ROOT}/guides/recipes/router/base.values.yaml \
+  -f ${REPO_ROOT}/guides/recipes/router/features/monitoring.values.yaml \
+  -f ${REPO_ROOT}/guides/<your-guide>/router/<your-guide>.values.yaml \
+  --set router.proxy.proxyType=agentgateway \
+  --set router.inferencePool.create=false \
+  --set router.proxy.agentgateway.service.name=<your-model-service-name> \
```
danehans: Ditto.

guides/optimized-baseline/README.md
abhay1999: Good catches — updated both examples. Removed the `service.name` flag since agentgateway creates the pseudo service automatically, and added `router.epp.flags.secure-serving=false` explicitly for documentation clarity (the chart already sets this automatically for `proxyType=agentgateway` in `_deployment.yaml`, but making it visible in the example documents the requirement).

One thing worth flagging: the current `_validations.tpl` in `llm-d-router` still validates that `service.name` is required when `proxyType=agentgateway` — so the updated example will fail chart validation as-is. Should the `llm-d-router` chart also be updated to adopt the pseudo-service behavior from GAIE#2905 (removing the `service.name` requirement), or is that a separate issue?

guides/recipes/router/README.md
abhay1999: Updated — removed `service.name` and added `router.epp.flags.secure-serving=false` explicitly. Same question applies here about whether the `llm-d-router` chart's `_validations.tpl` needs to be updated to remove the `service.name` requirement.

guides/recipes/router/README.md
abhay1999: Updated here as well.

ahg-g:
Is this still a draft or ready for review?

abhay1999:
> Is this still a draft or ready for review?

Hey @ahg-g , it's currently in draft because there's a blocker in the Helm chart validations. 

Currently, `llm-d-router`'s `_validations.tpl` strictly requires `service.name` to be set when `proxyType=agentgateway`. Because of this, the simplified Helm commands in these guides will fail chart validation.

I have a fix ready locally in the `llm-d-router` repo that defaults `service.name` to the release name and removes the empty validation check. Once I push that and open a PR there, we can mark this docs PR as ready. Let me know if that sounds good and I'll submit the chart PR.

guides/optimized-baseline/README.md
```diff
@@ -99,6 +99,20 @@ helm install ${GUIDE_NAME} \
...
+To use **agentgateway** as the sidecar proxy instead of Envoy, add the following flags. See [router recipes](../recipes/router/README.md) for full details.
+
+```bash
+helm install ${GUIDE_NAME} \
+    oci://ghcr.io/llm-d/charts/llm-d-router-standalone-dev \
+    -f guides/recipes/router/base.values.yaml \
+    -f guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml \
+    --set router.proxy.proxyType=agentgateway \
+    --set router.inferencePool.create=false \
+    --set router.epp.flags.secure-serving=false \
```
ahg-g: @danehans  are those two flags truly needed (disabling pool creation and secure serving)? Ideally we don't diverge much from the default.

guides/optimized-baseline/README.md
```diff
@@ -99,6 +99,20 @@ helm install ${GUIDE_NAME} \
...
 ```
+
+To use **agentgateway** as the sidecar proxy instead of Envoy, add the following flags. See [router recipes](../recipes/router/README.md) for full details.
+
+```bash
+helm install ${GUIDE_NAME} \
+    oci://ghcr.io/llm-d/charts/llm-d-router-standalone-dev \
+    -f guides/recipes/router/base.values.yaml \
+    -f guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml \
+    --set router.proxy.proxyType=agentgateway \
```
ahg-g: If the two flags below are not truly needed, then we can add agentgateway as follows to the above step (instead of having a separate/special one for agentgateway):

```
PROXY=envoy # options envoy|agentgateway
helm install ${GUIDE_NAME} \
    oci://ghcr.io/llm-d/charts/llm-d-router-standalone-dev \
    -f guides/recipes/router/base.values.yaml \
    -f guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml \
     --set router.proxy.proxyType=${PROXY} \
```

guides/optimized-baseline/README.md
danehans: @ahg-g thanks for the review. https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2895 includes these two flags. For this PR, I suggested to @abhay1999 that he compare the agentgateway flags between llm-d-router and GAIE standalone charts for equality. I added `secure-serving=false` to https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2895 because EPP was trying to create a TLS connection to the agentgateway sidecar proxy. I thought it was overkill to use TLS between the two containers. If you disagree, are you okay with us creating an issue and having @abhay1999 or me resolve this as a follow-up task?

`router.inferencePool.create=false`, is because agentgateway does not reconcile InferencePool kinds in standalone mode. https://github.com/agentgateway/agentgateway/issues/1644 tracks this feature upstream.

guides/optimized-baseline/README.md
danehans: @ahg-g Please see https://github.com/llm-d/llm-d/pull/1719#discussion_r3397154573 regarding the two flags.

abhay1999:
Quick update — #1600 just got merged in `llm-d-router`. It reworks the agentgateway standalone config to match GAIE's pseudo-service model, so `service.name`/`service.ports` are gone now and the backend gets picked up automatically from `modelServers.matchLabels`.

@ahg-g, about the two flags you asked on — I checked the merged `_validations.tpl` and both are still hard required when `proxyType=agentgateway`:
- `inferencePool.create=false` — standalone agentgateway only supports service-backed routing right now, chart fails without it
- `epp.flags.secure-serving=false` — also enforced, since agentgateway talks to EPP over plaintext gRPC on localhost

So this matches what @danehans explained earlier, it's not just a doc preference, the chart itself rejects it otherwise. Because of that we can't really merge it into one single `PROXY=envoy|agentgateway` command, since that would force these two settings onto the Envoy path too and change its default behavior.

I'd say let's keep the two separate snippets here, and if we want to revisit relaxing those validations later we can open a follow-up issue for it.

This PR's content already matches the chart post-#1600, so from my side it should be good to mark ready for review unless anyone has more concerns.

guides/recipes/router/README.md
```diff
@@ -4,17 +4,42 @@ llm-d uses the **llm-d Router** to make intelligent request routing decisions fo
...
+
 ```bash
 helm install <release-name> \
   oci://ghcr.io/llm-d/charts/llm-d-router-standalone-dev \
   -f ${REPO_ROOT}/guides/recipes/router/features/monitoring.values.yaml \
   -f ${REPO_ROOT}/guides/<your-guide>/router/<your-guide>.values.yaml \
-  --set provider.name=<gke|istio|none> \
+  -n ${NAMESPACE} \
+  --version ${ROUTER_CHART_VERSION}
```
danehans, line 30: I don;t see these env vars being set in this README.

guides/recipes/router/base.values.yaml
```diff
@@ -26,12 +26,20 @@ router:
...
+    proxyType: envoy
+
     # CLI args for the Envoy sidecar. Override the chart's defaults of
     # `--log-level trace` (very expensive under load) and no `--concurrency`
     # (Envoy spawns ~hardware_concurrency() worker threads, oversubscribing
     # the cgroup's CPU slice). `--concurrency 8` matches the resources.limits
     # equivalent below; trace-level logging would burn a meaningful share of
     # the sidecar's CPU at high QPS.
+    # Note: these args apply to the Envoy preset only. When proxyType=agentgateway,
+    # the agentgateway preset args are used instead.
```
danehans: I rendered the source chart and got the agentgateway image with Envoy args like `--service-node envoy-sidecar` and `-c /etc/envoy/envoy.yaml`. The comment saying these args apply only to Envoy is not true with the current values layering. Move these args to an Envoy-only overlay, or add an agentgateway overlay that replaces them.

guides/optimized-baseline/README.md
```diff
@@ -99,6 +99,20 @@ helm install ${GUIDE_NAME} \
     -f guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml \
     -n ${NAMESPACE} --version ${ROUTER_CHART_VERSION}
 ```
+
+To use **agentgateway** as the sidecar proxy instead of Envoy, add the following flags. See [router recipes](../recipes/router/README.md) for full details.
```
danehans: The optimized-baseline agentgateway snippet fails chart validation because `guides/optimized-baseline/router/optimized-baseline.values.yaml` sets `router.extraServicePorts[0].targetPort: 8081`. Agentgateway requires this targetPort to be omitted, http, or the listener port. The new snippet needs an agentgateway-specific override or values file.

guides/recipes/router/README.md
abhay1999: Added a Prerequisites section showing how to set these — sourcing `guides/env.sh` for REPO_ROOT/ROUTER_CHART_VERSION/etc, and `export NAMESPACE` explicitly since that one isn't set by env.sh.

guides/recipes/router/base.values.yaml
abhay1999: Good catch, you're right that the comment was wrong. Added `features/agentgateway-proxy.values.yaml` that clears `router.proxy.args` and layered it into both agentgateway install snippets. Tested it directly on Kind — without this overlay the agentgateway sidecar actually crash-loops with `error: unexpected argument '--service-node' found`, and with it the pod comes up clean and a request goes through end to end.

guides/optimized-baseline/README.md
abhay1999: Confirmed and fixed. Added `guides/optimized-baseline/router/optimized-baseline-agentgateway.values.yaml` which drops the targetPort so it falls back to the listener port, and wired it into the agentgateway snippet. Reproduced your exact validation error on Kind without it, confirmed it's gone with the override in place.

abhay1999:
Pushed the fixes for all three review comments above, plus pulled in #1651's new optimized-baseline scheduling config (token-load-scorer + prefix-cache-affinity-filter) and re-verified both proxy paths still work with it.

Ran a full e2e pass on Kind for both proxy types, model server was llm-d-inference-sim:

- **Envoy**: helm install, pod healthy, sent a real `/v1/completions` request through router → Envoy → EPP → sim, got back a real completion.
- **agentgateway**: same flow, also got a real completion back end to end.

One thing worth flagging separately, not a blocker for this PR: the chart's default agentgateway image tag is `latest-dev`, and that build has a bug where the proxy's data-plane listener never actually binds (pod shows healthy/ready, but the port is never reachable, confirmed by checking the open sockets inside the container directly). Switching to the `latest` tag fixed it completely. This is purely an agentgateway image issue, unrelated to the chart or these docs, but I'll file it upstream separately so it doesn't surprise anyone testing this guide with the default image.

This should be ready for another pass whenever you get a chance.

abhay1999:
Marked this ready for review now. @ahg-g @danehans whenever you have a chance to take another look.

guides/optimized-baseline/README.md
danehans: > Should the llm-d-router chart also be updated to adopt the pseudo-service behavior from GAIE#2905 (removing the service.name requirement), or is that a separate issue?

Yes, the llm-d-router chart should be updated.

guides/recipes/router/README.md
danehans: > Same question applies here about whether the llm-d-router chart's _validations.tpl needs to be updated to remove the service.name requirement.

Same response as https://github.com/llm-d/llm-d/pull/1719#discussion_r3477862180

guides/recipes/router/features/agentgateway-proxy.values.yaml
```diff
@@ -0,0 +1,8 @@
+## Clears the Envoy-specific sidecar args from base.values.yaml.
+## Layer this on top of base.values.yaml when proxyType=agentgateway,
+## since the chart passes router.proxy.args verbatim to whichever
+## sidecar container is deployed.
+
+router:
+  proxy:
+    args: []
```
danehans: `args: []` removes agentgateway’s required `-f /config/config.yaml` args. This overlay needs to preserve/set the agentgateway config args, not clear them to empty.

guides/recipes/router/README.md
```diff
@@ -2,19 +2,54 @@
...
+### Standalone with agentgateway
+
+agentgateway can be used as the sidecar proxy in place of Envoy. In this mode agentgateway runs alongside the EPP in the same pod and communicates with it over localhost via ext-proc — no Kubernetes Gateway API infrastructure required.
+
+> [!NOTE]
+> When using `proxyType=agentgateway`, set `router.inferencePool.create=false`.
+> agentgateway automatically creates a pseudo service for model workloads — no
+> explicit service name is required. TLS is disabled between EPP and agentgateway
+> automatically by the chart, but can be set explicitly via
+> `router.epp.flags.secure-serving=false`.
```
danehans: Instead, it should be stated that `true` is not supported.

guides/optimized-baseline/modelserver/gpu/vllm/gpt-oss/patch-vllm-gpt120b.yaml
```diff
@@ -14,6 +14,7 @@ spec:
             - "openai/gpt-oss-120b"
             - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
             - "--tensor-parallel-size=1"
+            - "--gpu-memory-utilization=0.95"
```
danehans: Why is this change needed to fix this issue?

danehans:
@abhay1999 PTAL at https://github.com/abhay1999/llm-d/pull/1 to merge the final bits needed to get this PR to the finish line.

guides/optimized-baseline/modelserver/gpu/vllm/gpt-oss/patch-vllm-gpt120b.yaml
abhay1999: Fair catch — that came along incidentally when I scoped the #1651 cherry-pick by directory instead of by content, it's not actually needed for the agentgateway work here. Dropped it, this PR no longer touches the gpt-oss patch file. It'll land properly whenever #1651 itself merges.

guides/recipes/router/README.md
abhay1999: Reworded — now states secure-serving=true is not supported with proxyType=agentgateway, instead of implying it as just an alternative setting.

guides/recipes/router/features/agentgateway-proxy.values.yaml
abhay1999: Merged your fix from abhay1999/llm-d#1 — args now set to ["-f", "/config/config.yaml"] instead of being cleared, and the chart override is in place for the dev chart until v0.10.0. Thanks for catching that, my own Kind testing earlier missed it.

guides/optimized-baseline/README.md
howardjohn: I think there must be some confusion on the InferencePool. Possibly on my end.

Agentgateway in standalone mode doesn't reconcile _anything_ - Service, Inference Pool, etc - it has no awareness. neither does envoy.

I am pretty sure we can support the InferencePool and can also support TLS if we want and agree with @ahg-g that we should keep the same flags

guides/recipes/router/base.values.yaml
```diff
@@ -21,12 +21,22 @@ router:
     protocol: http # http, grpc
 
   proxy:
+    # Sidecar proxy type. Options: envoy (default) or agentgateway.
+    # To use agentgateway instead of Envoy, override in your guide-specific
+    # values file or pass --set router.proxy.proxyType=agentgateway on the
+    # helm install command. See guides/recipes/router/README.md for details.
+    proxyType: envoy
```
howardjohn, line 28: Should we default to none, like the Gateway, and have users select it?

guides/recipes/router/base.values.yaml
```diff
@@ -21,12 +21,22 @@ router:
...
+
     # CLI args for the Envoy sidecar. Override the chart's defaults of
     # `--log-level trace` (very expensive under load) and no `--concurrency`
     # (Envoy spawns ~hardware_concurrency() worker threads, oversubscribing
     # the cgroup's CPU slice). `--concurrency 8` matches the resources.limits
     # equivalent below; trace-level logging would burn a meaningful share of
     # the sidecar's CPU at high QPS.
+    # Note: the chart passes these args verbatim to whichever sidecar container
```
howardjohn: I would make this a default in the envoy template, and make the default here `args: []` which would only be used if they are set explicitly.

guides/recipes/router/base.values.yaml
abhay1999: That makes sense as a direction — defaulting to `none` and making users opt in is cleaner than implying envoy is always the right choice. That said, changing the default proxyType feels like a separate change from what this PR is doing (docs only). Happy to open a follow-up issue to track it so it doesn't get lost?

guides/recipes/router/base.values.yaml
abhay1999: Agreed, that's architecturally cleaner — Envoy-specific args shouldn't live in the base values. Moving them into an `envoy-proxy.values.yaml` overlay and defaulting `args: []` here would make both proxy paths symmetric. Same as above — would it make sense to track this as a follow-up so this PR stays focused on the docs?

howardjohn: With my proposal you would not need any `envoy-proxy.values.yaml` or `agentgateway.values.yaml`.

The goal IMO should be I can do `helm install --set proxyType=envoy|agentgateway` alone and it works

abhay1999: Makes sense — cleaner for users too. For that to work the chart would need to inject proxy-specific args based on `proxyType` internally.

@danehans  — is that feasible in the standalone chart? Happy to update both chart and docs if so.

danehans: > cleaner for users too. For that to work the chart would need to inject proxy-specific args based on proxyType internally.

+1

guides/recipes/router/base.values.yaml
danehans: > --set router.inferencePool.create=false

@abhay1999 can you test setting this flag to `true` with `proxyType: agentgateway`? If inference routing works as expected, then this is an EPP-only setting.

abhay1999: Tested this on Kind with llm-d-inference-sim — routing works fine with inferencePool.create=true and proxyType=agentgateway. Pod came up healthy (2/2), sent a real /v1/completions request through agentgateway → EPP → sim and got a valid completion back. So yeah, it's an EPP-only setting, not agentgateway-specific. I'll remove it from the agentgateway section and update the docs accordingly.

abhay1999:
Updated the PR with a few things from the latest review round:

**`base.values.yaml` — `args: []`**

Changed `proxy.args` from a hardcoded list of Envoy flags to `args: []`. The standalone chart has a built-in preset system (`presets.envoy` / `presets.agentgateway` in `values.yaml`) that injects the right args for each `proxyType`. The old hardcoded args in `base.values.yaml` were silently overriding that preset, which meant `--set router.proxy.proxyType=agentgateway` deployed with Envoy args. With `args: []` the preset takes effect automatically. Users who want to customise the args can still supply explicit values — those override the preset.

**Companion chart fix — llm-d-router#1926**

The preset restore logic in `_helpers.tpl` also needed a fix (`mergeOverwrite` mutates its dst in-place, so preset args had to be captured before the merge). That's in the companion chart PR. The two PRs together make `helm install --set proxyType=envoy|agentgateway` work without any overlay file beyond `base.values.yaml`.

Tested both proxy types end-to-end on Kind — curl to `/v1/completions` returns valid responses for each.

danehans:
Requires https://github.com/llm-d/llm-d-router/pull/1926

abhay1999:
rebased on main now that llm-d-router#1926 is merged, so the blocker from the last comment is cleared. both proxy paths still work end to end on kind

@ahg-g should be good for another pass whenever you get a chance

ahg-g:
Sorry i missed this, can you pls resolve the conflicts?

abhay1999:
resolved, rebased on main and fixed the conflicts. two spots where an older commit was trying to re-add the temporary -dev chart override (llm-d-router-standalone-dev), didnt need that anymore since main already points at the stable chart and it supports proxyType=agentgateway directly now. just removed those bits and fixed up the chart names/description in artifacts.md to match. mergeable now

guides/recipes/router/base.values.yaml
```diff
@@ -21,26 +21,18 @@ router:
...
-      - "--log-level"
-      - "warn"
-      - "--concurrency"
-      - "8"
-      - "--drain-strategy"
-      - "immediate"
-      - "--drain-time-s"
-      - "60"
-      - "-c"
-      - "/etc/envoy/envoy.yaml"
```
ahg-g, line 42: There is one difference which is here we set `--concurrency` flag while in the chart it defaults to `--cpuset-threads` ; I believe this was changed based on concrete experiments, please update the default in the main chart [1] before changing the value here.

[1] https://github.com/llm-d/llm-d-router/blob/main/config/charts/llm-d-router-standalone/values.yaml#L256C13-L256C31

guides/optimized-baseline/router/optimized-baseline-agentgateway.values.yaml
```diff
@@ -0,0 +1,13 @@
+## agentgateway override for the optimized-baseline guide.
+## Layer on top of: optimized-baseline.values.yaml
+##
+## optimized-baseline.values.yaml sets extraServicePorts[0].targetPort: 8081,
+## but agentgateway requires that targetPort be omitted, "http", or equal to
+## the listener port (.port, 80 here). This override removes targetPort so
+## the chart's agentgateway validation passes.
+
+router:
```
ahg-g: We have a lot of dimensions to cover in the guides, accelerator types and modes and models. We should avoid increasing that with proxy-specific overlays.

guides/optimized-baseline/README.md
```diff
@@ -108,6 +108,21 @@ helm install ${GUIDE_NAME} \
 > For **TensorRT-LLM** (`trtllm-serve`), replace the last `-f` flag with the TensorRT-LLM
 > values file: `-f ${REPO_ROOT}/guides/${GUIDE_NAME}/router/${GUIDE_NAME}-trtllm.values.yaml`
 
+To use **agentgateway** as the sidecar proxy instead of Envoy, add the following flags. See [router recipes](../recipes/router/README.md) for full details.
```
ahg-g: The guides are complex, it is hard to fit every option on every guide. I don't think we should have this here, instead point to the guides in the router recipes only pls

ahg-g (commented):
This PR mistakenly changes the file name guides/optimized-baseline/modelserver/gpu/vllm/gpt-oss/patch-vllm-gpt120b.yaml

abhay1999:
fixed — renamed patch-vllm.yaml to patch-vllm-gpt120b.yaml to match the path referenced in kustomization.yaml. that rename got lost somewhere during the earlier rebase. should be clean now.

guides/optimized-baseline/router/optimized-baseline-agentgateway.values.yaml
ahg-g: not addressed yet

guides/optimized-baseline/README.md
ahg-g: not addressed

guides/recipes/router/base.values.yaml
ahg-g: not addressed

abhay1999:
@ahg-g  went through these:

- kept the envoy args in base.values.yaml instead of args: []. the chart preset uses --cpuset-threads and the pod has no cpu limit, so that ends up spawning threads off the whole node count. would've dropped --concurrency 8 which came from the v0.8.0 benchmark run, so i left the args alone. agentgateway gets its args from the features overlay instead

- dropped optimized-baseline-agentgateway.values.yaml

- took the agentgateway snippet out of the optimized-baseline guide, it just points to router recipes now

tested both with helm template against the standalone chart on main. envoy comes out the same as main, agentgateway gets -f /config/config.yaml and no envoy flags in there

abhay1999:
@ahg-g did the real e2e test now, not just helm template.

stood up a fresh kind cluster, ran the exact helm install commands from `guides/recipes/router/README.md` for both paths.

- standalone with envoy: pod came up 2/2 running, sent a real `/v1/completions` request through it, got a real completion back
- standalone with agentgateway: same, pod 2/2 running, real completion back

only thing i overrode was cpu/memory resources (the guide's defaults are 4 cpu / 8gi per component, meant for real clusters, too big for a kind node) and the agentgateway image tag, since the chart's built-in default tag doesn't exist in the registry, same issue i hit before on the llm-d-benchmark work. everything else ran exactly as documented in the guide.

both proxy paths work end to end, not just render clean yaml. description's testing section is updated too.

can you take another look when you get a chance?

howardjohn (approved):
would love to get rid of agentgateway-proxy.values.yaml  but this works

danehans:
> would love to get rid of agentgateway-proxy.values.yaml but this works

@abhay1999 can you create a tracker issue for this?

@ahg-g PTAL when you have a moment.

abhay1999:
opened llm-d/llm-d-router#2343 to track getting rid of the agentgateway overlay file, linked the concurrency/preset tension from earlier in this thread as the likely fix direction.

@ahg-g whenever you get a chance to take a look

ahg-g:
/approve

Thanks for addressing all the comments!

PR #1706, opened 2026-06-05
Fix precise guide config

liu-cong:
no_hit_lru should point to the precise producer, not the default approximate producer

Also make the REPO_ROOT pattern consistent across guides

PR #1723, opened 2026-06-08
Fix optimized baseline path in multimodal guide

rahulgurnani:
The path for optimized baseline is broken in multimodal guide.


rahulgurnani:
cc @liu-cong  @capri-xiyue

rahulgurnani:
seems the change https://github.com/llm-d/llm-d/pull/1706 fixed this guide as well.

PR #1725, opened 2026-06-08
Temporarily default all (benchmark) workloads to `sanity_random.yaml`

maugustosilva:
Also rename two workflows missed on a previous PR


diegocastanibm (approved):
LGTM

PR #1740, opened 2026-06-09
docs(agentic-serving): add Nemotron-3-Ultra on H200 deployment; make accelerator the dimension

Gregory-Pereira:
cc @tessapham


vMaroon:
Can we align with https://github.com/llm-d/llm-d/pull/1743?

guides/agentic/modelserver/gpu/vllm/base/patch-decode.yaml
```diff
@@ -0,0 +1,108 @@
...
+        - name: modelserver
+          command: ["vllm", "serve"]
+          args:
+            - "MiniMaxAI/MiniMax-M2.5"
+            - "--served-model-name=minimax-m2.5"
+            - "--trust-remote-code"
+            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
+            - "--tensor-parallel-size=8"
+            - "--enable-expert-parallel"
+            - "--max-num-seqs=512"
```
robertgshaw2-redhat: remove

guides/agentic/modelserver/gpu/vllm/base/patch-decode.yaml
```diff
@@ -0,0 +1,108 @@
...
+            - "--max-num-batched-tokens=32768"
```
robertgshaw2-redhat: definitely remove

guides/agentic/modelserver/gpu/vllm/base/patch-decode.yaml
```diff
@@ -0,0 +1,108 @@
...
+            - "--kv-cache-dtype=fp8"
+            - "--moe-backend=triton"
+            - "--attention-backend=FLASHINFER"
+            - "--enable-flashinfer-autotune"
+            - "--compilation-config"
+            - '{"mode":3,"cudagraph_mode":"PIECEWISE","pass_config":{"fuse_minimax_qk_norm":true}}'
+            - "--enable-auto-tool-choice"
+            - "--tool-call-parser=minimax_m2"
+            - "--reasoning-parser=minimax_m2"
+            - "--enable-prefix-caching"
```
robertgshaw2-redhat: remove

guides/agentic/modelserver/gpu/vllm/base/patch-decode.yaml
```diff
@@ -0,0 +1,108 @@
...
+          env:
+            # prevents OTel trace export errors when no collector is deployed
+            # TO FIX: tracing should be off by default and be configurable
+            - name: OTEL_SDK_DISABLED
+              value: "true"
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          args:
+            - "MiniMaxAI/MiniMax-M2.5"
```
robertgshaw2-redhat: old model

guides/agentic/modelserver/gpu/vllm/base/patch-prefill.yaml
```diff
@@ -0,0 +1,94 @@
...
+            - '{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_load_failure_policy":"fail"}'
+          env:
+            - name: HF_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: llm-d-hf-token
+                  key: HF_TOKEN
+            - name: PYTHONNOUSERSITE
+              value: "1"
+            - name: VLLM_USE_DEEP_GEMM
```
robertgshaw2-redhat: not needed

guides/agentic-serving/modelserver/gpu/vllm/nemotron-3-ultra/claude.env
liu-cong, line 1: remove?

guides/agentic/router/agentic.values.yaml
```diff
@@ -0,0 +1,94 @@
...
+        - type: prefix-cache-scorer
+          name: gpu-prefix-cache-scorer
+          parameters:
+            prefixMatchInfoProducerName: gpu-prefix-cache-producer
+        - type: prefix-cache-scorer
+          name: cpu-prefix-cache-scorer
+          parameters:
+            prefixMatchInfoProducerName: cpu-prefix-cache-producer
+        requestHandler:
+          parsers:
```
liu-cong: note there is a bug in the helm chart:  https://github.com/llm-d/llm-d-router/pull/1614

No need to specify the parser because by default we have both ant and openai parsers enabled.

guides/agentic-serving/router/agentic-serving-gpu.values.yaml
liu-cong, line 1: Can we directly use the the router config in the cpu offloading guide? It's basically the same.

guides/agentic/modelserver/gpu/vllm/base/patch-decode.yaml
```diff
@@ -0,0 +1,101 @@
...
+          args:
+            - "RedHatAI/NVIDIA-Nemotron-3-Ultra-550B-A55B-FP8-block"
+            - "--trust-remote-code"
+            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
+            - "--tensor-parallel-size=8"
+            - "--enable-expert-parallel"
+            - "--max-model-len=262144"
+            - "--max-num-seqs=32"
+            - "--max-num-batched-tokens=32768"
+            - "--enable-prefix-caching"
```
liu-cong: this is default, and probably a few others as well. Can we remove ?

guides/agentic-serving/modelserver/gpu/vllm/nemotron-3-ultra/claude.env
Gregory-Pereira: we need this - this is how we point claude code at the llm-d instance with a port forward. Theres nothing actually private here so im not leaking an env

guides/agentic-serving/router/agentic-serving-gpu.values.yaml
Gregory-Pereira: The biggest different between the two is PD configurations. Do you have CPU offloading with PD set of values I should use?

guides/agentic/router/agentic.values.yaml
Gregory-Pereira: I am not sure I understand, my current understanding is that because this is a string, you cannot inherit values from the base chart, if we want to override anything in the EPP config, we need to provide all the values. Please correct me if I am wrong

Gregory-Pereira: My mistake, I now see: https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/config/loader/defaults.go#L250-L256. Removing

guides/agentic-serving/benchmark-templates/agentic-serving-nemotron-3-ultra.yaml
```diff
@@ -0,0 +1,63 @@
...
+  parallelism: 1
+  wait_timeout: 6000
+  image: ghcr.io/llm-d/llm-d-benchmark:v0.6.0
+
+env:
+  - name: RAYON_NUM_THREADS
+    value: "4"
+
+workload:
+  agentic_shared_prefix:
```
achandrasekar, line 46: Can we use the same codegen benchmark template instead of a fixed shared-prefix?

guides/agentic-serving/README.md
```diff
@@ -42,6 +42,30 @@ this directory as sub-guides.
 
 ## Choosing an Option
 
+### Nemotron 3 Ultra (8 H200 Nodes, TP=8 + EP, PD Dissagregation, 1.6TB CPU offloading)
```
achandrasekar: Would be good to rebase this on top of the TPU guide and follow a similar format.

vMaroon: Updated, please take a look (cc @seanhorgan @Gregory-Pereira)

guides/agentic-serving/benchmark-templates/agentic-serving-nemotron-3-ultra.yaml
vMaroon: We should follow up with this, although I think we probably should avoid the same exact "traces source" of statistics. In order to take accelerator comparison confusions out of the picture.

vMaroon: I think as an additional followup, the two sub-guides should outline which traces they mimic.

Gregory-Pereira: +1 if we had more time before release I would agree but I think we should get something in at this point

seanhorgan: Can we plan to refresh this with the codegen benchmark right after the release?

vMaroon: I think we can merge and update on a best effort basis. Hopefully we can get something before the cut.

vMaroon (approved):

PR #1748, opened 2026-06-09
Add `consolidate-status` CI workflows to all tested variations of `optimized-baseline`

maugustosilva:
Update `optimized-baseline` guides with status badges.

Additionally, remove the **nightly** status badges from all guides.


diegocastanibm (approved):
it should be fine

PR #1727, opened 2026-06-08
Add agentic code generation workload well-lit path 

achandrasekar:
This change adds a new well-lit path combining different llm-d features and configurations focussing on agentic workloads with agentic code generation as the primary example. Note that this is only the first step with initial configurations. More tuning and improvements to follow.

Fixes https://github.com/llm-d/llm-d/issues/1558

PR #1761, opened 2026-06-10
guides: add HF token step to predicted-latency README, bump benchmark image

kaushikmitr:
## What

1. **predicted-latency-routing guide** — add the missing `llm-d-hf-token` secret creation step to the Prerequisites, mirroring the [optimized-baseline guide](../optimized-baseline/README.md). Without it, the reused optimized-baseline model server pods fail with `CreateContainerConfigError: secret "llm-d-hf-token" not found`.
2. **optimized-baseline benchmark template** — pin the harness image to the current nightly digest (`ghcr.io/llm-d/llm-d-benchmark:nightly@sha256:c5332b8...`).

## Why

The predicted-latency-routing guide reuses the optimized-baseline model server manifests (which mount `HF_TOKEN` from the `llm-d-hf-token` secret) but never told the operator to create that secret, so a fresh install gets stuck pulling models.

## Test

- Verified the secret step matches the optimized-baseline guide's working step.
- Verified the pinned digest resolves in ghcr and that the `nightly` tag currently points to it.

### DIFF
--- a/guides/optimized-baseline/benchmark-templates/guide.yaml
+++ b/guides/optimized-baseline/benchmark-templates/guide.yaml
@@ -18,7 +18,7 @@ harness:
   parallelism: 1                  # Number of parallel workload launcher pods to create.
   wait_timeout: 6000              # Time (in seconds) to wait for workload launcher pod to complete before terminating.
                                   # Set to 0 to disable timeout.
-  image: ghcr.io/llm-d/llm-d-benchmark:v0.5.2
+  image: ghcr.io/llm-d/llm-d-benchmark:v0.6.8.1
   # dataset_url: &dataset_url none
 
 env:
--- a/guides/predicted-latency-routing/README.md
+++ b/guides/predicted-latency-routing/README.md
@@ -56,6 +56,17 @@ Skip it when your pool is **heterogeneous** — mixed GPU types, model variants,
     kubectl create namespace ${NAMESPACE}
   ```
 
+- [Create the `llm-d-hf-token` secret in your target namespace with the key `HF_TOKEN` matching a valid HuggingFace token](../../helpers/hf-token.md) to pull models.
+<!-- llm-d-cicd:skip start -->
+  ```bash
+  export HF_TOKEN="your-huggingface-token"  # replace with a valid HuggingFace token
+  kubectl create secret generic llm-d-hf-token \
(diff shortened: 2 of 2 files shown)


guides/optimized-baseline/benchmark-templates/guide.yaml
```diff
@@ -18,7 +18,7 @@ harness:
   parallelism: 1                  # Number of parallel workload launcher pods to create.
   wait_timeout: 6000              # Time (in seconds) to wait for workload launcher pod to complete before terminating.
                                   # Set to 0 to disable timeout.
-  image: ghcr.io/llm-d/llm-d-benchmark:v0.5.2
+  image: ghcr.io/llm-d/llm-d-benchmark:nightly@sha256:c5332b81ac3de0ac09273e277abd649b9cddc65b75ec3d6259aa1847868e147e
```
ahg-g: this looks odd to do for a single guide, what is the reason we need to do this?

kaushikmitr: also need it here: https://github.com/llm-d/llm-d/pull/1727 Basically the latest released image of llm-d-benchmark points to an older version of inference-perf that does not support conversation-replay. Synced with @achandrasekar on this. Once released need to update the version here. Added a TODO

achandrasekar: @kaushikmitr Might make sense to add a comment there. We plan to switch it over to a released version once it is available. edit: Just noticed you added that. Thanks!

@maugustosilva should we do a patch version for now to get some of the agentic benchmarking capabilities? Or once the benchmark release is ready before we cut the llm-d release?

maugustosilva: We are in the process of updating **all** to refer to the stable version of `llm-d-benchmark`, starting with `optimized-baseline`. PR here https://github.com/llm-d/llm-d/pull/1760

maugustosilva: @kaushikmitr For now, please just pin to `v0.6.8.1`... when the aforementioned PR is finally merged, this will be no longer a problem.

kaushikmitr: done

maugustosilva (approved):

PR #1771, opened 2026-06-11
Use generic proxy terminology in router docs

danehans:
- Replace Envoy-specific proxy wording with generic proxy terminology.
- Normalize agentgateway casing in router docs.
- Update the standalone architecture diagram label to Proxy container.


ahg-g:
/approve

Thanks!

PR #1614, opened 2026-05-26
Encode disaggregation guide

roytman:
Add a guide for Encoder disaggregation

This guide extends the existing Prefill/Decode (P/D) disaggregation documentation by adding two new deployment scenarios: E/P/D and E/PD.

PR #1775, opened 2026-06-11
guides: add multimodal umbrella and precise-affinity flavor

### LINKED ISSUE #1531: [Router] Create Multi-Modal Guide
SUMMARY:
- [ ] create multimodal guide for cache affinity routing 
- [ ] [stretch] epd guide

hexfusion:
**What type of PR is this?**

/kind documentation

Adds the multimodal cache-aware routing guide (precise-affinity flavor).

- `guides/multimodal-serving/README.md`: umbrella entry point at an architectural altitude: aggregated multimodal serving (these guides) vs encode disaggregation (#1614). Links flat to the flavor sub-guides; the scheduling-flavor comparison lives in the flavor guides, not the umbrella.
- `guides/multimodal-serving/precise-affinity/`: the precise-affinity flavor, a sibling to the existing `optimized-baseline/`. Deterministic per-pod multimodal cache affinity via `token-producer` + `mm-embeddings-cache-producer` + `mm-embeddings-cache-scorer`, balanced by the load-aware scorers. The producer is self-contained (framework pod list + per-request affinity tracking); no endpoint-notification or data-layer wiring is required.

Topology: model server `Qwen/Qwen3-VL-32B-Instruct`, vLLM `v0.23.0`, 2 decode replicas (TP=2, 4 GPUs); render runs as a standalone CPU Service (`render/render.yaml`) that the `token-producer` calls, so it scales independently of the EPP.

**Benchmark**: the Benchmarking Report includes a representative decode-bound run (2 decode pods, 512-token outputs, render Service). Precise affinity matches a stock round-robin Kubernetes Service on throughput and holds ~30% lower TTFT p90 under saturation; render-replica scaling (knee 9/18/70 QPS across 1/2/8 replicas) is included as sizing guidance.

Builds on the existing `guides/multimodal-serving/optimized-baseline/` flavor; the optimized-baseline-vs-precise-affinity contrast lives in each flavor guide.

**Which issue(s) this PR fixes**:

Fixes #1531

**Release note**:

```release-note
NONE
```


guides/multimodal/precise-affinity/README.md
```diff
@@ -0,0 +1,261 @@
...
+}'
+
+curl -si -X POST http://${IP}/v1/chat/completions -H 'Content-Type: application/json' -d "$REQ" | grep -i x-inference-pod
+curl -si -X POST http://${IP}/v1/chat/completions -H 'Content-Type: application/json' -d "$REQ" | grep -i x-inference-pod
+```
+
+The first request picks a pod at random (cold cache, all endpoints tie at score 0). The second request, with the same image URL, should route to the same pod (score 1.0 on that pod, 0 elsewhere).
+
+### 3. Inspect the Cache-Hit Metric
+
```
rahulgurnani: We typically have benchmark with all our guides. It would make sense to checkin some kind of benchmark here. Thanks!

capri-xiyue:
Do you think it will be more useful to have encode cache and prefix cache plugin together and have a dataset and benchmark show the two plugins performance better than just using one plugin?

guides/multimodal/precise-affinity/README.md
hexfusion: yep in progress

guides/multimodal/README.md
```diff
@@ -0,0 +1,37 @@
+# Multimodal Guides
+
+llm-d serves multimodal models (vision, audio, video) on the same routing fabric as text-only models, with multimodal-aware scoring layered on top of the prefix-cache and load-aware foundation.
+
+Repeated multimodal content is a major reuse opportunity. A single 1080p image expands into roughly 1500 vision tokens before prefill; an audio clip of a few seconds expands into similar magnitudes. Routing a second request carrying the same image to the pod that already encoded it avoids redundant render and prefill, directly reducing time to first token (TTFT) and improving GPU utilization.
+
+The guides below cover two flavors of multimodal-aware routing. Both are workload-first deployments; pick the one whose tradeoffs match your traffic.
+
+## Choose a flavor
```
vMaroon: I think the scheduling flavors should be hidden inside the internal guides, and the higher level one should focus on different architectures, e.g., a disaggregated deployment and an aggregated one

hexfusion:
@capri-xiyue @vMaroon updated please take a look

capri-xiyue:
> @capri-xiyue @vMaroon updated please take a look

I didn't see the a lot of performance gain with the benchmark results.

I'm wondering if we can't find a real use case for encode cache for multimodal in aggregated everything (EPD) mode, we don't have to include it in the EPD multimodal guide.  As in real world, I think prefix cache aware plugin can handle most multimodal scenariosWe can just include encode cache aware plugin in e-disaggregation guide. What do you think of it?

hexfusion:
> > @capri-xiyue @vMaroon updated please take a look
> 
> I didn't see the a lot of performance gain with the benchmark results.
> 
> I'm wondering if we can't find a real use case for encode cache for multimodal in aggregated everything (EPD) mode, we don't have to include it in the EPD multimodal guide. As in real world, I think prefix cache aware plugin can handle most multimodal scenariosWe can just include encode cache aware plugin in e-disaggregation guide. What do you think of it?

30% TTFT improvement was observed. but as you said the EPD will show better. my understanding is that effort is blocked by the issue you raised.

ref. https://github.com/llm-d/llm-d-router/issues/1615

hexfusion:
> I'm wondering if we can't find a real use case for encode cache for multimodal in aggregated everything (EPD) mode

going to brainstorm on this

capri-xiyue:
> 30% TTFT improvement was observed. but as you said the EPD will show better. my understanding is that effort is blocked by the issue you raised.
> 
> ref. [llm-d/llm-d-router#1615](https://github.com/llm-d/llm-d-router/issues/1615)

I don't think it makes sense to have a guide that include encode cache only and exclude prefix cache in epd case. Basically I think we need to find a real use case that encode cache and prefix cache combined have better performance than prefix cache only. If prefix cache can handle cases we know in epd mode, then we just add prefix cache in epd mode.

capri-xiyue:
> > > @capri-xiyue @vMaroon updated please take a look
> > 
> > 
> > I didn't see the a lot of performance gain with the benchmark results.
> > I'm wondering if we can't find a real use case for encode cache for multimodal in aggregated everything (EPD) mode, we don't have to include it in the EPD multimodal guide. As in real world, I think prefix cache aware plugin can handle most multimodal scenariosWe can just include encode cache aware plugin in e-disaggregation guide. What do you think of it?
> 
> 30% TTFT improvement was observed. but as you said the EPD will show better. my understanding is that effort is blocked by the issue you raised.
> 
> ref. [llm-d/llm-d-router#1615](https://github.com/llm-d/llm-d-router/issues/1615)

I think for this release, we want to have e-disaggregation as a experimental guide. I think once https://github.com/llm-d/llm-d/pull/1614, you can include encode cache there for e phrase. As it's an experimental guide, I think currently leave benchmark blank and wait for next release to have the full benchmark should be fine?

vMaroon:
embeddings-cache affinity would be useful when hitting the same images in different prefixes. That's the kind of benchmark that needs to be checked. I think we should have both, for this case. I agree we should catch it.

This would then be especially useful when we add semantic affinity through the hashing logic for use-cases of doc/image processing, but these may be niches.

> I don't think it makes sense to have a guide that include encode cache only and exclude prefix cache

Agreed, I don't think any guide we recommend should exclude prefix-cache affinity unless absolutely justified.

ahg-g:
Is this expected to land today/tomorrow or should we push it to the next release?

hexfusion:
lets go ahead and push to 0.9 this is more of an experimental guide as per @capri-xiyue

capri-xiyue:
Hi @hexfusion Is there any update about this PR? v0.9.0 release will be cut within one week (July 30th 2026)

hexfusion:
closing the for now until I can reprio the effort

PR #1779, opened 2026-06-11
FMA guide

diegocastanibm:
## Summary

- Add an end-to-end Fast Model Actuation (FMA) guide integrating FMA with the llm-d router
- Author it in the **structured-guides format** from #1988 (machine-readable `guide.yaml` + README rendered/validated from it, kustomize manifests)
- Demonstrate inference through the router and the sleep/wake lifecycle to showcase FMA's speed benefit
- Verified end-to-end on a real OpenShift GPU cluster (deploy → inference → sleep → wake → teardown)

> [!NOTE]
> **Depends on #1988.** The render/validate scripts (`scripts/guide-{render,check-yaml,check-readme}.py`) and the shared `guides/env.sh` live on that PR. `env.sh` pins `ROUTER_CHART_VERSION=v0.9.0` and the stable `llm-d-router-standalone` chart. This branch rebases on top of #1988 once it merges.

## What this guide covers

1. **Full deployment**: FMA CRDs, controllers (Helm), RBAC, the llm-d router, and the FMA resources (ISC, LauncherConfig, LauncherPopulationPolicy, requesting pods)
2. **Inference verification**: send completions through the llm-d router to the vLLM instances managed by FMA
3. **Sleep/Wake demo**: scale the requesting Deployment to 0 (sleep — vLLM stays resident in GPU memory) and back up (wake — resumes without a cold model reload)
4. **Benchmarking**: `llmdbenchmark` section using the `nop` harness (stands the stack up/validates without driving synthetic load)

## Structured-guide format (#1988)

The guide is no longer hand-maintained prose with inline `kubectl apply -f - <<EOF` heredocs. It now consists of:

- **`guide.yaml`** — the single source of truth for every bash block (`env`, `prerequisites`, `deploy`, `verify`, `benchmark`, `cleanup`).
- **`README.md`** — prose with `<!-- guide:<path> start/end -->` markers; bash blocks are rendered from `guide.yaml` and checked in CI.
- **Kustomize trees** — `rbac/`, `modelserver/` (ISC + LauncherConfig + LPP), `requester/` (the requesting-pods Deployment), so the guide can be stood up with `standup_method: kustomize`.

## Key decisions

- Uses `Qwen/Qwen3-0.6B` (public model, no HF token needed) to reduce prerequisites
- Requesting pods are a **Deployment** (not a bare `ReplicaSet`) for compatibility with autoscalers such as WVA
- ISC enables sleep with `--enable-sleep-mode`; `HF_HOME=/tmp/hf_cache` (with a NOTE recommending a PVC for production/benchmarking so weights and compiled graphs persist)
- RBAC: the guide creates the `fma-node-viewer` ClusterRole and a namespace-scoped launcher pod-reader Role/RoleBinding; the **node-viewer ClusterRoleBinding is created by the Helm chart** via `--set global.nodeViewClusterRole=fma-node-viewer`
- `launcherCount: 1` per GPU node — only launchers bound to requesting pods start vLLM
- Documents the OpenShift-specific need for `runtimeClassName: nvidia`

## Tested on

- OpenShift (OCP) cluster with NVIDIA GPUs — deployed from scratch and validated end-to-end, then torn down
- **FMA v0.6.2**; llm-d router **`llm-d-router-standalone@v0.9.0`** (stable chart)
- Completion request served correctly through the router; **sleep/wake verified**: on scale-to-0 the controller unbinds the requester and the launcher keeps vLLM resident (GPU retained); on scale-up it re-binds and serves again without a model reload

## Files changed

- `guides/fast-model-actuation/guide.yaml` — **new**, machine-readable source of truth
- `guides/fast-model-actuation/README.md` — rewritten with `guide:` markers, rendered from `guide.yaml`
- `guides/fast-model-actuation/rbac/` — **new** kustomize tree (ClusterRole + launcher Role/RoleBinding)
- `guides/fast-model-actuation/modelserver/` — **new** kustomize tree (ISC, LauncherConfig, LauncherPopulationPolicy)
- `guides/fast-model-actuation/requester/` — **new** kustomize tree (requesting-pods Deployment)
- `guides/fast-model-actuation/router/fast-model-actuation.values.yaml` — router Helm values (now used with the stable chart)
- `guides/README.md` — add the FMA guide to the index

## Findings from live testing

- **`--enable-sleep-mode` works without `VLLM_SERVER_DEV_MODE=1`** — the guide's ISC is simpler than FMA's e2e `mkobjs-openshift.sh` and has no gap; the controller drives sleep/wake without vLLM dev mode
- **ClusterRoleBinding name collision** — the chart creates a cluster-scoped `<release-name>-node-view` binding; two FMA deployments in different namespaces sharing the default release name `fma` collide. Worked around in testing with a unique release name; flagged as an upstream fix
- Image tags are pinned to `v0.6.2` directly in the kustomize manifests (kustomize's `images:` transformer is unreliable against CRD-embedded pod templates); bump them alongside `FMA_VERSION`

## Known limitations (documented in guide)

- Wake only avoids a cold start if the new requesting pod is scheduled onto the same node/GPU as the sleeping instance (scheduler-dependent); most valuable in multi-GPU-per-node setups
- ISC `env_vars` does not support `secretKeyRef` — gated models need alternative secret injection
- The FMA Helm chart still does not ship the `fma-node-viewer` ClusterRole or the launcher pod-reader Role/RoleBinding (created by the guide)

## Follow-up work

- Nightly E2E CI workflow (`standup_method: kustomize`, `scenario_dir: guides`, `standup_scenario: fast-model-actuation`)
- `config/scenarios/guides/fast-model-actuation.yaml` in llm-d-benchmark + README status badge


guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,408 @@
+# Fast Model Actuation
+
+## Overview
+
+Fast Model Actuation (FMA) enables rapid model loading and switching for LLM inference on Kubernetes by exploiting vLLM sleep/wake and model swapping. It implements the "dual pods" technique:
```
jimcadden: We need a bit more explanation before "It implements the "dual pods" technique"

Something like:

"FMA uses the following `dual pod` technique to circumvent the 1-to-1 binding of gpus to pods in Kubernetes:"

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,408 @@
...
+
+- **Server-requesting Pods** reserve GPU resources via the Kubernetes scheduler but do not run inference themselves.
+- **Launcher Pods** (server-providing) run vLLM without requesting GPUs. They gain access to GPUs via `CUDA_VISIBLE_DEVICES`, directed by the FMA controller to the specific GPU(s) reserved by the requesting pod.
+- **FMA Controllers** manage the lifecycle: binding requesting pods to launchers, starting vLLM instances, and orchestrating sleep/wake.
+
+When a requesting pod is deleted, the controller puts the corresponding vLLM instance to sleep (model stays in GPU memory). When a new requesting pod arrives and gets assigned to the same GPU, the controller wakes the sleeping instance — resuming in seconds instead of cold-starting from scratch.
```
jimcadden: Before this, something about server-requesting pods are manipulated through replicatsets by higher-level k8s components (e.g., autoscalers) and the FMA controller converts the decisions of the sceduler into actions on the launcher pods and GPUs.

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,408 @@
...
+
+> [!NOTE]
+> Fast wake only occurs if the Kubernetes scheduler and device plugin assign the new requesting pod to the same GPU where the sleeping vLLM instance resides. In single-GPU-per-node configurations this is nearly guaranteed.
```
jimcadden: > In single-GPU-per-node configurations this is nearly guaranteed.

What? That's a bold claim...

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,408 @@
...
+    export GAIE_VERSION=v1.5.0
+    export ROUTER_CHART_VERSION=v0
+    export GUIDE_NAME="fast-model-actuation"
+    export NAMESPACE=llm-d-fast-model-actuation
+    export FMA_VERSION="0.6.0-alpha.13"
+    export FMA_CHART_INSTANCE_NAME="fma"
+    export REPO_ROOT=$(realpath $(git rev-parse --show-toplevel))
+  ```
+
+- Install the Gateway API Inference Extension CRDs:
```
jimcadden, line 76: Are these prereq steps part of another llm-d "Well Lit Path" document that we can reference instead of duplicating. In other words, assume the audience starts with a full llm-d cluster?

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,408 @@
...
+    kubectl apply -f https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/${GAIE_VERSION}/v1-manifests.yaml
+  ```
+
+- Create a target namespace for the installation:
+
+  ```bash
+    kubectl create namespace ${NAMESPACE} --dry-run=client -o yaml | kubectl apply -f -
+  ```
+
+## Installation Instructions
```
jimcadden, line 92: What cluster permissions are needed to run these installation steps?

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,408 @@
...
+              cpu: "4"
+              memory: "8Gi"
+            limits:
+              cpu: "8"
+              memory: "16Gi"
+EOF
+```
+
+> [!NOTE]
+> The launcher pod does **not** request GPU resources. It gains access to GPUs via `CUDA_VISIBLE_DEVICES`, set by the FMA controller to match the GPU reserved by the corresponding requesting pod. The `runtimeClassName: nvidia` is required on platforms (e.g., OpenShift) where GPU driver libraries are injected via the runtime class rather than the device plugin resource request.
```
jimcadden: This is a little vague. The launcher pod _uses_ GPU resources, but it does not request them from the kubernetes scheduler. Technically, it does request them from the CUDA runtime... I think

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,408 @@
...
+**Wait for ready and send another request:**
+
+```bash
+kubectl wait --for=condition=ready pod -l app=fma-requester -n ${NAMESPACE} --timeout=120s
+```
+
+Re-run the inference request from step 2 to confirm the model is serving again.
+
+> [!NOTE]
+> Wake latency depends on the Kubernetes scheduler assigning the new requesting pod to the same GPU where the sleeping vLLM instance resides. If a different GPU is assigned, a new vLLM instance starts from scratch (cold start). In single-GPU-per-node configurations, wake is nearly guaranteed.
```
jimcadden: > In single-GPU-per-node configurations, wake is nearly guaranteed.

In a single GPU-per-node, there's no good reason model server would be put to sleep, therefore no good use of wakeup. Right?

jimcadden (changes requested):
I like the structure and length, the text just needs some work. I did **not** test the commands or yaml

guides/fast-model-actuation/README.md
nicolexin: How does instant model switching work? The guide only talks about resuming.

guides/fast-model-actuation/README.md
nicolexin: When the requesting pod is deleted, I assume the GPU is also gone. How do you guarantee the GPU is not used by other workloads and continue holding the model?

guides/fast-model-actuation/README.md
nicolexin: What if you have multiple nodes in the node pool?

guides/fast-model-actuation/README.md
diegocastanibm: added some around your suggestion

guides/fast-model-actuation/README.md
diegocastanibm: Added paragraph explaining that requesting pods are managed by ReplicaSets/autoscalers and the FMA controller translates scheduler decisions into launcher actions

guides/fast-model-actuation/README.md
diegocastanibm: Replaced `nearly guaranteed` claim (bad term) with a better explanation.

guides/fast-model-actuation/README.md
diegocastanibm: Added intro to Prerequisites noting the guide assumes an existing llm-d cluster and that GAIE CRDs may already be installed

guides/fast-model-actuation/README.md
diegocastanibm: Added minimum permissions statement at the top of Installation Instructions

guides/fast-model-actuation/README.md
diegocastanibm: Clarified that the launcher pod bypasses the Kubernetes scheduler/device plugin but accesses the GPU via the CUDA runtime

guides/fast-model-actuation/README.md
diegocastanibm: Replaced single-GPU-per-node wake claim with a note that sleep/wake is valuable in multi-GPU-per-node configurations

guides/fast-model-actuation/README.md
diegocastanibm: Added explanation of instant model switching

guides/fast-model-actuation/README.md
diegocastanibm: Added explanation of GPU memory retention after requesting pod deletion and how the launcher holds the CUDA context

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,416 @@
...
+```bash
+kubectl apply -n ${NAMESPACE} -f - <<EOF
+apiVersion: fma.llm-d.ai/v1alpha1
+kind: InferenceServerConfig
+metadata:
+  name: fma-isc
+spec:
+  modelServerConfig:
+    port: 8000
+    options: "--model Qwen/Qwen3-0.6B --gpu-memory-utilization 0.9"
```
aavarghese: We are missing `--enable-sleep-mode` here. 
Maybe good to be in sync with `https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/main/test/e2e/mkobjs-openshift.sh`?

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,416 @@
...
+  ```
+
+- Set the following environment variables:
+
+  ```bash
+    export GAIE_VERSION=v1.5.0
+    export ROUTER_CHART_VERSION=v0
+    export GUIDE_NAME="fast-model-actuation"
+    export NAMESPACE=llm-d-fast-model-actuation
+    export FMA_VERSION="0.6.0-alpha.13"
```
aavarghese: Can we upgrade to FMA's latest which is currently "0.6.1"?

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,416 @@
...
+  ```bash
+    export branch="main" # branch, tag, or commit hash
+    git clone https://github.com/llm-d/llm-d.git && cd llm-d && git checkout ${branch}
+  ```
+
+- Set the following environment variables:
+
+  ```bash
+    export GAIE_VERSION=v1.5.0
+    export ROUTER_CHART_VERSION=v0
```
aavarghese: Benchmarking I believe uses `v0.9.0` for ROUTER

aavarghese:
Benchmarking parses the guide README using these prefixes: https://github.com/llm-d/llm-d-benchmark/blob/f181dcbee1d3beae624e770de679cb3dbd3e2259/llmdbenchmark/kustomize/readme_parser.py#L142-L153
So some of the command steps in the FMA guide may be ignored including block lines after the kubectl apply. We can modify this for a benchmarking deployable PoC step by step...

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,416 @@
...
+metadata:
+  name: fma-isc
+spec:
+  modelServerConfig:
+    port: 8000
+    options: "--model Qwen/Qwen3-0.6B --gpu-memory-utilization 0.9"
+    env_vars:
+      VLLM_LOGGING_LEVEL: "INFO"
+      HF_HOME: "/tmp/hf_cache"
+      TRANSFORMERS_CACHE: "/tmp/hf_cache"
```
aavarghese: TRANSFORMERS_CACHE is deprecated

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,416 @@
...
+kind: InferenceServerConfig
+metadata:
+  name: fma-isc
+spec:
+  modelServerConfig:
+    port: 8000
+    options: "--model Qwen/Qwen3-0.6B --gpu-memory-utilization 0.9"
+    env_vars:
+      VLLM_LOGGING_LEVEL: "INFO"
+      HF_HOME: "/tmp/hf_cache"
```
aavarghese: I'm not sure if we want to use /tmp or specify a PVC mount path for the guide so weights and compiled graphs persist across launchers and we're not wasting time recompiling

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,416 @@
...
+> `launcherCount` is **per matching node**. Setting `launcherCount: 1` creates one launcher pod on each node that has `nvidia.com/gpu.present: "true"`. Only launchers that get bound to a requesting pod will actually start a vLLM instance.
+
+#### Requesting Pods
+
+Create the server-requesting pods that reserve GPUs and trigger model loading:
+
+```bash
+kubectl apply -n ${NAMESPACE} -f - <<EOF
+apiVersion: apps/v1
+kind: ReplicaSet
```
aavarghese: In benchnmarking, we have moved completly to using DEPLOYMENTS instead of RS to work with WVA autoscaling. Maybe that's what we should encourage to use in the guide for future proofing.

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,416 @@
...
+kind: ClusterRole
+metadata:
+  name: fma-node-viewer
+rules:
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
aavarghese: We don't need to create this CRB. Helm install for FMA should take care of creating this as long as you pass this `nodeViewClusterRole` value in `helm upgrade --install` in Step 2 above

aavarghese: `--set global.nodeViewClusterRole=vcp-node-viewer`

aavarghese: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/main/docs/e2e-recipe.md#setup

aavarghese:
My understanding for our first round of FMA guide might be something like this:

- Addressing the comments in the PR coz that will fix some issues
- Restructure the guide's CRs into files to work with parser mirroring `guides/optimized-baseline/{modelserver,router}/`
- Add a new nightly flow in llm-d something like `nightly-e2e-fast-model-actuation-ibm-acc-gpu-vllm-x.yaml` with the right params `scenario_dir: guides, standup_method: kustomize,  standup_scenario: fast-model-actuation` using `nop` harness
- Add a new scenario in llm-d-benchmarking like `config/scenarios/guides/fast-model-actuation.yaml` similar to `https://github.com/llm-d/llm-d-benchmark/blob/main/config/scenarios/guides/optimized-baseline.yaml`
- Add a new badge for FMA


All of this should eventually somehow replace our current FMA nightly workflow in benchmarking that uses Python code to deploy/run workload with manual scaling/undeploy

aavarghese:
Looks like if you add a `Benchmarking` section to our FMA guide, we can run benchmarking load harness plus analysis on FMA as part of the workflow Reference: https://github.com/llm-d/llm-d/tree/main/guides/optimized-baseline#benchmarking

guides/fast-model-actuation/README.md
diegocastanibm: The `llm-d-router-standalone-dev` chart used here only publishes the v0 tag, and all other well-lit-path guides pin `ROUTER_CHART_VERSION=v0`, so I've kept it consistent. I believe v0.9.0 refers to a different router component in the benchmarking harness rather than this standalone router chart? (not sure). Let me know if you meant a specific image/chart and I'll switch it.

guides/fast-model-actuation/README.md
diegocastanibm: Bumped `FMA_VERSION` to 0.6.1

guides/fast-model-actuation/README.md
diegocastanibm: Removed the manually-created `ClusterRoleBinding` and added `--set global.nodeViewClusterRole=fma-node-viewer` to the Helm install in Step 2, so the chart creates the binding. The `fma-node-viewer` ClusterRole itself still has to be created (the chart only creates the binding), so Step 3 now just creates the ClusterRole + the launcher pod-reader Role/RoleBinding. Thanks for the e2e-recipe pointer.

guides/fast-model-actuation/README.md
diegocastanibm: Added `--enable-sleep-mode` to the ISC options, in sync with `mkobjs-openshift.sh`.

guides/fast-model-actuation/README.md
diegocastanibm: Kept `/tmp` to keep the guide self-contained, but added a NOTE explaining that for production or repeated benchmarking you should back `HF_HOME` with a PVC so weights and compiled graphs persist across launchers and aren't recomputed on each start.

guides/fast-model-actuation/README.md
diegocastanibm: Removed it

guides/fast-model-actuation/README.md
diegocastanibm: I thought that WVA, HPA (and I suppose KEDA too) are able to scale RS. However, switched the requesting pods from a ReplicaSet to a Deployment for future-proofing with WVA autoscaling. Updated the manifest, the sleep/wake kubectl scale commands, the cleanup step, and added a short note explaining the choice.

@lionelvillard , can you confirm if the WVA can handle RS?

diegocastanibm:
> My understanding for our first round of FMA guide might be something like this:
> 
>     * Addressing the comments in the PR coz that will fix some issues
> 
>     * Restructure the guide's CRs into files to work with parser mirroring `guides/optimized-baseline/{modelserver,router}/`
> 
>     * Add a new nightly flow in llm-d something like `nightly-e2e-fast-model-actuation-ibm-acc-gpu-vllm-x.yaml` with the right params `scenario_dir: guides, standup_method: kustomize,  standup_scenario: fast-model-actuation` using `nop` harness
> 
>     * Add a new scenario in llm-d-benchmarking like `config/scenarios/guides/fast-model-actuation.yaml` similar to `https://github.com/llm-d/llm-d-benchmark/blob/main/config/scenarios/guides/optimized-baseline.yaml`
> 
>     * Add a new badge for FMA
> 
> 
> All of this should eventually somehow replace our current FMA nightly workflow in benchmarking that uses Python code to deploy/run workload with manual scaling/undeploy

Thanks @aavarghese for laying this out. Just a couple of adjustments based on where we've landed with the team:

1. Addressing the PR comments. Agreed, and done. I've gone through all the review comments in this PR and pushed the fixes.

2. Guide structure. Rather than mirroring guides/optimized-baseline/{modelserver,router}/, we're going to follow the new structured-guides format that Tyler is introducing in #1988 (machine-readable guide.yaml + README kept in sync via the checker/renderer scripts). That's becoming the standard schema for all well-lit paths, so it makes sense to conform to it directly instead of the older layout. #1988 is expected to merge between the 15th and the 25th, so I'll rebase the CR restructuring on top of it once it lands.

3. Duplication between benchmark and guide is fine. We don't need this to replace the current FMA nightly workflow in benchmarking. It's OK to have duplication. Your existing Python-based deploy/scale/undeploy workflow should stay in benchmarking, since it gives us better/more flexible experimentation there. The guide, on the other hand, needs to stay deliberately simpler.

4. Harness. Per @maugustosilva , we keep the same harness we're currently using for FMA rather than switching to nop. Manuel set that one up on purpose because we need metrics that we can't get out of `inference-perf`, so we want to preserve that.

guides/fast-model-actuation/README.md
lionelvillard: AFAIK no guides are using RS. WVA may not support RS

MikeSpreitzer:
FMA release 0.6.2 is available, and it fixes some important bugs that were present up through 0.6.1 (e.g., https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/629). I recommend updating this to use FMA release 0.6.2.

guides/fast-model-actuation/README.md
MikeSpreitzer: The latest is now 0.6.2, and it is noticeably better than 0.6.1

guides/fast-model-actuation/modelserver/kustomization.yaml
MikeSpreitzer, line 1: Why is kustomize involved at all? I see it doing nothing here.

guides/fast-model-actuation/rbac/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+# The namespaced Role/RoleBinding (and the ServiceAccount subject reference)
+# are pinned to the guide's namespace so the binding is valid. The ClusterRole
+# is cluster-scoped and is left un-namespaced by kustomize.
+# If you deploy the guide into a different namespace, update this value to match.
+namespace: llm-d-fast-model-actuation
```
MikeSpreitzer, line 8: This is the world's most painful way to control the destination namespace. To change it, you have to modify a version-controlled file. Why not just tell people to `kubectl apply -n ${NAMESPACE} ... `, as is done in `guides/fast-model-actuation/modelserver/kustomization.yaml` ?

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,303 @@
+# Fast Model Actuation
+
+## Overview
+
+Fast Model Actuation (FMA) enables rapid model loading and switching for LLM inference on Kubernetes by exploiting vLLM sleep/wake and model swapping. In Kubernetes, GPUs are bound 1-to-1 to pods: a pod that requests a GPU holds it exclusively for its lifetime. FMA uses the following **dual pod** technique to circumvent this constraint:
```
MikeSpreitzer, line 5: "rapid model loading and switching" is, technically, entirely wrong (from the point of view of a vllm instance).

FMA does nothing about model loading in vllm. You might be thinking of using sleep and wake to avoid model loading and CUDA graph compilation entirely, in lucky scenarios.

FMA is specifically and explicitly designed to be orthogonal to vllm's model loading, so that FMA combines out-of-the-box with technologies that accomplish fast (or whatever other attribute(s) you want) model loading in vllm.

FMA does not "switch" or "swap" models in a vllm instance. FMA's launcher makes one contribution to reducing startup time: pre-loading the Python modules.

Having a collection of vllm instances bound to the same GPU, with at most one instance awake at a time, accomplishes fast switching or swapping of models on that GPU. This generalizes to multiple GPUs on one node.

Perhaps something like the following would be better.

> Fast Model Actuation (FMA) addresses vllm startup time using two technologies. One is vLLM sleep and wake, which can entirely avoid model loading and CUDA graph compilation in lucky scenarios. The other is running multiple vllm processes as children of a launcher process that has already done the Python module loading. FMA can manage multiple vllm instances bound to one GPU, with one instance awake at a time, to accomplish fast switching of which model that GPU is used for; this generalizes to multiple GPUs of one node.

MikeSpreitzer:
I see that this document is being asked to do explanatory work that should be done by documentation in the FMA repo.

guides/fast-model-actuation/README.md
MikeSpreitzer, line 5: No, GPUs are not necessarily bound 1:1 to Pods. In Kubernetes scheduling and in the kubelet's device plugin, multiple GPUs can be bound to one Pod.

guides/fast-model-actuation/guide.yaml
```diff
@@ -0,0 +1,129 @@
+name: fast-model-actuation
+
+env:
+  static:
+    BRANCH: main
```
Vezio, line 5: Thanks you for being an early adopter of https://github.com/llm-d/llm-d/pull/1988! This already looks like it's doing it's purpose to cleanly and reproducibility render README.md for guides.

CC: @liu-cong

aavarghese (approved):

jimcadden (approved):

Gregory-Pereira (approved):
/lgtm

Lets get this in so we can continue iterating on it

PR #1782, opened 2026-06-11
docs: add AGENTS.md with agent coding conventions

rahulgurnani:
Addresses #1783

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,17 @@
+# AGENTS.md
+
+**Do:**
+- Edit `README.md` files in guides to fix accuracy, improve clarity, or update version numbers.
+- Update Helm values files (`values.yaml`, `values-*.yaml`) in `guides/*/router/` and `guides/*/modelserver/`.
+- Update Kustomize overlays (`kustomization.yaml`, patch files) in guide directories.
+- Fix env var declarations in `docker/scripts/` shell scripts to satisfy the linter.
+- Keep code blocks in guides consistent when environment variables or version pins change.
+
+**Do not:**
+- Commit without a DCO sign-off (`git commit -s`).
+- Remove or rename environment variables exported in guide `README.md` files without updating every downstream reference in the same guide.
+- Introduce new Docker base images or add external dependencies without maintainer discussion.
+- Mark experimental guides as stable without approval.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,260 @@
+# AGENTS.md
```
gyliu513, line 1: I think you can also add a CLAUD.md with content as `@AGENTS.md`, refer to https://github.com/kubernetes-sigs/kueue/blob/main/CLAUDE.md as a reference

vMaroon:
I think this file should be more instructional and stateless. E.g., no need to list guides here directly, or dive into niche specifics.

I also suggest distilling https://github.com/llm-d/llm-d-router/blob/main/AGENTS.md into here. It has valuable instructions on preferences/styling/language .. etc

AGENTS.md
rahulgurnani: Doesn't CLAUDE automatically take AGENTS.md as a context?

gyliu513: Not currently — Claude Code only auto-loads CLAUDE.md, not AGENTS.md.

rahulgurnani:
> I think this file should be more instructional and stateless. E.g., no need to list guides here directly, or dive into niche specifics.
> 
> I also suggest distilling https://github.com/llm-d/llm-d-router/blob/main/AGENTS.md into here. It has valuable instructions on preferences/styling/language .. etc

Made it more stateless in the last commit. I think distilling the language styling conventions here is not particularly valuable here since llm-d-router has mostly go code and in llm-d we mostly have configs and documentation. Thanks!

vMaroon:
>I think distilling the language styling conventions here is not particularly valuable here since llm-d-router has mostly go code and in llm-d we mostly have configs and documentation. 

I think some are still valuable, especially around verbosity, clarity and avoiding temporal language which agents often do when injecting conversation bias.

AGENTS.md
rahulgurnani: I see, updated to add CLAUDE.md as well.

rahulgurnani:
> > I think distilling the language styling conventions here is not particularly valuable here since llm-d-router has mostly go code and in llm-d we mostly have configs and documentation.
> 
> I think some are still valuable, especially around verbosity, clarity and avoiding temporal language which agents often do when injecting conversation bias.

added some of these.

PR #1785, opened 2026-06-12
Fix file path in tiered prefix guide 

liu-cong:


guides/tiered-prefix-cache/router/tiered-prefix-cache-cpu.values.yaml
```diff
@@ -18,13 +18,10 @@ router:
         - type: no-hit-lru-scorer
         - type: approx-prefix-cache-producer
           name: gpu-prefix-cache-producer
-          parameters:
-            maxPrefixTokensToMatch: 40000 # enough to cover the ~20k input length in the benchmark.
```
liu-cong, line 22: we increased the default, no need for this now.

Issue #1582, opened 2026-05-21
Recommendations to mitigate envoy ext_proc overflow


jneuff:
Hi and thanks for this amazing project!

I am not reporting an issue with llm-d, but want to get some community knowledge about an issue I have been seeing in my setup.

The setup is Envoy Gateway and AI Gateway in conjunction with llm-d as EPP. For quite modest amounts of requests envoy returns HTTP 500s with the following response code details:

```
ext_proc_error_gRPC_error_14{upstream_connect_error_or_disconnect/reset_before_headers._reset_reason:_overflow{upstream_reset_before_response_started{overflow}}}
```

It seems that it hits the maximum amount of connections to the ext_proc process. A colleague of mine raised this in the envoy slack and apparently there is no way to currently raise the limit, but a fix is being worked on.

Scaling the issue away with more envoy replicas seems not feasible to me, as I am seeing quite a few of these with a modest amount of requests (say ~2000 concurrent requests, more than half returning HTTP 500s, 8 envoy replicas). I think part of the issue is that LLM requests can be long-running, thus the limit gets saturated quite quickly.

So the reason I am raising this here: I imagine others have run into this problem and would like to hear about mitigation approaches.

If you rather not have this discussion here, please don't hesitate to close the issue.

Thanks :)

robertgshaw2-redhat:
yeah, that does not seem right (I personally have run Istio with 10000+ concurrent request for WideEP jobs) --- I asked the networking experts in the llm-d slack https://llm-d.slack.com/archives/C08SBNRRSBD/p1779364106446319 for their advise

robertgshaw2-redhat:
Okay, I was looking back in my notes and actually recalling running into this before with kgateway and Istio --- caveat, im not an expert in this stuff

```
[2025-07-20 22:10:33.829][210][debug][http] [external/envoy/source/common/http/async_client_impl.cc:199] async http request response headers (end_stream=true):
':status', '200'
'x-envoy-overloaded', 'true'
'content-type', 'application/grpc'
'grpc-status', '14'
'grpc-message', 'upstream connect error or disconnect/reset before headers. reset reason: overflow'

[...]

PR #1788, opened 2026-06-12
guides: add Envoy AI Gateway installation instructions

nacx:
Adds a guide explaining how to install [Envoy AI Gateway](https://aigateway.envoyproxy.io/)

### DIFF
--- a/docs/api-reference/artifacts.md
+++ b/docs/api-reference/artifacts.md
@@ -37,7 +37,7 @@ llm-d Router is deployed via Helm. We offer a chart both Standalone and Gateway 
 | Chart | Version | OCI Registry | Description |
 |-------|---------|--------------|-------------|
 | **Standalone Mode** | v0 | `oci://ghcr.io/llm-d/charts/llm-d-router-standalone-dev` | Deploys an InferencePool and EPP with a standalone Envoy proxy as sidecar in EPP pod  |
-| **Gateway Mode** | v0 | `oci://ghcr.io/llm-d/charts/llm-d-router-gateway-dev` | Deploys an InferencePool and EPP for use with an existing Kubernetes Gateway (e.g. Istio, AgentGateway, GKE) |
+| **Gateway Mode** | v0 | `oci://ghcr.io/llm-d/charts/llm-d-router-gateway-dev` | Deploys an InferencePool and EPP for use with an existing Kubernetes Gateway (e.g. Istio, AgentGateway, Envoy AI Gateway, GKE) |
 
 The charts are currently published by the Gateway API Inference Extension (GAIE) project (see [standalone mode source](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/config/charts/standalone) and [gateway mode source](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/config/charts/inferencepool)). Each well-lit path guides provides values files on top of the chart defaults to enable the functionality implemented in EPP.
 
--- a/docs/infrastructure/gateway/envoy-ai-gateway.md
+++ b/docs/infrastructure/gateway/envoy-ai-gateway.md
@@ -173,6 +173,7 @@ curl -X POST http://${IP}/v1/completions \
 ## Cleanup
 
 ```bash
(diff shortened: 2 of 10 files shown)


robertgshaw2-redhat:
QQ @nacx - can you be on point to maintain this integration moving forward?

nacx:
Yes, happy to do that. I'm one of the maintainers of Envoy AI Gateway.

robertgshaw2-redhat:
> Yes, happy to do that. I'm one of the maintainers of Envoy AI Gateway.

Awesome, thank you very much

robertgshaw2-redhat:
cool

Is there anything we want to change about the default configuration of Envoy AI Gateway? We have seen some users reporting issues with default circuit breakers with the default setup here

see for example: https://github.com/llm-d/llm-d/issues/1582

nacx:
Good point. I don't think that concrete case of the max connections is something I'd suggest out of the box, but I've added a common setting [we suggest in our examples](https://github.com/envoyproxy/ai-gateway/blob/61b4cc81564959b1750331ec0b1cfdcb4f003ce4/examples/basic/basic.yaml#L30-L43) to better adjust some default buffer values.

docs/api-reference/artifacts.md
```diff
@@ -37,7 +37,7 @@ llm-d Router is deployed via Helm. We offer a chart both Standalone and Gateway
 | Chart | Version | OCI Registry | Description |
 |-------|---------|--------------|-------------|
 | **Standalone Mode** | v0 | `oci://ghcr.io/llm-d/charts/llm-d-router-standalone-dev` | Deploys an InferencePool and EPP with a standalone Envoy proxy as sidecar in EPP pod  |
-| **Gateway Mode** | v0 | `oci://ghcr.io/llm-d/charts/llm-d-router-gateway-dev` | Deploys an InferencePool and EPP for use with an existing Kubernetes Gateway (e.g. Istio, AgentGateway, GKE) |
+| **Gateway Mode** | v0 | `oci://ghcr.io/llm-d/charts/llm-d-router-gateway-dev` | Deploys an InferencePool and EPP for use with an existing Kubernetes Gateway (e.g. Istio, AgentGateway, Envoy AI Gateway, GKE) |
```
nacx, line 40: BTW, @robertgshaw2-redhat could you point me to the source code for this `llm-d-router-gateway-dev` chart? I opened a PR to the `llmd-infra` repo in https://github.com/llm-d-incubation/llm-d-infra/pull/287, but I'm not sure that's the right place to have EAIGW added.

robertgshaw2-redhat: the charts are here --- https://github.com/llm-d/llm-d-router/tree/main/config/charts

docs/infrastructure/gateway/envoy-ai-gateway.md
```diff
@@ -0,0 +1,237 @@
...
+
+# Install the CRDs first, skipping the Gateway API CRDs installed in the previous step.
+# 
+# Note 1: We’re using helm template piped into kubectl apply instead of helm install due to aknown Helm limitation related
+# to large CRDs in the templates/ directory: https://github.com/helm/helm/pull/12277
+# Note 2: We filter the output of the Helm template to remove offending lines until this is fixed: https://github.com/helm/helm/pull/32217
+helm template eg-crds oci://docker.io/envoyproxy/gateway-crds-helm \
+  --version ${ENVOY_GATEWAY_VERSION} \
+  --set crds.gatewayAPI.enabled=false \
+  --set crds.envoyGateway.enabled=true \
```
alexagriffith, line 37: can you add a note as to why we use these two flags? They aren't what's used on the quickstart so I think this context would help users

alexagriffith: I see that its skipping install of CRDs from install script but maybe explicitly noting that this is what the flag is for/why we are doing it?

nacx: This is mentioned just above in line 29. Do we need to add more comments about it?

```
# Install the CRDs first, skipping the Gateway API CRDs installed in the previous step.
```

PR #1796, opened 2026-06-12
docs: added hf token part in multimodal optimized baseline guide

capri-xiyue:
added hf token part in multimodal optimized baseline guide


capri-xiyue:
assign @liu-cong @ahg-g

maugustosilva (approved):

PR #1801, opened 2026-06-12
docs(flowcontrol): map shutdown-drained requests to 503

LukeAVanDrie:
## Summary

Updates the EPP Flow Control architecture doc to reflect the behavior change in llm-d/llm-d-router#1612, which remaps requests drained during a graceful shutdown from `500 Internal` to a retryable `503 Service Unavailable` (with `x-llm-d-request-dropped-reason: rejected-shutting-down`).

## Changes

- Error-mapping table: split the old `RejectedOther`/`EvictedOther` → 500 row into graceful-drain → 503 vs. genuine internal error → 500; added an `x-llm-d-request-dropped-reason` header column documenting the drop-reason contract.
- State diagram: added `Drained` (shutdown → 503) and `Failed` (internal → 500) terminal states.
- Non-persistence note: graceful shutdown now documented as retryable 503, not 500.
- New operator note: clarifies these mappings only apply while the EPP is reachable — under the router recipe's default `FailOpen`, an unreachable EPP is bypassed and requests route directly to the pool with no flow control.

## Related

- Documents llm-d/llm-d-router#1612 (merged)


LukeAVanDrie:
cc: @RishabhSaini

RishabhSaini:
LGTM

PR #1817, opened 2026-06-15
agentic-serving with token-load scorer

kaushikmitr:
## Summary

Optimizes the **agentic-serving** guide's router configuration for prefix-cache-aware routing on KV-offloaded deployments, and makes the benchmark reproducible per concurrency level. On the agentic code-generation workload (Qwen3-Coder-480B-FP8, 8× TPU v7x), the optimized config sustains **~3× the throughput** (~120K vs ~40K tok/s peak) and **dramatically lower TTFT** vs the plain k8s-Service baseline.

## What changed

- **Router config (`router/agentic-workloads.values.yaml`)** — switch to `token-load-scorer` + `inflight-load-producer` + `approx-prefix-cache-producer` + `prefix-cache-affinity-filter`. The producer's LRU is sized for the **KV-cache offload tier** (`autoTune: false`, `lruCapacityPerServer` set manually); `peakPrefillThroughput` is calibrated for this model/hardware. (See note re: the LRU workaround below.)
- **Chart** — deploy via `llm-d-router-standalone-dev` (`v0`) and add the missing **GAIE CRD install** step.
- **Benchmark template (`benchmark-templates/guide.yaml`)** — parameterize the workload on `${CONCURRENCY_LEVEL}` / `${NUM_REQUESTS}` / `${SEED}` (one stage per run) so each concurrency level is an isolated run; bump the benchmark image to `v0.6.8.1`.
- **Calibration tooling (`calibration/`, new)** — `calibrate.sh` + measurement Job + values template + README to measure `peakPrefillThroughput` (it's hardware/model-specific).
- **Results** — refreshed `benchmark-results/` charts (now through conc=80) and the summary table.

## Why a separate seed per concurrency level

The benchmark uses `conversation_replay` with **`num_conversations = ${CONCURRENCY_LEVEL}`**, and each concurrency level is run as a **separate** benchmark against the **same** running deployment. With a *fixed* seed, the generated conversations are **nested** — the seeded RNG produces conversations `1…N`, so conc=40's prompts are an exact **subset** of conc=80's (the first 40 are identical).

Because the deployment caches prefixes — and with KV-cache offloading those prefixes persist in host memory across runs — a later, higher-concurrency run would hit caches **already warmed by the earlier, lower-concurrency runs** that shared prompts with it. That inflates the prefix-cache hit rate and produces optimistic, **cross-run-contaminated** TTFT/throughput for the higher concurrencies.

Setting a **distinct seed per concurrency** (`SEED = 7 + CONCURRENCY_LEVEL`) makes every run draw an **independent** prompt set, eliminating cross-run cache carryover so each concurrency point is measured cleanly.


guides/agentic-serving/agentic-code-generation.md
```diff
@@ -144,12 +163,14 @@ envsubst < guide.yaml > config.yaml
...
+
 ### Summary with 40 concurrent coding sessions:
 | Metric | k8s Service | llm-d-optimized | Δ Improvement |
 | :--- | :--- | :--- | :--- |
-| **TTFT P50 (ms)** | 17391 | 1314 | ⬇️ 92.4% |
-| **Input tokens / sec** | 36987 | 78065 | ⬆️ 111.1% |
-| **Output tokens / sec** | 436.8 | 888.5 | ⬆️ 103.4% |
+| **TTFT P50 (ms)** | 17391 | 1064 | ⬇️ 93.9% |
+| **Input tokens / sec** | 36987 | 92705 | ⬆️ 150.6% |
+| **Output tokens / sec** | 436.8 | 647.5 | ⬆️ 48.2% |
```
vMaroon, line 175: Looking at the results, I'm not too sure this is better / apples to apples. Can the baseline configuration be ran with this benchmark configuration?

guides/recipes/router/calibration/calibrate.sh
```diff
@@ -0,0 +1,138 @@
+#!/bin/bash
```
ahg-g, line 1: can we place this script under `https://github.com/llm-d/llm-d/tree/main/guides/recipes/router`? we will probably use it for other guides in the future, right?

guides/agentic-serving/agentic-code-generation.md
kaushikmitr: yes this is the comparision. The token load scorer is indeed doing better and we can puch conc to upto 80: 
image

We deployed fresh vllm, ran benchmark with the default configmap, restarted vllm and reran benchmark with the new configmap to make sure the warmed caches from the first run dont impact the 2nd run.

ahg-g, line 175: pls put the total throughput as well to show that it is higher even though the output tokens/sec is lower for this configuration compared to the old one.

kaushikmitr: added

guides/recipes/router/calibration/calibrate.sh
kaushikmitr: yes, that the plan. I moved it to https://github.com/llm-d/llm-d/tree/main/guides/recipes/router

kaushikmitr:
@ahg-g can you help merge, it seems auto merge is disabled

PR #1824, opened 2026-06-16
Benchmark dynamic model

WenjiaoYue:
## Summary

This PR updates benchmark templates to make the served model configurable through the `MODEL` environment variable instead of hard-coding model names in each template.

The change covers benchmark templates for:
- optimized-baseline
- precise-prefix-cache-routing
- pd-disaggregation, including the TPU benchmark template

It also updates the related benchmark documentation to show how to set `MODEL` with guide-specific defaults before running `envsubst`, so the generated benchmark config continues to match the deployed stack while allowing overlays or alternate stacks to use different models.

## Changes

- Replace hard-coded `endpoint.model` values in benchmark templates with `${MODEL}`.
- Add `MODEL` exports with sensible defaults to the affected guide benchmark instructions.
- Document `MODEL` in the shared benchmark helper docs and include it in the environment sanity check output.
- Clarify that `envsubst` now resolves `MODEL` along with the existing namespace and gateway variables.

### DIFF
--- a/guides/optimized-baseline/README.md
+++ b/guides/optimized-baseline/README.md
@@ -253,6 +253,7 @@ export IP=$(kubectl get gateway llm-d-inference-gateway  -n ${NAMESPACE} -o json
 </details>
 
 ```bash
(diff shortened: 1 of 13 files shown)


maugustosilva:
Hi @WenjiaoYue. Please do note we moved all workloads to `llm-d-benchmark`, as workload [templates](https://github.com/llm-d/llm-d-benchmark/tree/main/workload/profiles)

PR #1829, opened 2026-06-16
Upgrading vllm-tpu image from 0.18.0 to 0.19.0 as 0.18.0 does not support gemma-4

gushob21:


guides/optimized-baseline/modelserver/tpu-v6/vllm/kustomization.yaml
```diff
@@ -8,7 +8,7 @@ namePrefix: optimized-baseline-tpu-v6-vllm-
 images:
   - name: REPLACE_MODEL_SERVER_IMAGE
     newName: vllm/vllm-tpu
-    newTag: v0.18.0
+    newTag: v0.19.0
```
liu-cong, line 11: please also update the tpu-v7 overlay. There are a few other places as well. As a follow-up, we can create a base overlay for all TPU variants and update the image there.

gushob21: I have made the changes in all the files that were using v0.18.0. These files were changed:

guides/optimized-baseline/modelserver/tpu-v6/vllm/kustomization.yaml
guides/optimized-baseline/modelserver/tpu-v7/vllm/kustomization.yaml
guides/pd-disaggregation/modelserver/tpu/v6/vllm/kustomization.yaml

PR #1830, opened 2026-06-16
Re-organized the well-lit paths docs into three groupings

ahg-g:
Completely restructured our well-lit paths decoupled pillars:
- Core Capability Building Blocks
- Workloads
- Operational Excellence

As part of this, batch well-lit paths moved under workloads; no-k8s deployment moved under operational excellence


vMaroon:
Organization makes total sense to me

liu-cong:
+1 on the reorg as well

chcost:
+1, I also like this re-org.

Ibrahim2595:
Minor suggestions:
- `Capability Building Blocks` or `Core Building Blocks` instead of `Core Capability Building Blocks`
- `Operational Patterns` instead of `Operational Excellence`

maugustosilva:
Fair warning: the moment this gets merged, **all** CICD workflows will immediately fail. I propose we coordinate this merging with a separate PR, where we deal with the github actions.

chcost:
my understanding is that workflows reference guides/ by directory name (scenario_dir: 'guides', standup_scenario: 'optimized-baseline'). Those paths resolve to guides/optimized-baseline/ which contains the kustomize manifests. This PR doesn't move anything under guides/ itself. But, But any workflow or check that does doc validation (linting, link checking, Docusaurus builds), like on the docs website will break immediately because the files moved but every consumer of those paths (sync-docs.sh, release branch snapshots, other docs that reference them) still points to the old flat structure.
So yes, we should make sure we also have the separate PRs dealing with any breakage we expect from this.

maugustosilva:
@chcost Right, I have confirmed with @Vezio, who wrote the "`README.md` parser" and he states that we are not following the `docs` directly, but rather solely the `guides`. So, **I jumped the gun on pointing out a problem** here. Given the goal on having this for this release, I will merge now, and confirm that no "github actions workflows" were harmed in the making of this PR.

PR #1835, opened 2026-06-17
add missing consolidation-* workflows for past status of tiered prefix

amirfr3:
…x cache across multiple configurations


amirfr3:
@liu-cong These ci workflows are currently missing for the cpu offloading (that's why the badges are missing in the guide)

liu-cong (approved):

liu-cong:
Thanks!

Issue #1750, opened 2026-06-10
Review benchmark config for optimized baseline with sglang


liu-cong:

rahulgurnani:
Addressed in https://github.com/llm-d/llm-d/pull/1840 . @liu-cong can you please close this one? I don't have permissions.

Issue #1749, opened 2026-06-10
Review benchmark configurations for sglang


liu-cong:
Some of the sglang benchmarks show significant different with their vllm counter part, e.g. https://github.com/llm-d/llm-d/pull/1720. This could be due to different configurations such as kv cache capacity. This task is to review those and suggest either updating sglang recipes or the benchmark templates to better reflect sglang performance.

The task is to:
* run existing benchmark with sglang  and compare with vllm results
* if there are significant differences, investigate the cause and suggest updates

rahulgurnani:
One observation I have is that sglang cache hit seems lower than what I see for vLLM when running these benchmarks. So for vLLM its 90+ percent and sglang its ~20%. It may be something to do with the benchmark or sglang config, still investigating this further.

Also, found a gap in inference-perf while running these benchmarks, that I have documented in https://github.com/kubernetes-sigs/inference-perf/issues/548

PR #1840, opened 2026-06-17
[SGLang]: Increase the memory used for KV cache and update benchmark results in optimized baseline

rahulgurnani:
Earlier, benchmarks for SGLang were using mem-fraction-static to be default (~82%) which was causing early saturation for sglang deployment with the same benchmark. The --mem-fraction-static parameter in [SGLang](https://sgl-project.github.io/references/faq.html) controls the portion of available GPU memory dedicated to the static memory pool, which primarily hosts the KV cache.
In this change, the mem-fraction-static is bumped up to 90%. vLLM deployments use 95% as the limit but SGLang deployment at 95% fails with out of memory errors while loading CUDA kernels. 

SGLang [docs](https://docs.sglang.io/docs/advanced_features/hyperparameter_tuning#tune-mem-fraction-static-to-increase-kv-cache-pool-capacity) recommends to gradually increase the mem-fraction-static. I think 90% is a safe spot upon trying different numbers as above 92%, I started seeing OOM errors.

Related issues: #1750 #1749


guides/optimized-baseline/README.md
```diff
@@ -322,36 +322,36 @@ Graphs below compare optimized-baseline routing to a stock Kubernetes Service th
...
-| 35   | 5,890      | 14,026       | 24.162        | 0.117           | 45.190       | 0.150          |
-| 40   | 6,336      | 16,041       | 68.673        | 0.153           | 126.238      | 0.216          |
-| 43   | 6,588      | 16,339       | 72.429        | 0.254           | 130.275      | 0.218          |
-| 46   | 6,459      | 16,665       | 70.084        | 0.154           | 129.810      | 0.220          |
-| 49   | 6,265      | 16,126       | 70.659        | 0.151           | 133.718      | 0.209          |
-| 52   | 6,303      | 16,474       | 74.326        | 0.152           | 134.981      | 0.219          |
-| 55   | 6,290      | 16,854       | 72.564        | 0.153           | 134.034      | 0.215          |
-| 57   | 6,089      | 16,641       | 72.329        | 0.153           | 135.023      | 0.217          |
-| 60   | 6,551      | 17,064       | 75.586        | 0.154           | 138.663      | 0.217          |
+| Rate | k8s Output | vLLM Output | SGLang Output | vLLM TTFT mean | SGLang TTFT mean | vLLM TTFT p90 | SGLang TTFT p90 |
```
ahg-g: this table removed the k8s columns, can we keep them pls?

rahulgurnani: yes, added them back. thanks!

guides/optimized-baseline/README.md
```diff
@@ -322,36 +322,36 @@ Graphs below compare optimized-baseline routing to a stock Kubernetes Service th
...
 Summary across the full ladder (rates 3 → 60):
 
-| Metric              | k8s service (RR) | llm-d Optimized | Δ% vs k8s |
-| :------------------ | :--------------- | :-------------- | :-------- |
-| Output tokens/sec   | 5,722            | 13,163          | +130.0%   |
-| Requests/sec        | 35.87            | 36.38           | +1.4%     |
-| TTFT mean (s)       | 58.10            | 0.156           | −99.73%   |
-| TTFT p90 (s)        | 107.43           | 0.206           | −99.81%   |
-| ITL mean (ms)       | 44.0             | 47.0            | +6.8%     |
+| Metric            | k8s service (RR) | llm-d (vLLM) | llm-d (SGLang) | Δ% (vLLM vs k8s) | Δ% (SGLang vs k8s) |
```
liu-cong: note it looks like sglang vs k8s is actually llm-d+sglang vs. k8s+vllm, which isn't quite meaningful

liu-cong: I suggest keeping the existing vllm untouched, and add a new table for sglang.

rahulgurnani: makes sense, updated the PR.

rahulgurnani: yeah.. updated to segregate the two into different tables

liu-cong (approved):

PR #1844, opened 2026-06-17
Nightly llm-d-router resource usage test

jtechapps:
This PR creates a new nightly test that runs inference-perf job and records llm-d-router resource usage to a markdown file in the repo.

The test logic is in a python helper file which is used to run inference-perf job against the current kubectl context.

The pipeline supports the following configurable parameters (via manual triggers and environment variables):
- `guide_name` (e.g. `optimized-baseline`): the recipe name to use for llm-d-router configuration.
- `test_name`: the name to use in the output results.
-  `perf_job`: the path to the inference-perf configuration to use for the benchmark job.
-  `sim_replicas`: the number of llm-d-simulator pods to run to simulate model server backends.

Example usage of running the python helper file to generate resource usage data:
```
python3 scripts/run_nightly_perf.py \
    --guide-name optimized-baseline \
    --test-name optimized-baseline-job1 \
    --perf-job scripts/perf-configs/shared_prefix_job1.yaml \
    --sim-replicas 10 \
    --gcp-project llm-d-scale \
    --epp-cpu=3 \
    --epp-memory=6Gi \
    --router-machine-family=e2
```

I configured a single nightly workflow to run resource usage test with optimized-baseline llm-d-router recipe with 10k input/1k output shared prefix job.

In the future, different configurations will be extended that will reuse this test framework.


jtechapps:
I am moving this load testing logic into llm-d-router repo.

PR #1846, opened 2026-06-17
fix: standardize engine-type label to llm-d.ai/engine-type

liu-cong:


liu-cong:
Consolidating into a single PR.

Issue #1360, opened 2026-05-02
[CI/CD] Add automation to nightly testing of guides


robertgshaw2-redhat:
SUMMARY:
* the ci/cd runs commands from the guides
* these are manually done
* we should do this automatically

maugustosilva:
@robertgshaw2-redhat We (@achandrasekar @jjk-g @Vezio) are working on an end-to-end solution to this (FYI @liu-cong @vMaroon @chcost @ahg-g)

Once `v0.7.0` is cut, we will work to provide a proof-of-concept using `optimized-baseline`

More details will follow soon.

liu-cong:
cc @rlakhtakia

maugustosilva:
As promised, `llm-d-benchmark` added a new "standup method" - `kustomize` - and has now the ability to run the commands from a guide directly (since [this commit](https://github.com/llm-d/llm-d-benchmark/commit/1234ada6347053feb0638ae1213f42deb77da2b6), merged in 05/15/26). This is achieved by parsing a guide's `README.md`. As a proof of concept, we started with the [optimized-baseline](https://github.com/llm-d/llm-d/blob/main/guides/optimized-baseline/README.md) ).

When the command `llmdbenchmark --spec guides/optimized-baseline standup -t kustomize` is invoked, `llmdbenchmark will take care of `cloning` this repository, "parsing" the `README.md` looking for commands and then execute these. During the standing up operation, all the usual per-`pod` health-checking and per-stack "smoketest" functionalities will be leveraged, greatly reducing the complexity of the github CI/CD workflow.

maugustosilva:
The next steps are:

1) Create, in `llm-d-infra`, a new set of `reusable-nightly-<cloud provider>` workflows, shared between `llm-d` and `llm-d-benchmark` (undergoing)

2) Extend this new CI/CD for the other guides. In our initial assessment, we have noticed that not all the guides are easily parseable, and **modifications on the structure of the guides might be required**

maugustosilva:
Each nightly CI/CD collects and stores data on a (shared) GCS bucket. From this data source, we will be able to track not only failures/successes, but also comparative performance (e.g., detecting performance regressions). Our plan is to make this data available to [llm-d-prism](https://github.com/llm-d/llm-d-prism) in order to provide an interface for data visualization.

chcost:
this is great @maugustosilva! Thanks for putting it together. I also agree with the next steps.

maugustosilva:
[...]

PR #1865, opened 2026-06-18
Add async-processor operations guide

shimib:
## Summary

Adds an operations guide for the **Async Processor** under `docs/operations/`. The focus is **how to size the llm-d async container**, with concurrency as the central decision.

- `docs/operations/async-processor.md` — throughput model, concurrency sizing, container resources, horizontal scaling.
- `docs/operations/README.md` — link entry for the new guide.

## Concurrency sizing is backed by a measured sweep

A closed-loop sweep (each worker holds one in-flight request for its full duration — exactly an Async Processor worker) was run against a live llm-d stack: **Llama-3.1-8B on vLLM (`--max-num-seq 1024`), a single H100-80GB**, ~256-token outputs, dispatched through the inference gateway + EPP.

| Concurrency | Throughput (req/s) | Mean latency (s) | GPU KV used | Notes |
| :--- | :--- | :--- | :--- | :--- |
| **8 (default)** | **4.2** | 1.9 | <1% | GPU essentially idle |
| 64 | 24.1 | 2.6 | 5% | |
| 128 | 35.8 | 3.6 | 12% | approaching saturation |
| 192 | 40.4 | 4.7 | 16% | knee |
| 256 | 44.3 | 5.8 | 20% | past knee |
| 512 | 61.4 | 9.7 | 51% | unstable (errors recurred) |

Results:
- The **default `concurrency: 8` leaves the GPU ~90% idle** (~4 req/s) — ~10× below what the same hardware sustains.
- The **saturation knee is ~128–192** (16–24× the default).
- **Little's Law held exactly**: `in-flight = throughput × latency` matched the worker count at every stable point, validating `concurrency ≈ throughput × latency × 1.3`.
- A single replica destabilized past ~256 → argues for capping per-replica concurrency and scaling out.

## Notes / open items (WIP)

- CPU/memory figures in the Container Resource Sizing section are **architecture-derived** (clearly labeled), not yet measured from a real async-processor deployment.
- Benchmark covers a single 8B/H100 backend at ~256-token outputs; broader profiles (long-output, large agentic payloads) are extrapolated.

### DIFF
--- a/docs/operations/async-processor.md
+++ b/docs/operations/async-processor.md
@@ -1,0 +1,188 @@
+# llm-d Async Processor Operations Guide
+
+This guide covers operational best practices, scaling behavior, and container sizing recommendations for the [Async Processor](https://github.com/llm-d/llm-d-async).
+
+The Async Processor is a **lightweight dispatch agent**, not an inference engine. It pulls requests from a message queue (Redis or GCP Pub/Sub), passes them through dispatch gates, and forwards them to the llm-d Router over HTTP. Because the heavy lifting (token generation) happens on the model servers, the processor is almost entirely **I/O-bound**: each worker spends the vast majority of its lifetime blocked on a long-running HTTP call to the Router. This shapes every sizing decision below — the container is small, and the constraint is concurrency and payload buffering, not compute.
+
+> [!NOTE]
+> The **concurrency** guidance in [§2](#2-choosing-concurrency-the-most-important-sizing-decision) is backed by the measured sweep in that section. The **CPU/memory** figures in [§3](#3-container-resource-sizing) are architecture-derived starting points — validate them against your own workload and prefer your own measurements where you have them.
+
+---
+
+## 1. Throughput Model
+
+Understanding how the processor scales is a prerequisite to sizing it.
+
+- **Workers are the unit of in-flight concurrency.** Each worker pulls one request, dispatches it, and blocks until the Router returns a result (or the deadline expires). The total number of requests the processor can have in flight at once is the sum of its worker counts.
+- **The bottleneck is downstream, not the processor.** Dispatch rate is ultimately limited by inference-server capacity and the [dispatch gates](../architecture/advanced/batch/async-processor.md#dispatch-gates), not by the processor's CPU. Adding workers beyond what the inference pool can absorb only grows queue backlog and memory, not throughput.
+- **Scaling is horizontal and stateless.** The processor is a pull-based consumer. Running N replicas against the same queue multiplies effective concurrency: `total in-flight = replicas × workers-per-replica`. Queue backends distribute messages across all consumers, so no leader election or coordination is required.
+
+### Concurrency Configuration
+
+In the helm chart, concurrency is controlled by `ap.concurrency` (default `64`), or per-pool via `ap.workerPools` when you need independent parallelism for different queues/topics:
+
+```yaml
+ap:
+  concurrency: 64       # global worker count (default)
+  # OR, for per-queue control:
+  workerPools:
+    - id: "high-priority"
+      workers: 16
+    - id: "bulk"
+      workers: 8
+```
+
+When `workerPools` is set, the per-pool `workers` value overrides the global `concurrency` for that pool. Total in-flight concurrency for the replica is the sum of all pool worker counts.
+
+---
+
+## 2. Choosing Concurrency (the most important sizing decision)
+
+Concurrency is the single highest-impact knob. The default of `64` is a sane starting point — enough to keep a modest pool busy — but it is **not** automatically right for yours: a single large GPU still has headroom above it, and high-throughput or long-output workloads need considerably more. The sizing rule is Little's Law:
+
+```
+required_workers ≈ target_throughput (req/s) × avg_request_latency (s)
+```
+
+Because LLM requests are slow (seconds to minutes end-to-end), the latency multiplier is large: with a 2 s average request, `concurrency: 64` caps you at `64 / 2 = 32 req/s`, no matter how much GPU sits behind the Router (the old default of `8` capped you at just `4 req/s`).
+
+### Measured sweep
+
+We ran a closed-loop sweep (each worker holds one in-flight request for its full duration — exactly an Async Processor worker) against a live llm-d stack: **Llama-3.1-8B on vLLM (`--max-num-seq 1024`), a single H100-80GB**, ~256-token outputs, dispatched through the inference gateway + EPP. Throughput and vLLM KV-cache utilization vs. concurrency:
+
+| Concurrency | Throughput (req/s) | Mean latency (s) | GPU KV used | Notes |
+| :--- | :--- | :--- | :--- | :--- |
+| 8 | 4.2 | 1.9 | <1% | old default — GPU essentially idle |
+| 32 | 14.7 | 2.2 | 3% | |
+| **64 (default)** | **24.1** | 2.6 | 5% | current default |
+| 128 | 35.8 | 3.6 | 12% | approaching saturation |
+| 192 | 40.4 | 4.7 | 16% | **knee — best throughput/latency** |
+| 256 | 44.3 | 5.8 | 20% | past knee, latency climbing |
+| 512 | 61.4 | 9.7 | 51% | unstable: errors in repeated runs |
+
+Takeaways:
+
+- **The `64` default reaches ~60% of peak throughput.** At `concurrency: 64` this H100 sustains ~24 req/s (KV 5%) — a solid start, but still short of the ~40 req/s at the knee. The old default of `8` managed only ~4 req/s (KV <1%), roughly **10× below** the knee.
+- **The knee for this backend is ~128–192 — about 2–3× the `64` default.** Beyond it, latency rises faster than throughput (256→512 doubled latency for ~40% more throughput).
+- **Little's Law held exactly.** In-flight requests (`throughput × latency`) matched the worker count at every stable point — workers are never idle in the closed loop, so concurrency directly sets the load on the backend.
+- **A single replica destabilized past ~256.** At very high concurrency a lone processor overwhelmed the gateway/EPP connection path (stalls and errors recurred). This is a reason to **scale out** rather than push one replica arbitrarily high (see [§4](#4-horizontal-scaling)).
+
+### Applying it
+
+1. Estimate (or measure) your workload's average end-to-end request latency `W` at the Router.
+2. Set `concurrency ≈ ceil(target_throughput × W × 1.3)` (the 1.3 is headroom).
+3. Cap per-replica concurrency at a level your gateway/EPP handles cleanly (≈128–256 in the test above) and add replicas beyond that.
+4. **With a saturation/budget gate, you can set concurrency above this number and let the gate throttle dispatch** — but remember a gated worker still holds its pulled message in memory, so don't set it absurdly high.
+
+Quick anchor (derived from the sweep, ~256-token outputs):
+
+| Scenario | Target tput | ~Latency | Suggested concurrency | Topology |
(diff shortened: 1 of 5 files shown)


guides/asynchronous-processing/README.md
```diff
@@ -10,70 +10,245 @@ Async Processor integrates with llm-d to:
...
+  export branch="main" # branch, tag, or commit hash
+  git clone https://github.com/llm-d/llm-d.git && cd llm-d && git checkout ${branch}
+  ```
+
+- Set the following environment variables:
+
+  ```bash
+  export REPO_ROOT=$(realpath $(git rev-parse --show-toplevel))
+
+  # Async Processor
```
ahg-g: instead of listing the env vars explicitly in each guide, we have https://github.com/llm-d/llm-d/blob/main/guides/env.sh that gets sourced by all guides, see for example: https://github.com/llm-d/llm-d/tree/main/guides/optimized-baseline#prerequisites

guides/asynchronous-processing/README.md
```diff
@@ -10,70 +10,245 @@ Async Processor integrates with llm-d to:
...
+# Store it in a secret under the key `url`.
+kubectl create secret generic redis-creds -n ${NAMESPACE} \
+    --from-literal=url="redis://:${REDIS_PASSWORD}@redis-master.redis.svc.cluster.local:6379"
+```
+
+In `guides/asynchronous-processing/redis/values.yaml`, remove the plaintext `ap.redis.url` and instead set `ap.redis.secretName: redis-creds` and `ap.redis.secretKey: url` before installing.
+
+</details>
+
+### 3. Resolve the inference gateway endpoint
```
ahg-g: please use the term `llm-d Router` throughout the guide instead of inference gateway

guides/asynchronous-processing/README.md
shimib: Done in bdbcc1d6 — the guide (and the kind-e2e harness) now `source ${REPO_ROOT}/guides/env.sh` and use `${ROUTER_STANDALONE_CHART}` / `${ROUTER_CHART_VERSION}` instead of redeclaring them. Dropped the per-guide `GAIE_VERSION`/`ROUTER_CHART_VERSION`.

guides/asynchronous-processing/README.md
shimib: Done in bdbcc1d6 — renamed the section to "Resolve the llm-d Router endpoint" and switched the remaining "inference gateway"/"baseline gateway" wording to "llm-d Router" throughout (kept `llm-d-inference-gateway` only where it's the literal Gateway resource name in the Gateway Mode block).

guides/asynchronous-processing/test/kind-e2e.sh
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env bash
```
ahg-g, line 1: this needs to be invoked from the github [workflows](https://github.com/llm-d/llm-d/tree/main/.github/workflows), are you planning to do that in a follow up PR?

shimib: Done in this PR (674cd604) rather than a follow-up — added `.github/workflows/ci-async-kind-e2e.yaml`. It runs this script from zero on a self-contained kind cluster (optimized-baseline on a CPU model server + Redis + Async Processor) and asserts the smoke test, on PRs touching the guide + `workflow_dispatch`. It's a GPU-free single-runner check (~6.5 min end-to-end locally); the per-provider/accelerator nightly e2e stays with llm-d-benchmark per #1360.

One note: the harness pins the llm-d Router chart to the latest **published** release (`v0.9.0`, env-overridable) instead of sourcing `guides/env.sh`, because `env.sh` currently pins `ROUTER_CHART_VERSION=v0.9.1` which isn't published yet (`ghcr.io/llm-d/charts/llm-d-router-standalone:v0.9.1` 404s anonymously). The guide README still sources `env.sh`. Heads-up that the unpublished `v0.9.1` pin affects every guide that sources `env.sh` — happy to bump it to `v0.9.0` here or leave it for the router release, whichever you prefer.

ahg-g: env.sh actually uses v0.9.0: https://github.com/llm-d/llm-d/blob/main/guides/env.sh, but we will revert it to use the main now that the release is done to ensure that the main branch in llm-d/llm-d is testing the latest released image from the router.

ahg-g:
The new test is failing

docs/operations/async-processor.md
```diff
@@ -0,0 +1,186 @@
+# llm-d Async Processor Operations Guide
+
+This guide covers operational best practices, scaling behavior, and container sizing recommendations for the [Async Processor](https://github.com/llm-d-incubation/llm-d-async).
+
+The Async Processor is a **lightweight dispatch agent**, not an inference engine. It pulls requests from a message queue (Redis or GCP Pub/Sub), passes them through dispatch gates, and forwards them to the llm-d Router over HTTP. Because the heavy lifting (token generation) happens on the model servers, the processor is almost entirely **I/O-bound**: each worker spends the vast majority of its lifetime blocked on a long-running HTTP call to the Router. This shapes every sizing decision below — the container is small, and the constraint is concurrency and payload buffering, not compute.
+
+> **Note:** The **concurrency** guidance in [§2](#2-choosing-concurrency-the-most-important-sizing-decision) is backed by the measured sweep in that section. The **CPU/memory** figures in [§3](#3-container-resource-sizing) are architecture-derived starting points — validate them against your own workload and prefer your own measurements where you have them.
```
ahg-g:
```suggestion
> [!NOTE]
> The **concurrency** guidance in [§2](#2-choosing-concurrency-the-most-important-sizing-decision) is backed by the measured sweep in that section. The **CPU/memory** figures in [§3](#3-container-resource-sizing) are architecture-derived starting points — validate them against your own workload and prefer your own measurements where you have them.
```

docs/operations/async-processor.md
```diff
@@ -0,0 +1,186 @@
...
+
+- **Workers are the unit of in-flight concurrency.** Each worker pulls one request, dispatches it, and blocks until the Router returns a result (or the deadline expires). The total number of requests the processor can have in flight at once is the sum of its worker counts.
+- **The bottleneck is downstream, not the processor.** Dispatch rate is ultimately limited by inference-server capacity and the [dispatch gates](../architecture/advanced/batch/async-processor.md#dispatch-gates), not by the processor's CPU. Adding workers beyond what the inference pool can absorb only grows queue backlog and memory, not throughput.
+- **Scaling is horizontal and stateless.** The processor is a pull-based consumer. Running N replicas against the same queue multiplies effective concurrency: `total in-flight = replicas × workers-per-replica`. Queue backends distribute messages across all consumers, so no leader election or coordination is required.
+
+### Concurrency Configuration
+
+Concurrency is controlled by `ap.concurrency` (default `8`), or per-pool via `ap.workerPools` when you need independent parallelism for different queues/topics:
```
ahg-g:
```suggestion
In the helm chart, concurrency is controlled by `ap.concurrency` (default `8`), or per-pool via `ap.workerPools` when you need independent parallelism for different queues/topics:
```

docs/operations/async-processor.md
```diff
@@ -0,0 +1,186 @@
...
+
+Quick anchor (derived from the sweep, ~256-token outputs):
+
+| Scenario | Target tput | ~Latency | Suggested concurrency | Topology |
+| :--- | :--- | :--- | :--- | :--- |
+| Single small GPU (dev) | ~30 req/s | ~3–5 s | **128** | 1 replica |
+| Saturate 1× H100 (8B) | ~40 req/s | ~5 s | **~192** | 1 replica |
+| Larger fleet | scale linearly | per workload | `tput × W × 1.3` | N replicas × ≤256 |
+
+> For long-output workloads, `W` grows proportionally (a 5k-token completion is ~10–20× a 256-token one), so the required concurrency grows with it — long-output pipelines often need concurrency in the high hundreds to low thousands, split across replicas.
```
ahg-g, line 86:
```suggestion
> [!NOTE]
> For long-output workloads, `W` grows proportionally (a 5k-token completion is ~10–20× a 256-token one), so the required concurrency grows with it — long-output pipelines often need concurrency in the high hundreds to low thousands, split across replicas.
```

guides/asynchronous-processing/README.md
ahg-g: should we remove the experimental designation?

guides/asynchronous-processing/README.md
```diff
@@ -10,70 +10,242 @@ Async Processor integrates with llm-d to:
...
+  export BASELINE_NAMESPACE=llm-d-optimized-baseline
+  export MODEL=Qwen/Qwen3-32B     # must match the model served by the baseline stack
+  export ACCELERATOR_TYPE=gpu     # options: gpu, amd, xpu, hpu, tpu/v6, tpu/v7, cpu
+  export INFRA_PROVIDER=base      # base | gke  (GPU only; leave unset for other accelerators)
+  export MODEL_SERVER=vllm        # options: vllm, sglang
+  export MODELSERVER_OVERLAY=${ACCELERATOR_TYPE}/${MODEL_SERVER}/${INFRA_PROVIDER}
+  ```
+
+  > [!NOTE]
+  > The `${INFRA_PROVIDER}` segment exists only for GPU overlays (`gpu/vllm/base`, `gpu/vllm/gke`). For other accelerators the overlay has no such subfolder, so set `MODELSERVER_OVERLAY` accordingly — e.g. `export MODELSERVER_OVERLAY=cpu/vllm` for a CPU-only test.
```
ahg-g:
```suggestion
> [!NOTE]
> The `${INFRA_PROVIDER}` segment exists only for GPU overlays (`gpu/vllm/base`, `gpu/vllm/gke`). For other accelerators the overlay has no such subfolder, so set `MODELSERVER_OVERLAY` accordingly — e.g. `export MODELSERVER_OVERLAY=cpu/vllm` for a CPU-only test.
```

docs/operations/async-processor.md
shimib: Applied in the reworked guide (using the `[!NOTE]` callout).

docs/operations/async-processor.md
shimib: Applied — with one correction: the chart default is now **`64`**, not `8` (`charts/async-processor/values.yaml`). The whole guide is reframed around `64` as the (reasonable) default rather than treating `8` as a broken one.

docs/operations/async-processor.md
shimib: Applied (as a `[!NOTE]`).

guides/asynchronous-processing/README.md
shimib: Done — dropped `[Experimental]` from the title.

guides/asynchronous-processing/README.md
shimib: Good catch. I've reworked this PR to be **additive** — it keeps main's lean guide README (already migrated) and adds the depth in a dedicated ops guide + the kind-e2e harness, rather than re-expanding the README. Your overlay note now lives as a comment in `test/kind-e2e.sh` where the CPU overlay is selected (the harness is CPU-only by design, so the GPU `${INFRA_PROVIDER}` split doesn't apply there).

shimib:
Rebased onto current main and reworked into a clean **additive** PR (new ops guide + kind-e2e harness + CI); it no longer rewrites the guide README/redis/pubsub files (main already owns those). Verified end-to-end locally (twice): kind → optimized-baseline (CPU) → Redis → async-processor `0.7.4`, smoke test returns a live result and tears down cleanly. Also fixed the harness to pull from `ghcr.io/llm-d` `0.7.4` (was stale `llm-d-incubation`/`0.7.2`). The ops guide's concurrency guidance is reframed around the current `ap.concurrency=64` default. Inline suggestions addressed above.

PR #1871, opened 2026-06-18
guides/precise-prefix-cache-routing: run the renderer as a Service, not an EPP sidecar

vMaroon:
The tokenizer renderer was injected as a per-EPP-pod `vllm-render` sidecar by the standalone chart (`router.tokenizer.enabled: true`). That couples render capacity to the EPP replica count and duplicates the tokenizer in every EPP pod.

Replace it with a standalone, horizontally-scalable render Service:
- Add `render/` kustomize overlay: a `vllm launch render` Deployment (3 replicas)
  + ClusterIP Service, mirroring the chart's sidecar container spec.
- Disable the sidecar (`router.tokenizer.enabled: false`) and point the EPP `token-producer` plugin at the Service (`http://precise-prefix-cache-routing-render:8000`).
- Keep the `llm-d.ai/guide` label off render pods — it is the InferencePool / model-server selector, so labeled render pods would be treated as routable model servers (and the EPP would try to ZMQ-subscribe to their nonexistent :5556 KV-event socket).
- README: document deploying/cleaning up the render Service + the label caveat.
- CI: dry-run the new render overlay; correct the stale sidecar comments.

Pin the GPU vLLM model server and the renderer to vllm/vllm-openai{,-cpu}:v0.23.0.

Validated on 16xH100 (8x Qwen3-32B, TP=2) with EPP image ghcr.io/llm-d/llm-d-router-endpoint-picker:v0.9.0-rc.2: the full guide_precise-prefix-cache-routing_1.yaml ladder (rates 3->60) ran with 0 request failures; the 3 render pods served ~17k tokenization requests, evenly load-balanced, with 0 pod restarts.

### DIFF
--- a/guides/precise-prefix-cache-routing/render/deployment.yaml
+++ b/guides/precise-prefix-cache-routing/render/deployment.yaml
@@ -11,16 +11,23 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/component: vllm-render
+      app.kubernetes.io/part-of: precise-prefix-cache-routing
   template:
     metadata:
       labels:
         app.kubernetes.io/component: vllm-render
+        # Guide-scoped (but NOT `llm-d.ai/guide`, which is the InferencePool /
+        # model-server selector) so the Service below cannot cross-select a
+        # `vllm-render` pod from another guide in a shared namespace.
+        app.kubernetes.io/part-of: precise-prefix-cache-routing
     spec:
+      # The renderer only tokenizes over HTTP; it needs no Kubernetes API access.
+      automountServiceAccountToken: false
       containers:
         - name: vllm-render
           # GPU-less vLLM image — `vllm launch render` only tokenizes, no model
           # weights are loaded. Mirrors the chart's sidecar image/command.
-          image: vllm/vllm-openai-cpu:v0.23.0
+          image: docker.io/vllm/vllm-openai-cpu:v0.23.0
           imagePullPolicy: IfNotPresent
           command: ["vllm", "launch", "render"]
           args:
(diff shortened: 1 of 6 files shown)


guides/precise-prefix-cache-routing/render/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+                  name: llm-d-hf-token
+                  key: HF_TOKEN
+            # Disable vLLM usage telemetry — its writes to /.config fail under
+            # restricted SecurityContexts (matches the model server patch).
+            - name: DO_NOT_TRACK
+              value: "1"
+          resources:
+            requests:
+              cpu: "4"
+              memory: 8Gi
```
zdtsw: gonna be quite resource to run replica 3

guides/precise-prefix-cache-routing/render/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
...
+# NOTE: the guide label goes on resource metadata ONLY, never on the pod
+# template or selectors. `llm-d.ai/guide: precise-prefix-cache-routing` is the
+# InferencePool / model-server selector (router modelServers.matchLabels); if
+# render pods carried it the EPP would treat them as routable model servers and
+# try to ZMQ-subscribe to their (nonexistent) :5556 KV-event socket.
+labels:
+  - pairs:
+      llm-d.ai/guide: precise-prefix-cache-routing
+    includeSelectors: false
+    includeTemplates: false
```
zdtsw: these are default ones on "labels", aint they? probably can skip setting here, but keep comments above

guides/precise-prefix-cache-routing/router/precise-prefix-cache-routing.values.yaml
```diff
@@ -1,23 +1,20 @@
...
   tokenizer:
-    enabled: true
-    # Required by llm-d-router chart v0 when tokenizer is enabled;
-    # must match the deployed model server and the token-producer modelName below
-    modelName: Qwen/Qwen3-32B
-    env:
-      - name: HF_HOME
-        value: /tmp/hf/hub
+    # Sidecar disabled — the renderer runs as a separate Service (../render/).
+    enabled: false
```
zdtsw: why not just take away  this totally? which should be equal to set to false

ahg-g:
Will this have a chance to land today/tomorrow?

vMaroon:
Addressed Wen's comments, @albertoperdomo2 is running a final benchmarking run to make sure the numbers are good.

vMaroon:
Good to go, we verified numbers but it was on H200s and not H100s. Will keep current set.

Gregory-Pereira (approved):
/lgtm

PR #1881, opened 2026-06-20
docs(quickstart): add port to Proxy's IP curl

CefBoud:
The quickstart verification step was using the proxy’s default HTTP port 80, but the standalone chart exposes the user-facing  Envoy listener on [8081 by default](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/release-1.5/config/charts/standalone/values.yaml#L129) (hardcoded).

This PR updates the example request to `http://${IP}:8081/v1/completions` so it matches the exposed port.

### DIFF
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -81,7 +81,7 @@ kubectl run curl-debug --rm -it \
 Inside the shell, send a completion request:
 
 ```bash
-curl -X POST http://${IP}/v1/completions \
+curl -X POST http://${IP}:8081/v1/completions \
     -H 'Content-Type: application/json' \
     -d '{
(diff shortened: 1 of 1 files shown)


CefBoud:
Just realized  that the [following values](https://github.com/llm-d/llm-d/blob/092879417b1c0ddc25134f5ce28aaa092d3b5f02/guides/optimized-baseline/router/optimized-baseline.values.yaml#L4-L9) are intended to expose port 80:

```yaml
router:
  extraServicePorts:
    - name: http
      port: 80
      protocol: TCP
      targetPort: 8081
```

However, this runs into the same GAIE vs llm-d router values mismatch described in [this issue](https://github.com/llm-d/llm-d/issues/1882).

PR #1907, opened 2026-06-23
build llm-d image with vllm patch for DP supervisor

tessapham:
This image pulls in a hotfix for vllm's DP-aware scheduling: https://github.com/neuralmagic/vllm/commit/51f799c1a0c8a0476faf7e17eeb6a77983cdd778 and removes `/compat` from `LD_LIBRARY_PATH`, which was breaking deployments on CKS.


tessapham:
@chcost This has been tested with CUDA deployments on GKE.

tessapham:
@chcost This should point to vllm upstream when the [fix](https://github.com/vllm-project/vllm/pull/46628) is in, but will have wait for next version. I'll open an issue to note.

Opened tracking issue https://github.com/llm-d/llm-d/issues/1934.

PR #1916, opened 2026-06-23
Updated guides on how to enable monitoring on llm-d router

ahg-g:
also removed the special GKE note to reduce clutter


chcost:
great addition, thanks a lot @ahg-g! /approve

PR #1929, opened 2026-06-24
ci: let guide OWNERS trigger /test-nightly for their guide

kfirtoledo:
## Problem

`/test-nightly` requires repo **write/admin** access to run. Guide maintainers who own a guide but aren't repo collaborators (e.g. tiered-prefix-cache owners) currently can't trigger their guide's nightly on a PR — only admins can.

## Change

Extend the permission gate in `slash-test-nightly.yaml` so a commenter is allowed if **either**:
- they are a maintainer (write/admin) — unchanged, runs any nightly; **or**
- they are listed in the OWNERS file of **every** matched nightly's guide.

A nightly maps to its guide via its `standup_scenario` (→ `guides/<scenario>/OWNERS`). Access stays scoped: a tiered-prefix-cache owner can run `tiered-prefix-cache-*` but not other guides' nightlies. Nightlies with no guide mapping remain maintainer-only.

The gate now reads as intent:
```js
if (await isMaintainer() || await ownsAllMatchedGuides()) return 'ok';
```

### DIFF
--- a/.github/workflows/slash-test-nightly.yaml
+++ b/.github/workflows/slash-test-nightly.yaml
@@ -196,30 +196,86 @@ jobs:
         if: steps.parse.outputs.result == 'ok'
         id: perm
         uses: actions/github-script@v9
+        env:
+          MATCHES: ${{ steps.parse.outputs.matches }}
         with:
           result-encoding: string
           script: |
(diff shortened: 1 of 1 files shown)


kfirtoledo:
@liu-cong, @maugustosilva, could you please review?

maugustosilva (approved):

PR #1930, opened 2026-06-24
ci(tiered-prefix-cache): add nightlies for FS storage offloading

amirfr3:
## Summary

Adds a nightly E2E for the FS (storage) offloading path of `tiered-prefix-cache` on the IBM and GKE cluster. Small guide changes needed to make that path deployable with the existing ci flow.

## Changes

- CI flows added.
- Parameterized `${VARIANT}` in both vLLM-native commands so a single `cpu`/`fs` override flips between them.
- Storage class instructions now point at the in-tree `pvc.yaml` instead of an `envsubst` template.
- **PVC bundling.** The FS overlay (`modelserver/gpu/vllm/native/fs/base`) now pulls in the PVC as a kustomize resource so `kubectl apply -k` creates it together with the modelserver.


guides/tiered-prefix-cache/manifests/pvc/pvc.yaml
```diff
@@ -8,4 +8,5 @@ spec:
   resources:
     requests:
       storage: 9000Gi
-  storageClassName: ${STORAGE_CLASS} # Specify the storage class here, or remove this line to use the default storage class.
+  # Uncomment the following line to specify a storage class for the PVC. If you leave it commented out, the default storage class will be used.
```
kfirtoledo, line 11: Why did you remove it?

amirfr3: The ci won't run the "envsubst" command that puts the correct storage class in that, so I had to change it so it is valid without that command and the readme instructs the user to change the file themselves. Also, the gke cluster does not like it when the storage class is just an empty string, and omitting it works.

amirfr3:
@kfirtoledo @liu-cong @maugustosilva PTAL.
We tested the flows by overwriting the cpu offloading flows because new flows can't be tested in a pr

kfirtoledo:
lgtm,
Can you take a look at this PR @liu-cong, @Gregory-Pereira, @maugustosilva, @vMaroon?
Because to test the CI you have to actually run it. @amirfr3 tested it by editing the CPU files, and after it passed, it created a new file for storage.

PR #1937, opened 2026-06-24
bump wide-ep images

Gregory-Pereira:


liu-cong:
/test-nightly wide-ep-lws-gke-acc-gpu-vllm-x decode_pods=auto, prefill_pods=auto, harness=inference-perf

liu-cong (approved):

PR #1949, opened 2026-06-25
Update on-disk layout and add NOTE for llmdbenchmark installation

cheng-hsiang-chiu:
There are two updates and one NOTE added.

1. Update the on-disk layout in [benchmark.md](https://github.com/llm-d/llm-d/blob/main/helpers/benchmark.md) as the analysis folder is at the same level as results folder, as shown below.
Screenshot 2026-06-25 at 10 36 08 AM

2. Update the generated PNG names in the analysis folder as shown below.
Screenshot 2026-06-25 at 10 36 55 AM

3. Add a NOTE for the llmdbenchmark installation process. As the llmdbenchmark requires several dependencies, and all dependencies should have executable permission (`+x`). Otherwise, the installation would terminate early without clear error message, which leads to the installation failure. Then the subsequent command, `llmdbenchmark --version`, would fail to find `llmdbenchmark`.


Vezio (changes requested):
@cheng-hsiang-chiu the `README.md` updates look good!

I do have some additional concerns:

Please correct me if I'm wrong - but it appears the guide templates you've changed are only usable if one is still using the _phased out_ `run_only.sh`. 

Since we are now only using `llmdbenchmark`, those benchmark guide templates as they exist here today, would not be compatible. Additionally, benchmark guide templates should now live in `llm-d-benchmark` as we've seen with the other guides.

I believe the `agentic` benchmark guide template is one of the last remaining guides that need to be moved to `llm-d-benchmark` and that need to be converted to the appropriate configuration.

In the meantime, I'm ok with merging the `README.md` changes.

In a subsequent PR - the benchmark template guide changes should be merged to `llm-d-benchmark` so that transition gap is closed.

Thanks for this @cheng-hsiang-chiu !

CC: @maugustosilva

cheng-hsiang-chiu:
@Vezio, 

Thank you for your comments. 

This PR is meant to update the helper/benchmark.md only. But I mistakenly included updates from [PR#2024](https://github.com/llm-d/llm-d/pull/2024). I will close this PR. 

For the `run_only.sh` concern, I am using it because I would like to keep both agentic-serving guides aligned (https://github.com/llm-d/llm-d/blob/main/guides/agentic-serving/qwen3-coder-480b-tpu.md and https://github.com/llm-d/llm-d/blob/main/guides/agentic-serving/nemotron-3-ultra-550b-h200.md). As the `qwen3-coder-480b-tpu` is using `run_only.sh`, for now I would follow the usage.

cheng-hsiang-chiu:
This PR is closed because the main updates are covered in [PR#2024](https://github.com/llm-d/llm-d/pull/2024).

PR #1959, opened 2026-06-26
Add program-aware fairness scheduling and OTel trace replay benchmark to agentic-serving guide

pavanipenumalla:
## Summary

- Integrate `agent-identity` and `program-aware-fairness` plugins into both
  `agentic-serving.values.yaml` and `agentic-serving-gpu.values.yaml`.
- Add `benchmark-templates/otel-trace-replay.yaml`, an `inference-perf` config that replays
  real agentic traces from `Exgentic/agent-llm-traces`, carrying per-session identity
  headers to exercise the fairness stack end-to-end.
- Update README with program-aware fairness in the optimization stack table and a reference to
  the benchmark template.

## Related

- [agent-identity plugin](https://github.com/llm-d/llm-d-router/pull/1244) 
- [Program-aware fairness README](https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware)


guides/agentic-serving/README.md
```diff
@@ -55,4 +56,8 @@ cross-turn prefix reuse is actually exercised rather than assumed. The exact pre
 deployment (model, accelerator, and serving topology) rather than forced to be identical — see
 each sub-guide for its workload. Deployments are compared on program-level metrics — whole-session
 completion time and task throughput alongside TTFT and ITL. Replaying real agentic traces (program
-structure and tool-call timing from OpenTelemetry) is the direction for program-level evaluation.
+structure and tool-call timing from OpenTelemetry) is the direction for program-level evaluation —
+[`benchmark-templates/otel-trace-replay.yaml`](benchmark-templates/otel-trace-replay.yaml) replays
```
liu-cong, line 60: do we have benchmark results to share?

guides/agentic-serving/router/agentic-serving-gpu.values.yaml
```diff
@@ -46,6 +47,12 @@ router:
           name: cpu-prefix-cache-scorer
           parameters:
             prefixMatchInfoProducerName: cpu-prefix-cache-producer
+        - type: program-aware-fairness
+        featureGates:
+        - flowControl
```
liu-cong: do we have confidence that this generally works better than the existing config? We should benchmark the new config with the existing benchmark template and compare. 

If not, I suggest we add a new router config, and explain that this new config works better for the `Exgentic` workload.

guides/agentic-serving/README.md
praveingk: Hi @liu-cong , We did benchmark of Otel traces. The setup and benchmark results are here https://github.com/pavanipenumalla/agentic-bench/blob/experiments-v2/README.md

guides/agentic-serving/router/agentic-serving-gpu.values.yaml
praveingk: @liu-cong We are currently unable to benchmark with the existing config due to multiple issues :
1) Unavailability of the exact setup to benchmark.
2) Current workload in agentic serving guide does not have any agentic session-id. We are working on enabling that in inference-perf (https://github.com/kubernetes-sigs/inference-perf/issues/574)

As you have suggested, we can add a new config for exgentic traces and benchmark on a custom setup (4x replicas instead of 2x replicas we currently used).

vMaroon: Agreed that we should show iterative improvements through this kind of optimizations, @liu-cong can someone run the TPU guide with this approach? I'll schedule something for the Nemotron guide.

guides/agentic-serving/README.md
```diff
@@ -24,6 +24,7 @@ specific pressure of the agentic workload:
 | **[Tiered KV offloading](../tiered-prefix-cache/README.md)** | Offload KV cache beyond accelerator memory across tiers, so idle sessions restore on resume instead of recomputing prefill. |
 | **[Precise prefix-cache routing](../precise-prefix-cache-routing/README.md)** — advanced | An exact, global view of cache state, enabling session-centric orchestration and non-naive (beyond-LRU) KV-cache offloading & retention. |
 | **[P/D disaggregation](../pd-disaggregation/README.md)** — large models / interactivity | Separate prefill and decode pools so heavy prefill never stalls token generation, stabilizing ITL. |
+| **[Program-aware fairness](https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware)** — multi-session contention | Identify each agent session and enforce per-session LAS (Least Attained Service) queuing so that bursty sessions cannot starve concurrent ones, even under uneven arrival rates. |
```
vMaroon: I think the entry here should be on a higher level as there are several emerging "program-aware scheduling" techniques, and this is a major optimizations subdomain. The specific algorithm or technique used can be in the guide/deployment artifacts.

guides/agentic-serving/benchmark-templates/otel-trace-replay.yaml
```diff
@@ -0,0 +1,42 @@
...
+  request_timeout: 900
+  stages:
+    - concurrent_sessions: 600
+      num_sessions: 1000
+      session_rate: 10
+  worker_max_concurrency: 12000
+
+server:
+  type: vllm
+  model_name: Qwen/Qwen3.6-35B-A3B
```
vMaroon: This doesn't match any of the existing guides, we need to run benchmarks on them with the proposed configuration. If we want to change the benchmark harness/config, it should match the model/deployment currently established.

Lets coordinate on the slack channel.

praveingk:
@liu-cong @vMaroon I have added a separate guide for **program-aware** in https://github.com/llm-d/llm-d/pull/1959/commits/b568fd3aa1678384f473823bb4962f85e90d8fd9
This is an initial-cut with experiments on a single replica of 4 x H100s.

PR #1963, opened 2026-06-26
docs: remove experimental positioning from batch gateway guide

j-mok-dev:
## Why is this PR needed?
Batch Gateway has graduated, but the umbrella docs still had a few places that positioned it as experimental.
## What does this PR do?
- moves Batch Gateway out of the Experimental Guides section
- removes `[Experimental]` from the Batch Gateway guide title
- updates the root Batch Processing link to the workloads section
- updates the Batch Gateway OCI chart reference to `oci://ghcr.io/llm-d/charts/batch-gateway`
Async-related references are intentionally left unchanged for a later pass.
## How was this tested?
Manually verified the affected markdown files and cross-checked the chart path against the sibling `llm-d-batch-gateway` repo's chart and release metadata.
## Related Issues
Related to llm-d/llm-d-batch-gateway#501

Signed-off-by: Jooyeon Mok <jmok@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Model servers like [vLLM](https://docs.vllm.ai) and [SGLang](https://github.com/
 * **[Advanced KV-Cache Management:](https://llm-d.ai/docs/guides#advanced-kv-cache-management)** Increase the effective "working set size" for multi-turn requests with tiered offloading to CPU or disk and precise global indexing of the KV cache state.
 * **[Serving Large Models:](https://llm-d.ai/docs/guides#serving-large-models)** Optimize massive models (e.g., DeepSeek-R1, GPT-OSS) using prefill/decode disaggregation and wide expert-parallelism over fast accelerator interconnects.
 * **[Operational Excellence:](https://llm-d.ai/docs/guides#operational-excellence)** Ensure production stability with intelligent flow control for multi-tenant serving and proactive, SLO-aware autoscaling based on real-time inference signals.
-* **[Batch Processing:](https://llm-d.ai/docs/guides#experimental)** Efficiently manage large-scale offline inference with OpenAI-compatible Batch APIs and asynchronous processing to maximize hardware utilization.
+* **[Batch Processing:](https://llm-d.ai/docs/guides#workloads)** Efficiently manage large-scale offline inference with OpenAI-compatible Batch APIs and asynchronous processing to maximize hardware utilization.
 
 For a complete list of tested recipes and architectural patterns, see our [well-lit path guides](https://llm-d.ai/docs/guides). These guides provide benchmarked recipes and Helm charts to start serving quickly with best practices common to production deployments. Our intent is to eliminate the heavy lifting common in tuning and deploying generative AI inference on modern accelerators.
 
--- a/guides/README.md
+++ b/guides/README.md
@@ -33,12 +33,12 @@ We currently offer the following:
 Workload-centric guides — each provides the recommended, cohesive deployment for serving a workload, composing the capability guides above. See the [workload narratives](../docs/well-lit-paths/workloads/README.md) for overviews.
 
 * [Agentic Serving](./agentic-serving/README.md) - serve long, multi-turn, tool-using agentic workloads (e.g. coding agents) by composing prefix-aware routing, KV-cache offloading, and P/D disaggregation.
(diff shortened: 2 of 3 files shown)


lioraron:
@ahg-g @robertgshaw2-redhat Can you please review this PR?

maugustosilva:
This guide is not part of the CI/CD nightly benchmarks yet. The standard operating procedure adopted so far has been keeping the `[Experimental]` tag until a guide is tested within the CI/CD infrastructure on the project.

chcost:
@j-mok-dev changes to doc look good. For the guide itself, +1 to Marcio's comment. We should have it part of the nightly tests before we remove the experimental label.

lioraron:
@maugustosilva @chcost Thanks for the context.
Batch-gateway already has nightly e2e and regression tests running in its own repo ([ci-integration-tests.yml](https://github.com/llm-d/llm-d-batch-gateway/blob/main/.github/workflows/ci-integration-tests.yml) — these deploy the full stack (llm-d Router, vLLM simulator, PostgreSQL, Redis, batch-gateway) on a Kind cluster and run e2e scenarios.

As far as I saw the guides that have graduated (optimized-baseline, flow-control, workload-autoscaling, etc.) have nightly tests only in the `llm-d/llm-d` infra framework and not in their own repos, while batch-gateway has the opposite — nightly tests in its own repo but not in the infra framework.

Would the existing batch-gateway nightly tests satisfy the requirement for removing `[Experimental]`, or do we need to be added to the `llm-d/llm-d` infra framework?

j-mok-dev:
Thanks @maugustosilva @chcost for the feedback. @lioraron created an issue on the batch gateway side(llm-d/llm-d-batch-gateway#519).

## The challenge

The `reusable-ci-nightly-benchmark.yaml` workflow is designed for a two-component stack (model server + Router). Batch-gateway requires additional infrastructure (PostgreSQL, Redis, object storage, and the batch-gateway services which are apiserver, processor and gc)  and the reusable workflow has no mechanism to deploy extra Helm charts beyond the Router.

Additionally, the "run" harnesses (`inference-perf`, `guidellm`, etc.) test online inference. Batch-gateway needs a different validation: submit a batch job → poll to completion → verify results.

## Proposal

I'd like to create a **standalone nightly workflow** in this repo that:
1. Deploys the full batch-gateway stack on GKE via Helm (same approach as our existing [nightly integration tests](https://github.com/llm-d/llm-d-batch-gateway/blob/main/.github/workflows/ci-integration-tests.yml), but on GKE instead of Kind)
2. Runs a deployment-correctness sanity test (batch job lifecycle)
3. Integrates with `reusable-update-badge.yaml` and `reusable-query-success-past-runs.yaml` for badge/status tracking

This gives the same observability (nightly pass/fail badges, history) as other graduated guides while actually testing batch-gateway end-to-end.

## Question

Does a standalone workflow with badge/status integration satisfy the graduation requirement? Or must it specifically call `reusable-ci-nightly-benchmark.yaml`?

maugustosilva:
Hi @j-mok-dev thanks for the reply, this is **exactly** the discussions we should be having with all guide owners!

So, the goal with a **nightly** benchmark is four-fold. 
a) Ensure that guide, as written (verbatim or as close as possible) results in a "functional" `llm-d` stack (i.e., the model-serving `pods` do not crash and can respond to requests)
b) Ensure that, in addition to "basic functionality", the `llm-d` stack just stood up does behave in accordance with the features highlighted on the guide (e.g., KV cache blocks are exchanged between `pods` in case of `pd-disaggregation`)
c) Track actual performance regressions on the different components which form an stack
d) Provide data for analysis and visualization on `llm-d-prism`. 

That being said, any "standalone" limited testing done by owners of individual repositories are not sufficient. The core idea here is to be able to answer the question "is the guide, as presented, broken?". Given the fact that all other guides are following the standard set by `reusable-ci-nightly-benchmark.yaml`, we need to further discuss and understand why this particular guide cannot follow the same procedure.

One last thing. As you very pertinently pointed out, each guide owner should also provide a "feature" test too. It could be something as simple as a shell script, or some specific workload coupled with some sort of parsing of the logs. Be as it may, there is the need for this additional component. As of today, virtually no other guide owner has provided us with such tooling, but it is our goal to make it also the standard operating procedure.

ahg-g:
@lioraron can you please advice if this is ready to graduate?

PR #1969, opened 2026-06-28
fix: Add missing `update_badge` parameters to ci workflows

amirfr3:
in #1951, A new parameter was added to the /test-nightly default parameters sent to the workflow dispatch. A couple of the workflows did not have that parameter added to them and so running the `/test-nightly` command on them fails. This adds the parameter to the rest of the workflows.


amirfr3:
@diegocastanibm @maugustosilva PTAL

maugustosilva (approved):

PR #1984, opened 2026-07-01
fix: Move the router's grafana dashboard to llm-d/llm-d

gyliu513:
Fixed part of https://github.com/llm-d/llm-d-router/issues/1855

Will create another PR to remove the dashboard in llm-d-router repo.

/cc @ahg-g @LukeAVanDrie


ahg-g:
/approve

Thanks, so this will just work, no need to add it to any script, right?

gyliu513:
@ahg-g Yes, that's right, no changes to `load-llm-d-dashboards.sh` are needed. The script already loads every `*.json` file under `guides/recipes/observability/grafana/dashboards/`, so placing `llm-d-inference-gateway.json` there is sufficient.

I also added it to the dashboard table in `docs/operations/observability/setup.md` for discoverability.

I'll follow up with a separate PR in llm-d-router to remove the duplicate dashboard there (llm-d/llm-d-router#1855).

PR #1987, opened 2026-07-01
Verification scripts for ci workflows

amirfr3:
Add nightly E2E verification scripts, for running in ci workflow through https://github.com/llm-d/llm-d-infra/pull/215.

# What it does

Runs between the harness `Run` and `Teardown` steps in `reusable-ci-nightly-benchmark.yaml`, while pods and PVCs are still up. Per-scenario `verify.py` scripts run pass/fail checks against the run's `metrics_summary.json`, check the cluster state, etc. the exit code gates the workflow.

# Layout

- `verify_helpers.py` - shared library. MetricsSummary wraps metrics_summary.json and exposes helpers for metric comparison, Check dataclass for readable assertions, kubectl helpers.
- `_template/verify.py` - minimal scaffold; copy into <scenario>/verify.py and edit.
- `tiered-prefix-cache/verify.py` - Follows verification logic in the tiered-prefix-cache guide:
   - For storage offloading - asserts the offload PVC is Bound, kv-cache files landed on it. 
   - For both offloading types - vllm:kv_offload_{store,load}_bytes counters moved on at least one pod. 
- `README.md` - API reference and how to wire up a new scenario.
- `verify-locally.sh` - run the verification scripts locally - for testing purposes


amirfr3:
@maugustosilva PTAL

.github/scripts/nightly-e2e-verification/verify_helpers.py
```diff
@@ -0,0 +1,305 @@
...
+    Only `llmdbenchmark run` invocations produce a `results/` directory whose
+    parent is the timestamped `<user>-YYYYMMDD-HHMMSS-mmm` workspace subdir
+    llmdbenchmark mints per invocation. Newest by mtime wins.
+    """
+    if not workspace:
+        error("workspace not set (LLMDBENCH_WORKSPACE)")
+        return None
+    ws = Path(workspace)
+    for results in sorted(ws.rglob("results"), key=lambda p: -p.stat().st_mtime):
+        if not results.is_dir() and not results.parent.name.startswith(getpass.getuser()):
```
kfirtoledo: The skip condition doesn't match the intent in the docstring/comment ("only a `results` **dir** whose parent is `<user>-<ts>`"). With `and`, a decoy `results` entry that is newest-by-mtime but is a file, or whose parent isn't user-prefixed, is **not** skipped — it then either hits the `break` and shadows the real results dir, or raises `NotADirectoryError` on `.iterdir()`. Should be `or`:

```suggestion
        if not results.is_dir() or not results.parent.name.startswith(getpass.getuser()):
```

.github/scripts/nightly-e2e-verification/README.md
```diff
@@ -0,0 +1,168 @@
...
+Verification runs between the harness `Run` step and `Teardown` in
+`reusable-ci-nightly-benchmark.yaml` (llm-d-infra), so pods and PVCs are
+still up and `kubectl` is configured against the test cluster.
+
+## Layout
+
+```
+nightly-e2e-verification/
+├── README.md
+├── run-locally.sh            ← wrapper: sets env vars, runs a verify.py on your laptop
```
kfirtoledo: The script is `verify-locally.sh`, but the README (lines 14, 36, 42, 46, 58) and the script's own usage/`--help` header (`verify-locally.sh:12,15,18,23`) all call it `run-locally.sh`. Following the docs (`./run-locally.sh …`) fails. Please rename the references (or the file) so they match — looks like a leftover from the "script name change" commit.

.github/scripts/nightly-e2e-verification/README.md
```diff
@@ -0,0 +1,168 @@
...
+# Reads metrics.aggregated[metric][aggregate], threshold-checks it.
+metrics.check_aggregated("vllm:time_to_first_token_seconds", "p99", "<=", 2.0)
+# -> Check(name="vllm:time_to_first_token_seconds.p99", passed=..., detail="1.85 <= 2.00")
+
+# Pulls metrics.per_pod[pod][metric][aggregate] for every pod, combines with
+# combine(values), threshold-checks the result. Default combine funciton is max
+# ("did any pod hit the bound?"). Pass any callable that consumes an iterable
+# of floats: max, min, sum, statistics.mean, or a lambda around functools.reduce.
+metrics.check_per_pod("vllm:kv_offload_store_bytes_total", "max", ">", 0.0)
+metrics.check_per_pod("vllm:kv_cache_usage_perc", "mean", "<=", 80.0, reduce=statistics.mean)
```
kfirtoledo: `check_per_pod`'s parameter is `combine=`, not `reduce=` — this exact line raises `TypeError: unexpected keyword argument 'reduce'`. (Verified against the signature at `verify_helpers.py:184`.)

```suggestion
metrics.check_per_pod("vllm:kv_cache_usage_perc", "mean", "<=", 80.0, combine=statistics.mean)
```

.github/scripts/nightly-e2e-verification/verify_helpers.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+def get_vllm_version(namespace: str, pod: str) -> tuple[int, ...] | None:
+    out = kubectl(
+        ["exec", "-n", namespace, pod, "--", "python3", "-c",
+         "import vllm; print(vllm.__version__)"],
+        timeout=10,
+    ).strip()
+
+    version = tuple(int(x) for x in out.split(".") if x.isdigit())
```
kfirtoledo: `tuple(int(x) for x in out.split(".") if x.isdigit())` drops non-numeric components, so `0.24.0rc1` → `(0, 24)`, which compares `< (0, 24, 0)` → treated as **old** metrics. Release builds are fine, but an rc/dev image would silently pick the wrong metric name. Consider parsing with `packaging.version` or stripping the suffix off the patch component. Non-blocking.

.github/scripts/nightly-e2e-verification/README.md
```diff
@@ -0,0 +1,168 @@
+# Nightly E2E verification scripts
+
+Verification scripts for running in the verification step in the nightly e22 ci workflow.
```
kfirtoledo: A few typos worth a quick sweep: `e22` → `e2e` here (README:3), `funciton` → `function` (README:108), and the table header `TRESH` → `THRESH` in `verify_helpers.py:233`. Also the tiered-prefix-cache docstring (`tiered-prefix-cache/verify.py:18`) says `vllm:kv_offload_store_bytes.max` but the code checks `vllm:kv_offload_store_bytes_total`.

.github/scripts/nightly-e2e-verification/verify_helpers.py
```diff
@@ -0,0 +1,305 @@
+"""Helpers for nightly E2E verification scripts.
```
kfirtoledo, line 1: This PR is verification tooling but ships no tests of its own. Most of `verify_helpers.py` is pure and trivially unit-testable, and two of the bugs in this PR live exactly there — the `find_results_dirs` boolean filter (line 76) and the `get_vllm_version` parser (line 94, e.g. `0.24.0rc1` → `(0, 24)` → mis-classified as old metrics). A small `test_verify_helpers.py` covering:
- `check_aggregated` / `check_per_pod` — invalid op, missing metric, missing aggregate, and the `combine=` reduction,
- `MetricsSummary` view-building from a sample `metrics_summary.json` (incl. the `_info.status == "no_data"` path),
- `get_vllm_version` across `0.11.0` / `0.24.0` / `0.24.0rc1`,
- `find_results_dirs` with a `tmp_path` fixture (real dir, decoy file, non-user parent),

would have caught both bugs and pins the check semantics the scenario scripts depend on. Since there's no pytest in the repo yet, stdlib `unittest` keeps it dependency-free — happy to defer the CI wiring to a follow-up.

.github/scripts/nightly-e2e-verification/tiered-prefix-cache/verify.py
```diff
@@ -0,0 +1,206 @@
...
+    version_str = ".".join(str(x) for x in vllm_version)
+    checks.append(v.Check(
+        name=f"vLLM version - {version_str}, kv-cache metrics (new/old)",
+        passed=True,
+        detail=f"{'old' if old_metrics else 'new'}",
+    ))
+    if old_metrics:
+        checks.append(metrics.check_per_pod("vllm:kv_offload_total_bytes_total", "max", ">", 0.0))
+    else:
+        checks.append(metrics.check_per_pod("vllm:kv_offload_store_bytes_total", "max", ">", 0.0))
```
kfirtoledo, line 211: Per the PR description, the current workload doesn't trigger offloading, so this metric assertion (`vllm:kv_offload_store_bytes_total > 0`) can't pass yet, and the load-side counter is commented out (line 454). That leaves the smoke check (`num_requests_running.count > 0`) as effectively the only thing gating the nightly. Is the plan to keep `verify_script` unset in the caller until a triggering workload lands, so this doesn't turn the nightly red — and is there a tracking issue for enabling the real checks?

.github/scripts/nightly-e2e-verification/tiered-prefix-cache/verify.py
```diff
@@ -0,0 +1,206 @@
...
+        v.error(f"Failed to get vllm version from pod {pods[0]}")
+        return 1
+    old_metrics = vllm_version < (0, 24, 0)
+    version_str = ".".join(str(x) for x in vllm_version)
+    checks.append(v.Check(
+        name=f"vLLM version - {version_str}, kv-cache metrics (new/old)",
+        passed=True,
+        detail=f"{'old' if old_metrics else 'new'}",
+    ))
+    if old_metrics:
```
kfirtoledo, line 208: The nightly always runs the latest vLLM image, so this `old_metrics` (`< 0.24.0`) branch never executes in CI — `kv_offload_total_bytes_total` is dead code in the nightly context. It's also not the same quantity as the new branch: old `total` aggregates all offload traffic, while new `store` is only the write side (the schema split `total` into `store` + `load`). Suggest either dropping the old branch or marking it `# TODO: remove once all CI targets are >= 0.24.0` so it's clear it only exists for local runs against older builds.

.github/scripts/nightly-e2e-verification/tiered-prefix-cache/verify.py
```diff
@@ -0,0 +1,206 @@
...
+    under KV_CACHE_DIR from inside the pod, and report their total size.
+
+    Mirrors the guide's storage-offload verification step.
+    """
+    out = v.kubectl(
+        ["exec", "-n", namespace, pod, "--", "sh", "-c",
+         f"find {KV_CACHE_DIR} -type f 2>/dev/null | wc -l; "
+         f"du -sh {KV_CACHE_DIR} 2>/dev/null | awk '{{print $1}}'"],
+        timeout=30,
+    ).strip().splitlines()
```
kfirtoledo: `find … 2>/dev/null` + `int(out[0])` reports `0 files` for three different cases: empty dir, missing `KV_CACHE_DIR`, and a failed/timed-out `kubectl exec`. Consider at least guarding `[ -d $KV_CACHE_DIR ]` so a missing mount is distinguishable from a genuinely empty one. (Largely resolved if `kubectl()` signals failure — see the helper comment.)

.github/scripts/nightly-e2e-verification/tiered-prefix-cache/verify.py
```diff
@@ -0,0 +1,206 @@
...
+        return 1
+    # PVC-side checks only apply in storage mode. The PVC is RWX and shared
+    # across every model pod, so we only need to check one of them. 
+    if storage_mode and pods:
+        checks.append(check_pvc_is_bound(namespace, pods[0]))
+        checks.append(check_pvc_has_data(namespace, pods[0]))
+
+    # Metric checks: both counters should have moved on at least one pod.
+    vllm_version = v.get_vllm_version(namespace, pods[0])
+    if vllm_version is None:
```
kfirtoledo, line 196: `get_vllm_version` is only used to pick a metric name, but returning `None` here `return 1`s and reds the whole nightly on a flaky `kubectl exec … import vllm`. A metric-name selector shouldn't be able to fail the run — prefer defaulting to the new metrics when the version is unknown, e.g. treat `None` as new (`old_metrics = vllm_version is not None and vllm_version < (0, 24, 0)`) and make the `version_str`/Check below tolerate `None` rather than aborting.

.github/scripts/nightly-e2e-verification/verify_helpers.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+# ---------------------------------------------------------------------------
+# kubectl helpers
+# ---------------------------------------------------------------------------
+
+def kubectl(args: list[str], timeout: int = 30) -> str:
+    try:
+        r = subprocess.run(["kubectl", *args], capture_output=True, text=True, timeout=timeout)
+        return (r.stdout or r.stderr or "").rstrip()
```
kfirtoledo: `return (r.stdout or r.stderr or "")` never checks `r.returncode`, so a failed `kubectl` (exit 1, message on stderr) returns that stderr text as if it were normal output — and callers (`get_vllm_version`, `find_offload_pvc`, `get_model_pods`, `check_pvc_*`) then `.split()`/`int()` it as data. Signal failure distinctly:

```suggestion
        return r.stdout.rstrip() if r.returncode == 0 else ""
```
This also cleans up the PVC-check conflation noted on `tiered-prefix-cache/verify.py`.

kfirtoledo (changes requested):
looks good, minor comments

.github/scripts/nightly-e2e-verification/tiered-prefix-cache/verify.py
amirfr3: I was wrong, i need to change the PR desc. The current workflow should cause offloading, so it will pass. loading will not pass as a lot of tokens are needed to get eviction to happen (and also, the llmdbenchmark workloads don't let you do something specific like sending some prefix, then a bunch of random tokens to get the prefix evicted, then that token again)

I think that in general, if the verify script does not have any added value, it should just be left unset.

.github/scripts/nightly-e2e-verification/tiered-prefix-cache/verify.py
amirfr3: I think that None should be the old version - as the old version is still available in 0.24.0, it will always work (until it is removed)

.github/scripts/nightly-e2e-verification/tiered-prefix-cache/verify.py
amirfr3: As i understand it, currently the guide yamls pin the vllm version to 0.23.0, and the nightly take whatever the guides use. the tpu version is a bit weirder (its like a branch name, but then its safer to assume old metrics.
check [guides/recipes/modelserver/components/images/gpu-vllm/kustomization.yaml](https://github.com/llm-d/llm-d/blob/main/guides/recipes/modelserver/components/images/gpu-vllm/kustomization.yaml#L6)

kfirtoledo:
lgtm, @liu-cong , @Gregory-Pereira @maugustosilva , can you review (this link to  a commit in the infra https://github.com/llm-d/llm-d-infra/pull/215)

kfirtoledo (commented):
lgtm

kfirtoledo (approved):

liu-cong:
We are experimenting with a mechanism to make the user guide machine-readable: https://github.com/llm-d/llm-d/pull/1988

We should add verifications such as verifying the pvc is bound in the guide itself, because a user would need to verify that as well. 

I believe that as a principle, we should add any verifications required in the guide itself, so that both the CI/CD and the guide is consistent. I would like to learn if there is a case that the CI/CD needs to do additional verification than the guide. 

@maugustosilva @Vezio

maugustosilva (approved):

PR #1988, opened 2026-07-01
[feat] Structured Guides for Machine and Human Readable Format

Vezio:
## Problem Solved

- Consistent user experience across guides. Every well-lit-path guide will now conform to a shared schema, validated by companion `guide-check-yaml.py` and `guide-check-readme.py` scripts. Structural drift can't slip past review.

- A programmatic interface for automation. The machine-readable guide.yaml lets CI runners, deployment tooling, and other non-human consumers install a chosen guide by parsing structured data instead of scraping bash out of dynamic prose that is not consistent, and subject to drift.

- Sets up an entry point to have in-line `AI Agent Hints per guide` for autonomous agents deploying well-lit-paths by easily integrating this inside of our `YAML` schema per guide. This will directly supplement skills.

## Provided in this PR

Introduces a machine-readable representation for guides paired with the existing human-readable READMEs, plus tooling to keep them in sync and validated.  Every bash code block in a guide's README is now emitted from a companion `guide.yaml`. The YAML is now the source of truth for installation commands and loose variables **not** already included in the `env.sh`. The README owns the prose that surrounds them.

### Three tools
- **`guide-check-yaml.py`** — validates the YAML against the schema: known keys, well-formed steps, `when:` references against declared variables, sensitive var placeholders, `default:`/`values:` consistency.
- **`guide-render.py`** — fills the README from the YAML. Supports `--dry-run` (preview) and `--check` (non-zero exit if the README is stale) for CI.
- **`guide-check-readme.py`** — validates the rendered README: marker pairing, path resolution against the YAML, well-formed ` ```bash ` fences with optional CI-skip wrappers.

Each of the above scripts are idempotent, and will report errors, updates, and other status as appropriate to the user.

### Already Supports Incremental Adoption

Existing guides can adopt the marker system one block at a time. Any bash fence *without* markers stays untouched by the renderer, so migration proceeds section by section rather than as a single rewrite.

See the documentation https://github.com/Vezio/llm-d/blob/structured-guide-implementation/guides/templates/README.md and example using `optimized-baseline` as a first implementation.

## The two-file guide model explained

Each guide becomes:

- **`guide.yaml`** — declarative schema describing environment variables, prerequisites, deploy steps, verification, benchmarking, and cleanup as structured data.

- **`README.md`** — a free-form prose with paired HTML-comment markers identifying where bash blocks belong. Content outside markers is preserved byte-for-byte on render.

Bash blocks appear between paired markers of the form:

```
<!-- guide:<yaml-path> start -->
…rendered from YAML…
<!-- guide:<yaml-path> end -->
```

Path notation follows the YAML shape — `env.static`, `deploy.standalone`, `verify.endpoint.gateway`, `prerequisites.secrets`, and so on.

## Documentation

`guides/templates/README.md` is the authoring guide. It covers the two-file model, the marker system, converting existing guides to markers, step and filter syntax with rendered examples, all three scripts, and a schema quick reference — every rule anchored to a concrete snippet from `optimized-baseline/`. A short pointer in `guides/README.md` directs authors to the doc.

- See: https://github.com/Vezio/llm-d/blob/structured-guide-implementation/guides/templates/README.md

## "Gross" Stuff in `YAML`

So today there are some `commented` sections within bash - after speaking with @liu-cong and @maugustosilva - these will soon go away once we refactor directory structures as well as properly push various control orchestration kustomize and helm (for example,  in `llm-d router` to arguments for `monitoring` should be parameterized via `--set` etc...).

## Demonstration of Tooling

```
scripts/guide-check-yaml.py    guides/optimized-baseline/guide.yaml
scripts/guide-render.py        --yaml guides/optimized-baseline/guide.yaml --readme guides/optimized-baseline/README.md
scripts/guide-check-readme.py  --yaml guides/optimized-baseline/guide.yaml --readme guides/optimized-baseline/README.md
scripts/guide-render.py        --yaml guides/optimized-baseline/guide.yaml --readme guides/optimized-baseline/README.md --check
```


guides/README.md
```diff
@@ -75,6 +75,21 @@ images:
     newTag: nightly-20260601
 ```
 
+## Authoring a Guide
```
liu-cong, line 79: I suggest referencing the templates/README.md directly here to avoid duplicates/out-of-sync

guides/templates/README.md
```diff
@@ -0,0 +1,368 @@
...
+mkdir guides/my-guide
+cp guides/templates/guide.template.yaml   guides/my-guide/guide.yaml
+cp guides/templates/README.template.md    guides/my-guide/README.md
+
+# edit both files: name, commands, prose
+$EDITOR guides/my-guide/guide.yaml guides/my-guide/README.md
+
+# validate + render
+scripts/guide-check-yaml.py    guides/my-guide/guide.yaml
+scripts/guide-render.py        --yaml guides/my-guide/guide.yaml --readme guides/my-guide/README.md
```
liu-cong: can we have the render script also call the validation scripts by default, and fail the render if validation fails? So in a common workflow I only call `guide-render.py`

guides/templates/README.md
```diff
@@ -0,0 +1,368 @@
...
+```yaml
+name: <guide-name>              # required
+
+env:                            # required
+  static:                       # required — variable declarations
+    VAR:                        # constant (scalar form)
+    VAR: { default: <v> }       # overridable default
+    VAR: { default: <v>, values: [<v1>, <v2>] }   # categorical (values gates when:)
+    VAR: { default: PLACEHOLDER, sensitive: true } # secret — compiler fails if unset
+  source: ["<path>", …]         # optional — `source <path>` lines (verbatim)
```
liu-cong, line 427: We should enforce the shared `env.sh` for every guide. I think we should not expose this and keep the schema minimal. The render script will simply apply the `source env.sh`, and the schema should simply have a `env` section with key/value pairs

guides/templates/README.md
```diff
@@ -0,0 +1,368 @@
...
+
+prerequisites:                  # optional — flat list or map of named sub-groups
+  - run: <bash>
+  # or:
+  namespace:
```
liu-cong, line 432: is it important to have a `namespace` field, or can we just use a `run` command with `kubectl create ns $NAMESPACE`?

guides/templates/README.md
```diff
@@ -0,0 +1,368 @@
...
+    - run: <bash>
+  secrets:
```
liu-cong, line 434: same question, is `secrets` a special field or can we just use a `run` command?

guides/templates/README.md
```diff
@@ -0,0 +1,368 @@
...
+    - run: <bash>
+
+deploy:                         # required — sub-groups (mode-named or common)
+  standalone:                   # mode names should match verify.endpoint keys
```
liu-cong, line 438: Instead of hard coding the proxy options, I think we should use the generic `when` condition to match dimensional values such as accelerator type, proxy type, etc. 

My vision is that eventually we will move to contextual guides instead of one giant guide for every environment, when users picks their deployment environment, we can customize the guide to only information relevant to their environment.

guides/optimized-baseline/guide.yaml
```diff
@@ -0,0 +1,164 @@
+name: optimized-baseline
+
+_lists:
+  non_gpu: &non_gpu [amd, xpu, hpu, "tpu/v6", "tpu/v7", cpu]
+
+env:
+  static:
+    BRANCH: main
```
liu-cong, line 8: This can be a follow up but can we also support extending from a "base yaml" that define all the common configuration such as `BRANCH, REPO_ROOT, HF_TOKEN, BENCHMARK_HARNESS` and pretty much the entire `prerequisites` section. This not only reduces complexity per guide, but also helps enforce consistency across guides.

guides/optimized-baseline/README.md
```diff
@@ -46,65 +46,138 @@ This guide includes configurations for the following accelerators:
...
+
+<!-- guide:env.static start -->
+```bash
+export BRANCH=main
+export REPO_ROOT=$(realpath $(git rev-parse --show-toplevel))
+export GUIDE_NAME=optimized-baseline
+export NAMESPACE=llm-d-optimized-baseline
+export HF_TOKEN=HF_TOKEN_PLACEHOLDER
+export MONITORING_VALUES=
+export PROVIDER_NAME=gke # options: none, gke, agentgateway, istio
```
liu-cong, line 78: I suggest merging this and the GATEWAY_CLASS below into one: `ROUTER_TYPE=standalone # options: standalone, gw_gke, gw_agentgateway, gw_istio`

liu-cong:
For this to be successful, we need to 1) make it simple and intuitive enough so the guide writers can effortlessly learn this. 2) the render logic needs to be simple enough so it does not add complexity to our long-term maintenance. Can we consider simplifying it to `env` `prereq` `install` and `verification` blocks that are just free form bash scripts, with optional `when` conditional switch, and eliminate other fields?

On the other side, can we investigate whether there are any tools out there that solve similar problems? A quick prompting with AI and it suggests Markdoc/Docusaurus/Cog for conditional content rendering and Runme for executable doc (without needing to maintain 2 files). I haven't tried any of these but it's worth looking into.

mamy-CS:
Really like this. The generated README direction makes drift structurally impossible, which is stronger than any lint we'd write, I think. 
Two requests from the runner side, both cheap now and a migration later. @lionelvillard was building a guide runner that executes a guide's steps locally in kind and asserts the outcome. That matters more than it sounds. 8 of our 16 guides have no nightly at all, and the 8 that do need GPUs on a self-hosted cluster. So today, most guides have no automated validation, and no guide can be validated pre-merge. (For example, our WVA guide sat in main for weeks recommending a KEDA ScaledObject that had never been applied by anything.)

For the schema to serve that:
1. MODEL_SERVER needs a sim value. It's values: [vllm, sglang, trtllm] today, so guide-check-yaml.py rejects when: { MODEL_SERVER: [sim] } and the GPU-free path can't be expressed at all. The whole local story rests on llm-d-inference-sim.
2. An environment/profile axis. skip_in: [ci] is the only context tag. Kind-specific steps (helm-install KEDA locally vs. assume the OCP operator) need somewhere to live — either ENV: { default: existing, values: [kind, existing, ocp] } in env.static with when: filters, or generalizing skip_in into context tags.

mamy-CS:
I converted `README.wva.md` to a guide.yaml as a stress test, it's our most complicated guide. Good news: it passed `guide-check-yaml.py` first try, renders cleanly, round-trips through guide-check-readme.py. `WVA_NAMESPACE: ${NAMESPACE}` works, the `PLATFORM: [k8s, ocp]` split works, and a 15-line polling assertion with exit 1 drops into `verify.tests` untouched. Expressiveness is not the problem.

But it surfaced one thing worth deciding now. `When:` is lossy in the rendered `README`. Mutually exclusive steps flatten into a single fence:
For example in wva readme:
```
# only when SCALER=keda:
kubectl apply -k .../optimized-baseline-autoscaling/keda/${PLATFORM} -n ${NAMESPACE}

# only when SCALER=hpa:
kubectl apply -k .../optimized-baseline-autoscaling/hpa -n ${NAMESPACE}

```
A human picks one. A runner scraping the rendered markdown runs both, which for this guide is destructive (the guide explicitly warns against it). So any runner has to consume `guide.yaml` directly, where `when:` is still machine-readable. Worth stating explicitly in the authoring doc, since `skip_in: [ci]`'s comment implies README-scraping runners are the expected consumer.

Smaller findings from the conversion:
- `verify.endpoint` - the doc says it's required whenever `verify: exists`, but the checker is happy with just `verify.tests`. The checker's right (WVA has no endpoint to curl), so the doc just needs fixing.
- No way to say "install X first." WVA sits on top of `optimized-baseline` right now for example and nothing in the schema expresses that.
- No "edit this file" step. The guide asks you to update serverAddress in the ScaledObject before applying - a runner can't do that. It'd have to become a yq command, which is probably a healthy forcing function maybe. 
- Ordering is implicit. For example on this guide, KEDA's CRDs have to exist before the WVA controller boots, and the only thing enforcing that is list order.

cc: @lionelvillard

guides/templates/README.md
Vezio: As per discussion via slack - yes this is important for logical separation.

guides/templates/README.md
Vezio: As per discussion via slack - yes this is important for logical separation.

guides/templates/README.md
Vezio: This can be done as a future item as the guides currently today do not support this - this PR is intended to be a minimal enhancement that brings **structure** and **reproducibility** - as well as a very important API that CICD and non-human consumers can easily utilize - to the guides as they're written today.

Future enhancements to clean up the guides and maintain them will be added once merged.

guides/optimized-baseline/guide.yaml
Vezio: Yes! Totally agree - this will be a future feature.

guides/optimized-baseline/README.md
Vezio: While they both have similar fields, it's not a one-to-one yet because of a few options in `llm-d-benchmark`. We can fix this eventually once the two are aligned.

guides/templates/README.md
Vezio: Agree - but again I must say:

This can be done as a future item as the guides currently today do not support this - this PR is intended to be a minimal enhancement that brings **structure** and **reproducibility** - as well as a very important API that CICD and non-human consumers can easily utilize - to the guides as they're written today.

Future enhancements to clean up the guides and maintain them will be added once merged.

guides/templates/README.md
Vezio: I will add that in shortly today.

Vezio:
@liu-cong RE:

>For this to be successful, we need to 1) make it simple and intuitive enough so the guide writers can effortlessly learn this. 2) the render logic needs to be simple enough so it does not add complexity to our long-term maintenance. Can we consider simplifying it to env prereq install and verification blocks that are just free form bash scripts, with optional when conditional switch, and eliminate other fields?

As per our discussion last week - we can not allow large free form bash blocks within the YAML directly - instead guide owners will leverage subsequent `run` marks as defined in the example of converting Optimized Baseline - the reason for this many but here are a few:

- This allows us to have a way to ensure _each_ command is run successfully from a non-human consumer perspective
- This allows clear separation of concerns
- Even though, in the YAML, a maintainer will need to specify `run` - when the guide is finally rendered into it's `README.md` format, using the tools provided here, those `run` sections are then collapsed into a free-form bash field like we currently have on existing guides. The `run` statements offer a way for automation and non-humans to easily and reliably pull units of execution. Without it - a maintainer would be effectively building `in-line` bash scripts into a YAML - and at that point .... the maintainer should actually just maintain `.sh` scripts rather than embed them (AKA an installer).

We have our first consumer today here https://github.com/llm-d/llm-d/pull/1779 and managed to very quickly and cleanly generated a guide for FMA that is reproducible (and clearly structured) and consumable by automation.

Vezio:
@mamy-CS RE:

>Really like this. The generated README direction makes drift structurally impossible, which is stronger than any lint we'd write, I think.
Two requests from the runner side, both cheap now and a migration later. @lionelvillard was building a guide runner that executes a guide's steps locally in kind and asserts the outcome. That matters more than it sounds. 8 of our 16 guides have no nightly at all, and the 8 that do need GPUs on a self-hosted cluster. So today, most guides have no automated validation, and no guide can be validated pre-merge. (For example, our WVA guide sat in main for weeks recommending a KEDA ScaledObject that had never been applied by anything.)

>For the schema to serve that:

>    MODEL_SERVER needs a sim value. It's values: [vllm, sglang, trtllm] today, so guide-check-yaml.py rejects when: { MODEL_SERVER: [sim] } and the GPU-free path can't be expressed at all. The whole local story rests on llm-d-inference-sim.
>    An environment/profile axis. skip_in: [ci] is the only context tag. Kind-specific steps (helm-install KEDA locally vs. assume the OCP operator) need somewhere to live — either ENV: { default: existing, values: [kind, existing, ocp] } in env.static with when: filters, or generalizing skip_in into context tags.


Thanks for this feedback - I appreciate you diving in and applying this to `WVA`. I'll take those points and fix them in a subsequent commit.

Vezio:
@mamy-CS RE:

> But it surfaced one thing worth deciding now. When: is lossy in the rendered README. Mutually exclusive steps flatten into a single fence:
For example in wva readme:

> # only when SCALER=keda:
> kubectl apply -k .../optimized-baseline-autoscaling/keda/${PLATFORM} -n ${NAMESPACE}

> # only when SCALER=hpa:
> kubectl apply -k .../optimized-baseline-autoscaling/hpa -n ${NAMESPACE}

> A human picks one. A runner scraping the rendered markdown runs both, which for this guide is destructive (the guide explicitly warns against it). So any runner has to consume guide.yaml directly, where when: is still machine-readable. Worth stating explicitly in the authoring doc, since skip_in: [ci]'s comment implies README-scraping runners are the expected consumer.

> Smaller findings from the conversion:

>    verify.endpoint - the doc says it's required whenever verify: exists, but the checker is happy with just verify.tests. The checker's right (WVA has no endpoint to curl), so the doc just needs fixing.
    No way to say "install X first." WVA sits on top of optimized-baseline right now for example and nothing in the schema expresses that.
    No "edit this file" step. The guide asks you to update serverAddress in the ScaledObject before applying - a runner can't do that. It'd have to become a yq command, which is probably a healthy forcing function maybe.
    Ordering is implicit. For example on this guide, KEDA's CRDs have to exist before the WVA controller boots, and the only thing enforcing that is list orde
    

So to your first point about the `when` blocks, for the `human-readable` guide, that is actually normal. The automation is meant to pull from `YAML` so it would detect `when` and appropriately render the correct command.

The README.md will display _all of the options_ since the README.md isn't a functional script a user could interact with, and otherwise, these options would be hidden.

So the `"runner scraping the rendered readme"` is actually backwards, it should scrape the YAML. I will have a skill developed for this btw.

maugustosilva (approved):

PR #1992, opened 2026-07-02
[Fix] Add new config required by `actions/checkout@v7`

maugustosilva:
The `allow-unsafe-pr-checkout: true input is a safety toggle introduced in actions/checkout v7. It allows a workflow to bypass security blocks and check out untrusted pull request code from public forks.

This is required for our CI/CD testing.


diegocastanibm (approved):
LGTM

PR #1994, opened 2026-07-02
fix: WVA HPA guide fails to install in non-default namespaces

### LINKED ISSUE #1327: bug: HPA+WVA guide fails to install WVA when using non-default namespaces
The [HPA+WVA guide](https://github.com/llm-d/llm-d/blob/main/guides/workload-autoscaling/README.wva.md) fails to install WVA when the default namespaces are [overridden](https://github.com/llm-d/llm-d/blob/main/guides/workload-autoscaling/README.wva.md#set-namespaces). It should not.

Currently this guide supports 2 platforms, k8s and openshift, namespace-scoped installation. This bug applies to both.

willmj:
Fixes llm-d/llm-d-workload-variant-autoscaler#1327

## Why step 6 didn't work
Step 6 installed the controller with `kubectl apply -k <overlay> -n ${WVA_NAMESPACE}`. The platform overlays a fixed namespace via a kustomize `namespace:` transformer (`wva-system` on ocp, `llm-d-optimized-baseline` on k8s). `-n` can't override that, so apply failed with:

```
the namespace from the provided object "wva-system" does not match the namespace "..."
```

## Fix

- `platform/ocp`: add `namespace`: `llm-d-optimized-baseline` so it matches `platform/k8s` and the documented default.
- Guide install/cleanup: replace `-n` with `kustomize edit set namespace ${WVA_NAMESPACE}` before apply/delete. The transformer retargets every resource, the CRB subjects, and the bundled Namespace object, so any `WVA_NAMESPACE` works on both platforms.

## Why this approach

Kustomize can't read env vars, and a `${WVA_NAMESPACE}` placeholder in the committed overlays would fail CI's `--dry-run=server` (invalid namespace name). Setting the namespace at install time keeps the committed overlays valid (CI/nightly untouched) and keeps `WVA_NAMESPACE` the single knob.

## Trade-off

`kustomize edit set namespace` modifies the tracked `platform/${PLATFORM}/kustomization.yaml` in place, so a non-default `WVA_NAMESPACE` leaves the file modified and it must be reverted with `git checkout`. It also adds a `kustomize` CLI dependency.

## Testing 

Deployed on OpenShift into a non-default namespace: install succeeds with no namespace-mismatch errors, all resources and CRB subjects land in the target namespace, and controller runs.

### DIFF
--- a/guides/workload-autoscaling/README.wva.md
+++ b/guides/workload-autoscaling/README.wva.md
@@ -49,7 +49,7 @@ export REPO_ROOT=$(realpath $(git rev-parse --show-toplevel))
 ```
 
 > [!NOTE]
-> **Namespaced-Scoped Installation**: this guide installs WVA to watch resources only in the `llm-d-optimized-baseline` namespace. For cluster-wide autoscaling, set `--watch-namespace=""` in the controller deployment.
+> **Namespaced-Scoped Installation**: this guide installs WVA to watch resources only in the `${WVA_NAMESPACE}` namespace. For cluster-wide autoscaling, set `--watch-namespace=""` in the controller deployment.
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


lionelvillard:
@willmj can you please rebase?

willmj:
@lionelvillard done!

ahg-g:
sorry, this fell through the cracks, can you please rebase again?

lionelvillard:
@mamy-CS do you still need this?

mamy-CS:
> @mamy-CS do you still need this?

yes, @willmj please rebase and let's land it. Thanks.

willmj:
@mamy-CS @lionelvillard @ahg-g it has been rebased!

lionelvillard (approved):

PR #1885, opened 2026-06-21
v0.8.0 release prep: swap to current official images

Gregory-Pereira:
Changes:
- Consume common image style in multi-modal recipe
- Bump container images to most recent RCs / release images
- Correct outdated --connector refs

cc @llm-d/release-crew

PR #1731, opened 2026-06-09
Support tiered prefix cache on Intel XPU with LMCache connector

XinyuYe-Intel:
Support tiered prefix cache on Intel XPU with LMCache connector.

PR #1995, opened 2026-07-03
Fix: vLLM startup failure in XPU tiered prefix cache path

weizhoublue:
The new XPU tiered-prefix-cache vLLM deployment path, misses the USER=llm-d workaround for the
  XPU + vLLM + Kubernetes/OpenShift guide scenario.

  This fix adds the same USER environment variable to the new XPU tiered-prefix-cache base deployment so the LMCache overlay inherits it.
  
  
  As reference, the shared XPU vLLM component already documents that `USER` is required for this runtime:

  https://github.com/llm-d/llm-d/blob/main/guides/recipes/modelserver/components/images/xpu-vllm/kustomization.yaml

  ```yaml
  # Workaround for https://github.com/vllm-project/vllm/issues/44548, required with vLLM v0.20.0+ due to torch 2.11+
  # See: https://github.com/vllm-project/vllm/blob/main/requirements/xpu.txt#L15
  value:
    name: USER
    value: llm-d
```

  llm-d applied this workaround broadly in PR #1885. The underlying issue is documented in:

  - https://github.com/llm-d/llm-d-benchmark/pull/1573
  - https://github.com/vllm-project/vllm/issues/44548

  the specific issue is following:

 During startup, vLLM triggers PyTorch Inductor/cache-related logic that reads the current username. In environments such as OpenShift, Pods may run with a random UID that is not present in /etc/passwd. The getpass.getuser()/pwd.getpwuid() path can then raise `KeyError: uid not found`, causing the vLLM container to fail during startup.

### DIFF
--- a/guides/tiered-prefix-cache/modelserver/xpu/vllm/base/patch-vllm.yaml
+++ b/guides/tiered-prefix-cache/modelserver/xpu/vllm/base/patch-vllm.yaml
@@ -20,6 +20,8 @@ spec:
           securityContext:
             privileged: true
           env:
+            - name: USER
+              value: llm-d
             - name: HF_TOKEN
               valueFrom:
                 secretKeyRef:


weizhoublue:
@XinyuYe-Intel  hi , I submit this follow-up PR for your PR #1731
Could you help check , thanks

XinyuYe-Intel:
> @XinyuYe-Intel hi , I submit this follow-up PR for your PR #1731 Could you help check , thanks

Hi @weizhoublue , thanks for catching the USER=llm-d requirement for the XPU runtime! 

In the https://github.com/llm-d/llm-d/pull/1230/, I've included [xpu-vllm/kustomization.yaml](https://github.com/llm-d/llm-d/blob/main/guides/recipes/modelserver/components/images/xpu-vllm/kustomization.yaml) in [base](https://github.com/llm-d/llm-d/pull/1230/changes#diff-432f2a04da445a460112c163d8c7b0e98a3d9a47e6cebd988d29440d41684199), I think this would solve the issue you mentioned, right?

weizhoublue:
> In the #1230, I've included [xpu-vllm/kustomization.yaml](https://github.com/llm-d/llm-d/blob/main/guides/recipes/modelserver/components/images/xpu-vllm/kustomization.yaml) in [base](https://github.com/llm-d/llm-d/pull/1230/changes#diff-432f2a04da445a460112c163d8c7b0e98a3d9a47e6cebd988d29440d41684199), I think this would solve the issue you mentioned, right?

Yes, that's right. Once the PR #1230 is merged , the issue could be fixed.
I close this PR to avoid duplication .

PR #2007, opened 2026-07-07
Update README.md with two Docs links

Ibrahim2595:
Update the two GitHub links to point to the Docs pages now that the pages are live in the 0.8 release. Keeps all links in this section consistent.


Ibrahim2595:
@robertgshaw2-redhat update the two links now that the pages are live in docs.

robertgshaw2-redhat (approved):

PR #2011, opened 2026-07-07
guides: add SLO-driven autoscaling (KEDA + predicted latency)

kaushikmitr:
A guide to autoscale a vLLM decode pool against latency SLOs using only off-the-shelf components: the EPP's predicted-latency histograms, a Prometheus recording rule for the saturation signal, and a KEDA ScaledObject whose expr-lang formula computes the desired replica count and drives a standard HPA. No custom controller.

Includes the control-law math (signal chain, hysteresis, in-flight credit), the deployable manifests (recording rules, ScaledObject, kube-state-metrics), the EPP plugin config, and a reproducible benchmark harness with results (95.8% combined SLO attainment at 5.93 avg replicas vs 100% at a static 8).

Addresses: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1018


lionelvillard:
thanks @kaushikmitr for doing this! I have a general comment: all autoscaling guides are under the [workload-autoscaling](https://github.com/llm-d/llm-d/tree/main/guides/workload-autoscaling) directory. There is already a HPA (soon to be replaced by KEDA)+EPP section that, with this PR, references 2 sub-guides (the existing one and the one in this PR). 

The latency predictor guide should also be updated to reference the autoscaling guide (something like `Enable Autoscaling (optional)`)

kaushikmitr:
> thanks @kaushikmitr for doing this! I have a general comment: all autoscaling guides are under the [workload-autoscaling](https://github.com/llm-d/llm-d/tree/main/guides/workload-autoscaling) directory. There is already a HPA (soon to be replaced by KEDA)+EPP section that, with this PR, references 2 sub-guides (the existing one and the one in this PR).
> 
> The latency predictor guide should also be updated to reference the autoscaling guide (something like `Enable Autoscaling (optional)`)

Done!
For this "The latency predictor guide should also be updated to reference the autoscaling guide (something like `Enable Autoscaling (optional)`)" I will do it in a follow up PR. Note that this slo driven autoscaling guide will work with any llm-d scheduling guide as long as there are ttft and tpot slos. Predicted Latency signal is good to have but not necessary, it falls back to actual latency signals. Only for p/d we will need to make some changes to this guide as ttft will drive prefill pods and tpot will drive decoderpods.

guides/workload-autoscaling/slo-aware/prometheus-rule.yaml
lionelvillard, line 1: I'm wondering how generally applicable this rule is. The hard-coded TTFT/TPOT SLO targets indicate there is potentially one set of recording rules per variant. If that is the case then we will have to think about how identify variant at the EPP level. For instance, the same model can be served by two different variants, one running on accelerator type 1, the other on accelerator type 2. These variants can potentially belong to different inference pools.

guides/workload-autoscaling/slo-aware/prometheus-rule.yaml
```diff
@@ -0,0 +1,64 @@
...
+    # or set `ruleSelector: {}` on the Prometheus to match every PrometheusRule.
+    role: recording-rules
+spec:
+  groups:
+  - name: epp-saturation
+    interval: 15s
+    rules:
+    - record: epp:latency_ttft_p90:seconds
+      expr: >-
+        ((histogram_quantile(0.9, sum(rate(inference_objective_request_predicted_ttft_seconds_bucket[1m])) by (le)) >= 0)
```
lionelvillard: `inference_objective_request_predicted_ttft_seconds_bucket` should be `llm_d_epp_request_predicted_ttft_seconds` (across the entire PR).

guides/workload-autoscaling/slo-aware/prometheus-rule.yaml
kaushikmitr: yes this scheme overall will not work when we have multi variant or multiple SLOs. Only one set of TTFT/TPOT SLO  and one variant that the llm-d epp targets

guides/workload-autoscaling/slo-aware/prometheus-rule.yaml
```diff
@@ -0,0 +1,64 @@
...
+#
+# Each raw sample is clamped (C) BEFORE smoothing, then the clamped signal is
+# averaged — so one knee-region spike cannot be averaged in at full magnitude and
+# pull the mean up to the ceiling. avg_over_time[45s] at 15s evaluation approximates
+# an EMA (alpha 0.6 at a ~30s cycle has a ~20s mean age; a 45s window matches).
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: slo-driven-autoscaling
+  namespace: monitoring
```
lionelvillard: Depending on the question/comment above, these set of recording rules should be in the application/tenant namespace.

guides/workload-autoscaling/slo-aware/scaledobject.yaml
```diff
@@ -0,0 +1,78 @@
...
+        : (sat < 0.40 ? n - floor(n * (1 - sat / 0.40)) : n)
+  triggers:
+  - type: prometheus
+    name: sat
+    metadata:
+      serverAddress: http://prometheus-server.monitoring.svc:80
+      query: epp:saturation:smoothed
+      threshold: "1"
+  - type: prometheus
+    name: "n"
```
lionelvillard: you may want to use more meaningful names to make the formula above easier to understand without having to go through the documentation.

guides/workload-autoscaling/slo-aware/prometheus-rule.yaml
kaushikmitr: Done , moved the PrometheusRule into the workload namespace (llm-d-optimized-baseline) so the tenant owns their scaling rules, and documented that the operator needs its ruleNamespaceSelector to match.

guides/workload-autoscaling/slo-aware/scaledobject.yaml
kaushikmitr: agreed, updated see now

guides/workload-autoscaling/slo-aware/prometheus-rule.yaml
kaushikmitr: agreed, updated

guides/workload-autoscaling/slo-aware/prometheus-rule.yaml
kaushikmitr: clarified this point in the doc. Currently the only scenaio where epp can target multiple varinats is when we have disagg (prefill and decode pods). That can be fixed, may be in a follow up PR.

asm582:
Can make it clear in the doc that this implementation works on static SLOs attached to requests.

asm582:
For benchmarking, can we add steps to generate plots?

asm582:
As discussed SLO-Aware Autoscaling should be under EPP+HPA path based on the current setup that we have.

guides/workload-autoscaling/slo-aware/decode-warmup-patch.yaml
```diff
@@ -0,0 +1,43 @@
+# Recommended (not required): faster pod warmup for the vLLM inference pool.
```
asm582, line 1: is the current benchmarking numbers consuming this optimization? should there be a mention about the impact on SLOs if one does not use this optimization?

guides/workload-autoscaling/slo-aware/benchmark-templates/plot_run.py
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env python3
```
asm582, line 1: Users trying to use this script should also have correct python libraries installed on the user system, if this is the correct place to add this visualization code, can you also add requirements.txt file?

asm582 (changes requested):

kaushikmitr:
> Can make it clear in the doc that this implementation works on static SLOs attached to requests.

i already did: 
Scope: one variant, one SLO pair. The recording rules bake in a single TTFT/TPOT SLO target and aggregate the EPP's latency histograms pool-wide, so this guide assumes one inference pool serving one model variant against one SLO pair. It does not yet handle multiple variants of the same model (e.g. the same model on two accelerator types, or split across separate inference pools) or multiple SLO tiers — those would need per-variant recording rules and a way to identify the variant at the EPP level. For a disaggregated (prefill/decode) deployment, TTFT would drive the prefill pool and TPOT the decode pool; adapting the signal chain for that is future work.

guides/workload-autoscaling/slo-aware/decode-warmup-patch.yaml
kaushikmitr: yes it does. i will add that.

kaushikmitr: done

guides/workload-autoscaling/slo-aware/benchmark-templates/plot_run.py
kaushikmitr: done

kaushikmitr:
@asm582 @lionelvillard i think all comments have been addressed. PTAL. Only thing remainign is moving the inference-perf benchmarking guide in this PR to llm-d-bennchmarking which i plan to do in a follow up.

kaushikmitr:
@asm582 @lionelvillard as suggested using update the benchmark guides to use llm-d-benchmark

guides/workload-autoscaling/README.slo-aware.md
```diff
@@ -0,0 +1,276 @@
...
+
+3. **kube-state-metrics.** The control law reads the target Deployment's
+   provisioned and ready replica counts, which kube-state-metrics exposes to
+   Prometheus. Any standard install works; if you don't already run it, a
+   minimal deployments-only install is in
+   [`slo-aware/kube-state-metrics.yaml`](#files).
+
+4. **KEDA** ≥ 2.15 (we ran 2.20.1): `kubectl apply --server-side -f https://github.com/kedacore/keda/releases/download/v2.20.1/keda-2.20.1.yaml`.
+
+## How it works — the control law
```
lionelvillard: For an operator whose main question is "which knobs do I turn," this is a lot — ~100 lines of derivation with histogram_quantile, the piecewise formula, the credit factor, etc. It reads more like a reference than a how-to.

Every other guide here keeps "How it works" to a few conceptual lines (flow-control, precise-prefix), and predicted-latency-routing just provides a link to the doc. Could we do the same? Something like:

- short conceptual blurb, no formulas — "SLOs → one saturation signal s; KEDA scales up when it's high, down when low, dead band to avoid flapping, warmup credit to avoid over-scaling"
- move the full derivation to a linked advanced doc (in https://github.com/llm-d/llm-d/tree/main/docs/architecture/advanced/autoscaling)
- use the tunable table below for the operator's actual job — maybe a "symptom → knob → direction" column

kaushikmitr: Done in a99fac11 — "How it works" is now a short conceptual paragraph (signal → thresholds → dead band → warmup credit) with a link to a new advanced doc, `docs/architecture/advanced/autoscaling/slo-aware-keda.md`, which holds the full derivation. The Tunables table is reworked into symptom → knob → direction form so it reads as the operator's runbook.

lionelvillard (approved):

asm582 (approved):

PR #1643, opened 2026-05-31
feat: add guide for SGLang in precise-prefix-cache-routing

zdtsw:
# Changes
- add new guide on SGLang for precise-prefix-cache (this is now based on the vllm render tokenizer)
- update dry-run in CI to verify render manifests but no add nightly e2e for that
- fix lint in CI
- set to use `llm-d.ai/engine-type` than old` inference.networking.k8s.io/engine-type` , the rest guide will be updated in PR  https://github.com/llm-d/llm-d/pull/1642 to follow up

ref https://github.com/llm-d/llm-d/issues/1632

# Test
on a small cluster with only 1node8GPU, trim replicas from 8 to 2 for sglang pods
1. basic test precise prefix routing -- following README.md steps 
```
export GUIDE_NAME=precise-prefix-cache-routing
export NAMESPACE=wenzhou-dev
IP=$(kubectl get service ${GUIDE_NAME}-epp -n ${NAMESPACE} -o jsonpath='{.spec.clusterIP}')
 kubectl run curl-debug --rm -it \
    --image=cfmanteiga/alpine-bash-curl-jq \
    --env="IP=$IP" --env="NAMESPACE=$NAMESPACE" \
    -- /bin/bash
```
then send long prompt (more than 64 to match - "--page-size=64") 3-4 times but with some different appended words in each curl 
then check 2 sglang pods log, 
> k logs pod/precise-prefix-cache-routing-gpu-sglang-decode-f6f6d4cf6-z77mw | grep cached-token
```
...
[2026-06-03 11:38:45 TP0] Prefill batch, #new-seq: 1, #new-token: 64, #cached-token: 512, token usage: 0.00, #running-req: 0, #queue-req: 0, #pending-token: 0, cuda graph: True, input throughput (token/s): 33.26
[2026-06-03 11:38:48 TP0] Prefill batch, #new-seq: 1, #new-token: 64, #cached-token: 512, token usage: 0.00, #running-req: 0, #queue-req: 0, #pending-token: 0, cuda graph: True, input throughput (token/s): 21.08
[2026-06-03 11:38:51 TP0] Prefill batch, #new-seq: 1, #new-token: 64, #cached-token: 512, token usage: 0.00, #running-req: 0, #queue-req: 0, #pending-token: 0, cuda graph: True, input throughput (token/s): 21.06
[...]

PR #2012, opened 2026-07-08
Add engine-type: sglang label to optimized-baseline kustomization

Amit-Berman:
EPP defaults to vLLM metric names when engine-type is unset, causing all core-metrics extraction to fail on SGLang pods.


sagearc (changes requested):
@Amit-Berman Thanks! Did this work out of the box without any other changes to `patch-sglang.yaml`? Also noticed it has drifted from the vLLM optimized-baseline equivalent, might be worth aligning the gaps so we can do a fair comparison.
ref https://github.com/llm-d/llm-d/pull/1643

cc @zdtsw

Amit-Berman:
Hi @sagearc, thanks for the review.

I addressed your comments and also ran a benchmark. I got the following results:

image

I can add these results to the documentation, but I don't have the corresponding results for the Kubernetes Service configuration as it exists today.

maugustosilva (approved):

PR #2015, opened 2026-07-08
Proposal: llm-d-perf-simulator

jgchn:
As llm-d grows into a platform for intelligent distributed inference, simulation should become a core capability rather than a standalone research tool. This document proposes a core llm-d Simulator: an officially supported simulation framework that enables rapid, reproducible experimentation across the llm-d ecosystem.


sagearc:
> simulation should become a core capability rather than a standalone research tool

I agree, but I don't think replacing the llm-d simulator with an external simulator like https://github.com/inference-sim/inference-sim would achieve that goal. Personally, my main issue with external simulators is that they tend to lag behind where new features emerge.

I recently used https://github.com/vllm-project/vllm/pull/47922 for hybrid attention behavior directly in vllm, and I think this direction addresses that concern more generally, with less work and a softer integration path.

wseaton:
I would like to also highlight awareness around other simulator types/architectures. 

Perhaps the project doesn't need to endorse "one simulator to rule them all". Our project https://github.com/neuralmagic/vllm-vcr is a different simulator design that allows for byte and timing identical replay of vllm sessions by doing engine trace capture, and offers similar features for training learned profiles off of real benchmark runs.

Notably this allows very accurate replication of things like prefill/decode contention, tail latencies, and all slowdowns or bugs caused by frontend behavior or metrics bookkeeping are preserved because all requests flow through the real vllm frontend (including tool parser bugs!). This also allows for durable storage of extracted traces (ie. into object storage) so you can do things like 'only run on real GPUs once' cache CI, no knob tuning needed.

This new simulator proposal seems to carry some of the same downsides of the old sim by completely reimplementing the vllm frontend. I would like to understand what the plan is to make sure the simulator stays up to date w/ new frontend features or quirks? 

> Personally, my main issue with external simulators is that they tend to lag behind where new features emerge.

+1  to @sagearc's comment

jgchn:
Thanks @sagearc and @wseaton for the thoughtful feedback. We've updated the [Google Docs](https://docs.google.com/document/d/1TRdmK2DcKxZ39vxbsjx0Zef1KZmOPFojVTyXtS1WwNw/edit?tab=t.0) proposal with two new sections (Scope, Maintaining Parity with llm-d Updates, Alternatives) that address your concerns directly, full version in the Google Doc.

The core distinction worth highlighting: BLIS is not a functional emulator. It's an inference performance estimator. Given cluster- and engine-level configurations, it predicts latency and throughput by advancing a simulation clock event-to-event, without executing real inference. This means the question of "will it lag behind vllm frontend features?" doesn't apply in the same way: BLIS is not modeling API-surface behavior, request parsing, or frontend quirks. It's modeling timing, how admission and flow control, routing, batching, KV-cache, and prefill/decode placement interact.

Functional approaches like trace capture/replay are the right tool when frontend fidelity is the goal. BLIS is the right tool when the goal is comparing hundreds of configurations or evolving policies across large search spaces, workflows that are inherently bounded by wall-clock time in replay-based approaches. The two are complementary, not competing.

Let us know what you think about the proposal before we update the PR!

sagearc:
> Maintaining Parity with llm-d Updates

Great, this helps. My concern was more about whether external BLIS will stay aligned with vllm's engine internals: model-derived cache config, kvcache groups, per-layer attention types, hybrid full/sliding attention, eviction policies, tiered cache policies, quantization, etc. Those pieces change quickly in vllm. If they're simplified or manually mirrored, I worry simulation will mostly catch up after deployment instead of helping us evaluate changes before they land.

The small vllm-native simulator in vllm#47922 already gets a lot of that behavior for free by running through the real scheduler/kvcache path. Maybe we could build around that kind of native hook?

jgchn:
Thanks @sagearc! We've updated the [proposal doc](https://docs.google.com/document/d/1TRdmK2DcKxZ39vxbsjx0Zef1KZmOPFojVTyXtS1WwNw/edit?tab=t.0) to make the points more clear with a concrete example that shows the benefits of BLIS and highlighted why a cluster-level performance estimator is needed. Please let up know your thoughts!

jgchn:
Hi @wseaton and @sagearc curious if you had a chance to go over the revised proposal? We shortened the proposal significantly and made several points more clear.

jgchn:
Hi @wseaton and @sagearc any more feedback on the proposal? Thanks for your time in advance!

PR #2038, opened 2026-07-13
fix(wva): migrate nightly + guide from Prometheus Adapter to KEDA

mamy-CS:
Moves the WVA nightlies and guide off Prometheus Adapter and onto KEDA.
WVA repo already made this switch - [llm-d-workload-variant-autoscaler#1399](https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1399) removed the adapter entirely, leaving KEDA as the only supported scaler backend. Our guide already said KEDA was recommended, but the nightlies still deployed the plain HPA + adapter path, and the KEDA manifest that came with the docs had never been run by CI.

PR #2039, opened 2026-07-13
fix(wva): authenticate KEDA to Thanos and assert the metric path

mamy-CS:
The nightly that shipped with #2038 went green while autoscaling was dead. KEDA was getting a 401 from Thanos on every poll, suppressing it, and serving `fallback: replicas: 2 - pods Ready, ScaledObject Ready=True and Active=True`, and WVA was driving nothing.

Both nightlies now assert the metric pipeline instead of just liveness. KEDA's fallback turns a broken metrics path into a healthy-looking steady state, so "the stack came up" says nothing about whether WVA is driving replicas. The scripts
require the ScaledObject's Fallback condition to hold False across consecutive polls (it reads False before KEDA's first query, so a single sample proves nothing), and dump the KEDA operator errors and fail if it does not.

Verified on live cluster
Via /test-nightly on this PR:
```
FALLBACK    False                      # was True (fallback replicas)
ScalingActive  True / ValidMetricFound  # HPA reading the external metric
KEDA 401s   none
wva_desired_replicas = 1               # real value from WVA, not the fallback's 2
```
The value being 1 rather than 2 is the tell: under fallback KEDA pinned it at 2 regardless of load. This is the first run where the deployment is genuinely scaled by WVA.

**Note for cluster owners** cc @maugustosilva 
`v1beta1.external.metrics.k8s.io` is owned by `openshift-keda` serves nothing and has not since Jul 8, no APIService conflict, but it's dead weight worth deleting (no Helm release owns it, needs a manual kubectl delete).


clubanderson:
### Nightly E2E Result: `workload-autoscaling` (OpenShift)

| | |
|---|---|
| Result | :x: **Failed** |
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/29274897780) |
| Ref | `41d083d` |

clubanderson:
### Nightly E2E Result: `workload-autoscaling` (OpenShift)

| | |
|---|---|
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/29276323545) |
| Ref | `56fc12c` |

lionelvillard (approved):

clubanderson:
### Nightly E2E Result: `workload-autoscaling` (OpenShift)

| | |
|---|---|
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/29277714202) |
| Ref | `2c5f3ec` |

mamy-CS:
ready for merge @lionelvillard

PR #2052, opened 2026-07-14
[docs] Add DisaggregatedSet variant of the P/D disaggregation guide

BenjaminBraunDev:
## What

Adds a third variant of the pd-disaggregation guide (alongside the default Deployment-based guide and the TPU variant) that manages the whole P/D topology with a single LWS DisaggregatedSet resource:

1. `guides/pd-disaggregation/README.ds.md`: DS prerequisites, deploy, slice operations (scale slices, per-role replicas, per-slice rolling updates, placement policy), verification, benchmarking, cleanup.
2. `guides/pd-disaggregation/modelserver/gpu/vllm-ds/`: Kustomize overlay with the DS manifest, prefill/decode roles carrying the same vLLM args, routing-proxy sidecar, probes, and labels as the Deployment-based overlay; slices: 2 (2 × [4 × TP1 prefill + 1 × TP4 decode], same 16-GPU aggregate as the main guide). Images resolved via the shared gpu-vllm / routing-sidecar components.
- `guides/pd-disaggregation/README.md`: variants blurb + backends-table row linking the new guide.

## Why

The Deployment-based guide manages prefill and decode as two seperate workloads with no version coupling. This results in uncoordinated rollouts and duplicated manifests. DisaggregatedSet makes the P/D topology one declarative unit with version-synchronized per-slice rollouts. It also adds the slices feature: allowing for an arbitrary number of `slices: N` independent complete copies of each LWS role, and (upcoming) per-domain placement policy (see link below). Router integration is unchanged as EPP discovers pods by the same labels regardless of workload controller.

## Requirements / caveats

- Requires the LWS controller with DisaggregatedSet enabled; spec.slices is merged upstream and ships in the next LWS release (> v0.9.0).
- spec.placementPolicy is documented but not yet merged in LWS (kubernetes-sigs/lws#848). The guide and manifest carry explicit notes on this, and the field is commented out.
- As mentioned before, EPP schedules across slices as flat pools. DS is not concerned with routing.

## Testing

- kubectl kustomize builds cleanly, DS fan-out, naming (`<ds>-<slice>-<revision>-<role>`), labels, and per-slice Services validated on GKE.
- Slices scale up/down at constant revision with zero disruption to existing slices
- per-role replica scaling applied per-slice without a rollout; independent per-slice rolling updates; cascade cleanup.
- tested e2e (with 16 × A100, guide-shape topology, NIXL over TCP): disaggregated serving through the llm-d router in both standalone and GKE Inference Gateway modes, heterogeneous TP1→TP4 KV transfer confirmed (TransferTopology tp_ratio=4), and benchmarked with inference-perf (8 qps: 720/720 success, TTFT p90 680 ms, ITL p50 4.8 ms).

> NOTE: Testing surfaced two upstream LWS rolling-update planner issues:
> 1. Instant old-revision drain when replicas change with a template
> 2. rollout deadlock on fully-allocated clusters

> These are being looked at and neither are issues for the sake of this guide. There's a note on them at the bottom of the readme and they should be removed once fixed.

### DIFF
--- a/guides/pd-disaggregation/README.ds.md
+++ b/guides/pd-disaggregation/README.ds.md
@@ -136,8 +136,8 @@ A stuck slice degrades only itself; the other slices keep serving and finish the
 
 ### Pin slices to accelerator domains (placement policy)
 
-> [!WARNING]
-> `spec.placementPolicy` is **not yet merged or released** in LWS — it is under development (tracked in [lws#848](https://github.com/kubernetes-sigs/lws/issues/848)). Unlike `slices`, which is merged upstream and ships in the next LWS release, the example below only works with a controller built from the in-progress placement-policy branch. On released controllers the field is rejected by the CRD schema.
+> [!NOTE]
+> `spec.placementPolicy` is merged upstream ([lws#916](https://github.com/kubernetes-sigs/lws/pull/916), [KEP-848](https://github.com/kubernetes-sigs/lws/tree/main/keps/848-disaggregatedset-placement-policy)) and, like `slices`, ships in the next LWS release after `v0.9.0`. Controllers at `v0.9.0` or older reject the field.
 
 To confine each slice to one topology domain and spread slices across domains, set `placementPolicy` (a commented example ships in `modelserver/gpu/vllm-ds/disaggregatedset.yaml`):
 
--- a/guides/pd-disaggregation/README.ds.md
+++ b/guides/pd-disaggregation/README.ds.md
@@ -102,53 +102,15 @@ Follow the [monitoring section of the main guide](./README.md#3-enable-monitorin
 
 ## Operating the DisaggregatedSet
 
-### Scale the number of slices
+Day-to-day mechanics (scaling slices, scaling role replicas, rolling updates, placement policy) are DisaggregatedSet features rather than anything llm-d specific, and are documented in the [LWS DisaggregatedSet docs](https://lws.sigs.k8s.io/docs/examples/disaggregatedset/). The notes that matter for this guide:
 
-Add a third complete P/D copy (4 more prefill, 1 more decode). Because `slices` is excluded from the revision hash, this stands up slice 2 at the current revision **without touching slices 0 and 1**:
-
-```bash
-kubectl patch disaggregatedset pd-disagg -n ${NAMESPACE} --type merge -p '{"spec":{"slices":3}}'
-```
-
-Scaling down deletes the highest-indexed slices' resources directly; lower slices are untouched.
-
-### Scale replicas within a role
-
-Per-role `replicas` applies **per slice**. With 2 slices, raising prefill replicas from 4 to 6 yields 12 prefill instances in total:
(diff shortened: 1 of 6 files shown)


BenjaminBraunDev:
Had to add a kustomize dryrun CRD install step to pass all checks since to reference a DS version with slices which is not yet in an official release, this will need to be updated when v0.10.0 of lws releases

ahg-g:
/assign @liu-cong 

Deferring to Cong to advise on how we want to structure the guide given the new deployment option.

guides/pd-disaggregation/README.ds.md
```diff
@@ -0,0 +1,203 @@
...
+```bash
+kubectl get leaderworkerset -n ${NAMESPACE} \
+  -l disaggregatedset.x-k8s.io/name=pd-disagg,disaggregatedset.x-k8s.io/slice=0 -w
+```
+
+A stuck slice degrades only itself; the other slices keep serving and finish their own rollouts.
+
+### Pin slices to accelerator domains (placement policy)
+
+> [!WARNING]
```
liu-cong: can we add this section after the pr is merged upstream? Is there any urgency adding this caveats here with cognitive burden?

guides/pd-disaggregation/README.ds.md
```diff
@@ -0,0 +1,203 @@
...
+```bash
+kubectl get leaderworkerset -n ${NAMESPACE} \
+  -l disaggregatedset.x-k8s.io/name=pd-disagg,disaggregatedset.x-k8s.io/slice=0
+```
+
+### 3. Enable Monitoring (optional)
+
+Follow the [monitoring section of the main guide](./README.md#3-enable-monitoring-optional) — the model server pods carry the same labels as the Deployment-based variant, so the same monitoring resources apply.
+
+## Operating the DisaggregatedSet
```
liu-cong, line 103: We should have this somewhere in the DS repo. This is not strictly required as part of this guide (similar to how you scale a deployment, for example).

guides/pd-disaggregation/README.ds.md
BenjaminBraunDev: Agree, though placement policy just merged upstream in kubernetes-sigs/lws#916, so we are good to go there.

guides/pd-disaggregation/README.ds.md
BenjaminBraunDev: Agree. Trimmed down this section to the llm-d specific notes with a link to the LWS docs for the generic mechanics. I'm also going to update the upstream DisaggregatedSet page to cover these operations, since today it only has deploy examples.

BenjaminBraunDev: (opened https://github.com/kubernetes-sigs/lws/pull/951 for updated docs)

BenjaminBraunDev:
@liu-cong Good to merge this?

liu-cong (approved):

PR #2054, opened 2026-07-15
add benchmark results for P/D vllm 0.24

DolevAdas:
Added benchmark results to the P/D disaggregation guide with benchmark results from llm-d main and vLLM v0.24.0, run on H200 cluster.
Added an llm-d-baseline measurement (same 16 GPUs with the full llm-d EPP and routing stack, but in aggregated mode) alongside the plain Kubernetes Service baseline, so the results cleanly attribute the gains to P/D disaggregation itself. 
The guide YAMLs have also been updated to match the tested configuration.

### DIFF
--- a/guides/pd-disaggregation/modelserver/gpu/vllm/base/patch-decode.yaml
+++ b/guides/pd-disaggregation/modelserver/gpu/vllm/base/patch-decode.yaml
@@ -17,6 +17,7 @@ spec:
             - "--kv-transfer-config"
             - '{"kv_connector":"NixlConnector", "kv_role":"kv_both"}'
             - "--no-disable-hybrid-kv-cache-manager"
+            - "--max-model-len=8192"
             - "--port=8200"
           env:
             - name: HF_TOKEN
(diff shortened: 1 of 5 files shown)


rachelt44:
@vMaroon can you take a look?

guides/pd-disaggregation/modelserver/gpu/vllm/base/patch-decode.yaml
```diff
@@ -17,6 +17,7 @@ spec:
             - "--kv-transfer-config"
             - '{"kv_connector":"NixlConnector", "kv_role":"kv_both"}'
             - "--no-disable-hybrid-kv-cache-manager"
+            - "--max-model-len=8192"
```
ahg-g, line 20: why are we adding this?

guides/pd-disaggregation/modelserver/gpu/vllm/base/patch-decode.yaml
```diff
@@ -28,11 +29,18 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: UCX_MAX_RMA_RAILS
+              value: "1"
```
ahg-g, line 33: what does this mean?

ahg-g:
Pls see https://github.com/llm-d/llm-d/pull/2072

/cc @kaushikmitr

guides/pd-disaggregation/modelserver/gpu/vllm/base/patch-decode.yaml
DolevAdas: The benchmark workload uses ISL=5000 / OSL=250 (~5,250 tokens max per request). Without this flag, vLLM defaults to the model's full native context window, which increases the activation memory footprint during GPU profiling and allows each sequence to consume proportionally more KV cache blocks, leaving less room for concurrent sequences. Setting --max-model-len=8192 caps the maximum sequence length to match the actual workload, maximizing the number of concurrent sequences that fit in GPU memory.

guides/pd-disaggregation/modelserver/gpu/vllm/base/patch-decode.yaml
DolevAdas: UCX_MAX_RMA_RAILS=1 limits UCX to a single network rail per RDMA memory transfer. Without it, UCX tries to stripe KV cache transfers across multiple rails by default. In a Kubernetes pod with exactly one RDMA device allocated (rdma/ib: "1"), UCX attempting multi-rail can cause NUMA/PCIe misalignment and, in severe cases, out-of-order delivery of KV tensor blocks. Setting this to 1 aligns UCX's behavior with the single device actually available to the pod and was required for stable NIXL KV transfers in our tests.

PR #2035, opened 2026-07-13
fix: CI dry-run and optimized-baseline guide

zdtsw:
# Notes
- dry-run did not cover all manifests (only 2 layers deep before the fix)
- the patch name does not match whats set in the kustomizaiton, rename the file
- this leads to we have a bug in the optimized guide set model-storage but not declared since it is not in used, just remove it.

fixes #2027 

cc @diegocastanibm @maugustosilva

PR #2063, opened 2026-07-16
update(CI): convert wva dry-run job to do recursive find for kustomization

zdtsw:
# Notes
The dry-run-workload-autoscaling job still used the old single-level loop (for kust_dir in .../platform/*/),
converge it onto the recursive pattern already used by dry-run-precise-prefix-cache-routing
so the whole workflow is consistent and nested overlays are validated automatically

ref  comments from https://github.com/llm-d/llm-d/pull/2035#pullrequestreview-4715414406

cc @diegocastanibm


diegocastanibm (approved):
LGTM

PR #2081, opened 2026-07-20
multimodal-serving: adopt token-based routing, refresh benchmark report

### LINKED ISSUE #2069: [Feature]: Standardize multimodal-serving router on token-load-aware routing (affinity filter + token-load / active-request scorer)
### Feature Area

Other

## Background

The `optimized-baseline` and `agentic-serving` guides now route with the **token-based**
stack — `prefix-cache-affinity-filter` + `token-load-scorer` (fed by
`approx-prefix-cache-producer` + `inflight-load-producer`). This replaced the older
weighted-scorer stack (`queue-scorer` / `kv-cache-utilization-scorer` /
`prefix-cache-scorer` / `no-hit-lru-scorer` at weights 2/2/3/2).

The `multimodal-serving` guides still ship the older stack, so they no longer match the
reference routing. This issue tracks converging them so every non–feature-specific guide
is evaluated on the **same routing setup**, which is a prerequisite for comparable
benchmark numbers across guides.

Routing choice depends on what work the profile actually does:
- **Aggregated (single pool):** `prefix-cache-affinity-filter` selects the endpoint set and
  `token-load-scorer` picks within it — the load signal is *queued prefill token load*.
- **Any profile that prefills:** same stack — affinity filter + `token-load-scorer`. In
  `e-pd-disaggregation` prefill and decode are colocated, so its `decode` profile prefills too
  and therefore uses this stack (not active-request).
- **Pure decode** (the separate `decode` profile in `e-p-d-disaggregation`): scores on
  **`active-request-scorer`** only — decode is bound by concurrent in-flight requests, not
  prefill throughput.
- **Encode:** unchanged in every config.

## Current state

| Config | Profiles | Scorers today |
| :-- | :-- | :-- |
| `aggregation/router/aggregation.values.yaml` | `default` | queue 2 / kv 2 / prefix 3 / no-hit-lru 2 (+ `token-producer`, `approx-prefix-cache-producer`) |
| `e-disaggregation/router/e-pd-disaggregation.values.yaml` | `encode`, `decode` | encode: queue 2 · decode: active-request 2 / prefix 3 / kv 2 |
| `e-disaggregation/router/e-p-d-disaggregation.values.yaml` | `encode`, `prefill`, `decode` | encode: queue 2 · prefill: prefix 3 / queue 2 / kv 2 · decode: queue 2 / kv 2 / prefix 3 / no-hit-lru 2 |

Note the multimodal-specific `token-producer` (image → token-count estimation) must be
retained wherever `token-load-scorer` is used, since that scorer needs per-request token
counts and image inputs have no text length to read.

### 1. Aggregation (single pool) — token-load-aware
```yaml
plugins:
- type: token-producer            # keep: multimodal image-token estimation
  parameters: { ... existing image estimate params ... }
- type: inflight-load-producer
- type: approx-prefix-cache-producer
- type: prefix-cache-affinity-filter
- type: token-load-scorer
schedulingProfiles:
- name: default
  plugins:
  - pluginRef: prefix-cache-affinity-filter
  - pluginRef: token-load-scorer
```

### 2. e-pd-disaggregation — decode also prefills, so it gets the prefill stack
Prefill and decode are colocated on the `decode` profile, so it is routed like a prefill node:
affinity filter + `token-load-scorer` (**not** active-request). Encode unchanged.
```yaml
schedulingProfiles:
- name: encode          # unchanged
  plugins:
  - pluginRef: encode-filter
  - pluginRef: queue-scorer
  - pluginRef: max-score-picker
- name: decode          # runs prefill+decode -> token-load-aware, same as prefill
  plugins:
  - pluginRef: decode-filter
  - pluginRef: prefix-cache-affinity-filter
  - pluginRef: token-load-scorer
  - pluginRef: max-score-picker
```

### 3. e-p-d-disaggregation — three separate profiles; pure decode scores active-request only
Prefill and decode are separate. Prefill is token-load-aware; the pure `decode` profile uses
`active-request-scorer` **only** (no affinity filter, no token-load). Encode unchanged.
```yaml
schedulingProfiles:
- name: encode          # unchanged
  plugins: [ encode-filter, queue-scorer, max-score-picker ]
- name: prefill
  plugins:
  - pluginRef: prefill-filter
  - pluginRef: prefix-cache-affinity-filter
  - pluginRef: token-load-scorer              # prefill = token-load-aware
  - pluginRef: max-score-picker
- name: decode
  plugins:
  - pluginRef: decode-filter
  - pluginRef: active-request-scorer          # pure decode = active-request only
  - pluginRef: max-score-picker
```

Producers (`inflight-load-producer`, `approx-prefix-cache-producer`, `token-producer`) are
added to the top-level `plugins:` list in each config as needed by the filters/scorers above.

## Benchmarks to redo

`aggregation/benchmark-results/` publishes `throughput_vs_qps.png`, `latency_vs_qps.png`,
and `ttft_p90_vs_qps.png`. These were measured on the old weighted-scorer stack and **must
be regenerated** against the new routing, alongside a fresh stock-k8s-Service baseline on the
**same model-server config** (matching how the optimized-baseline pages were redone), so the
Δ% is legitimate.

- [ ] Re-run `aggregation` benchmark (3 charts + summary table) on token-load-aware routing.
- [ ] Regenerate the k8s round-robin baseline on the identical server config.
- [ ] Report latency as **median (p50)** — TTFT p50/p90 + ITL p50 — to match the other pages.
- [ ] e-disaggregation: no published results today (templates only); re-run only if we decide to publish disagg numbers.

## Acceptance criteria

- [ ] `aggregation.values.yaml` uses `prefix-cache-affinity-filter` + `token-load-scorer`, retaining `token-producer`.
- [ ] `e-pd-disaggregation.values.yaml` `decode` profile (prefill+decode colocated) uses `prefix-cache-affinity-filter` + `token-load-scorer`.
- [ ] `e-p-d-disaggregation.values.yaml` `prefill` uses affinity filter + `token-load-scorer`; pure `decode` uses `active-request-scorer` only.
- [ ] `encode` profiles left unchanged in both disagg configs.
- [ ] `aggregation` benchmark pages regenerated with p50 latency + valid same-config baseline.
- [ ] Guide READMEs updated to describe the routing change.

## Open questions

- Do we keep the multimodal `token-producer` estimation params as-is, or recalibrate the
  image→token `factor`/`defaultResolution` for the new scorer?
- `peakPrefillThroughput` for `prefix-cache-affinity-filter`: use the default (15928, calibrated
  for Qwen3-32B/H100/TP=2) or measure a multimodal-specific value via
  `guides/recipes/router/calibration`?
- Update benchmark guide to a documented multi modal workload that stresses the system

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

kaushikmitr:
Fixes #2069.

Converges the multimodal-serving router configs on the reference token-based stack (`prefix-cache-affinity-filter` + `token-load-scorer`), matching optimized-baseline and agentic-serving, so every non-feature-specific guide is evaluated on the same routing setup.

**Config changes** (per the issue's proposal):
- **aggregation** — affinity filter + token-load-scorer on the single pool; the multimodal `token-producer` (image → token-count estimation) stays, since the scorer needs per-request token counts and image inputs have no text length to read.
- **e-pd** — the colocated prefill+decode profile is routed like a prefill pool (affinity + token-load, **not** active-request); `token-producer` added (was missing). Encode unchanged.
- **e-p-d** — prefill on affinity + token-load; the pure decode profile scores on `active-request-scorer` only. Encode unchanged.

**Benchmark refresh** (aggregation): re-measured on the new stack with a fresh same-fleet k8s round-robin baseline — 16 × H200 (8 × TP2), harness v0.7.0, back-to-back runs on identical hardware. Tables now report median (p50) TTFT alongside p90, matching the other guide pages. Headline: at parity throughput below saturation, llm-d cuts TTFT p50 by up to ~5× (98 ms vs 501 ms at 20 req/s) and finished the ladder with 0 failed requests vs 27 for the plain Service.

Open item tracked in the issue: `peakPrefillThroughput` is left at the plugin default (calibrated for text-only Qwen3-32B/H100); the values file comments point to `guides/recipes/router/calibration` for vision-prefill calibration.


kaushikmitr:
Addressed the issue's open calibration question: measured `peakPrefillThroughput` for Qwen3-VL-32B-Instruct on H200 141 GB (TP=2) with `guides/recipes/router/calibration` → **15,751 tok/s**, now set explicitly on the `prefix-cache-affinity-filter` in all three values files and added to the configuration matrix. The measured value is within 1.1% of the plugin default (15,928) the benchmark ran with, so the published numbers are unaffected (the filter's TTFT gate shifts by ~1%, well inside run-to-run noise). Note the calibration workload measures text prefill; vision-encoder prefill is not exercised by it.

kaushikmitr:
with the current workload, the weighted set of scorers work better however based on the optimized baseline and agentic workload results i think this is a better overall setup that generalizes better.

capri-xiyue:
Since we use a prefix cache heavy dataset for our optimized baseline with benchmark results available here https://github.com/llm-d/llm-d/tree/main/guides/optimized-baseline. We should stick to this convention until we adopt a more realistic dataset. Alternatively, if we need to move away from it, we should store it in a separate folder rather than deleting it entirely

ahg-g:
The drop in perf seems significant, right? any idea why?

kaushikmitr:
@ahg-g @capri-xiyue see the updated charts. The main take away is it seems tokens in flight is underestimated for multi modal. I would suggest two main follow ups:
1. See if we can autotune tokens in flight for images and videos. Ideally track non text tokens separately
2. Have a more realistic workload like an agentic workload. 

For this PR, with the params for affinity filter tuned, results are  same as current router and I think it will generlize better (though we dont have data yet to show that yet except that it generlizes better in the non multi modal case). 

Final tuning update (see the amended commit): alongside `affinityThreshold: 0.6` (since the way the workload is designed, the overlaps at abouut 70% or nothing), updated MaxTTFTPenalty to 36s (2x) to take into account that images and video tokens are more expensive than test tokens so affinity pays more. 


**Result:** with both parameters set, the token stack holds sub-second TTFT p90 (0.15–0.40 s) across the whole 5→40 req/s ladder and scales to 11,128 tok/s at rate 40 — +44% throughput over the same-fleet k8s Service with ~125× lower median TTFT at the top of the ladder, 2 vs 27 failures. This also matches the previous weighted-scorer stack head-to-head on the same fleet (within 1% at every rate), so the routing convergence carries no multimodal performance regression.

capri-xiyue:
```
The main take away is it seems tokens in flight is underestimated for multi modal. I would suggest two main follow ups:
See if we can autotune tokens in flight for images and videos. Ideally track non text tokens separately
Have a more realistic workload like an agentic workload.
```


I thought the tokens in flight will use the same token producer that approximate prefix cache aware plugin used, could you share the context why tokens in flight is underestimated for multi modal? @kaushikmitr

kaushikmitr:
> ```
> The main take away is it seems tokens in flight is underestimated for multi modal. I would suggest two main follow ups:
> See if we can autotune tokens in flight for images and videos. Ideally track non text tokens separately
> Have a more realistic workload like an agentic workload.
> ```
> 
> I thought the tokens in flight will use the same token producer that approximate prefix cache aware plugin used, could you share the context why tokens in flight is underestimated for multi modal? @kaushikmitr

i think number of tokens not underestimated if we set the confguraiton correctly (which we are doing here). BBut we use inflight load  to estimate how much prefill work is queued in the endpoint. If its too much work we break prefix affinity, Now an inflight token load of 3000 images tokens and 3000 text tokens is not same as 6000 text tokens. We are not making that distinction, and so we are underestimating the work. One way to compensate for this is increase the maxttftpenaltyMS which we did here.

capri-xiyue (approved):

PR #2086, opened 2026-07-21
docs(guides): use lighter model (Qwen3.5-4B) and reduce replicas for quickstart

haitwang-cloud:
### Problem

The quickstart guide is intended to provide the easiest path for users to evaluate llm-d. However, the quickstart guide currently deploys `Qwen/Qwen3-32B` with `tensor_parallel_size=2` and 8 replicas, requiring **16 NVIDIA GPUs** in total. This is a very high hardware barrier for a getting-started guide and makes it difficult for most users to try llm-d for the first time.

### Changes

- **Switch model** from `Qwen/Qwen3-32B` (~64GB weights) to `Qwen/Qwen3.5-4B` (~8GB weights).
- **Reduce `tensor_parallel_size`** from 2 to 1 (single GPU is sufficient for a 4B model).
- **Reduce replicas** from 8 to 2 (enough to demonstrate load-aware routing).
- **Adjust resource requests/limits**: 1 GPU, 20Gi→40Gi memory (down from 96Gi→128Gi).
- **Update quickstart doc** and kustomization label to reflect the new model name.

### Impact

Total GPU requirement drops from **16 GPUs** to **2 GPUs**, making the quickstart accessible to a much wider audience. Users who want a production-scale deployment can still use the other configurations under `guides/optimized-baseline/modelserver/`.


maugustosilva (dismissed):

guides/optimized-baseline/modelserver/gpu/vllm/base/kustomization.yaml
```diff
@@ -10,7 +10,7 @@ components:
 
 labels:
   - pairs:
-      llm-d.ai/model: Qwen3-32B
+      llm-d.ai/model: Qwen3.5-4B
```
zdtsw, line 13: isnt change here gonna change optimized-baseline benchmark as well?

haitwang-cloud: @zdtsw Could you please help provide more details? I am confused about how I should address your comments.

zdtsw: by switching model used in optimized guide, wouldnt that change what had been benchmarked already? e.g https://github.com/llm-d/llm-d/tree/main/guides/optimized-baseline/benchmark-results/vllm-qwen3-32b-h100 
would you benchmark and provider new data for Qwen3.5-4B as well?

haitwang-cloud: @zdtsw I C, good point. It is okay for me to provide the new benchmark for Qwen3.5-4B. I will add it later

haitwang-cloud:
@zdtsw I ran a benchmark for the proposed reduced-footprint configuration, Qwen/Qwen3.5-4B with TP=1 and 2 replicas, and added the data to this branch.

  The result confirms your concern: on the smaller 2-GPU RTX PRO 6000 setup, this workload saturates early and does not show a clear EPP acceleration benefit over a plain Kubernetes Service. So
  changing the optimized-baseline base configuration would invalidate the existing Qwen3-32B/H100 benchmark reference, while the new reduced-footprint data is not a strong replacement for
  demonstrating optimized-baseline routing performance.

  Given that, I do not think this PR should be merged in its current form. I will close it for now. If we revisit this, we should separate the goals more clearly: keep the optimized-baseline benchmark/reference configuration intact, and introduce any smaller model setup explicitly as a quickstart or reduced-footprint path with its limitations documented.

zdtsw:
> @zdtsw I ran a benchmark for the proposed reduced-footprint configuration, Qwen/Qwen3.5-4B with TP=1 and 2 replicas, and added the data to this branch.
> 
> The result confirms your concern: on the smaller 2-GPU RTX PRO 6000 setup, this workload saturates early and does not show a clear EPP acceleration benefit over a plain Kubernetes Service. So changing the optimized-baseline base configuration would invalidate the existing Qwen3-32B/H100 benchmark reference, while the new reduced-footprint data is not a strong replacement for demonstrating optimized-baseline routing performance.
> 
> Given that, I do not think this PR should be merged in its current form. I will close it for now. If we revisit this, we should separate the goals more clearly: keep the optimized-baseline benchmark/reference configuration intact, and introduce any smaller model setup explicitly as a quickstart or reduced-footprint path with its limitations documented.

thanks for the efforts, appreciated and agreed @haitwang-cloud

PR #2096, opened 2026-07-22
[ROCM] add update_badge input to precise and predicted nightly

vcave:
Add missing update_badge workflow_dispatch input and guard the update-badge job condition


Gregory-Pereira (approved):
/lgtm

PR #2098, opened 2026-07-22
test(autoscaling): validate the KEDA EPP guide with the WVA Kind lifecycle

nourey:
Supports [#1431](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1431) and contributes to the validation path required before [#1326](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1326).

Depends on [llm-d/llm-d-workload-variant-autoscaler#1472](https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1472).

## Background

This work started from the discussions around [#1326](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1326), where we need reliable KEDA and EPP autoscaling coverage before promoting the path into nightly qualification.

The first version of this PR introduced a standalone Kind-based validation runner inside llm-d. That implementation successfully established the direct KEDA and EPP metric contract and the bounded `1 -> 2` scaling behavior but it also added approximately ~2,5k lines of guide specific lifecycle code for:
* Kind cluster creation and cleanup
* Prometheus and KEDA installation
* readiness polling
* diagnostics
* request generation
* artifact handling

Following @mamy-CS 's guidance, we revisited that structure through #1431.

The conclusion was that llm-d should continue to own the canonical guide configuration, while the reusable Kind execution lifecycle should live in the workload-variant-autoscaler repository. Maintaining both implementations would duplicate infrastructure behavior and create a long term risk of drift between the two validation paths. WVA PR llm-d/llm-d-workload-variant-autoscaler#1472 was therefore created to make the existing WVA Kind lifecycle reusable for this contract.

This PR has now been refactored to use that lifecycle instead of maintaining a second runner in llm-d.

### Removed the guide-local runner
The following standalone runner has been removed:

`guides/workload-autoscaling/scripts/test-keda-epp-kind.sh`

This removes ~2.5k lines of duplicated lifecycle and validation code

### Reworked the CI workflow

`.github/workflows/ci-keda-epp-kind.yaml` is now a thin llm-d caller of the reusable WVA lifecycle

The workflow:

1. Checks out llm-d

2. Checks out WVA at the immutable commit

   `8ff1ce41b1765843b99e2039c2616927d825b4fe`

3. Creates a unique one node Kind cluster through WVA

4. Uses the existing llm-d certificate generator to create the Prometheus TLS and CA material

5. Installs Prometheus and KEDA through WVA with:

   `DEPLOY_WVA=false`

6. Installs the canonical llm-d router values and Kind Kustomization

7. Runs only the label-isolated WVA contract:

   `make test-e2e-keda-epp-guide`

8. Collects sanitized diagnostics

9. Deletes only the isolated cluster created for the run

### Preserved llm-d ownership

The canonical guide configuration remains in llm-d and was not moved into WVA

llm-d continues to own:
* the `ScaledObject`
* the `TriggerAuthentication`
* router values
* monitoring and Prometheus values
* KEDA values
* the Kind Kustomize overlay
* the Prometheus TLS resources
* the Kind only CA compatibility configuration
* the CI workflow inputs

WVA owns only the reusable execution mechanics:
* Kind lifecycle
* Prometheus and KEDA installation helpers
* polling and readiness logic
* bounded request stimulus
* diagnostics
* artifact sanitation
* request-Pod cleanup
* isolated-cluster cleanup
* the direct `keda-epp-guide` contract

## What the contract validates
The test verifies that:

* the guide `ScaledObject` becomes `Ready=True`
* the guide `ScaledObject` becomes `Active=True`
* KEDA creates the expected HPA
* the HPA contains exactly two External metrics
* both metrics use `AverageValue`
* deterministic request load causes the first bounded `1 -> 2` transition
* three stable exact-two samples are observed
* no sampled HPA or Deployment value exceeds two
* KEDA produces no `x509`, unknown-authority, or `TriggerError` messages

This PR intentionally does not cover:
* sustained load
* scale down behavior
* performance qualification
* OpenShift specific behavior
* post scale inference validation
* nightly promotion

Those belong to the later [#1326](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1362) qualification layer.

## Local validation
The refactored workflow and retained guide resources were validated locally with:

* `git diff --check`
* `actionlint .github/workflows/ci-keda-epp-kind.yaml`
* YAML parsing
* Bash syntax validation for all embedded workflow blocks
* canonical Kustomize rendering
* Kind Kustomize rendering
* KEDA `2.20.0` Helm rendering
* kube-prometheus-stack `62.7.0` rendering
* router `v0.9.0` rendering
* workflow contract assertions
* Secret metadata and artifact safety fixtures

A complete fresh cluster execution also passed against WVA commit:

`8ff1ce41b1765843b99e2039c2616927d825b4fe`

Observed runtime results:

* exactly `1 of 73` Ginkgo specs selected
* first desired `1 -> 2` transition observed
* three stable exact two samples observed
* no sampled replica value exceeded two
* `1 passed`
* `72 skipped`
* `x509=0`
* `unknown_authority=0`
* `trigger_error=0`
* no PEM or private-key material retained
* no kubeconfig retained
* no Secret payload retained
* isolated Kind cluster cleanup passed

## Dependency
This PR temporarily pins the immutable head commit of WVA PR:
[llm-d/llm-d-workload-variant-autoscaler#1472](https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1472)

Once that PR is merged, this workflow should be updated to reference the immutable upstream commit before this PR is merged.


mamy-CS:
@asm582 you can add your input here if any. Thanks

mamy-CS:
Thank you for the thorough work @nourey. One high-level question before this grows into a gate: I was wondering, since the WVA repo already has a [CPU-only kind](https://github.com/llm-d/llm-d-workload-variant-autoscaler/tree/main/deploy/kind-emulator) + KEDA + EPP e2e harness, and its kind setup is already designed to consume llm-d guides directly. Could we point that existing harness at this guide's kustomization instead of maintaining a separate ~2.5k-line runner here?

Maybe we can export the WVA kind lifecycle as a reusable workflow_call workflow and add one small direct-keda spec that asserts on the guide's own ScaledObject that keeps the test the shipped artifact property you've got here, but with a fraction of the surface to maintain? Or focus on testing the critical path, as I see a lot of self-tests on the runner.

nourey:
Small CI follow up: the initial runs failed before the WVA backed contract started because the workflow requested a nonexistent Kind `.sha256` asset. Commit now consumes Kind’s published `.sha256sum` record directly with `sha256sum --check`. Tool versions and the KEDA EPP contract logic are unchanged. 

a fresh Actions run is now pending maintainer approval

.github/workflows/ci-keda-epp-kind.yaml
```diff
@@ -0,0 +1,384 @@
...
+        uses: actions/checkout@v7
+        with:
+          path: llm-d
+          persist-credentials: false
+
+      - name: Checkout WVA lifecycle
+        uses: actions/checkout@v7
+        with:
+          repository: llm-d/llm-d-workload-variant-autoscaler
+          ref: 8ff1ce41b1765843b99e2039c2616927d825b4fe
```
lionelvillard, line 54: the ref should be main

PR #1902, opened 2026-06-22
Add a guide for LLM Cold Start Optimization

nicolexin:
This PR introduces a guide for LLM Cold Start Optimization under guides/fast-startup/.

**[Note: GKE only]** This guide demonstrates how to combine GKE Image Streaming and GKE Pod Snapshots to eliminate the two main bottlenecks of LLM container deployment (image pull and model engine initialization). Together, these GKE-exclusive features reduce the total cold-start latency of a LLM container (**meta-llama/Llama-3.1-8B-Instruct** serving via vLLM on a **single NVIDIA L4 GPU**) from **nearly 7 minutes to just 15 seconds** (a 96.3% latency reduction).

Deployment Scenario | Image Pull Time | Model/Engine Initialization | Total Startup Time | Latency Reduction
-- | -- | -- | -- | --
Standard Cold Start (Baseline) | ~214s | ~195s | ~409s (6.8 mins) | Baseline
GKE Image Streaming Only | ~4s | ~195s | ~199s (3.3 mins) | ~51.6%
Pod Snapshots Only | ~214s | ~11s | ~225s (3.8 mins) | ~45%
Optimized Stack (Streaming + Snapshots) | ~4s | ~11s | ~15s (15 seconds) | ~96.3%

This PR includes:
- a detailed README file with overview, benchmarking result above, and instructions on how to optimize cold start problem on GKE.
- a recipe that specifically designed to run with pod snapshot feature on GKE.

Limitations & Plans for future:
- **GPU support**: This guide is tested on single GPU using 1 X NVIDIA L4. Multi-GPU support with inter-GPU communication (TP>1) will be tested and added later (e.g. 8 X NVIDIA L4). There is currently a cuda checkpoint limitation on NCCL P2P and NCCL SHM support so cross-GPU communication with A100 and above is limited. TPUs currently not supported. 
- **Duplicate Weights (affecting performance)**: Before pod snapshot trigger, we remove duplicate model weights from disk to reduce the snapshot size. Currently the recipe is only optimized for model weights downloaded from hugging face. If other source is used, the duplicate model weights removal needs to be adjusted accordingly. We are working on integrating the snapshotting feature in vllm so that users do not need to make adjustment in the future.

PR #2105, opened 2026-07-23
feat(snapshot): add GKE fast pod snapshot provider and vLLM wrapper

RyanRosario:
## Summary
Adds `docker/scripts/snapshot` providing GKE Fast Pod Snapshotting and model weight cache clearing for vLLM.
### Key Changes
* **Snapshot Providers (`providers.py`)**: Implements `GKESnapshotProvider` to trigger gVisor sandbox checkpoints via `/proc/gvisor/checkpoint` and clear on-disk weight caches (`~/.cache/huggingface/hub`) to reduce snapshot image size from 40GB+ down to <2GB.
* **vLLM Integration (`vllm/wrapper.py`)**: Provides `patch_vllm_lifespan()` to hook into vLLM's FastAPI lifespan manager, calling `engine.sleep()` to unmap VRAM, triggering the gVisor checkpoint, and calling `engine.wake_up()` upon container restore before uvicorn binds TCP ports.
* **Unit Testing (`test_providers.py`)**: Comprehensive test suite covering provider factory options, POSIX `/proc` descriptor operations, error handling, and lifespan context hooks.
## Testing Done
* `python3 -m unittest docker/scripts/snapshot/test_providers.py` (12/12 tests passing).


RyanRosario:
Tracking Issue created for maintainer discussion and feedback: #2106

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+def patch_vllm_lifespan(app, snapshot_provider: Optional[GKESnapshotProvider] = None):
+    """
+    Patches the FastAPI app lifespan context manager for vLLM snapshotting (single-rank scope).
+
+    Args:
+        app: The vLLM FastAPI application instance.
+        snapshot_provider: Snapshot provider instance. Defaults to GKESnapshotProvider().
+    """
+    if snapshot_provider is None:
+        if os.getenv("SNAPSHOT_PROVIDER", "").lower() in ("none", "disabled", "false", "0"):
```
nicolexin: Can we land on one keyword for gke/gvisor, and treat rest of it as no snapshot provider? This is the reverse.

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+
+    original_lifespan = app.router.lifespan_context
+
+    @asynccontextmanager
+    async def patched_lifespan(app_inst):
+        async with original_lifespan(app_inst) as state:
+            engine = getattr(app_inst.state, "engine_client", None)
+
+            # Release physical VRAM maps (keeps virtual memory addresses stable)
+            if engine and hasattr(engine, "sleep"):
```
nicolexin: I think we can log an error here if the engine does not support sleep.

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+                logger.info("Executing engine.sleep(level=1) to release physical VRAM...")
+                await engine.sleep(level=1)
+
+            # Trigger Checkpoint (if provider is enabled)
+            if snapshot_provider is not None:
```
nicolexin, line 60: If provider is none, we can also skip the sleep and wake up.

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+                await engine.wake_up()
+
+            yield state
+
+    app.router.lifespan_context = patched_lifespan
+    logger.info("Successfully patched vLLM FastAPI lifespan context for GKE snapshotting.")
+    return app
+
+
+if __name__ == "__main__":
```
nicolexin, line 101: Have we tested this wrapper against a real vllm server? This block looks suspicious. 
- I am not sure in which cases we would import the app but might be getting the module not installed error. 
- The host and port is hardcoded. 
- Looks like there is a unicorn server running but I don't think we are parsing the arguments or setup the server / engine properly.

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,109 @@
...
+    Clears cached weight files and triggers container checkpointing by writing to /proc/gvisor/checkpoint.
+    """
+
+    def __init__(
+        self,
+        proc_path: str = GVISOR_CHECKPOINT_PATH,
+        cache_dir: Optional[str] = None,
+    ) -> None:
+        self.proc_path = proc_path
+        self.cache_dir = (
```
nicolexin: Similar to the GKE plugin approach, we should just land on one environment variable for the cache removal and there no need to guess whether the model is hugging face vs GCS vs other provider.

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,109 @@
...
+            or os.path.expanduser("~/.cache/huggingface/hub")
+        )
+
+    def clear_cache(self) -> None:
+        """Clear out cached weight directories if a cache directory exists."""
+        if not self.cache_dir:
+            logger.debug("No cache directory specified to clear.")
+            return
+
+        target_path = pathlib.Path(os.path.expanduser(self.cache_dir))
```
nicolexin, line 71: I think we can simplify this a lot more. See above comments, we should only remove model weights when the clear cache environment variable is given.

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,109 @@
...
+            raise SnapshotError(f"Could not delete locally stored weights at {target_path}: {e}") from e
+
+    def trigger(self) -> None:
+        """Execute weight cache clearing followed by gVisor checkpoint triggering and barrier sync."""
+        self.clear_cache()
+
+        try:
+            fd = os.open(self.proc_path, os.O_RDWR)
+        except PermissionError as e:
+            raise SnapshotError(f"gVisor checkpoint file is not writable: {self.proc_path}") from e
```
nicolexin, line 110: We should not raise an error if permission is denied. It just means pod snapshot is not available. We should log it and exit the function early. Ideally we should check before clearing the cache because there is no need to clear the cache if we are not running snapshot/restore.

docker/scripts/snapshot/test_providers.py
```diff
@@ -0,0 +1,223 @@
...
+    SnapshotError,
+    patch_vllm_lifespan,
+)
+
+CHECKPOINT_PATH = "/proc/gvisor/checkpoint"
+MOCK_FD = 42
+NONEXISTENT_PATH = "/nonexistent/path/to/cache"
+VALID_PROVIDERS = ("gke", "gke_sandbox", "gvisor")
+DISABLED_PROVIDERS = ("none", "disabled", "false", "0")
+
```
nicolexin: May not be relevant anymore but these are not used.

docker/scripts/snapshot/test_providers.py
```diff
@@ -0,0 +1,223 @@
...
+        err.errno = errno.EINVAL
+        mock_read.side_effect = err
+
+        provider = GKESnapshotProvider()
+        with self.assertRaises(SnapshotError) as ctx:
+            provider.trigger()
+        self.assertIn("Snapshot was never initiated", str(ctx.exception))
+
+
+class TestGKEVllmWrapper(unittest.TestCase):
```
nicolexin, line 163: I saw mock provider is created but not passed to the patch_vllm_lifespan at line 199, and I assume the test still pass. But anyways - I am not sure if the test provide much value here since the logic here is fairly simple and straightforward.

nicolexin: I think we can have some CI/CD pipeline eventually for the E2E story, doesn't have to in this PR though.

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,109 @@
...
+    ) -> None:
+        self.proc_path = proc_path
+        self.cache_dir = (
+            cache_dir
+            or os.getenv("HF_HUB_CACHE")
+            or os.getenv("HUGGINGFACE_HUB_CACHE")
+            or os.path.expanduser("~/.cache/huggingface/hub")
+        )
+
+    def clear_cache(self) -> None:
```
nicolexin, line 56: Clear cache must be paired with eager loading (--safetensors-load-strategy eager) otherwise during initialization vllm may just map to the model weights fds and not copy them. We will definitely specify this in the recipe but I am wondering if there is a way to check it explicitly here.

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,109 @@
...
+                            if mdir.is_file() or mdir.is_symlink():
+                                mdir.unlink()
+                            else:
+                                shutil.rmtree(mdir)
+                    else:
+                        logger.debug("No model weight directories (models--*) found under: %s", target_path)
+            else:
+                logger.debug("Cache path does not exist: %s", target_path)
+        except OSError as e:
+            raise SnapshotError(f"Could not delete locally stored weights at {target_path}: {e}") from e
```
nicolexin, line 87: What happens if the error is raised? Does it mean we crash the vllm startup? We should have some error catching at upper level, log the error and handle this gracefully.

docker/scripts/snapshot/vllm/wrapper.py
nicolexin: Also what's the plan for wiring? I don't see any dockerfile change.

docker/scripts/snapshot/vllm/wrapper.py
RyanRosario: This assumes GKE is the default, which I think is fine for now. Either the provider is gke or it's None. The other keywords were left over and were meant to be deleted. There are only two now: "gke" or None.

docker/scripts/snapshot/providers.py
RyanRosario: Yeah. This is now implemented like the plugin. I think this must have been older code from before I made the changes to the plugin.

docker/scripts/snapshot/test_providers.py
RyanRosario: They're actually used in the unit tests, but they are not needed since "gke" is the only valid provider right now.

docker/scripts/snapshot/vllm/wrapper.py
RyanRosario: Added.

docker/scripts/snapshot/vllm/wrapper.py
RyanRosario: Added.

docker/scripts/snapshot/providers.py
RyanRosario: Right, thanks for the correction. The issue isn't a race condition, which was my original communication about this, but rather that deleting the weight files while vLLM has active mmap file descriptors invalidates the memory mapping. This leads to crashes during checkpointing/restoration. I'll need to study how the kernel handles this scenario more closely.

RyanRosario: It seems that we do not want to fail here, in the event that a model doesn't use safetensors models, or a config file but it seems that the standard is using safetensor models. What's unclear here is if we should fail this process if the flag isn't specified for safetensor models.

docker/scripts/snapshot/providers.py
RyanRosario: Added exception handling around the trigger function. This is the best place to put it since we want to abort checkpointing if there is an error when clearing the cache.

We've discussed what should happen when triggering fails. We agreed that we should fallback to normal operation without checkpointing. But if feels icky here. I feel the user should specify what happens on exception:

1. continue without snapshotting
2. abort

docker/scripts/snapshot/providers.py
RyanRosario: I believe this is completed per the previous comments.

docker/scripts/snapshot/providers.py
RyanRosario: Done.

docker/scripts/snapshot/test_providers.py
RyanRosario: Agreed. There is a way to test this without a Dockerfile, but for E2E testing it may be necessary.

docker/scripts/snapshot/vllm/wrapper.py
RyanRosario: Updated the entrypoint block to use 

`runpy.run_module("vllm.entrypoints.openai.api_server", run_name="__main__")`

It now uses VLLM standard argument parsing.

We actually do not need a Dockerfile to test this. Because this acts as a drop-in wrapper around vLLM's standard server module, deployments can enable snapshotting simply by setting the container command/entrypoint to 

`python -m docker.scripts.snapshot.vllm.wrapper <standard vLLM args...>`

without needing a separate Dockerfile.

Because of this, I think we can merge this PR (when ready) without a Dockerfile for the time being. That would be next on the list.

docker/scripts/snapshot/test_providers.py
RyanRosario: Cleaned up the test by removing the unused mock_provider. Instead of checking trigger on a disconnected mock, the test now sets SNAPSHOT_PROVIDER="none" and asserts that both engine.sleep and engine.wake_up are skipped on the VRAM engine.

docker/scripts/snapshot/vllm/wrapper.py
nicolexin: When provider is None or unset, I expect us to not snapshot (i.e. snapshot feature not enabled) instead of defaulting to GKE. This may not happen but just in case this script get packaged into llm-d official vllm docker images, I don't want the default to blow up all the well-lit-path with a forced attempt of snapshot.

Also can we change the name to something like "gke_gvisor"? This is also for future proofing, there might be other gke snapshotting mechanism that does not rely on gvisor in the future.

docker/scripts/snapshot/vllm/wrapper.py
nicolexin: Yeah as long as we can test it end to end, it's fine!

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,150 @@
...
+            os.close(fd)
+
+
+def get_snapshot_provider() -> Optional[GKESnapshotProvider]:
+    """
+    Returns the appropriate snapshot provider instance based on SNAPSHOT_PROVIDER env var.
+    Defaults to GKESnapshotProvider if unset or set to 'gke'.
+    Returns None if SNAPSHOT_PROVIDER is set to any other value.
+    """
+    provider_name = os.getenv("SNAPSHOT_PROVIDER", "gke").lower() or "gke"
```
nicolexin: Snapshot unset still means "to snapshot", let's change it to empty string here instead of "gke". Snapshot feature should be opt in. GKE should not be the default, "not snapshotting" would be the default.

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,150 @@
...
+    # Fetch the current snapshot provider, but if it's not GKE, reset to None.
+    if provider_name == "gke":
```
nicolexin: This was probably missed last week but can we change it to "gke_gvisor"? I think there might be other variant of gke snapshot in the future which does not rely on gvisor.

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def clear_cache(self) -> None:
+        """Clear out cached weight files and directories if a cache directory exists."""
+        if not self.cache_dir:
+            logger.debug("No cache directory specified to clear.")
+            return
+
+        # Log a warning instead of raising an error if eager loading is not detected in sys.argv.
+        # Clearing cached weights without eager loading can break memory-mapped file descriptors,
+        # but we do not fail hard to avoid false positives for non-safetensors models or config files.
```
nicolexin: I think we can remove the "or config files" here. These are small anyways and they are never memory mapped.

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,150 @@
...
+        if not _is_eager_loading_configured():
+            logger.warning(
+                "Clearing model cache without '--safetensors-load-strategy eager' detected. "
+                "Ensure eager loading is enabled so model weights are copied to memory rather than memory-mapped from disk."
```
nicolexin: nit: Make sure eager loading is enabled "for safetensor models"

docker/scripts/snapshot/test_providers.py
nicolexin, line 1: It's good to have unit test but I am not sure if this runs. Have you manually ran the test? Unenforced test can rot very easily and lost it purpose. Can you look into this project and see if we can trigger it in the CI?

 ci-pr-checks.yaml already runs pre-commit and I think we can add the unit test there.

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -0,0 +1,136 @@
...
+import time
+from contextlib import asynccontextmanager
+from typing import Optional
+
+from ..providers import (
+    GKESnapshotProvider,
+    get_snapshot_provider,
+)
+
+_handler = logging.StreamHandler(sys.stdout)
```
nicolexin: Can we use vllm.logger.init_logger instead of starting a StreamHandler?

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -0,0 +1,136 @@
...
+
+    original_lifespan = app.router.lifespan_context
+
+    @asynccontextmanager
+    async def patched_lifespan(app_inst):
+        async with original_lifespan(app_inst) as state:
+            engine = getattr(app_inst.state, "engine_client", None)
+
+            if snapshot_provider is not None:
+                # Release physical VRAM maps (keeps virtual memory addresses stable)
```
nicolexin: Can we check availability (gvisor signal file accessible) before sleep? That way we don't have to sleep and wake up if snapshot is not going to happen.

You can add something like provider.is_available(): https://github.com/nicolexin/llm-d/blob/nccl-test/docker/scripts/snapshot/providers.py#L43.

You can also remove the check inside the provider because it's already checked outside.

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -0,0 +1,136 @@
...
+                if engine and hasattr(engine, "sleep"):
+                    logger.info("Executing engine.sleep(level=1) to release physical VRAM...")
+                    await engine.sleep(level=1)
+                else:
+                    logger.error("vLLM engine does not support sleep(level=1); physical VRAM maps were not released before snapshot.")
+
+                logger.info("Triggering snapshot checkpoint...")
+                snapshot_start_time = time.time()
+                try:
+                    snapshot_provider.trigger()
```
nicolexin: Can we wrap this in asyncio.to_thread() just in case it's long running operation? That way background tasks can still run.

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -0,0 +1,136 @@
...
+                logger.info("TOTAL TIME FROM INITIAL START: %.2f seconds", total_startup_sec)
+
+            yield state
+
+    app.router.lifespan_context = patched_lifespan
+    logger.info("Successfully patched vLLM FastAPI lifespan context for GKE snapshotting.")
+    return app
+
+
+if __name__ == "__main__":
```
nicolexin: reopening here - I suggest to hook build_app instead of monkeypatching uvicorn.

There are two places you can intercept: where the app is created, or where the app is consumed. The current code intercepts at consumption (uvicorn), which means it has to correctly model every way vLLM might call uvicorn (and both vllm and uvicorn are moving targets). The snapshot can also run twice: _patched_uvicorn_run wraps the original lifespan, _patched_server_init wraps a second time.

Alternative (reference https://github.com/nicolexin/llm-d/blob/nccl-test/docker/scripts/snapshot/launcher.py for multiGPU):

```
# docker/scripts/snapshot/launcher.py   (must NOT be named __main__.py)
  import sys

  from vllm.entrypoints.openai import api_server

  from .vllm.wrapper import patch_vllm_lifespan

  _orig_build_app = api_server.build_app


  def _build_app(*args, **kwargs):
      return patch_vllm_lifespan(_orig_build_app(*args, **kwargs))


  # Module level, above the __main__ guard, so it also applies in spawned children.
  api_server.build_app = _build_app

  if __name__ == "__main__":
      from vllm.entrypoints.cli.main import main

      sys.argv = ["vllm", "serve", *sys.argv[1:]]
      main()
```
  Run with python3 -m docker.scripts.snapshot.launcher <vllm serve args...>

This is better because:
- build_app is the constructor of the app and is called exactly once per API per process. No duplicate snapshots.
- It's vllm's own function so we depend on only vllm, and not uvicorn.
- Probably not relevant to this PR at all but I'd like to apply some NCCL communication rendezvous logic into every child process for multiGPU and this design can adapt easily.

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -0,0 +1,136 @@
...
+                    logger.info("Executing engine.wake_up() to restore VRAM...")
+                    await engine.wake_up()
+                else:
+                    logger.error("vLLM engine does not support wake_up(); physical VRAM maps were not re-allocated after snapshot.")
+
+                restore_to_ready_sec = time.time() - restore_start_time
+                total_startup_sec = time.time() - PROCESS_START_TIME
+
+                logger.info("TIME TO READY FROM RESTORE: %.2f seconds", restore_to_ready_sec)
+                logger.info("TOTAL TIME FROM INITIAL START: %.2f seconds", total_startup_sec)
```
nicolexin: Can we remove the time tracking? I don't they are meaningful, at least for the duration tracking.
- restore_to_ready_sec does not really count from where the snapshot is being pulled and process being restored (it only counts wakeup(), which probably already have a log line). Also, k8s events already has when the container started vs being restored as pod snapshot events.
- total_startup_sec counts from when the pod is being frozen. So if you checkpoint today and restore days later you get a number that's not really meaningful for users (I'd say for your own testing this is probably helpful).

docker/scripts/snapshot/providers.py
```diff
@@ -0,0 +1,150 @@
...
+    def trigger(self) -> None:
+        """Execute weight cache clearing followed by gVisor checkpoint triggering and barrier sync."""
+        if not os.access(self.proc_path, os.W_OK):
+            logger.warning("Pod snapshot trigger not available (checkpoint file '%s' is not writable).", self.proc_path)
+            return
+
+        self.clear_cache()
+        try:
+            import gc
+            import ctypes
```
nicolexin: Maybe I missed this earlier - what are we trying to do here? The sleep() release the KV cache and the model weights are cleared above, I am just confused why we need this block for gc.collect and malloc_trim.

RyanRosario: This has to do with reducing the host process memory footprint within the pod, not the KV cache, GPU VRAM or file system. `clear_cache` does not address memory, only the weights on disk.

- `gc.collect()` forces immediate garbage collection of dead objects and circular references in process memory created during model deserialization and weight loading. It is Python specific.
- `malloc_trim(0)` is for memory at the C / OS level. Even after Python clears objects from memeory, heap pages stay mapped in the process rather than returning them to the ernel. `malloc_trim(0)` forces glibc to release that unused heap memory back to the OS.

This has to do with Python holding references to objects that have been deleted (when moved to VRAM or KV cache) and not the weights on the filesystem. 

While not strictly required for the snapshot mechanics, it reduces the memory footprint and helps avoid OOMs when running vLLM on tighter/borderline memory instances, like I was.

docker/scripts/snapshot/vllm/wrapper.py
RyanRosario: How do you suggest we do time tracking? What if I fixed the timing and the names of the metrics to be more precise? Here we would lose that logging output.

docker/scripts/snapshot/providers.py
RyanRosario: Ack

docker/scripts/snapshot/providers.py
RyanRosario: Ack

docker/scripts/snapshot/providers.py
RyanRosario: IIRC I was given the option, and to just pick one. I like your idea of just using gke_gvisor.

Updated the provider name from gke to gke_gvisor to distinguish it from potential future GKE snapshot implementations that do not use gVisor.

docker/scripts/snapshot/providers.py
RyanRosario: Sorry about that. I think you had two request on the previous review and I missed the second one. If there is no `SNAPSHOT_PROVIDER` env variable, set it to "". The overall default is no snapshotting. It's purely opt-in.

docker/scripts/snapshot/vllm/wrapper.py
RyanRosario: I'll create a separate branch that keeps the logging, but have removed it here.

docker/scripts/snapshot/vllm/wrapper.py
RyanRosario: Thanks for the suggestion. The synchronous implementation does work since the engine is asleep and the HTTP port is not yet bound, but offloading the snapshotting to a separate thread via `asyncio.to_thread()` seems to be best practice to avoid starving the main asyncio event loop if any other processing needs to be done while the snapshot is taken. Updated in the latest commit.

docker/scripts/snapshot/vllm/wrapper.py
RyanRosario: Ah. I see where this can trigger twice. Also I like setting this up for multiple. There is also the problem of Uvicorn possibly being updated and this breaking. The launcher moves import execution above `main` at the module level so that child processes also inherit the patching.

docker/scripts/snapshot/vllm/wrapper.py
RyanRosario: Makes total sense. Added `provider.is_available()` to check whether the checkpoint file is writable before calling `engine.sleep()`. If it's not available, it logs a warning and skips the snapshot entirely so we don't do an unnecessary sleep/wake cycle. Updated in the latest commit.

docker/scripts/snapshot/test_providers.py
RyanRosario: Yes, I manually ran the test suite uding `python3 -m unittest discover -t . -s docker/scripts/snapshot` and verified all 27 unit tests pass. 

I've added a test execution step to `.github/workflows/ci-pr-checks.yaml` so GitHub Actions automatically runs the tests on every PR.

RyanRosario:
@nicolexin If necessary, I'd like to split out the user guide to prevent delaying a review cycle for it. I can add it to a new PR. Let me know your thoughts.

docker/scripts/snapshot/README.md
```diff
@@ -0,0 +1,349 @@
+# GKE Fast Pod Snapshot Utility for vLLM
```
nicolexin, line 1: Let's limit the scope of the README for this PR to explain this wrapper only (for developers) and leave the user guide to a separate PR. The user guides needs to live in the /guide folder and should be user centric.

docker/scripts/snapshot/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+```bash
+# Create cluster with Pod Snapshots enabled
+gcloud container clusters create "my-snapshot-cluster" \
+  --region="us-east4" \
+  --release-channel=rapid \
+  --machine-type="e2-standard-4" \
+  --num-nodes=1 \
+  --workload-pool="<PROJECT_ID>.svc.id.goog" \
+  --enable-pod-snapshots
```
nicolexin: We do not need exact steps for cluster creation, node pool creation, bucket creation etc. Refer to the public documentation if available and point out specific caveat if needed.

docker/scripts/snapshot/README.md
```diff
@@ -0,0 +1,349 @@
...
+## Configuration & Environment Variables
+
+### Snapshot Variables
+
+| Variable | Description | Default |
+| :--- | :--- | :--- |
+| `SNAPSHOT_PROVIDER` | Snapshot provider backend. Set to `gke_gvisor` to enable GKE gVisor snapshotting. If unset or any other value, snapshotting is disabled. | `""` (disabled) |
+| `MODEL_CACHE_DIR` | Path to the local model weight cache directory to purge before triggering the checkpoint (e.g. `/root/.cache/huggingface/hub`). | `None` |
+
+### gVisor & NCCL Compatibility Settings
```
nicolexin: I am a little skeptical about some of the flags here. Not urgent since we are tracking user guide separately but just fyi I only need to set VLLM_HOST_IP="127.0.0.1". We should explain why these flags are needed.

docker/scripts/snapshot/README.md
```diff
@@ -0,0 +1,349 @@
...
+  triggerConfig:
+    type: workload       # Allows the Python container to trigger the snapshot
+    postCheckpoint: resume
+```
+
+### 3. Model Server Deployment
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
```
nicolexin: This is a standalone deployment. I am expecting a llm-d style kustomize model server deployment. That way we inherit all the standard settings and only need to overwrite things that's necessary for pod snapshot. See https://github.com/llm-d/llm-d/pull/1902.

docker/scripts/snapshot/README.md
```diff
@@ -0,0 +1,349 @@
...
+When a new pod restores from the snapshot, the vLLM logs show an instant wake-up:
+
+```logs
+(APIServer pid=1) [llm-d.snapshot.wrapper] INFO: Process restored from snapshot checkpoint. Resuming engine...
+(APIServer pid=1) [llm-d.snapshot.wrapper] INFO: Executing engine.wake_up() to restore VRAM...
+(EngineCore pid=55) INFO: It took 0.002859 seconds to wake up tags {'kv_cache', 'weights'}.
+(APIServer pid=1) INFO: Application startup complete.
+```
+
+### 3. Networking & Testing in gVisor
```
nicolexin: Related to this - I think we should include the standard llm-d stack (helmchart to install EPP etc in standalone / gateway mode). Even though this feature can work on its own, I think we should still showcase the value with the full llm-d stack. That way you don't have to make workarounds for the validation steps.

docker/scripts/snapshot/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Key Components
+
+- **[`GKESnapshotProvider`](file:///usr/local/google/home/ryanrosario/Code/llm-d/docker/scripts/snapshot/providers.py)**: Manages gVisor checkpoint trigger via `/proc/gvisor/checkpoint`, cache purging, and memory trimming.
+- **[`patch_vllm_lifespan`](file:///usr/local/google/home/ryanrosario/Code/llm-d/docker/scripts/snapshot/vllm/wrapper.py)**: Async lifespan wrapper that handles the `sleep(1)` -> checkpoint -> `wake_up()` sequence.
+- **[`launcher.py`](file:///usr/local/google/home/ryanrosario/Code/llm-d/docker/scripts/snapshot/launcher.py)**: Intercepts FastAPI app creation to apply the lifespan patch before delegating to `vllm.entrypoints.cli.main`.
+
+---
+
+## Running Unit Tests
```
nicolexin, line 75: nit: can you also mention this is checked in CI/CD?

RyanRosario:
@nicolexin I've copied down your comments on the README. Thank you and thanks for going through them with me today. I've removed the `README.md` from this PR but will make these changes in that PR.

docker/scripts/snapshot/providers.py
RyanRosario: We've agreed to remove the garbage collection for now. There are reasons to include it, but we need to experiment how much memory is being saved by this, and also whether or not the assumption should be that the end user will just pick a larger instance.

RyanRosario:
- [ ] Test the memory saved by using garbage collection and malloc trim vs not

docker/scripts/snapshot/README.md
RyanRosario: Acknowledged and completed!

docker/scripts/snapshot/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+This package provides a drop-in launcher and snapshot provider that hooks into vLLM's FastAPI application lifespan to:
+1. Initialize the vLLM engine and compile CUDA graphs during container cold start.
+2. Put the vLLM engine to sleep (`engine.sleep(level=1)`) to release physical GPU VRAM while preserving virtual memory mappings.
+3. Purge cached model weight files from disk to minimize the checkpoint storage footprint.
+4. Trigger a gVisor userspace checkpoint via `/proc/gvisor/checkpoint`.
+5. Restore the container and wake up the engine (`engine.wake_up()`) to re-allocate physical GPU VRAM upon restoration before binding HTTP ports and serving traffic.
+
+> [!NOTE]
+> This README is intended for developers maintaining and integrating this snapshot utility. For the user-facing deployment guide, cluster configuration, and end-to-end benchmark walkthroughs, refer to the [`/guides`](../../../guides) directory.
```
nicolexin: Thanks for adding this! Can we say the user guide is upcoming (since it does not exist now)?

docker/scripts/snapshot/README.md
```diff
@@ -0,0 +1,321 @@
...
+  - `clear_cache()`: Recursively deletes downloaded model weight files from `MODEL_CACHE_DIR` to reduce checkpoint size.
+  - `is_available()`: Checks if the procfs checkpoint trigger file is writable.
+  - `trigger()`: Clears the cache, opens `/proc/gvisor/checkpoint`, writes `1` to initiate checkpointing, and blocks on a 1-byte read until the container is restored.
+
+- **[`providers.py`](providers.py) (`get_snapshot_provider`)**:
+  Factory function that returns a `GKESnapshotProvider` if `SNAPSHOT_PROVIDER=gke_gvisor`, or `None` if unset/disabled.
+
+---
+
+## Configuration & Environment Variables
```
nicolexin: I think we can remove everything below, except keeping the unit test section since the user guide is going to be elsewhere.

docker/scripts/snapshot/README.md
RyanRosario: Done.

docker/scripts/snapshot/README.md
RyanRosario: Removed all user guide and configuration sections below the component layout. The README now only covers developer/component details and unit tests.

RyanRosario:
Requested changes to logger and README completed.

docker/scripts/snapshot/README.md
RyanRosario: Changed made previously.

docker/scripts/snapshot/vllm/wrapper.py
```diff
@@ -24,16 +24,12 @@
...
-)
-_llm_d_logger = logging.getLogger("llm-d")
-_llm_d_logger.setLevel(logging.INFO)
-if not _llm_d_logger.handlers:
-    _llm_d_logger.addHandler(_handler)
+try:
+    from vllm.logger import init_logger
 
-logger = logging.getLogger("llm-d.snapshot.wrapper")
+    logger = init_logger("llm-d.snapshot.wrapper")
```
nicolexin: The logger name's first segment must start with vllm - that's what makes it a child of the configured logger and gets it a handler.

RyanRosario: Fixed!

docker/scripts/snapshot/launcher.py
```diff
@@ -0,0 +1,40 @@
...
+    from vllm.entrypoints.openai import api_server
+
+    _orig_build_app = api_server.build_app
+
+    def _build_app(*args, **kwargs):
+        return patch_vllm_lifespan(_orig_build_app(*args, **kwargs))
+
+    api_server.build_app = _build_app
+except ImportError:
+    api_server = None
```
nicolexin: Can we add one more thing here? The api server has just recently moved: [#52131 [Frontend] Move api_server.py out openai folder](https://github.com/vllm-project/vllm/pull/52131) and this path will eventually get deprecated. We pin to a released vllm version so this is going to be fine for a while. 

Just for precaution, can we add a log line here in case we run into the import error we should at least log a warning and say api server is not importable and because of that snapshot feature is not available.

nicolexin: This is now resolved, thanks!

nicolexin:
LGTM! All the open comments are for a future user guide that we decided out of scope for this PR, those will be address later.

This is ready for merge, pending CI checks. @ahg-g Can you please take a look? Thank you!

RyanRosario:
@nicolexin CI appears to be blocked waiting for approval of the build-approval environment for the image build. Would a maintainer be able to approve the pending run? Thanks!

RyanRosario:
@nicolexin @Gregory-Pereira  The CI issue was resolved and all checks have passed.

Gregory-Pereira:
> @nicolexin @Gregory-Pereira The CI issue was resolved and all checks have passed.

Nothing actually built here - it skipped the cuda build: https://github.com/llm-d/llm-d/actions/runs/33737147469/job/101534182603?pr=2105. If you would like it to build you would need to extend the cuda build filter to honor your snapshot changes: https://github.com/llm-d/llm-d/blob/main/.github/workflows/build-image.yaml#L158-L167

nicolexin:
Hi @Gregory-Pereira I think by CI, @RyanRosario actually means the newly added unit tests. They were failing before and now they are passing. For the first iteration, our plan is to inject these scripts into the container using config maps (so they will not be built directly into the container yet). That way they are only going to be consumed by exactly the well-lite-path that needs it. If they looks good to you, can you please help approve this change? Thanks a lot!

Gregory-Pereira (approved):
/lgtm

nicolexin:
/lgtm
/approve
@ahg-g can you help merge this? Thanks!

PR #2108, opened 2026-07-23
fix: correct the name of status workflow for wide ep lws

rahulgurnani:
Name for status workflow to update status badge in the README's is incorrect. This PR fixes it.


liu-cong (approved):
thanks for fixing this!

PR #2109, opened 2026-07-23
feat(sglang): add sglang-decode-hicache component for P/D disaggregation

tyuchn:
## What this changes

Adds an opt-in Kustomize component `guides/pd-disaggregation/modelserver/components/sglang-decode-hicache` that enables decode-side prefix caching with a host-DRAM (L2) tier for SGLang P/D disaggregation by layering it over any provider overlay:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - guides/pd-disaggregation/modelserver/gpu/sglang/gke  # or base | coreweave | aws
components:
  - guides/pd-disaggregation/modelserver/components/sglang-decode-hicache
```

When included, the component configures:
- `--disaggregation-decode-enable-radix-cache`: SGLang forces `disable_radix_cache=True` on decode servers in P/D mode unless this flag is passed. Without it, a decode pod keeps no prefix state and every turn of a conversation re-transfers its whole prefix from prefill.
- `--enable-hierarchical-cache`: adds a host-DRAM tier beneath HBM.
- `--hicache-size=64`: sets the host KV pool to 64 GB per TP rank (128 GB per pod at TP=2).
- `requests.memory: 256Gi` and `limits.memory: 256Gi`: reserves the host pool upfront to prevent node memory overcommit (eliminating cgroup OOMs on high-HBM cards like H200/B200) and protects against kubelet memory-pressure eviction ranking (the container cannot exceed its request), while providing ~128 GB headroom above the pinned host pool for model weights, engine runtime, and the 16Gi `/dev/shm` tmpfs. On an 8-GPU node with 4 TP=2 pods, 4 × 256Gi = 1,024 GiB (≤55% of node RAM on `a3-megagpu-8g` or `p5.48xlarge`), so GPUs remain the binding constraint without orphaning memory.

The base manifests and provider overlays (`base`, `aws`, `coreweave`, `gke`) remain completely untouched and default to standard P/D serving without prefix caching.

## Results

2× prefill + 2× decode, all TP=2, Qwen2.5-7B-Instruct, ~12k-token prompts, TTFT p50, on GKE with 4× `a3-megagpu-8g` (NVIDIA H100 Mega 80GB). Routing is **pinned, not observed** — the decode replica is addressed through its routing-proxy sidecar and prefill is pinned with `x-prefiller-host-port` — so prefill-side caching is held constant:

| | before | **after** |
|---|---|---|
| turn 1 (cold) | 1.247 s | 1.271 s |
| turn 2, replica holding the prefix | 1.235 s | **0.262 s** |
| turn 2, a different decode replica | 1.258 s | 1.203 s |

Flushing the prefill radix cache too (forcing full prefill recompute), turn 2 on the replica holding the prefix goes 1.372 s → **0.417 s** (3.3× win). The baseline is not a strawman: on turn 2 prefill reported `#cached-token: 12032` of 12160 (~99% hit) and TTFT barely moved — prefill caching is not the lever, avoiding decode-side KV transfer is.

Stability: 4 clean runs in this configuration, 0 crashes and 0 restarts.

## Routing and Incompatibilities

- **Replica-local**: the router's decode scheduling profile selects the least-busy pod without prefix-cache affinity (unlike the prefill profile, which has affinity). A follow-up turn hits the warm decode replica only when scheduled there; otherwise behavior matches the pre-change baseline.
- **Incompatibilities**: SGLang's decode radix cache currently conflicts with speculative decoding (`--speculative-algorithm`), `--enable-hisparse`, and `--disaggregation-transfer-backend fake`. Keeping this as an opt-in component ensures `base/` remains compatible with speculative decoding.

## Note on L3 (shared filesystem offload)

A shared `ReadWriteMany` L3 tier (`--hicache-storage-backend file`) was also implemented and benchmarked, and is **not** included here. In P/D disaggregation, the alternative to reading a prefix from shared storage is re-transferring KV from a prefill pod rather than recomputing prefill (~0.17 s for 12k tokens). Shared storage must beat the P/D interconnect rather than GPU compute, which is generally disadvantageous.


rahulgurnani:
Thanks for the change @tyuchn , from what I understand, the offloading story for vLLM with pd disaggregation is also missing. I think we should add support for vLLM offloading as well if we are adding support here. Would love to hear from @tlrmchlsmth on this as well.

guides/pd-disaggregation/README.md
```diff
@@ -252,17 +252,29 @@ kubectl apply -n ${NAMESPACE} -k ${REPO_ROOT}/guides/${GUIDE_NAME}/modelserver/g
...
-export INFRA_PROVIDER=base # base | coreweave | gke | cks-mooncake
+export INFRA_PROVIDER=base # base | coreweave | gke | aws
 
 kubectl apply -n ${NAMESPACE} -k ${REPO_ROOT}/guides/${GUIDE_NAME}/modelserver/gpu/sglang/${INFRA_PROVIDER}
 ```
 
 SGLang-specific notes:
 
 * **Engine flags**: prefill and decode pods launch with `--disaggregation-mode={prefill,decode}` and `--disaggregation-transfer-backend=nixl`. The decode pod's routing-proxy sidecar is configured with `--kv-connector=sglang`.
+* **Decode-side prefix caching (opt-in)**: to enable decode-side prefix caching with a host-DRAM (L2) tier, layer the `sglang-decode-hicache` component on top of your provider overlay in a `kustomization.yaml`:
```
ahg-g, line 266: why do we want this as an opt-in instead of just enabling on both prefill and decode for all p/d guides?

tyuchn: Good question — this is my thinking process to structure this as an opt-in component focused on SGLang decode:
1. **Why not on prefill:**
   - **Prefill already has prefix caching enabled in GPU HBM by default** (`disable_radix_cache=False`), and the router's `prefix-cache-affinity-filter` already steers matching prompts to the warm prefill replica.
   - **Low HBM pressure:** Unlike decode (which must hold KV pages across hundreds of generation steps for concurrent streams), prefill frees prompt KV immediately after transferring it to decode, so its HBM radix tree turns over much faster without needing host-DRAM spillover.
   - **Host RAM savings:** Skipping it on prefill saves 128 GB of pinned host DRAM per pod (`256Gi` request vs `64Gi`).

2. **Why a dedicated opt-in component instead of default on all P/D manifests:**
   - **Avoids breaking speculative decoding:** SGLang's decode radix cache in P/D mode currently conflicts with speculative decoding (`--speculative-algorithm`), `--enable-hisparse`, and `--disaggregation-transfer-backend fake`. Making it default in `base/` would break those configurations.
   - **Memory footprint in `base/`:** Reserving the pinned host pool requires `256Gi` requests/limits per pod (`1,024 GiB` for 2P+2D). While this fits comfortably on 8-GPU production nodes (`a3-megagpu-8g` / `p5.48xlarge`), it would prevent the default base manifests from scheduling on smaller dev/test clusters.
   - **Decode router has no prefix affinity:** The router's decode profile balances load via `active-request-scorer` (least-busy) with no cache affinity, so decode hits are replica-local (~1/N). Workloads that are single-turn or use large decode pools shouldn't pay the host RAM penalty by default.

If we prefer batteries-included defaults for production overlays (e.g. enabling it by default in `gke/` where nodes have 1.8 TB+ RAM), happy to move it there.

PR #2116, opened 2026-07-26
chore: remove explicit --vllm-port=8200 from sidecar configs

zdtsw:
## What

Remove the now-redundant `--vllm-port=8200` from all routing-sidecar configs.

## Why

- llm-d/llm-d-router#1723 changed the disagg sidecar's `--vllm-port` default  from 8001 to 8200 (`defaultVLLMPort` in `pkg/sidecar/proxy/options.go`), so  the explicit flag is a no-op.
- Across all guides, every decode server paired with a sidecar listens on 8200,  so the flag is not needed anywhere.


rahulgurnani:
thanks for sending this out, looks good to me.

Gregory-Pereira (approved):
/lgtm

PR #2119, opened 2026-07-27
Coordinator's guide

roytman:
Add a well-lit guide for Coordinator-based disaggregation


guides/coord-disaggregation/coordinator/patch-pd-only.yaml
```diff
@@ -0,0 +1,37 @@
...
+data:
+  coordinator.yaml: |
+    log_level: 2
+    server:
+      listen_addr: ":8080"
+      read_timeout: 30s
+      write_timeout: 120s
+
+    gateway:
+      address: "http://llm-d-inference-gateway-${PROVIDER_NAME}.${NAMESPACE}.svc:80"
```
elevran: This hardcodes `llm-d-inference-gateway-${PROVIDER_NAME}.${NAMESPACE}.svc`, but the base configmap.yaml uses `${GATEWAY_SERVICE}` specifically because that value differs between Gateway mode and Standalone Mode (`${GUIDE_NAME}-epp`). The comment above claims this line is "copied whole from configmap.yaml" but it isn't - applying this patch after choosing PD-only + Standalone Mode will point the coordinator at a Gateway service that doesn't exist, breaking all outbound calls. Should use `${GATEWAY_SERVICE}` here too.

guides/coord-disaggregation/README.md
```diff
@@ -0,0 +1,517 @@
...
+  role's pods via `modelServers.matchLabels`. Each EPP runs a single `default`
+  scheduling profile picked implicitly by the chart's `single-profile-handler`
+  (no profile-selection plugin needed, since a role-scoped EPP never has to
+  choose between profiles) — no custom image either. Each role's file does
+  configure its own scorer plugins (see below); that's just picking which
+  endpoint within the role, not which profile to run. The Gateway's `HTTPRoute`
+  (not the EPP) is what dispatches each `EPP-Profile` value to the right EPP's
+  InferencePool.
+
+## Default Configuration (will be updated after finalizeng the guide)
```
elevran: nit: 
"(will be updated after finalizeng the guide)" has a typo and reads like a draft note that shouldn't ship.

guides/coord-disaggregation/coordinator/configmap.yaml
```diff
@@ -0,0 +1,43 @@
...
+kind: ConfigMap
+metadata:
+  name: llm-d-coordinator-config
+data:
+  coordinator.yaml: |
+    log_level: 2
+    server:
+      listen_addr: ":8080"
+      read_timeout: 30s
+      write_timeout: 120s
```
elevran: nit: 
`write_timeout here` is 120s, but the Gateway/HTTPRoute request timeouts allow 300s. For a full encode-prefill-decode pipeline (especially with a cold render container), the coordinator could truncate a response before the Gateway's own timeout kicks in. Worth aligning these.

guides/coord-disaggregation/coordinator/deployment.yaml
```diff
@@ -0,0 +1,134 @@
...
+      app: llm-d-coordinator
+  template:
+    metadata:
+      labels:
+        app: llm-d-coordinator
+    spec:
+      terminationGracePeriodSeconds: 30
+      containers:
+      - name: coordinator
+        image: ghcr.io/roytman/coordinator:07657e922
```
elevran: nit:
pinned to a personal registry (ghcr.io/roytman/coordinator). The vLLM image override has a `# TODO: Remove once upstream...` comment explaining why; this one doesn't say when/if it moves to an official image.

I suggest a top level README note that some images are from private repo's due to pending PRs not merged yet (would be good to list them).

guides/coord-disaggregation/router/coord-disaggregation.values.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 8081
+
+  epp:
+    # llm-d-router#2134 (single EPP for encode/prefill/decode) is merged but may
+    # not be in an official release image yet; this is a build of llm-d-router
+    # containing it. Swap back to the chart default once it is.
+    image:
```
elevran, line 34: nit: same as the coordinator image - ghcr.io/roytman/llm-d-router-endpoint-picker is a personal registry. The comment above notes it's pending release but doesn't flag ongoing maintenance risk if that branch moves.

elevran (commented):
Nice work on this guide - it's thorough and the tradeoffs between the single-EPP and 3-EPP topologies, plus the security note on EPP-Profile not being a trust boundary, are well explained. Found one deployment-breaking bug in the PD-only patch path, plus a few smaller items below.
The use of personal repo images warrants a top level README note that clearly stands out and explain which PRs are pending.

guides/coord-disaggregation/README.md
roytman: yes, it is a draft node. During development and initial testing, I don't want to create heavy deployments.

guides/coord-disaggregation/coordinator/configmap.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: v1
```
capri-xiyue, line 1: I'm wondering should we have a guides/disaggregation/coord and guides/disaggregation/sidecar, and both of them should have pd and epd?

guides/coord-disaggregation/coordinator/deployment.yaml
```diff
@@ -0,0 +1,134 @@
...
+      - name: vllm-render
+        image: vllm/vllm-openai-cpu:v0.21.0
+        imagePullPolicy: IfNotPresent
+        # "vllm launch render" runs only the preprocessing/postprocessing HTTP layer
+        # (no GPU/engine warmup, no encoder-cache profiling) -- "vllm serve" boots
+        # the full inference engine, which for this model hangs for 10+ minutes on
+        # a CPU-backend warmup pass profiling a synthetic max-resolution image,
+        # blowing well past the startup probe below.
+        command: ["vllm", "launch", "render"]
+        args: ["Qwen/Qwen3-VL-2B-Instruct", "--port=8000"]
```
capri-xiyue: For the default model, should we use the same model in multimodal routing guide if it is for epd. If it is pd, should we use the same model as https://github.com/llm-d/llm-d/tree/main/guides/pd-disaggregation

guides/coord-disaggregation/modelserver/gpu/vllm/common/encode-deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      llm-d.ai/role: encode
+  template:
+    metadata:
+      labels:
+        llm-d.ai/role: encode
+    spec:
+      serviceAccountName: sa
```
capri-xiyue, line 17: nits: should we use a service account name that is more specific

guides/coord-disaggregation/router/coord-disaggregation-encode.values.yaml
```diff
@@ -0,0 +1,45 @@
...
+  extraServicePorts:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 8081
+
+  epp:
+    pluginsConfigFile: "coordinator-plugins.yaml"
+    pluginsCustomConfig:
+      coordinator-plugins.yaml: |
```
capri-xiyue, line 29: For the encode plugins, should we make the mm-cache-affinity-scorer as the default?

guides/coord-disaggregation/router/coord-disaggregation.values.yaml
```diff
@@ -0,0 +1,85 @@
...
+        - type: approx-prefix-cache-producer
+        - type: inflight-load-producer
+        - type: prefix-cache-affinity-filter
+        - type: token-load-scorer
+        - type: active-request-scorer
+        - type: max-score-picker
+        schedulingProfiles:
+        - name: encode
+          plugins:
+          - pluginRef: encode-filter
```
capri-xiyue, line 70: Same here, should we use mm-cache-affinity-scorer here

guides/coord-disaggregation/coordinator/configmap.yaml
roytman: It's possible, but I think the Coordinator needs a bit more maturity first. At this stage, it's probably better not to mix it with the sidecar.

guides/coord-disaggregation/modelserver/gpu/vllm/common/encode-deployment.yaml
roytman: Actually, the `serviceAccountName: sa` in the `encode/prefill` deployments is just the base name — the `namePrefix: coord-disaggregation-nvidia-gpu-vllm-` from the kustomization prepends to it, and the nameReference transformer rewrites the reference automatically.

So `sa` in source is changed to `coord-disaggregation-nvidia-gpu-vllm-sa` in the cluster.
This mechanism is shared among all guides.

guides/coord-disaggregation/README.md
```diff
@@ -0,0 +1,517 @@
...
+
+#### Single EPP (default)
+
+One llm-d Router release, EPP, and InferencePool cover all three roles.
+
+2. *Deploy the llm-d Router*:
+
+```bash
+export PROVIDER_NAME=gke # other: na, agentgateway, or istio
+export GATEWAY_SERVICE=llm-d-inference-gateway-${PROVIDER_NAME}
```
danehans: Agentgateway normally gives its generated proxy Service the Gateway’s name. It would be `llm-d-inference-gateway` instead of `llm-d-inference-gateway-agentgateway` ([xref](https://agentgateway.dev/docs/kubernetes/latest/setup/gateway/)).

Consider replacing the partially constructed service address with a complete address:
```
gateway:
  address: "${GATEWAY_ADDRESS}"
```

Then document:
```
# Agentgateway Gateway mode
export GATEWAY_SERVICE=llm-d-inference-gateway
export GATEWAY_ADDRESS="http://${GATEWAY_SERVICE}.${NAMESPACE}.svc:80"
export CLIENT_SERVICE=${GATEWAY_SERVICE}
export CLIENT_PORT=80
```

Standalone mode can set:

```
export GATEWAY_ADDRESS="http://${GUIDE_NAME}-epp.${NAMESPACE}.svc:80"
```

guides/coord-disaggregation/README.md
```diff
@@ -0,0 +1,517 @@
...
+     `EPP-Profile` → the Coordinator (client-facing inference calls).
+   - The same three paths **with** `EPP-Profile` → this router's EPP (the Coordinator's
+     own internal per-phase calls reuse those same paths, so both HTTPRoutes match
+     them at the same exact-path specificity; the header match then breaks the tie in
+     the router's favor — see the comments in both files for why path specificity
+     must match for this to work).
+   - Everything else without `EPP-Profile` (e.g. `/v1/models`, `/health`) → this
+     router's EPP too, which already falls back to its `decode` scheduling profile
+     when the header is absent.
+
```
danehans, line 229: Agentgateway supports this routing layout through standard Gateway API `HTTPRoute` matching and `InferencePool` backends. The client-facing route has no EPP-Profile match while the Coordinator’s internal calls match the more specific path-and-header rules and are sent to the appropriate EPP profile or role-specific InferencePool ([xref](https://agentgateway.dev/docs/kubernetes/main/llm/multiple-inference-pools/)).

guides/coord-disaggregation/README.md
```diff
@@ -0,0 +1,517 @@
...
+kubectl delete -n ${NAMESPACE} -k ${REPO_ROOT}/guides/${GUIDE_NAME}/modelserver/gpu/vllm/${INFRA_PROVIDER}
+```
+
+This deletes every resource the guide created, but leaves the namespace itself (and
+anything else in it, like the `llm-d-hf-token` secret) alone — `kubectl delete
+namespace ${NAMESPACE}` if you want it gone entirely.
+
+If you used Gateway mode and nothing else in your cluster still uses it, also remove
+the `llm-d-inference-gateway` Gateway by following [the gateway cleanup guide](../../docs/infrastructure/gateway/gke.md#cleanup).
```
danehans: These are GKE-specific cleanup instructions. Consider replacing with the selected provider’s guide, and add agentgateway troubleshooting commands:

```
kubectl get pods -n agentgateway-system
kubectl logs -n agentgateway-system deployment/agentgateway --tail=100
kubectl describe gateway llm-d-inference-gateway -n ${NAMESPACE}
```

danehans:
@roytman thanks for creating this PR and including "agentgateway" as a `provider.name` in the comments. I added a few specific review comments. Additionally, I would like to extend your work by:

1. Adding an agentgateway installation path by replacing the generic “follow the gateway guides” instruction with agentgateway as the demonstrated self-installed option.

2. Adding agentgateway-specific verification, i.e.:

      ```
      kubectl get gateway llm-d-inference-gateway -n ${NAMESPACE}
      kubectl get httproute coordinator ${GUIDE_NAME} -n ${NAMESPACE}
      kubectl describe httproute coordinator -n ${NAMESPACE}
      kubectl describe httproute ${GUIDE_NAME} -n ${NAMESPACE}
      kubectl get inferencepool -n ${NAMESPACE}
      ```

      The guide should require:
      - Gateway has `Programmed=True`.
      - Both routes have `Accepted=True` and `ResolvedRefs=True`.
      - A request without `EPP-Profile` reaches the Coordinator.
      - Coordinator-generated calls with `EPP-Profile` reach the EPP.
      - etc.

Thoughts?

guides/coord-disaggregation/coordinator/deployment.yaml
roytman: now it's ghcr.io/llm-d/llm-d-router-coordinator:main.
Later we should replace main by the release tag.

guides/coord-disaggregation/router/coord-disaggregation.values.yaml
roytman: now it's `llm-d-router-endpoint-picker:main`

guides/coord-disaggregation/coordinator/deployment.yaml
roytman: Yes, i replaced the model with `Qwen/Qwen3-VL-32B-Instruct`.
The model `openai/gpt-oss-120b` from the pd guide is much heavy and requires more TP. I'll leave it for the guides separation.

guides/coord-disaggregation/README.md
roytman: I replaced GKE specific instruction with istio and agentgateway instructions. 
The troubleshooting commands are part of each provider guides.

guides/coord-disaggregation/README.md
roytman: can you add it in a separate PR?

roytman:
> @roytman thanks for creating this PR and including "agentgateway" as a `provider.name` in the comments. I added a few specific review comments. Additionally, I would like to extend your work by:
> 
>     1. Adding an agentgateway installation path by replacing the generic “follow the gateway guides” instruction with agentgateway as the demonstrated self-installed option.
> 
>     2. Adding agentgateway-specific verification, i.e.:
>        ```
>        kubectl get gateway llm-d-inference-gateway -n ${NAMESPACE}
>        kubectl get httproute coordinator ${GUIDE_NAME} -n ${NAMESPACE}
>        kubectl describe httproute coordinator -n ${NAMESPACE}
>        kubectl describe httproute ${GUIDE_NAME} -n ${NAMESPACE}
>        kubectl get inferencepool -n ${NAMESPACE}
>        ```
>        The guide should require:
>        
>        * Gateway has `Programmed=True`.
>        * Both routes have `Accepted=True` and `ResolvedRefs=True`.
>        * A request without `EPP-Profile` reaches the Coordinator.
>        * Coordinator-generated calls with `EPP-Profile` reach the EPP.
>        * etc.
> 
> 
> Thoughts?

Hi Daneyon, @danehans,
I've addressed some of your feedback and would be happy to work together with you to validate that this works on `agentgateway`.
A few additional notes regarding your comments:

- Other guide e.g. [pd-disaggregation](https://github.com/llm-d/llm-d/tree/main/guides/pd-disaggregation#installation-instructions) also point directly to gateway documentation where users can find detailed setup instructions for their specific gateway provider. Those guides also cover provider-specific verification, troubleshooting, and cleanup steps.
- I intentionally avoided routing all requests lacking the `EPP-Profile` header to the Coordinator; instead, some of them are forwarded directly to the vLLM instances (Decoder). (But we can discuss it)

danehans:
@roytman thanks for the feedback. That separation makes sense. +1 on not duplicating agentgateway installation, provider verification, etc docs. I also agree that my statement about requests without EPP-Profile was too broad. Only the documented client-facing inference paths without that header should reach the Coordinator. I think one Coordinator-specific verification remains useful because the provider guide cannot validate this guide’s two-HTTPRoute routing setup. Could we add a short check after deploying the Coordinator:

```sh
kubectl get gateway,httproute,inferencepool -n ${NAMESPACE}
kubectl describe httproute coordinator ${GUIDE_NAME} -n ${NAMESPACE}
```

Both routes should report `Accepted=True` and `ResolvedRefs=True`. I’m also happy to help validate the single-EPP topology on agentgateway.

guides/coord-disaggregation/router/coord-disaggregation-encode.values.yaml
roytman: I guess you mean `mm-embeddings-cache-scorer`.
It was implemented to work in the monolith, un-disaggregated deployments, and will actually be a "no-op" for encoder profile.

guides/coord-disaggregation/router/coord-disaggregation.values.yaml
roytman: Same here; no, we should not.

roytman:
> @roytman thanks for the feedback. That separation makes sense. +1 on not duplicating agentgateway installation, provider verification, etc docs. I also agree that my statement about requests without EPP-Profile was too broad. Only the documented client-facing inference paths without that header should reach the Coordinator. I think one Coordinator-specific verification remains useful because the provider guide cannot validate this guide’s two-HTTPRoute routing setup. Could we add a short check after deploying the Coordinator:
> 
> ```shell
> kubectl get gateway,httproute,inferencepool -n ${NAMESPACE}
> kubectl describe httproute coordinator ${GUIDE_NAME} -n ${NAMESPACE}
> ```
> 
> Both routes should report `Accepted=True` and `ResolvedRefs=True`. I’m also happy to help validate the single-EPP topology on agentgateway.

@danehans we can add this check, but other guides just point to the gateways installation instructions, and each gateway instruction has a verify the installation part, e.g., for [agentgateway](https://github.com/llm-d/llm-d/blob/main/docs/infrastructure/gateway/agentgateway.md):
 

> Verify the installation:
> 
> kubectl get pods -n agentgateway-system
> kubectl get gatewayclass agentgateway
> 
> Expected output:
> 
> NAME           CONTROLLER                      ACCEPTED   AGE
> agentgateway   agentgateway.dev/agentgateway   True       30s

Why do you think we should add the validation instructions here?

roytman:
@elevran , @capri-xiyue , @danehans thank you for your reviews. 
I believe I've addressed most, if not all, of your comments.

I suggest merging this initial guide as a baseline, and then everyone can add or update instructions for specific platforms (e.g., GKE, agentgateway, etc.)

CC: @ahg-g

guides/coord-disaggregation/router/coord-disaggregation.values.yaml
```diff
@@ -0,0 +1,87 @@
...
+  create: false
+
+router:
+  extraServicePorts:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 8081
+
+  epp:
```
elevran, line 30: This profile config reuses inflight-load-producer plus prefix-cache-affinity-filter/token-load-scorer, the same plugins pd-disaggregation.values.yaml pins to `epp.replicas: 1` for (comment there: their state is local to each EPP process, so more than one replica would score against a partial view). This file relies on the chart default (also 1) instead of pinning it explicitly, so nothing is broken today, but there's no guard or explanation if someone scales the EPP later. Worth copying pd-disaggregation's explicit `replicas: 1` and comment here.

guides/coord-disaggregation/router/coord-disaggregation-prefill.values.yaml
```diff
@@ -0,0 +1,53 @@
...
+  create: false
+
+router:
+  extraServicePorts:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 8081
+
+  epp:
```
elevran, line 31: Same as the single-EPP values file: this profile also uses inflight-load-producer + prefix-cache-affinity-filter/token-load-scorer, which pd-disaggregation.values.yaml pins to `epp.replicas: 1` for (local-to-process state). Same suggestion applies here.

guides/coord-disaggregation/router/coord-disaggregation-decode.values.yaml
```diff
@@ -0,0 +1,45 @@
...
+  create: false
+
+router:
+  extraServicePorts:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 8081
+
+  epp:
```
elevran, line 26: Same as the other three values files here: active-request-scorer's in-flight accounting is also local to the EPP process. Worth pinning `replicas: 1` with the same rationale.

guides/coord-disaggregation/router/coord-disaggregation-encode.values.yaml
```diff
@@ -0,0 +1,45 @@
...
+  create: false
+
+router:
+  extraServicePorts:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 8081
+
+  epp:
```
elevran, line 26: Same as the other three values files here: active-request-scorer's in-flight accounting is also local to the EPP process. Worth pinning `replicas: 1` with the same rationale.

elevran (commented):
Thorough guide, and the previous round of feedback all looks addressed. One more thing worth a look before merge.

The guide reuses plugins from pd-disaggregation (inflight-load-producer feeding active-request-scorer/token-load-scorer/prefix-cache-affinity-filter) whose scoring state lives in the memory of a single EPP process, not in any shared store. 
If the EPP ever runs with more than one replica, each instance only sees the requests it personally handled, so its load/cache-affinity picture is partial and its scoring decisions become unreliable. `pd-disaggregation` guards against this explicitly: it sets `epp.replicas: 1` with a comment spelling out why. 
This guide's four values files use the exact same stateful plugins but they just happen to inherit `replicas: 1` from the chart's default. Nothing is broken today, but there's no comment or explicit pin stopping someone from scaling the EPP later and silently corrupting the scoring.

elevran (approved):
thanks @roytman! all my feedback was addresses - LGTM

guides/coord-disaggregation/router/httproute-3-epp.yaml
```diff
@@ -0,0 +1,114 @@
...
+        name: EPP-Profile
+        value: encode
+    - path:
+        type: Exact
+        value: /inference/v1/generate
+      headers:
+      - type: Exact
+        name: EPP-Profile
+        value: encode
+    timeouts:
```
capri-xiyue, line 50: Such timeout is not supported on httproute on GKE, can we make it a conditional only when the provided is not GKE, same for other timeout in this file.

guides/coord-disaggregation/router/httproute.yaml
```diff
@@ -0,0 +1,107 @@
...
+        value: /inference/v1/generate
+      headers:
+      - type: Exact
+        name: EPP-Profile
+        value: decode
+    - path:
+        type: PathPrefix
+        value: /
+    timeouts:
+      request: 300s
```
capri-xiyue, line 107: same here, such timeout is not supported on httproute on GKE, can we make it a conditional only when the provided is not GKE

guides/coord-disaggregation/coordinator/deployment.yaml
```diff
@@ -0,0 +1,138 @@
...
+            memory: 16Gi
+          limits:
+            cpu: "16"
+            memory: 32Gi
+        ports:
+        - name: http
+          containerPort: 8080
+        livenessProbe:
+          httpGet:
+            path: /healthz
```
capri-xiyue, line 44: To make it work on GKE,
We need to have a healthcheckpolicy as follows

```
apiVersion: networking.gke.io/v1
kind: HealthCheckPolicy
metadata:
  name: llm-d-coordinator-hcp
  namespace: llm-d-coord-disaggregation
spec:
  default:
    checkIntervalSec: 2
    timeoutSec: 2
    config:
      type: HTTP
      httpHealthCheck:
        port: 8080
        requestPath: /healthz
  targetRef:
    group: ""
    kind: Service
    name: llm-d-coordinator
```

guides/coord-disaggregation/README.md
```diff
@@ -0,0 +1,521 @@
...
+```bash
+export PROVIDER_NAME=istio # other: na, agentgateway
+# agentgateway's generated Service takes the Gateway's own name, with no
+# -<provider> suffix (unlike istio/gke/na) -- https://agentgateway.dev/docs/kubernetes/latest/setup/gateway/
+if [ "${PROVIDER_NAME}" = agentgateway ]; then
+  export GATEWAY_SERVICE=llm-d-inference-gateway
+else
+  export GATEWAY_SERVICE=llm-d-inference-gateway-${PROVIDER_NAME}
+fi
+export GATEWAY_ADDRESS="http://${GATEWAY_SERVICE}.${NAMESPACE}.svc:80"
```
capri-xiyue, line 204: gateway doesn't have such internal svc service when we use regional external load balancer, will need to differentiate among different gateway. 

For GKE external regional load balancer, need to use http://${IP}:${PORT}

As this GATEWAY_ADDRESS is only used when creating coordinator deployment and we won't know it until the gateway gets reconciled. Can we have this set up later when we need it and when we have such gateway address info?

capri-xiyue:
I went ahead and fixed the issues from my recent comments locally. I just tested it with GKE, and it successfully processed an inference request.

ahg-g:
@capri-xiyue pls follow up with the GKE-specific tunings

PR #2126, opened 2026-07-28
Add guides for 1p1d dp=ep=8, tp=1 for DSV3, runs only MORI-IO and intra node Wide-EP

shikamd123:
Single-pod 1P1D P/D disaggregation overlay for DeepSeek-V3 on AMD Instinct using the MoRI-IO KV connector (RDMA WRITE) and MoRI-EP Wide-EP backend (intra node ep) (DP=8, EP=8, TP=1). Follows the sibling NIXL overlay's base/ + amd-ci/ split.


vcave (changes requested):
Please move over to a wide-ep guide. Then clean-up unused env vars, properly isolate amd-ci cluster-specific changes under a patch-amd-ci.yaml, reduce comments clutter, clearly comment on which settings are temporary workarounds. You can check PR 2147 for reference but note we are looking into factorizing the pvc part.

PR #2135, opened 2026-07-29
docs(async/multitenant): make the saturation divisor a rendered SAT_CAP placeholder

shimib:
## What

Scenario C of the multitenant async guide ("priority under saturation") could never actually reach saturation with the guide's own worker-pool sizes.

Each pool's gate is:

```yaml
gate_type: wait-on-refuse
gate_params:
  gate: |
    {"gate_type":"prometheus-query","gate_params":{"query":"clamp(1 - sum(vllm:num_requests_running{inference_pool=\"POOL_A\"})/20, 0, 1)","fallback":"1"}}
```

`prometheus-query` closes at budget `<= 0` and `clamp(..., 0, 1)` floors the budget at 0, so the gate closes only once **20** requests are running on that model. But:

- the gates are per-model, and each pool has `workers: 8`;
- Scenario C's load is entirely async, so the only thing driving `vllm:num_requests_running` is that pool's own workers;
- a worker evaluates the gate while *holding a message it has not dispatched yet*, so at most `workers - 1` = **7** of the pool's requests are running at evaluation time.

The budget bottoms out around `1 - 7/20`, never reaches 0, `wait-on-refuse` never parks, and the run completes and looks healthy — the named scenario silently never happens.

## Change

Docs/values only; no chart or code changes.

- Replace the hardcoded `/20` divisor with a `SAT_CAP` placeholder in all three saturation overlays (`values/redis/saturation-prometheus.yaml`, `values/pubsub/saturation-prometheus.yaml`, `values/pubsub/saturation-gmp.yaml`), rendered by the guide's existing sed `render()` helper alongside `NAMESPACE`/`POOL_A`/`POOL_B`.
- Export a default `SAT_CAP=4` against `workers: 8` in the prerequisites block, and document the sizing rule: `SAT_CAP` must stay **below** the pool's `workers`, or async load alone can never close the gate. The margin is deliberate — `vllm:num_requests_running` counts only requests the server is actively running, not ones waiting in its queue.
- Say what the reader should observe: `model-a`'s budget hits **exactly 0** and stays there in stretches while `model-b` sits at `~1`; if it never reaches 0, the scenario is not happening. Added an in-flight query (`sum by (pool_name) (llm_d_async_async_inflight_requests)`) so parked pools are visible.
- Note that in production the divisor is a real capacity number (`ready pods × per-pod concurrency`), not bounded by this processor's workers, since the gate is back-pressure against all traffic on the pool including synchronous clients.
- Clarify in Observability that `async_dispatch_budget` is the **queue** gates' budget; the per-pool saturation gate's budget is the PromQL expression itself and must be queried directly.

## Verification

- `render()` leaves no unreplaced placeholders in any of the three overlays.
- Rendered YAML parses, and the embedded gate JSON parses, e.g. `{"query": "clamp(1 - sum(vllm:num_requests_running{inference_pool=\"pool-a\"})/4, 0, 1)", "fallback": "1"}`.
- `helm template ... --version 0.7.4 -f <rendered redis overlay>` succeeds and emits the expected `worker-pools.json`.
- No other guide uses this `clamp(...)` pattern; no CI workflow references the multitenant guide.

## Notes

Filed as llm-d/llm-d-async#365 but the defect lives here, in the guide.

A possible follow-up in `llm-d-async` (not needed for this fix, and unusable by the guide today since it pins published chart `0.7.4`): give the `prometheus-query` gate an explicit `threshold`/`baseline` param, so the closing point can be stated directly instead of relying on `clamp`'s floor coinciding with the hardcoded `0.0` threshold.

Ref: https://github.com/llm-d/llm-d-async/issues/365


shimib (commented):
mistake comment

ahg-g:
pls rebase and fix conflicts

PR #2141, opened 2026-07-29
docs(guides): Fix prerequisite ordering and improve readability in optimized-baseline guide

haitwang-cloud:
### Problem

  The optimized-baseline guide had two usability issues:

  1. Broken variable reference: The export BRANCH=main command appeared after the git clone ... && git checkout ${BRANCH} step, so users following the guide sequentially would hit a
  failure because ${BRANCH} was undefined at clone time.
  2. Unclear env.sh explanation: The note about source ${REPO_ROOT}/guides/env.sh was vague ("so the rest of the guide makes sense"), giving users no indication of which variables it
  defines or what would break if they skipped it.

 ### Changes

  - Move export BRANCH=main before the git clone command in both README.md and guide.yaml so that ${BRANCH} is defined when the clone step executes.
  - Replace the generic env.sh note with an explicit list of key variables it provides (GAIE_VERSION, ROUTER_CHART_VERSION, ROUTER_STANDALONE_CHART, ROUTER_GATEWAY_CHART) and a clear
  warning that skipping it causes unresolved variable errors downstream.
  - Add a resource prerequisite note reminding users to verify their cluster has enough accelerators (default: 16 GPUs) before proceeding.

### DIFF
--- a/guides/optimized-baseline/README.md
+++ b/guides/optimized-baseline/README.md
@@ -105,9 +105,8 @@ source ${REPO_ROOT}/guides/env.sh
 
 > [!NOTE]
 > This file defines shared variables required by subsequent steps, including
-> `GAIE_VERSION`, `ROUTER_CHART_VERSION`, `ROUTER_STANDALONE_CHART`, and
-> `ROUTER_GATEWAY_CHART`. If you skip this step, later commands will fail with
-> unresolved variable references.
+> `GAIE_VERSION`, `ROUTER_CHART_VERSION`, and the router chart reference for
+> the selected deployment mode.
 
 - Install the Gateway API Inference Extension CRDs:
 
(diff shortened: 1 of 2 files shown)


guides/optimized-baseline/README.md
```diff
@@ -101,7 +104,10 @@ source ${REPO_ROOT}/guides/env.sh
 <!-- guide:env.source end -->
 
 > [!NOTE]
-> Some environment variables are common amongst guides, to view these, please inspect the above file sourced so the rest of the guide makes sense.
+> This file defines shared variables required by subsequent steps, including
+> `GAIE_VERSION`, `ROUTER_CHART_VERSION`, `ROUTER_STANDALONE_CHART`, and
+> `ROUTER_GATEWAY_CHART`. If you skip this step, later commands will fail with
+> unresolved variable references.
```
zdtsw: do we really need to list env variables here? e.g you wont need `ROUTER_STANDALONE_CHART` and `ROUTER_GATEWAY_CHART` at the same time, isnt it?

haitwang-cloud: @zdtsw Fair point. The intent was to explain what `env.sh` provides, not imply both router charts are used together.
  I’ll rephrase this to mention the router chart reference for the selected deployment mode.

PR #2093, opened 2026-07-21
feat(sglang): add managed lustre L3 offloading support

tyuchn:
Part of https://github.com/llm-d/llm-d/issues/1967

### Summary
This PR adds support for SGLang native filesystem KV cache offloading (`--hicache-storage-backend file`) backed by GCP Managed Lustre, along with [client-side in-memory metadata caching](https://github.com/sgl-project/sglang/pull/29716) (`SGLANG_HICACHE_FILE_BACKEND_ENABLE_METADATA_CACHE=true`).

### Key Additions
1. **Kustomize Overlays:** Added base (`native/fs/base/`) and GKE (`native/fs/gke/`) overlays configuring SGLang HiCache file backend, 200 GiB host DRAM tiering, and PVC volume mounting. Added permission-fixing initContainer (`fix-lustre-permissions`) for group-writable GKE shared storage access.
2. **Empirical Benchmark Report:** Added [`sglang-qwen3-32b-h100-lustre.md`](./guides/tiered-prefix-cache/benchmark-results/sglang-qwen3-32b-h100-lustre.md) documenting empirical performance across 16 × H100 GPUs (8 replicas, TP=2) under the canonical 16K system prompt evaluation profile across 5.0, 10.0, 20.0, and 40.0 QPS.
3. **Documentation:** Updated tiered-prefix-cache guide README with SGLang filesystem offloading commands, summary table entries, and links to the new benchmark report.

### Key Architectural Insights & Observations
- **Peak Throughput Superiority at Concurrency Scale (40 QPS):** When request concurrency scales to 40.0 QPS, the working set (4,640,000 tokens / ~760 GB) exceeds host CPU DRAM capacity. Without an L3 filesystem tier, the system experiences LRU eviction thrashing and must recompute evicted prefix blocks from scratch. Adding Managed Lustre L3 offloading increases peak throughput from 308,780 tok/s to **332,945 tok/s** (an **188.4% gain** over HBM-only and **+7.8% gain** over CPU RAM alone).
[...]

PR #2143, opened 2026-07-29
Fix: Prevent SGLang GKE Pods from Stalling on Missing PVCs

weizhoublue:
No  PersistentVolumeClaim  named  `preprov-pvc`  is defined anywhere in the repository. 
Therefore, referencing  preprov-pvc  in  `guides/tiered-prefix-cache/modelserver/gpu/sglang/native/fs/gke/patch-sglang.yaml`  creates an invalid Deployment configuration.

The GKE patch should instead retain  `llm-d-kv-cache-storage` , the PVC defined in  `guides/tiered-prefix-cache/manifests/pvc.yaml` .

For comparison, the sibling base patch,  `guides/tiered-prefix-cache/modelserver/gpu/sglang/native/fs/base/patch-sglang.yaml` , correctly configures the same  files-storage  volume to use that PVC:

```
volumes:
  - name: files-storage
    persistentVolumeClaim:
      claimName: llm-d-kv-cache-storag
```

### DIFF
--- a/guides/tiered-prefix-cache/modelserver/gpu/sglang/native/fs/gke/patch-sglang.yaml
+++ b/guides/tiered-prefix-cache/modelserver/gpu/sglang/native/fs/gke/patch-sglang.yaml
@@ -15,4 +15,4 @@ spec:
       volumes:
         - name: files-storage
           persistentVolumeClaim:
-            claimName: preprov-pvc
+            claimName: llm-d-kv-cache-storage


weizhoublue:
@tyuchn  hi, I think there is typo error in PR #2093 
pls help check whether it makes sense, thanks

tyuchn:
> @tyuchn hi, I think there is typo error in PR #2093 pls help check whether it makes sense, thanks

Good catch. You are right, `preprov-pvc` was only used during my local benchmarks. We should use the PVC name already defined in llmd

tyuchn (approved):

PR #2153, opened 2026-07-30
[ROCM] optimized-baseline: persist model cache for AMD vllm via component

PrateekKumar1709:
## Problem

The AMD optimized-baseline overlay (`guides/optimized-baseline/modelserver/amd/vllm`) mounts `HF_HOME` on an `emptyDir`. `emptyDir` is ephemeral, so every run starts with an empty cache and re-downloads the model (`Qwen/Qwen3-32B`, ~65GB) from HuggingFace before the benchmark can start. This is slow and, on bandwidth-constrained networks, fragile - a stalled download can hang or fail the run.

## Change

Add a reusable **`model-cache` component** that backs the cache with a persistent `ReadWriteMany` PVC and points `HF_HOME` at it, so the model downloads once and is reused across runs (and across the decode replicas that share the volume).

- **New component** `guides/recipes/modelserver/components/model-cache/`:
  - `model-cache-pvc.yaml` - an RWX `PersistentVolumeClaim` (`model-cache`, 200Gi).
  - `kustomization.yaml` (`kind: Component`) - declares the PVC and patches the Deployment: `HF_HOME=/model-cache`, volumeMount, and the PVC-backed volume.
- **Isolated to CI, opt-in:** the guide base is moved to `amd/vllm/base/` and left as upstream (`emptyDir` - unchanged behavior). A new `amd/vllm/amd-ci/` overlay layers the component on top. This matches the existing `pd-disaggregation/modelserver/amd/vllm/{base,amd-ci,...}` structure.

Design notes:
- The volume is named generically (`model-cache`), independent of the model source - `HF_HOME` (the framework's cache var) simply points at it.
- The PVC omits `storageClassName` so it binds the cluster's **default** StorageClass, consistent with how the reusable benchmark CI workflow already probes for a default RWX StorageClass (`llm-d-infra` `reusable-ci-nightly-benchmark.yaml`). No storage class is hardcoded.
- The component is reusable: any other guide can include it from its own `amd-ci/` overlay. Per-guide namespace isolation means each guide gets its own PVC, so removing/refactoring one guide can't affect another's cached data.

## Test plan

- [x] `kubectl kustomize guides/optimized-baseline/modelserver/amd/vllm/base/` renders the upstream Deployment with no cache (emptyDir).
- [x] `kubectl kustomize guides/optimized-baseline/modelserver/amd/vllm/amd-ci/` renders the PVC + patched Deployment (`HF_HOME=/model-cache`, volumeMount, PVC-backed volume; PVC name matches claimName).
- [ ] On a cluster with an RWX default StorageClass: first run downloads the model and populates the PVC; a subsequent run starts without re-downloading (cache hit).
- [ ] Nightly AMD optimized-baseline benchmark completes successfully.


vcave (changes requested):
This is amd-ci/ specific so should go into an amd-ci/ provider. Also check if we could have a single location to declare the PVC and have each guide import it as a component.

PrateekKumar1709:
> This is amd-ci/ specific so should go into an amd-ci/ provider. Also check if we could have a single location to declare the PVC and have each guide import it as a component.

@vcave - addressed both points in 2c15885:

- Isolated to amd-ci/: the guide base is reverted to upstream (moved to base/, still emptyDir); a new amd-ci/ overlay layers the cache on top, matching the existing pd-disaggregation/.../amd/vllm/{base,amd-ci} structure.
- Single declaration as a component: the PVC + HF_HOME patch now live once in guides/recipes/modelserver/components/model-cache/, which any guide imports via components

Verified with kubectl kustomize on both paths (base = clean emptyDir, amd-ci = PVC + mount + HF_HOME)

maugustosilva:
This seems to be working properly. We can take care of the PVC creation and model download inside the reusable nightly CI workflow (in `llm-d-infra`) later, making it fully generalized.

maugustosilva (approved):

vcave (approved):

Issue #2125, opened 2026-07-27
[CI/CD] Nightly coverage gaps: lanes not producing verdicts (July 2026)


LukeAVanDrie:
Companion to #1565, covering lanes that are neither green nor red. These are invisible on the badge matrix because they are not failing — they are not reporting.

**Not running**

- [ ] `fast-model-actuation-ibm-acc-gpu-vllm-x` — registered in the Actions API but has never produced a scheduled run, and no workflow file exists under `.github/workflows/` on `main`. Either the file was removed without deregistering the workflow, or it only ever existed on a feature branch. Scheduled workflows fire only from the default branch.
- [ ] [`tiered-prefix-cache-gke-cpu-gpu-vllm-native`](https://github.com/llm-d/llm-d/blob/main/.github/workflows/nightly-e2e-tiered-prefix-cache-gke-cpu-gpu-vllm-native.yaml) — 6 of the last 7 scheduled runs were cancelled before concluding. The one run that concluded failed at `Install llmdbenchmark` ([run 30007081455](https://github.com/llm-d/llm-d/actions/runs/30007081455)). Suspect a concurrency group or a workflow-level timeout. Effectively zero coverage.

**Removed — confirm a replacement lane covers each**

Deleted from the repo but still listed by the Actions API; last ran 49 days ago.

- [ ] `wide-ep-lws-gke` — was 0/7 at removal
- [ ] `tiered-prefix-cache-cpu-offloading-gke` — was 6/7
- [ ] `tiered-prefix-cache-cpu-offloading-lmcache-gke` — was 6/7, no obvious successor in the current lane list
- [ ] `tiered-prefix-cache-cpu-offloading-ocp` — was 4/7
- [ ] `tiered-prefix-cache-gke-tpu` — was 7/7, the cleanest lane in the fleet

LukeAVanDrie:
cc @rlakhtakia and @nicolexin

PR #2161, opened 2026-08-01
fix(ci): stop tiered-prefix-cache GKE GPU nightly lanes from cancelling each other

sudoalok:
#2125 flagged tiered-prefix-cache-gke-cpu-gpu-vllm-native for having 6 of its last 7 scheduled runs cancelled. i think the cause is just in the workflow files.

the native and the lmcache lane both declare concurrency group nightly-e2e-tiered-prefix-cache-gke-gpu with cancel-in-progress: true, and both sit on cron '0 12 * * *'. so they fire together every night and whichever starts second kills the other. its the only duplicated concurrency group in the whole nightly fleet, every other lane has its own.

the shared group looks intentional though. both lanes use the same cluster_namespace, helm_release and workspace_dir and only differ by connector, so they really cant run side by side. so instead of splitting them up i set cancel-in-progress: false on both so the second one queues, and moved lmcache to 14:00 UTC so theyre not racing in the first place. 12:00 already has 5 lanes on it and 14:00 has 1. also left a comment on both files so the shared group doesnt look like a leftover to whoever reads it next.

one thing i didnt touch. both lanes pass the same badge_name (tiered-prefix-cache-gke-cpu-llmcache), so the native lane has no badge of its own and its result ends up on the lmcache one. thats probably the other half of the not producing verdicts thing here, but renaming it makes a badge nothing points at yet so i left it alone. can add it to this PR if you want.

Part of #2125

### DIFF
--- a/.github/workflows/nightly-e2e-tiered-prefix-cache-gke-cpu-gpu-vllm-lmcache.yaml
+++ b/.github/workflows/nightly-e2e-tiered-prefix-cache-gke-cpu-gpu-vllm-lmcache.yaml
@@ -58,15 +58,17 @@ on:
 #      - main
 
   schedule:
-    - cron: '0 12 * * *'
+    - cron: '0 14 * * *'  # 14:00 UTC daily (staggered after the "native" lane, same namespace)
 
 permissions:
   contents: write
   actions: read
 
 concurrency:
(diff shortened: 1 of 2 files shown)


Gregory-Pereira (approved):
/lgtm

Issue #2123, opened 2026-07-27
[Bug :bug:]: optimized-baseline trtllm decode pods exit Error, startup probe never passes


LukeAVanDrie:
### Contact Details

lukevandrie@google.com

### What happened?

The `optimized-baseline-gke-acc-gpu-trtllm-x` nightly lane is 3/7 over the last week. It is the only nightly lane where the workload starts and then dies, rather than failing to schedule, so it is the only one that currently looks like a product or image problem rather than cluster capacity.

On [run 30272097869](https://github.com/llm-d/llm-d/actions/runs/30272097869) both decode pods reached terminal `Error` at 2m10s. The TRT-LLM server never binds `:8000`, so the startup probe against `/v1/models` fails until the pod is killed.

Two candidate causes appear in the captured container output and I could not separate them:

1. `transformers 5.5.4 is incompatible with nvidia-modelopt` — a version conflict inside the vendor image `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc18`.
2. `You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN` — the job sets both `LLMDBENCH_HF_TOKEN` and `HF_TOKEN`, yet the container is making unauthenticated Hub requests. A rate-limited or failed weight download would explain the server never starting.

The second is the more suspicious. No Python traceback survives in the captured logs, so the fatal line is still unknown.

**Expected:** decode pods become Ready and serve `/v1/models`, as the vllm and sglang variants of the same guide do on the same cluster.

**Actual:** both decode pods exit with `Error` roughly two minutes in, every time the lane gets far enough to deploy.

**Next steps**

- Confirm whether `HF_TOKEN` is propagated into the trtllm modelserver pod spec the way it is for the vllm and sglang variants.
- Capture the full decode container log on failure. The current tail cuts off before the fatal error.
- Confirm `1.3.0rc18` is the intended pin — it is a release candidate, and the transformers/modelopt conflict is inside the vendor image.

Component versions on the failing run: router charts `v0`, endpoint-picker `v0.9.0`, disagg-sidecar `v0.9.0`, benchmark `v0.7.0`, GAIE `v1.5.0`. (The Version dropdown below has no `main` or v0.9.0 option; it needs updating.)

### Version

v0.6.0

### Relevant log output

```shell
# https://github.com/llm-d/llm-d/actions/runs/30272097869
NAME                                                           READY   STATUS   RESTARTS   AGE
[...]

PR #2162, opened 2026-08-01
fix(guides): add HF_TOKEN to the optimized-baseline trtllm decode pods

sudoalok:
checked the first next step on #2123, HF_TOKEN doesnt reach the trtllm pod.

patch-trtllm.yaml only sets HF_HOME and LD_LIBRARY_PATH. every other optimized-baseline variant pulls HF_TOKEN off the llm-d-hf-token secret and trtllm is the only one that doesnt, and nothing in guides/recipes injects it either. so the token the job sets stays on the runner, which matches the unauthenticated Hub warning in the container output.

this adds the same env block sglang uses.

not sure its the fatal cause though. Qwen3-32B is public so unauth pulls still work, just rate limited. worth a manual dispatch of the lane to see if it gets past the startup probe.

Part of #2123


ahg-g:
pls resolve the conflicts

sudoalok:
ah this is a  duplicate of #2167, closing.

PR #2164, opened 2026-08-03
ci: Fix XPU wide-ep failure

yuanwu2017:
Align the Intel XPU wide-ep-lws nightly with the other Intel XPU nightlies, which call the reusable benchmark workflow directly.
https://github.com/llm-d/llm-d/actions/runs/30785534890 Passed the wide-ep CI

The preflight gate duplicated checks that standup already performs: llm-d-benchmark step 02 installs the LWS controller when a scenario enables multinode. Its accelerator count also read the gpu.intel.com/xe extended resource, which is always 0 on clusters that publish Intel GPUs through DRA, so the gate failed before the benchmark job could run.

### DIFF
--- a/.github/workflows/nightly-e2e-wide-ep-lws-intel-acc-xpu-vllm-x.yaml
+++ b/.github/workflows/nightly-e2e-wide-ep-lws-intel-acc-xpu-vllm-x.yaml
@@ -64,45 +64,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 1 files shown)


yuanwu2017:
@Gregory-Pereira @maugustosilva This PR fix the failed wide-ep CI. Please help to review.

diegocastanibm (commented):
I'll approve it once we are done with the test

diegocastanibm (approved):
LGTM

PR #2175, opened 2026-08-04
Update feature table

asm582:
The saturation-based configuration needs to be tested to support scale-to/from-zero scenarios; hence, update the table to reflect the same.


mamy-CS:
Thanks @asm582, if you've validated scale-to-zero works, and I've opened #2183 to fix the OpenShift path it relies on: the saturation overlays/ocp didn't build, and `authenticationRef` was nested inside trigger metadata so it never authed to Thanos. #2183 wires proper bearer auth (`cluster-monitoring-view` SA token), verified live. Once it lands, saturation scale-to-zero is genuinely supported. So I think we can close this in favor of #2183.

maugustosilva (approved):

PR #2177, opened 2026-08-04
docs(optimized-baseline): fix helm values file variable expansion

asm582:
## Problem

The guide instructs users to store helm flags in shell variables and then expand them in helm commands. This pattern breaks with helm due to how shell argument expansion works, especially when combined with empty variables.

Users get: `open /path/base.values.yaml: no such file or directory`

## Solution

Store only the file path in variables (without the `-f` flag) and inline the flags directly in helm commands.

## Changes

- Remove `-f` from variable definitions (lines 143-152)
- Add `-f` flags inline in helm commands (lines 184-190, 206-214)
- Update setup notes to clarify this approach

PR #2180, opened 2026-08-05
docs(optimized-baseline): quote helm values file arguments

### LINKED ISSUE #2176: docs(optimized-baseline): helm values file variable expansion broken
## Problem

The guide at `guides/optimized-baseline/README.md` (lines 143-152) instructs users to define helm values using variables:

```bash
export ROUTER_BASE_VALUES="-f ${REPO_ROOT}/guides/recipes/router/base.values.yaml"
export ROUTER_VALUES="-f ${REPO_ROOT}/guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml"
```

Then pass them to helm:
```bash
helm install ${GUIDE_NAME}   ${ROUTER_GATEWAY_CHART}   ${ROUTER_BASE_VALUES}   ${MONITORING_VALUES}   ${ROUTER_VALUES}   ...
```

This fails with: `open /Users/.../guides/recipes/router/base.values.yaml: no such file or directory`

## Root Cause

Embedding the `-f` flag inside a shell variable and then expanding it as a positional argument doesn't work reliably with helm. The flag gets mishandled during argument expansion, particularly when there are empty variables like `${MONITORING_VALUES}=""` creating spacing issues.

## Solution

Either:
1. Store only the path in the variable (not the `-f` flag), or
2. Show users to inline the flags directly:

```bash
helm install ${GUIDE_NAME}   ${ROUTER_GATEWAY_CHART}   -f ${REPO_ROOT}/guides/recipes/router/base.values.yaml   -f ${REPO_ROOT}/guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml   ...
```

## Workaround

Use inline `-f` flags instead of variables.

## Affected File

- `guides/optimized-baseline/README.md` lines 143-152 (and similar patterns in other guides)

Solaris-star:
## Summary

The optimized-baseline guide stored Helm values-file variables with the `-f` flag embedded in the variable value, then expanded those variables unquoted in `helm install`. This breaks argument parsing and is especially fragile when optional values are empty.

## Changes

- Store values variables as paths only.
- Build the Helm values arguments as a quoted Bash array.
- Add the optional monitoring values file only when `MONITORING_VALUES` is non-empty.
- Keep `guide.yaml` as the source of truth and regenerate the corresponding README blocks.
- Apply the same argument handling to standalone and gateway installs.

## Verification

- `python scripts/guide.py render guides/optimized-baseline --check`
- `python scripts/guide.py check guides/optimized-baseline`
- `pre-commit run check-yaml --files guides/optimized-baseline/guide.yaml`
- `pre-commit run end-of-file-fixer --files guides/optimized-baseline/guide.yaml guides/optimized-baseline/README.md`
- `bash -n` on both rendered Helm snippets
- `shellcheck` on both rendered Helm snippets with repository-variable checks excluded
- Quoted-array simulation verified with empty and non-empty monitoring paths
- `git diff --check`

Fixes #2176

### DIFF
--- a/guides/optimized-baseline/README.md
+++ b/guides/optimized-baseline/README.md
@@ -180,11 +180,11 @@ This deploys the llm-d Router in [Standalone Mode](../../docs/architecture/core/
 # Assuming base-directory is the root of the llm-d repo
 helm_values=(
   -f "${ROUTER_BASE_VALUES}"
-  -f "${ROUTER_VALUES}"
 )
 if [[ -n "${MONITORING_VALUES:-}" ]]; then
   helm_values+=( -f "${MONITORING_VALUES}" )
 fi
+helm_values+=( -f "${ROUTER_VALUES}" )
 helm install ${GUIDE_NAME} \
   ${ROUTER_STANDALONE_CHART} \
   "${helm_values[@]}" \
@@ -204,11 +204,11 @@ To use a Kubernetes Gateway managed proxy rather than the standalone version, fo
 ```bash
 helm_values=(
   -f "${ROUTER_BASE_VALUES}"
-  -f "${ROUTER_VALUES}"
 )
 if [[ -n "${MONITORING_VALUES:-}" ]]; then
   helm_values+=( -f "${MONITORING_VALUES}" )
 fi
+helm_values+=( -f "${ROUTER_VALUES}" )
 helm install ${GUIDE_NAME} \
   ${ROUTER_GATEWAY_CHART} \
   "${helm_values[@]}" \
--- a/guides/optimized-baseline/README.md
+++ b/guides/optimized-baseline/README.md
@@ -136,11 +136,10 @@ kubectl create secret generic llm-d-hf-token \
 
 ### 1. Deploy the llm-d Router
 
-- Prepare the paths to the `helm` values files for `llm-d` router (used in the deployment commands below):
+- Define the `helm` values files for `llm-d` router:
 
 <!-- guide:deploy.router_values start -->
 ```bash
-# Paths to values files
 export ROUTER_BASE_VALUES="${REPO_ROOT}/guides/recipes/router/base.values.yaml"
 
 # only when MODEL_SERVER=vllm or sglang:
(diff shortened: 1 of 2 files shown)


Solaris-star:
Rewrote the PR commit with the registered SSH signing key and added the DCO sign-off. New head: `dfe0a1c004f05d443ff993952bdd95c61e0622a2`.

Remote GitHub verification: `verified: true`, `reason: valid`. The repository `signed-commits` and `DCO` checks are now passing.

ahg-g:
https://github.com/llm-d/llm-d/pull/2177 addresses this issue, right?

iacker (commented):
The array + explicit quoting is the right fix for paths with spaces, and dropping the inline `-f` from the exported vars is cleaner.

One behavior change worth a second look: the `-f` ordering shifts. Before it was BASE, MONITORING, ROUTER_VALUES; now it is BASE, ROUTER_VALUES, then MONITORING appended last. Helm merges values files left-to-right with the rightmost winning, so MONITORING now overrides ROUTER_VALUES instead of being overridden by it. In practice monitoring.values.yaml likely only touches monitoring keys so the overlap is probably empty, but since the stated goal here is just quoting, it would be safer to keep ROUTER_VALUES last to preserve the original precedence.

Solaris-star:
Fixed the review point in 5dba2ef: ROUTER_VALUES is appended after the optional MONITORING_VALUES, preserving the original Helm rightmost-wins precedence while keeping the quoted array argument handling.

Solaris-star:
Addressed the values-precedence review in the signed/DCO commit `22a4715`: `ROUTER_VALUES` is again the rightmost values file, after optional monitoring values. GitHub verifies the commit signature and the DCO sign-off is included.

Solaris-star:
Rebased onto current `main` and rewrote both PR commits with SSH signatures while preserving their DCO sign-offs. The optimized-baseline guide was re-rendered during conflict resolution; `guide.py render --check`, `guide.py check`, and `git diff --check` pass. New head: `8daaff8`.

Solaris-star:
Thanks for flagging #2177. It fixes the two unconditional router path variables in the rendered `README.md`, so it addresses the basic reproduction. I kept #2180 open because it covers the remaining cases that #2177 does not:

- `guide.yaml` is the source used to regenerate this guide, and #2177 changes only `README.md`;
- `MONITORING_VALUES` still embeds `-f` and is expanded conditionally;
- the quoted Bash array preserves the original `BASE, MONITORING, ROUTER_VALUES` precedence while avoiding an empty optional argument.

So #2180 is intended as a follow-up that makes the fix durable and covers the optional monitoring path, not a replacement for the already-merged correction in #2177.

PR #2182, opened 2026-08-05
feat(autoscaling): add OpenShift support to the KEDA+EPP queue guide

mamy-CS:
**What type of PR is this?**

/kind feature

Enables the queue-based KEDA + EPP autoscaling guide (`optimized-baseline-autoscaling/keda-epp-queue/`) to work on OpenShift. Follows #2181, which relocated the guide.

Previously the guide shipped only a bundled kube-prometheus-stack configuration with a CA-only `TriggerAuthentication`. On OpenShift, cluster metrics live behind Thanos Querier, whose `:9091` endpoint rejects unauthenticated queries with a 401 — and KEDA suppresses trigger errors and silently serves `fallback` replicas. The result is an autoscaler that reports healthy while never actually scaling.

This PR:

- **Restructures `keda-epp-queue/` into a `base/k8s/ocp/` sibling layout** (matching `optimized-baseline-autoscaling/keda/`). A nested `overlays/ocp` that references its parent (`../..`) hits a kustomize containment cycle and fails to build; the sibling layout is the pattern that builds cleanly. `base/` holds the ScaledObject (unchanged), `k8s/` is a passthrough for generic clusters, `ocp/` is the OpenShift overlay.
- **Adds the `ocp/` overlay:**
  - Points both triggers at `thanos-querier.openshift-monitoring.svc.cluster.local:9091` and enables `authModes: bearer`.
  - Provisions a dedicated `keda-epp-metrics-reader` ServiceAccount granted the `cluster-monitoring-view` ClusterRole, plus a `service-account-token` Secret. On OpenShift the service-ca operator injects `service-ca.crt` (the CA that signs Thanos's serving cert) into that Secret automatically, so no CA copying is required.
  - Repoints the `TriggerAuthentication` at the SA token Secret (`bearerToken` + `service-ca.crt`).
- **Updates `README.hpa-epp.md`**: platform-aware apply/delete paths (`/k8s` for generic clusters, `/ocp` for OpenShift) and a concrete OpenShift section.

The `k8s` overlay is an unchanged passthrough of `base` (bundled Prometheus, CA-only auth), so non-OpenShift behavior is preserved.

**Verification:**
- `kubectl kustomize` builds all three targets (`base`, `k8s`, `ocp`); the rendered `ocp` output has both triggers on Thanos with `authModes: bearer`, the TriggerAuthentication using the SA token Secret, and the SA/ClusterRoleBinding/token Secret present with the namespace propagated to the CRB subject; the `k8s` output is unchanged from the previous base.
- Live-validated on an OpenShift cluster: applied the `ocp` overlay against a dummy scale target with TLS verification on. The ScaledObject reached `Ready=True` and held `Fallback=False`, and the KEDA-generated HPA resolved a real external-metric value from Thanos (no 401 / no fallback). Confirmed the SA token Secret's `service-ca.crt` bundle includes `openshift-service-serving-signer`, the CA that signs Thanos's serving certificate.

**Which issue(s) this PR fixes**:

N/A — enables the KEDA+EPP queue autoscaling guide on OpenShift (follows #2181).

**Release note**:

```release-note
The KEDA + EPP queue-based autoscaling guide now supports OpenShift via a new `ocp` overlay that authenticates to Thanos Querier with a bearer token (a dedicated `cluster-monitoring-view` ServiceAccount; no CA copying needed). The guide is now applied per platform: `kubectl apply -k .../keda-epp-queue/k8s` for generic Kubernetes, or `.../keda-epp-queue/ocp` for OpenShift.
```


asm582:
/lgtm
/approve

Per the author, the PR has been validated using dummy vLLM servers. This makes sense since we are tracking pool-level metrics.

lionelvillard (approved):

PR #2184, opened 2026-08-05
Mooncake tiered offloading router values

Gregory-Pereira:
fixes:
- Deployment / operational: the mooncake client doesnt need to be in the mooncake namespace it can be wherever and should be treated as an application not infrastructure level component.
- Show comments on how to target a subset of nodes and not every node for offloading
- add mooncake tiered offloading values. As described currently the vLLM mooncake connector hardcodes every `medium="cpu"`, so basically llm-d controlls load signals and mooncake tent figures out cache hit wherever it may live
cc @vMaroon @ykwd @zdtsw @ahg-g @liu-cong 

I tested this succesfully. Logs attached, but we eventually see the prefix-cache-affinity-filter consider it a sticky endpoint, ie bubling up routing for cache hits to the router:
```
{"level":"debug","ts":1785963823.8774176,"caller":"prefixcacheaffinity/plugin.go:208","msg":"PrefixCacheAffinityFilter: narrowed to sticky","x-request-id":"ff3f3fa5-c641-46cf-a100-2b8f86228b3f","objectiveKey":"","incomingModelName":"Qwen/Qwen3-32B","targetModelName":"Qwen/Qwen3-32B","priority":0,"affinityThreshold":0.8,"sticky":1,"total":4}
```

[epp-log-mooncake-tiered-offload.log](https://github.com/user-attachments/files/30762742/epp-log-mooncake-tiered-offload.log)

### DIFF
--- a/guides/tiered-prefix-cache/router/mooncake-offloading.values.yaml
+++ b/guides/tiered-prefix-cache/router/mooncake-offloading.values.yaml
@@ -1,0 +1,75 @@
+## Mooncake Store tiered-offload router overrides (non-PD mode).
+## Layer on top of:
+##   guides/recipes/router/base.values.yaml
+##
+## Mooncake Store manages the shared CPU DRAM and SSD/NVMe tiers.
+## The router treats every prefix match the Mooncake transfer engine
+## reports as a single signal and trusts Mooncake to decide which tier
+## to pull from. The router owns the load-based routing signals (queue
+## depth, KV-cache utilization) and mixes them with prefix affinity.
+##
+## A single approximate prefix-cache producer and scorer are used
+## because MooncakeStoreConnector hardcodes medium="cpu" on all
+## BlockStored events regardless of actual backing tier (memory vs
+## disk). The EPP cannot distinguish offload tiers until vLLM exposes
+## the real storage medium:
+##   https://github.com/vllm-project/vllm/blob/8f158d0e/vllm/distributed/kv_transfer/kv_connector/v1/mooncake/store/worker.py#L896
+## When per-tier mediums are available, switch to
+## precise-prefix-cache-producer with per-tier kvCacheBackendConfigs
+## weights to score memory and disk hits differently.
+
+router:
+  extraServicePorts:
+    - name: http
+      port: 80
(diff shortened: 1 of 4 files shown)


guides/tiered-prefix-cache/router/mooncake-offloading.values.yaml
```diff
@@ -0,0 +1,68 @@
...
+          - type: token-load-scorer
+
+        schedulingProfiles:
+          - name: default
+            plugins:
+              - pluginRef: prefix-cache-affinity-filter
+              - pluginRef: token-load-scorer
+
+  modelServers:
+    type: vllm
```
liu-cong, line 64: this and the targetports are defaults, we can remove

guides/tiered-prefix-cache/router/mooncake-offloading.values.yaml
```diff
@@ -0,0 +1,68 @@
...
+            name: mooncake-prefix-cache-producer
+
+          # In-flight load feeds the affinity filter's TTFT estimation.
+          - type: inflight-load-producer
+
+          # Narrow to endpoints with high prefix cache scores, but
+          # break stickiness when sticky endpoints are too loaded.
+          # peakPrefillThroughput default (15928) is calibrated for
+          # Qwen3-32B on H100 80GB TP=2; adjust for your hardware.
+          - type: prefix-cache-affinity-filter
```
liu-cong, line 50: with mooncake I think we are overcounting the cache locality, so we may consider a lower `affinityThreshold`. But without a benchmark it's hard to tell.

liu-cong (commented):
We don't have any benchmark results for mooncake yet, right? If we have a baseline we should run a benchmark to compare.

If not, I am OK merging this but we should follow up with benchmark results

guides/tiered-prefix-cache/router/mooncake-offloading.values.yaml
Gregory-Pereira: This in particular is an excellent point - I am OK with waiting till I can run a benchmark to get this updated and merged

PR #2186, opened 2026-08-06
rename default wide-ep to deepseek-r1-0528

Gregory-Pereira:
Now that we have multiple wide-ep deployments, we should rename the orrigional one to be consistent structure with the rest
cc @robertgshaw2-redhat


ahg-g:
note that there has been some updates to that directory, adding disaggregatedSet support among other things: https://github.com/llm-d/llm-d/tree/main/guides/wide-ep-lws/modelserver/gpu/vllm

maugustosilva:
@Gregory-Pereira Please rebase

PR #2188, opened 2026-08-06
feat(precise): Serve render from the model server pods

albertoperdomo2:
## Summary

In the `precise-prefix-cache-routing` guide, the render (tokenizer) Service no longer has pods of its own by default. Instead of three GPU-less `vllm launch render` replicas, it selects the vLLM model server pods the guide already deploys, since `vllm serve` exposes `/v1/completions/render` and `/v1/chat/completions/render` out of the box.

The old renderer pool moves to `render/standalone/`, still required for SGLang (it doesn't implement vLLM's render endpoints). Both overlays publish the same Service name, so the router's `vllm.url` is unchanged. Apply exactly one.

### Why

Every request is tokenized before it's routed, so the render call sits inside TTFT. A separately-scheduled pool saturates before the model servers do as QPS climbs; fronting the model servers makes render capacity scale with the serving fleet.

At higher RPS, latency is more contained with this approach than with the CPU renderers. See below the benchmarking data for `Qwen3-32B` and [this workload](https://github.com/albertoperdomo2/benchflow/blob/main/profiles/benchmark/x-70r-6kSys.yaml). 

image

image

## Notes

- `modelName` **mismatches now fail loudly** instead of silently scoring against the wrong tokenizer.
- Added the render paths to `--disable-access-log-for-endpoints`, otherwise each model server logs a line per routed request.
- Unaffected by standalone vs. gateway mode, verified by templating both charts: identical `token-producer` params and InferencePool selectors.

### DIFF
--- a/guides/precise-prefix-cache-routing/README.md
+++ b/guides/precise-prefix-cache-routing/README.md
@@ -30,7 +30,7 @@ The routing decision combines precise cache knowledge with token-based load bala
 | Backend              | Directory                  | Default model                           | Notes                                                    |
 | -------------------- | -------------------------- | --------------------------------------- | -------------------------------------------------------- |
 | NVIDIA GPU           | `modelserver/gpu/vllm/`    | Qwen/Qwen3-32B                          | Default configuration                                    |
-| NVIDIA GPU (SGLang)  | `modelserver/gpu/sglang/`  | Qwen/Qwen3-32B                          | SGLang; `--page-size=64` matches the producer's `blockSizeTokens`      |
+| NVIDIA GPU (SGLang)  | `modelserver/gpu/sglang/`  | Qwen/Qwen3-32B                          | SGLang; `--page-size=64` matches the producer's `blockSizeTokens`; requires `render/standalone/` |
 | AMD GPU              | `modelserver/amd/vllm/`    | Qwen/Qwen3-32B                          | AMD GPU                                                  |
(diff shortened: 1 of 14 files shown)


guides/precise-prefix-cache-routing/README.md
```diff
@@ -157,7 +148,43 @@ export INFRA_PROVIDER=base # base | gke
 kubectl apply -n ${NAMESPACE} -k ${REPO_ROOT}/guides/${GUIDE_NAME}/modelserver/gpu/${MODEL_SERVER}/${INFRA_PROVIDER}/
 ```
 
-### 4. (Optional) Enable Monitoring
+### 4. Deploy the Render (Tokenizer) Service
+
+The EPP `token-producer` plugin tokenizes prompts by calling vLLM's `/v1/completions/render` endpoint. This guide serves that endpoint from a Service rather than a per-EPP-pod sidecar, so a single render pool is shared across EPP replicas and render capacity is decoupled from the EPP replica count.
```
vMaroon:
```suggestion
The EPP `token-producer` plugin tokenizes prompts by calling vLLM's `/v1/*/render` endpoints. This guide serves that endpoint from a Service rather than a per-EPP-pod sidecar, so a single render pool is shared across EPP replicas and render capacity is decoupled from the EPP replica count.
```

vMaroon (approved):

PR #2189, opened 2026-08-06
docs(e-disaggregation): rebalance E/P/D worker counts and TP, add TP=2 to encode workers

roytman:
## Summary
- Change E/P/D Prefill and Decode workers from 3 replicas to 4 replicas(`patch-prefill.yaml`, `patch-decode.yaml`)
- Add tensor-parallel-size=2 to the Encode workers (shared by E/PD and E/P/D) in the `README` file 
-  Update `README` worker-count tables for both topologies to match (also fixes a pre-existing mismatch where the `README` said "2 TP=4 Prefill/Decode Workers" while the manifests deployed 3 workers with TP=2)

### DIFF
--- a/guides/multimodal-serving/e-disaggregation/README.md
+++ b/guides/multimodal-serving/e-disaggregation/README.md
@@ -21,16 +21,16 @@ llm-d supports two encode-disaggregated topologies:
 
 In E/PD, dedicated encode workers handle multimodal processing while a single worker type handles both prefill and decode. Multiple encode workers enable parallel processing of multimodal entries within a single request:
 
-* 2 Encode Workers (multimodal encoding, parallelized across entries)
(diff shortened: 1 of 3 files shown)


guides/multimodal-serving/e-disaggregation/README.md
```diff
@@ -81,7 +81,6 @@ export GUIDE_PATH="multimodal-serving/e-disaggregation"
 export TOPOLOGY="e-p-d"
 export NAMESPACE="llm-d-e-p-d-disaggregation"
 export MODEL_NAME="Qwen/Qwen3-VL-32B-Instruct"
-export REPO_ROOT=$(realpath $(git rev-parse --show-toplevel))
```
ahg-g, line 84: why is this removed?

roytman: It was a duplication of line 77

PR #2191, opened 2026-08-06
feat: new WLP for KEDA+FMA

aavarghese:
New well-lit path for KEDA + FMA saturation autoscaling

Adds a KEDA-autoscaled variant of the fast-model-actuation guide: the fma-requester Deployment is scaled by a KEDA ScaledObject on the EPP pool-saturation metric (flowControl feature gate enabled), demonstrating warm-start and sleep→wake actuation under autoscaling.

README regenerated from guide.yaml; guide.py check and render --check pass.

**Summary**

Adds a new guide, `guides/fast-model-actuation-keda/`, that combines Fast Model Actuation (FMA) (../fast-model-actuation/README.md) with saturation-based autoscaling via KEDA. A KEDA ScaledObject scales the GPU-reserving FMA fma-requester Deployment on Endpoint-Picker (EPP) flow-control metrics; each new requester pod drives the FMA controllers to bring a vLLM instance online via a hot or warm start, and on scale-down the vacated instances sleep. It ships with its own nightly so the path is exercised, not just documented.

**Why Qwen3-32B**

The headline feature is saturation autoscaling, and a small model never triggers it — a fast model drains the queue instantly, so pool_saturation stays far below the 0.7 threshold and only the secondary running-requests trigger ever fires. This guide serves Qwen/Qwen3-32B, which saturates a single H100-80GB under load, so the saturation trigger does real work

**What's in the PR**

- `modelserver/` — reuses the base fast-model-actuation guide's model-server manifests as a kustomize base and patches only the KEDA-specific deltas, so the shared FMA plumbing is maintained once:
  - `InferenceServerConfig` → serve Qwen3-32B in vLLM dev mode (VLLM_SERVER_DEV_MODE=1, needed for the /sleep + /wake_up admin endpoints FMA hot-start uses), with the guide's labels.
  - `LauncherConfig` → removes the base guide's 16Gi container memory limit (correctness-critical): FMA sleeps a vacated instance with vLLM sleep level 1, offloading ~65 GB of 32B weights GPU→host RAM; the 0.6B-era 16Gi cap OOM-kills the container during offload, so scale-down tore instances down instead of sleeping them.
  - `requester` Deployment → starts at replicas: 1 (the KEDA floor).
- `keda/` — the ScaledObject (saturation trigger @ 0.7, running-requests trigger, minReplicaCount: 1 / maxReplicaCount: 2 capped at launcher-pool capacity, fast scale-up / gentle 300 s scale-down) + Prometheus TriggerAuthentication, with an overlays/ocp layer for OpenShift user-workload monitoring / Thanos.
- `router/…values.yaml` — enables the EPP flowControl feature gate so llm_d_epp_flow_control_pool_saturation is emitted (without it the saturation trigger reads 0), and binds the InferencePool to bound launchers by llm-d.ai/model: qwen3-32b.
- Nightly (.github/workflows/nightly-e2e-fast-model-actuation-keda-ibm-acc-gpu-vllm-x.yaml) + consolidate-status workflow — runs the same proven harness/workload as the FMA benchmark CI (inference-perf + shared_prefix_synthetic_heavy.yaml) with wait_timeout: 5400 / standup_timeout: 3600 sized for 32B warm load.
- Docs — README.md (rendered), foundations doc, guides/README.md entry, and a release nightly-matrix row.

**Companion PR**

Requires llm-d/llm-d-benchmark#1871 (https://github.com/llm-d/llm-d-benchmark/pull/1871), which switches the guide's benchmark scenario to Qwen3-32B + inference-perf

**Why this is a top-level sibling directory, not a child of `guides/fast-model-actuation/`**

Nesting the guide under `fast-model-actuation/` would force a `/` into **guideName**, and no single value satisfies all four roles: a slashed name is valid only as a path and breaks the Helm release name, the label value, and the Service DNS name; a flat name breaks the README/path lookup for a genuinely nested directory. This was verified empirically against the benchmark's kustomize standup — the Helm release name is the first hard failure. So the guide lives at the top level with a flat, slash-free name (`fast-model-actuation-keda`) and reuses the base guide's manifests as a kustomize base instead — which gives the "maintain shared FMA plumbing once" benefit that nesting would have been reaching for, without the identity-string breakage.


guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,476 @@
...
+# Created imperatively (not from a static manifest) so ${NAMESPACE} drives both the
+# binding's namespace and the subject ServiceAccount's namespace.
+kubectl create rolebinding fma-launcher-pod-state-writer \
+  --role=fma-launcher-pod-state-writer \
+  --serviceaccount=${NAMESPACE}:fma-launcher \
+  -n ${NAMESPACE} \
+  --dry-run=client -o yaml | kubectl apply -f -
+
+# ClusterRoleBinding (cluster-scoped): grant the EPP's ServiceAccount
+# system:auth-delegator.ß
```
diegocastanibm: a ß? iS this comming from `guide.yaml`?

guides/fast-model-actuation-keda/guide.yaml
```diff
@@ -0,0 +1,256 @@
...
+        # Created imperatively (not from a static manifest) so ${NAMESPACE} drives both the
+        # binding's namespace and the subject ServiceAccount's namespace.
+        kubectl create rolebinding fma-launcher-pod-state-writer \
+          --role=fma-launcher-pod-state-writer \
+          --serviceaccount=${NAMESPACE}:fma-launcher \
+          -n ${NAMESPACE} \
+          --dry-run=client -o yaml | kubectl apply -f -
+
+        # ClusterRoleBinding (cluster-scoped): grant the EPP's ServiceAccount
+        # system:auth-delegator.ß
```
diegocastanibm: yes it is!

diegocastanibm (commented):
sorry, I will continue with the review later on today

guides/fast-model-actuation-keda/README.md
aavarghese: lol WTH - good find!

guides/fast-model-actuation-keda/keda/base/scaledobject.yaml
```diff
@@ -0,0 +1,84 @@
...
+    kind: Deployment
+    name: fma-requester
+  minReplicaCount: 1
+  # Cap scale-out at the launcher pool's capacity. Each fma-requester replica
+  # reserves one GPU and drives exactly one vLLM instance, so maxReplicaCount
+  # must not exceed LauncherConfig.spec.maxInstances (2) -- extra requesters
+  # would have no launcher slot to bind and sit idle holding a GPU. It must
+  # also not exceed the number of GPUs available to this namespace. Raise all
+  # three together (maxInstances, GPU count, this value) to scale further.
+  maxReplicaCount: 2
```
diegocastanibm: you have a `4` in the readme file

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,476 @@
...
+
+> [!NOTE]
+> Hot start (vLLM `/wake_up`) only occurs if the Kubernetes scheduler assigns the new requesting pod to the same node (and GPU) as the sleeping vLLM instance. In a single-GPU-per-node pool the scheduler may land the pod on a different node, in which case FMA falls back to a warm start (new vLLM on an existing launcher) rather than a wake.
+
+## Configuration
+
+| Parameter                | Value                                                        |
+| ------------------------ | ------------------------------------------------------------ |
+| Model                    | [Qwen/Qwen3-0.6B](https://huggingface.co/Qwen/Qwen3-0.6B)    |
+| Requester replicas       | 1 (KEDA floor) … 4 (KEDA ceiling)                            |
```
diegocastanibm:
```suggestion
| Requester replicas       | 1 (KEDA floor) … 2 (KEDA ceiling)                            |
```

diegocastanibm: per my comment in scaledobject

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,476 @@
...
+  --namespace      "${NAMESPACE}" \
+  --harness        "${HARNESS}" \
+  --workload       "${WORKLOAD}" \
+  --analyze
+```
+<!-- guide:benchmark.execute end -->
+
+## Cleanup
+
+To remove all deployed components. **Order matters:** delete the KEDA `ScaledObject` first (so the HPA stops scaling the requester), then the model-server CRs and the requester Deployment, then wait for pods to drain while the dual-pods controller is still running to strip their finalizers, and only then remove the controllers. Removing the controller before the pods drain leaves finalizer-bound pods stuck in `Terminating` forever.
```
diegocastanibm: Maybe add the `Order matters` paragraph as a `warning` or `important` block

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,476 @@
+# Fast Model Actuation + KEDA Autoscaling
+
+[![E2E (OCP GPU)](https://github.com/llm-d/llm-d/actions/workflows/consolidate-status-fast-model-actuation-keda-ibm-acc-gpu-vllm-x.yaml/badge.svg)](https://github.com/llm-d/llm-d/actions/workflows/consolidate-status-fast-model-actuation-keda-ibm-acc-gpu-vllm-x.yaml)
+
+## Overview
+
+This guide combines [Fast Model Actuation (FMA)](../fast-model-actuation/README.md) with **saturation-based autoscaling via KEDA**. A KEDA `ScaledObject` scales the FMA server-requesting Deployment (`fma-requester`) on Endpoint Picker (EPP) flow-control metrics; each new requesting pod drives the FMA controllers to bring a vLLM instance online — either a **hot start** (waking a sleeping vLLM in seconds) or a **warm start** (creating a new vLLM on an existing launcher that has already loaded the Python modules).
```
diegocastanibm: add link also to the `saturation-based autoscaling` README file

diegocastanibm (commented):
another round

diegocastanibm (dismissed):

dumb0002:
@aavarghese, I don't understand the explanation of `Why Qwen3-32B?` in the description of this PR:  AFAIK, saturation is a ratio of offered load to capacity, not a function of model size.  A small model still saturates whenever arrival rate exceeds its service rate.

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,480 @@
+# Fast Model Actuation + KEDA Autoscaling
+
+[![E2E (OCP GPU)](https://github.com/llm-d/llm-d/actions/workflows/consolidate-status-fast-model-actuation-keda-ibm-acc-gpu-vllm-x.yaml/badge.svg)](https://github.com/llm-d/llm-d/actions/workflows/consolidate-status-fast-model-actuation-keda-ibm-acc-gpu-vllm-x.yaml)
+
+## Overview
+
+This guide combines [Fast Model Actuation (FMA)](../fast-model-actuation/README.md) with **[saturation-based autoscaling via KEDA](../workload-autoscaling/keda-epp-saturation/README.md)**. A KEDA `ScaledObject` scales the FMA server-requesting Deployment (`fma-requester`) on Endpoint Picker (EPP) flow-control metrics; each new requesting pod drives the FMA controllers to bring a vLLM instance online via a **hot or warm start**. (See the [FMA guide](../fast-model-actuation/README.md#overview) for what hot, warm, and cold start mean.)
```
dumb0002: "cold start" is to mentioned before, so do we really need it?

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+### How the autoscaling works
+
+- The **EPP** (with the `flowControl` feature gate enabled) emits `llm_d_epp_flow_control_pool_saturation` — a 0.0–1.0+ measure of how saturated the inference pool is — and `llm_d_epp_request_running`.
+- The **KEDA ScaledObject** queries these from Prometheus and scales `fma-requester` out when saturation crosses the threshold and in when it subsides.
+- Each **scale-up** replica reserves a GPU, which the FMA controller turns into a hot or warm start; each **scale-down** deletes a requesting pod, and the controller puts its vLLM to sleep. (The hot-vs-warm decision and its scheduling dependency are FMA behavior — see the [FMA guide](../fast-model-actuation/README.md#overview).)
+
+This is the recommended saturation signal for llm-d autoscaling (`llm_d_epp_flow_control_pool_saturation` + `llm_d_epp_request_running`); it requires the EPP flow-control feature gate, which this guide's router values enable.
```
dumb0002: I suggest to remove these paragraph. The paragraph below is already stating that [keda-epp-saturation](https://github.com/aavarghese/llm-d/blob/kedafma_guide/guides/workload-autoscaling/keda-epp-saturation/README.md) is being used. 

``
This is the recommended saturation signal for llm-d autoscaling (`llm_d_epp_flow_control_pool_saturation` + `llm_d_epp_request_running`); it requires the EPP flow-control feature gate, which this guide's router values enable.
``

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+The KEDA objects here — the [`ScaledObject`](keda/base/scaledobject.yaml) and its [`TriggerAuthentication`](keda/base/triggerauthentication.yaml) — are adapted from the [keda-epp-saturation](../workload-autoscaling/keda-epp-saturation/README.md) guide, reusing the same saturation signal and Prometheus trigger shape. The FMA-specific difference is that the `ScaledObject` targets the GPU-reserving `fma-requester` Deployment (not a vLLM server Deployment), caps `maxReplicaCount` at the launcher pool's capacity, and tunes the HPA `behavior` for prompt hot-wakes and gentle scale-down. The [`ScaledObject` file header](keda/base/scaledobject.yaml) spells out the full delta.
+
```
dumb0002: This sentence is just increasing the verbosity of this guide. Hence, it is not necessary: `The [ScaledObject file header](https://github.com/aavarghese/llm-d/blob/kedafma_guide/guides/fast-model-actuation-keda/keda/base/scaledobject.yaml) spells out the full delta.`

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,480 @@
...
+| Autoscaler               | KEDA `ScaledObject` → HPA on `fma-requester`                 |
+| Router                   | llm-d-router-standalone (EPP `flowControl` enabled)          |
+
+## Prerequisites
+
+This guide assumes you have a Kubernetes cluster with GPU nodes, the [llm-d router](../../guides/recipes/router/README.md) infrastructure, and a **Prometheus/monitoring stack that scrapes the EPP** (KEDA reads the saturation metric from it). If you are starting from an existing llm-d deployment, the Gateway API Inference Extension CRDs may already be installed and you can skip that step.
+
+- Have the [proper client tools installed on your local system](../../helpers/client-setup/README.md) to use this guide.
+
+- **KEDA** (or, on OpenShift, the Custom Metrics Autoscaler Operator) must be installed cluster-wide. See the [KEDA deployment docs](https://keda.sh/docs/latest/deploy/).
```
dumb0002: Can use similar instructions as here done here: https://github.com/llm-d/llm-d/tree/main/guides/workload-autoscaling/keda-epp-queue#prerequisites?

guides/fast-model-actuation-keda/README.md
aavarghese: nope -  dropped "cold". Thanks!

guides/fast-model-actuation-keda/README.md
aavarghese: sure - removed

guides/fast-model-actuation-keda/README.md
aavarghese: Removed.

aavarghese:
> @aavarghese, I don't understand the explanation of `Why Qwen3-32B?` in the description of this PR: AFAIK, saturation is a ratio of offered load to capacity, not a function of model size. A small model still saturates whenever arrival rate exceeds its service rate.

I see. That makes sense. I wasn't able to get `pool_saturation` to trigger with the smaller model (as expected?). Let me see how to reword why larger model more accurately.

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,480 @@
...
+> [!IMPORTANT]
+> On OpenShift the RBAC step (step 2) creates a `ClusterRoleBinding` to the
+> `cluster-monitoring-view` ClusterRole so KEDA can read the saturation metric
+> from Thanos. Creating that binding requires **cluster-admin at apply time**.
+> This is a one-time bootstrap concern, not a per-namespace one: enabling
+> user-workload monitoring (see [Prerequisites](#prerequisites)) and granting
+> `cluster-monitoring-view` are both cluster-admin actions, while everything else
+> in this guide is namespace-scoped. The rest of the RBAC/KEDA wiring
+> (`system:auth-delegator`, the EPP metrics-reader binding, the ScaledObject) is
+> created alongside it in the same step.
```
dumb0002: Do we need to describe this step here? Can we just use similar instructions in the other guides to install and configure KEDA?

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+### 1. Apply FMA CRDs
+
+<!-- guide:deploy.fma_crds start -->
+```bash
+# Consolidated CRD bundle (all three FMA CRDs in one file).
```
dumb0002: IMO, this comment is unnecessary.

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+> [!NOTE]
+> **Routing topology.** Only the **bound launcher** pods serve inference (vLLM on `:8000`); the `fma-requester` pods reserve the GPU and hold the pod identity but never listen on `:8000`. FMA stamps `llm-d.ai/guide` on *both* halves, so the router's `InferencePool` (which selects on that label with `targetPorts: 8000`) also selects `llm-d.ai/model` — a label the `InferenceServerConfig` applies to a launcher only once it is bound — so only bound launchers become routable EPP endpoints. Without it, requester pods would join the pool and the EPP would intermittently return `503 upstream connect error … Connection refused` (worst right after a KEDA scale-up, when a fresh requester joins the pool). Unbound launchers are already excluded for the same reason. Keep the `llm-d.ai/model` value in `router/fast-model-actuation-keda.values.yaml` in sync with the `InferenceServerConfig` patch in `modelserver/kustomization.yaml`. See [llm-d/llm-d#2212](https://github.com/llm-d/llm-d/pull/2212).
```
dumb0002: Do we really need this note? Also, it's unnecessary to have a link to PR in any llm-d guide.

dumb0002:
> > @aavarghese, I don't understand the explanation of `Why Qwen3-32B?` in the description of this PR: AFAIK, saturation is a ratio of offered load to capacity, not a function of model size. A small model still saturates whenever arrival rate exceeds its service rate.
> 
> I see. That makes sense. I wasn't able to get `pool_saturation` to trigger with the smaller model (as expected?). Let me see how to reword why larger model more accurately.

@aavarghese, maybe we did not use the heavy workload with the small model? I see saturation metrics for small models in the WVA WLP: https://github.com/dumb0002/llm-d/blob/update-benchmark-results/guides/workload-autoscaling/wva/README.md#benchmark-results

dumb0002:
@aavarghese, could we simplify this guide as following:
     1) Add in the prerequisites deployment of FMA, llm-d Router and Model server - basically just follow the FMA only guide. 
     2) Configure or patch the KEDA and FMA objects needed for this guide

For example:

## Prerequisites
Before proceeding, ensure you have:

- Monitoring stack with Prometheus over HTTPS — See [autoscaling prerequisites](https://github.com/llm-d/llm-d/blob/main/guides/workload-autoscaling/README.md#prerequisites) and [Prometheus Setup Guide](https://github.com/llm-d/llm-d/blob/main/docs/operations/observability/setup.md). This includes KEDA installation.

- EPP flow control enabled — The `llm_d_epp_flow_control_pool_saturation` metric requires the EPP flow control feature gate to be enabled in your Endpoint Picker configuration. This guide includes an epp-endpoint-picker-config.yaml that enables flow control and registers the optimized-baseline plugins. See [EPP Flow Control](https://github.com/llm-d/llm-d/blob/main/docs/architecture/core/router/epp/flow-control.md) for details on flow control behavior.

- FMA Deployment — Complete the [FMA guide](https://github.com/llm-d/llm-d/blob/main/guides/fast-model-actuation/README.md).

See a similar example for pre-requisites: https://github.com/llm-d/llm-d/tree/main/guides/workload-autoscaling/keda-epp-saturation#prerequisites

Basically, this guide would re-use the objects in the KEDA and FMA WLP and just focusing or patching the objects need for this current guide.

guides/fast-model-actuation-keda/keda/base/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+# Base KEDA bundle: the ScaledObject (saturation autoscaling of fma-requester)
+# and its Prometheus TriggerAuthentication. Applied with
+# `kubectl apply -n ${NAMESPACE} -k ...` on generic k8s (bundled Prometheus).
+# OpenShift users apply overlays/ocp instead, which patches the serverAddress.
+resources:
+  - scaledobject.yaml
+  - triggerauthentication.yaml
```
dumb0002: Could we just re-use the same objects as in the KEDA saturation WLP and just patch them here for our scenario? It would make it easier for us to maintain this WLP in sync with updates in the KEDA WLP.

aavarghese: Done

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,441 @@
...
+
+- Have the [proper client tools installed on your local system](../../helpers/client-setup/README.md) to use this guide.
+
+- **Monitoring stack with Prometheus over HTTPS** — See [autoscaling prerequisites](../workload-autoscaling/README.md#prerequisites) and [Prometheus Setup Guide](../../docs/operations/observability/setup.md). This includes [KEDA installation](../workload-autoscaling/README.md#kubernetes-metrics-adapter).
+
+- **EPP flow control enabled** — The `llm_d_epp_flow_control_pool_saturation` metric KEDA scales on requires the EPP flow control feature gate. This guide's router values enable the `flowControl` gate for you (applied in [step 4](#4-deploy-the-llm-d-router-epp-flow-control-enabled)), so no extra action is needed. See [EPP Flow Control](../../docs/architecture/core/router/epp/flow-control.md) for details on flow control behavior.
+
+- **Fast Model Actuation** — This guide reuses the base [Fast Model Actuation guide](../fast-model-actuation/README.md)'s manifests (`LauncherConfig`, `LauncherPopulationPolicy`, the requester Deployment) as a kustomize base and stands up its own FMA controllers. You do not need to deploy it separately.
```
dumb0002: We're already installing FMA in the steps below. Then, we can remove this text from the prerequisites: We could either remove this content completely or just add it as a "note".

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,441 @@
...
+  -f ${REPO_ROOT}/guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml \
+  -n ${NAMESPACE} --version ${ROUTER_CHART_VERSION}
+```
+<!-- guide:deploy.standalone end -->
+
+### 5. Deploy the Model Server (Dual Pods)
+
+Apply the FMA custom resources — `InferenceServerConfig`, `LauncherConfig`, and `LauncherPopulationPolicy` — **together with** the server-requesting `fma-requester` Deployment **and the KEDA autoscaling layer** in a single `kubectl apply -k modelserver/`. This guide's `modelserver/` reuses the base [`fast-model-actuation`](../fast-model-actuation) guide's manifests as a kustomize base — so the shared FMA plumbing (`LauncherConfig`, `LauncherPopulationPolicy`, the requester Deployment) is maintained once — and patches only the KEDA-specific deltas: the `InferenceServerConfig` serves Qwen3-32B in dev mode, and the requester starts at 1 replica (the KEDA floor). The autoscaler is pulled in via `../keda/overlays/ocp`. Bundling KEDA here — rather than as a separate apply — is what lets the benchmark's kustomize standup bring up the autoscaler; the overlay is namespace-agnostic, so all of these resources land in `${NAMESPACE}` from the `-n` flag:
```
dumb0002: In the sentence `... servers Qwen3-32B in dev mode ...` - are you're referring to vLLM dev mode to enable sleep/wake? If so, I suggest to remove the text `in dev mode` because it's required more explanation to put `dev mode` in the right context for the user to understand.

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,441 @@
...
+<!-- guide:deploy.modelserver start -->
+```bash
+kubectl apply -n ${NAMESPACE} -k ${REPO_ROOT}/guides/${GUIDE_NAME}/modelserver/
+
+kubectl rollout status deployment/fma-requester -n ${NAMESPACE} --timeout=300s
+```
+<!-- guide:deploy.modelserver end -->
+
+> [!NOTE]
+> This guide uses [Qwen/Qwen3-32B](https://huggingface.co/Qwen/Qwen3-32B), which is publicly accessible and does not require a HuggingFace token. Saturation is a ratio of offered load to serving capacity, so any model saturates given enough load — the choice of a 32B model here is practical, not definitional: at the request rate this guide's workload generates on a single 80 GB GPU, a small model's service rate is high enough that `pool_saturation` stays near zero and never crosses the `0.7` threshold, so only the secondary `llm_d_epp_request_running` trigger would ever fire. A 32B model's lower service rate and larger KV footprint push saturation past the threshold under the same load, so the primary saturation trigger — the feature this guide exists to demonstrate — actually does the scaling.
```
dumb0002: IMO, this note can be removed - we don't need to explain here why, we selected 32B. This seems to me just over-explaining.

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,441 @@
...
+kubectl run curl-test --rm -i --restart=Never \
+  --image=${CURL_TEST_IMAGE} \
+  --namespace="${NAMESPACE}" \
+  --env="IP=${IP}" \
+  --env="MODEL=${MODEL}" \
+  -- /bin/sh -c 'curl -sS -X POST "http://${IP}/v1/completions" -H "Content-Type: application/json" -d "{\"model\": \"${MODEL}\", \"prompt\": \"How are you today?\"}"'
+```
+<!-- guide:verify.tests.request end -->
+
+### 3. Demonstrate Warm Start
```
dumb0002: Why to  we need to demonstrate `FMA warm-start` here without using KEDA? IMO, the `warm-start` scenario described here is better suited for the FMA only WLP.

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,441 @@
...
+
+kubectl scale deployment fma-requester -n ${NAMESPACE} --replicas=2
+
+time kubectl rollout status deployment/fma-requester -n ${NAMESPACE} --timeout=300s
+
+kubectl logs -n ${NAMESPACE} -l app.kubernetes.io/component=dual-pods-controller --tail=500 | grep -i "create_instance" || true
+```
+<!-- guide:verify.tests.warm_start end -->
+
+### 4. Demonstrate Hot Start
```
dumb0002: Same here: why to  we need to demonstrate `FMA hot-start` here without using KEDA? IMO, the `hot-start` scenario described here is better suited for the FMA only WLP.

guides/fast-model-actuation-keda/README.md
aavarghese: Changing to a note

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,381 @@
...
+
+## Configuration
+
+| Parameter                | Value                                                        |
+| ------------------------ | ------------------------------------------------------------ |
+| Model                    | [Qwen/Qwen3-0.6B](https://huggingface.co/Qwen/Qwen3-0.6B)    |
+| Requester replicas       | 0 (KEDA floor, idle) … 2 (KEDA ceiling)                      |
+| Launcher count           | 1 (per matching GPU node)                                    |
+| GPUs per requester pod   | 1                                                            |
+| Scale metric             | `llm_d_epp_flow_control_queue_size` (threshold `5`, activation `0`) |
```
dumb0002: Here `threshold` value should be `1` here - see https://github.com/llm-d/llm-d-benchmark/blob/main/config/scenarios/cicd/ocp-keda-fma-warmstart.yaml#L192-L197

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,381 @@
...
+## Overview
+
+This guide combines [Fast Model Actuation (FMA)](../fast-model-actuation/README.md) with **scale-from-zero autoscaling via KEDA**. A KEDA `ScaledObject` scales the FMA server-requesting Deployment (`fma-requester`) on Endpoint Picker (EPP) flow-control metrics — all the way down to **zero** when the pool is idle, and back up on the first queued request. Each requesting pod reserves a GPU and drives the FMA controllers to bring a vLLM instance online via a **hot or warm start**; scaling to zero releases the GPU and puts the vLLM to sleep. (See the [FMA guide](../fast-model-actuation/README.md#overview) for what hot and warm start mean.)
+
+## Configuration
+
+| Parameter                | Value                                                        |
+| ------------------------ | ------------------------------------------------------------ |
+| Model                    | [Qwen/Qwen3-0.6B](https://huggingface.co/Qwen/Qwen3-0.6B)    |
+| Requester replicas       | 0 (KEDA floor, idle) … 2 (KEDA ceiling)                      |
```
dumb0002: The ceiling for KEDA should be one. Meaning: `scale-up: 0 --> 1 & scale-down: 1 --> 0`

guides/fast-model-actuation-keda/README.md
dumb0002, line 17: Here threshold value should be 1 here - see https://github.com/llm-d/llm-d-benchmark/blob/main/config/scenarios/cicd/ocp-keda-fma-warmstart.yaml#L192-L197

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,381 @@
...
+```
+<!-- guide:deploy.standalone end -->
+
+### 5. Deploy the Model Server (Dual Pods)
+
+Apply the FMA custom resources — `InferenceServerConfig`, `LauncherConfig`, and `LauncherPopulationPolicy` — **together with** the server-requesting `fma-requester` Deployment **and the KEDA autoscaling layer** in a single `kubectl apply -k modelserver/`. This guide's `modelserver/` reuses the base [`fast-model-actuation`](../fast-model-actuation) guide's manifests as a kustomize base — so the shared FMA plumbing (`LauncherConfig`, `LauncherPopulationPolicy`, the requester Deployment) is maintained once — and patches only the KEDA-specific deltas: the `InferenceServerConfig` serves Qwen3-0.6B, and the requester is applied at 1 replica so the stack comes up serving — KEDA then owns the count, scaling it to 0 once the flow-control queue is empty and back to 1 on the next request. The autoscaler is pulled in via `../keda/overlays/ocp`. Bundling KEDA here — rather than as a separate apply — is what lets the benchmark's kustomize standup bring up the autoscaler; the overlay is namespace-agnostic, so all of these resources land in `${NAMESPACE}` from the `-n` flag:
```
dumb0002, line 204: This is no longe this delta, isn't it? "the `InferenceServerConfig` serves Qwen3-0.6B"?

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,381 @@
...
+
+<!-- guide:deploy.modelserver start -->
+```bash
+kubectl apply -n ${NAMESPACE} -k ${REPO_ROOT}/guides/${GUIDE_NAME}/modelserver/
+
+kubectl rollout status deployment/fma-requester -n ${NAMESPACE} --timeout=300s
+```
+<!-- guide:deploy.modelserver end -->
+
+### 6. Enable Scale-from-Zero Autoscaling (KEDA)
```
dumb0002, line 214: We should create also the KEDA scaledObject here - its creation seems to be coupled with the RBAC creation in `step-2`. `Step-6` is the most appropriate location to create and verify the KEDA `ScaledTarget` object.

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,381 @@
...
+      if [ "${code}" = "200" ]; then cat /tmp/b; echo; jq -e ".choices[0].text | length > 0" /tmp/b >/dev/null; exit 0; fi
+      i=$((i+1)); sleep 15
+    done
+    echo "no 200 after retries"; cat /tmp/b 2>/dev/null || true; exit 1'
+```
+<!-- guide:verify.tests.request end -->
+
+## Benchmarking
+
+This guide uses [`llmdbenchmark`](https://github.com/llm-d/llm-d-benchmark) — the supported standard CLI for llm-d performance benchmarking. The commands below run the `inference-perf` harness with the `shared_prefix_synthetic_heavy.yaml` workload, which drives sustained load through the EPP so you can watch the flow-control queue rise and KEDA scale the requester up from zero, then drain and scale it back down; the richer, FMA/KEDA-specific experimentation workflow lives in [`llm-d-benchmark`](https://github.com/llm-d/llm-d-benchmark) itself.
```
dumb0002, line 297: Since, we're using now a smaller model `Qwen/Qwen3-32B -->Qwen/Qwen3-0.6B`  - the content in this paragraph needs to be updated.

guides/fast-model-actuation-keda/keda/base/kustomization.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - ../../../workload-autoscaling/keda-epp-saturation/optimized-baseline/base
```
dumb0002, line 5: `keda/base/kustomization.yaml:5` pulls `keda-epp-saturation/optimized-baseline/base`, then patches the saturation trigger into a queue-size trigger, renames it, and deletes trigger 1. But `keda-epp-queue/optimized-baseline/base` already is a **queue-size** ScaledObject.

guides/fast-model-actuation-keda/keda/overlays/ocp/kustomization.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - patch: |-
+      - op: replace
+        path: /spec/secretTargetRef
+        value:
+          - parameter: bearerToken
+            name: keda-epp-metrics-reader-token
+            key: token
+          - parameter: ca
+            name: keda-epp-metrics-reader-token
+            key: service-ca.crt
```
dumb0002, line 49: No `k8s/` overlay. `keda-epp-queue` ships a 9-line passthrough so non-OpenShift users have something to apply. FMA has none, and `modelserver/kustomization.yaml:6` hardcodes `../keda/overlays/ocp` — so every standup, OpenShift or not, gets Thanos + service-ca wiring; also the directory naming (`overlays/ocp vs ocp)` also breaks the sibling convention.

guides/fast-model-actuation-keda/keda/overlays/ocp/epp-servicemonitor.yaml
```diff
@@ -0,0 +1,37 @@
...
+      app.kubernetes.io/name: fast-model-actuation-keda-epp
+  jobLabel: fast-model-actuation-keda-epp
+  endpoints:
+    - port: http-metrics
+      path: /metrics
+      interval: 10s
+      authorization:
+        credentials:
+          name: fast-model-actuation-keda-epp-metrics-token
+          key: token
```
dumb0002, line 37: We should use the same approach used in others WLP to enable prometheus monitoring for EPP. For example: 

     Example-1: https://github.com/aavarghese/llm-d/tree/kedafma_guide/guides/predicted-latency-routing#3-enable-monitoring-optional
     Example-2: https://github.com/aavarghese/llm-d/tree/kedafma_guide/guides/coord-disaggregation#5-optional-enable-monitoring


This should be configure during the installation of the EPP router. For example:

```
helm install ${GUIDE_NAME} \
  ${ROUTER_STANDALONE_CHART} \
  -f ${REPO_ROOT}/guides/recipes/router/features/monitoring.values.yaml \   # <-- add
  -f ${REPO_ROOT}/guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml \
```

guides/fast-model-actuation-keda/keda/overlays/ocp/keda-metrics-reader-serviceaccount.yaml
```diff
@@ -0,0 +1,11 @@
...
+# Querier on OpenShift. Inlined here (rather than pulled from the shared
+# recipes/autoscaling/metrics-reader component) because this overlay is
+# namespace-agnostic: it carries no `namespace:` transformer, so it cannot use
+# the component's transformer-dependent ClusterRoleBinding. The matching
+# cluster-monitoring-view binding is created imperatively in the guide's RBAC
+# step.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: keda-epp-metrics-reader
```
dumb0002, line 11: The comment **below** applies to both files `keda-metrics-reader-serviceacount.yaml` & `keda-metrics-reader-token.yaml`:

**Duplicated — should be deleted and replaced with the component**

`keda-metrics-reader-serviceaccount.yaml` and `keda-metrics-reader-token.yaml` are byte-for-byte the same resources as `guides/recipes/autoscaling/metrics-reader/{serviceaccount,token-secret}.yaml` — same names (`keda-epp-metrics-reader`, `keda-epp-metrics-reader-token`), same annotation, same type. Even the comments are copied nearly verbatim.

Only the ClusterRoleBinding is transformer-dependent — the SA and token Secret are plain namespaced resources that inherit the namespace from `kubectl apply -n`. So even accepting the namespace-agnostic constraint, the right move is to pull in the component and drop the CRB, not to hand-copy all three. 

Net: the guide creates the CRB imperatively in guide.yaml anyway, so importing the component and patching out the CRB (or suffixing it) gives the same result with 2 fewer files and matches the pattern used in other WLP.

guides/fast-model-actuation-keda/README.md
```diff
@@ -0,0 +1,381 @@
...
+  -f "https://raw.githubusercontent.com/llm-d-incubation/llm-d-fast-model-actuation/v${FMA_VERSION}/config/crds.yaml"
+kubectl wait --for=condition=Established crd/inferenceserverconfigs.fma.llm-d.ai --timeout=120s
+kubectl wait --for=condition=Established crd/launcherconfigs.fma.llm-d.ai --timeout=120s
+kubectl wait --for=condition=Established crd/launcherpopulationpolicies.fma.llm-d.ai --timeout=120s
+```
+<!-- guide:deploy.fma_crds end -->
+
+### 2. Grant RBAC Permissions
+
+The FMA controllers need cluster-level access to list nodes (for the launcher-populator) and namespace-level access for launcher pods to read their own pod spec:
```
dumb0002, line 120: This step is also creating RBAC for KEDA and EPP - so this paragraph needs to be updated to reflect that. I recommend doing so after addressing this comment: https://github.com/llm-d/llm-d/pull/2191/changes#r3985254428

dumb0002 (approved):
All my comments have been addressed. /LGTM

diegocastanibm (approved):

PR #2196, opened 2026-08-06
feat(precise): enable KV event replay buffer for index recovery

RishabhSaini:
## Summary

- Enable vLLM's ZMQ replay buffer on the GPU model server backend so the router can rebuild its KV-block index after an EPP restart without waiting for live traffic
- Bump GPU vLLM image from v0.25.0 to v0.26.0 (includes the replay topic frame fix, vllm-project/vllm#45177)
- Add `replaySocketPort: 5557` to the router's `podDiscoveryConfig` so the EPP connects to each pod's ROUTER socket for replay requests

## Details

Each GPU model server now binds a ZMQ ROUTER socket on port 5557 alongside the existing PUB socket on 5556. On first connect (or after an EPP restart), each subscriber requests a replay of buffered events from the model server to rebuild the prefix-cache index without waiting for live traffic. vLLM retains the last 10,000 event batches in memory by default.

Non-GPU backends (CPU, AMD ROCm, XPU, TPU) remain unchanged — their vLLM images are on v0.23.0 or earlier which lack replay support. Replay will be enabled for those backends when their images are updated to include vllm-project/vllm#45177.

Depends on: llm-d/llm-d-router#1887 (merged)

## Test plan

- [ ] Deploy precise-prefix-cache-routing guide on GPU cluster with v0.26.0 image
- [ ] Verify KV events flow normally (prefix affinity routing works)
- [ ] Kill the EPP pod, verify it recovers and rebuilds the index via replay (check EPP logs for "Replay complete")


RishabhSaini:
## Live Cluster Validation

Tested on a GPU cluster with 1 decode replica (Qwen/Qwen3-32B, TP=2, vLLM v0.26.0) and a custom EPP image built from llm-d-router main (includes PR llm-d/llm-d-router#1887).

### Setup
- EPP configured with `replaySocketPort: 5559` and `socketPort: 5556`
- vLLM configured with `replay_endpoint: tcp://*:5559`

### First boot (empty replay buffer)
EPP connects to model server, requests proactive replay, receives 0 events (buffer empty):
```
zmq-subscriber  "Connected subscriber socket"  endpoint=tcp://10.0.1.132:5556
zmq-subscriber  "Requesting proactive replay on connect"  replayEndpoint=tcp://10.0.1.132:5559
zmq-replay      "Replay complete"  replayed=0  startSeq=0
```

### After sending requests + EPP restart
Sent 10 requests with distinct prompts (~300 tokens each, blockSize=64) to populate vLLM's KV cache and replay buffer, then force-deleted the EPP pod. The replacement pod replayed 20 event batches:
```
zmq-subscriber  "Requesting proactive replay on connect"  replayEndpoint=tcp://10.0.1.132:5559
zmq-replay      "Replay complete"  replayed=20  startSeq=0
```

The KV-block index was rebuilt from the replay buffer without waiting for new live traffic.

Note: an earlier test with 5 identical prompts showed `replayed=1` because vLLM's prefix cache hit on requests 2-5 — only the first request allocated new blocks. Switching to 5 distinct prompts confirmed `replayed=20` (each prompt generating ~4 block-allocation event batches).

### What this covers
- **Proactive replay on EPP restart**: subscriber connects, requests replay from seq 0, rebuilds index from buffered events
- **Mid-stream gap replay** (unit-tested only): if the subscriber detects `seq > lastSeq + 1`, it requests replay from `lastSeq + 1` to fill the gap. This path cannot be triggered on a live cluster without overflowing ZMQ's 100k HWM — TCP retransmission guarantees ordered delivery at the transport layer. The gap path uses the same `requestReplay()` function proven above; only the trigger condition differs.

vMaroon (approved):
Great work

PR #2205, opened 2026-08-07
fix: batch-gateway guide

zdtsw:
# Changes
- fix chart url
- missing PVC for the steps
- wrong API port on 8000 than health port
- drop secret as default one to set
- fix lint


lioraron (approved):

zdtsw:
cc @Gregory-Pereira seems need someone with write permissions

PR #2206, opened 2026-08-07
updating images preparing for v0.9.0 release

Gregory-Pereira:
cc @zdtsw @wseaton @tlrmchlsmth

changes:
- remove gb200 images
- remove offloading connector as its landed in vllm
- remove lmcache and inifistore build (exists in vLLM kvconnectors requirements / image)
- bump images to match vllm v0.26.0
- drop old nvshmem patches that are no longer applicable (kept the one for 3.4.5 as we sometimes build against that target)
- create new components for different images
- version bumps


docker/Dockerfile.cuda
```diff
@@ -31,44 +31,35 @@ ARG TORCH_CUDA_ARCH_LIST="8.0;8.6;8.9;9.0;9.0a;10.0;12.0+PTX"
...
 
-# GB200-specific DeepEP variant (built alongside standard, selected via --target runtime-gb200)
-ARG DEEPEP_GB200_REPO="https://github.com/tlrmchlsmth/DeepEP"
-ARG DEEPEP_GB200_VERSION="sgl-gb200-blog-pt2"
 ARG DEEPGEMM_REPO="https://github.com/deepseek-ai/DeepGEMM"
 ARG DEEPGEMM_VERSION="477618cd51baffca09c4b0b87e97c03fe827ef03"
-ARG FLASHINFER_REPO="https://github.com/flashinfer-ai/flashinfer.git"
-ARG FLASHINFER_VERSION="v0.6.12"
+
+ARG FLASHINFER_VERSION="0.6.14"
```
zdtsw, line 62: should not this revmoed as well?

guides/recipes/modelserver/components/images/cpu-vllm/kustomization.yaml
```diff
@@ -3,7 +3,7 @@ kind: Component
 images:
   - name: REPLACE_MODEL_SERVER_IMAGE
     newName: ghcr.io/llm-d/llm-d-cpu
-    newTag: v0.7.0
+    newTag: v0.8.1
```
zdtsw, line 6: this is from previous release, isnt it? so need bump to 0.9.0 once released

docker/Dockerfile.cuda
```diff
@@ -537,6 +498,7 @@ COPY scripts/warn-vllm-precompiled.sh /opt/
     # 1) install vllm from source totally editably - dev option, supports fork and commit swapping
     # 3) install vllm from source with pulling precompiled binaries (shared libraries) from the vllm wheels index - less flexible dev option, may or may not work swapping shas but faster than full editable
     # 2) install vllm from a wheel on the vllm wheels index - prod option, no flexibility
+ARG FLASHINFER_VERSION
```
zdtsw, line 501: and here

Gregory-Pereira: We need this, we still install flashinfer we just dont allow the option to build it form source because it times out

docker/Dockerfile.cuda
Gregory-Pereira: We want flashinfer as it gets resolved in deps but we just dont build from source

maugustosilva (approved):

PR #2209, opened 2026-08-07
fix(flow-control): repair user-facing guide breakages

LukeAVanDrie:
1. **`objectives.yaml` binds to a pool that does not exist.** All three InferenceObjectives hardcode `poolRef.name: default-pool`; this guide's InferencePool is `flow-control`. Priority and fairness headers resolve to nothing and every request lands in band `priority="0"` (observed: 450/450 burst requests dispatched at priority 0 despite `best-effort-traffic` headers), so Use Case 1 silently no-ops. After the fix, a premium request lands in `priority="100"`.

2. **The CRD manifests URL 404s.** The command interpolates `ROUTER_CHART_VERSION` (`v0`), a floating OCI chart channel with no matching GitHub release. `env.sh` gains a distinct `ROUTER_RELEASE_VERSION=v0.9.0` and the README prerequisite uses it.

3. **The verification grep matches nothing.** `grep "Flow Control enabled"` has no counterpart in the EPP. Current `main` logs `Initializing Flow Control layer` (the `experimental` qualifier was dropped at GA graduation), and the disabled path logs `Flow Control layer is disabled...`. The doc now greps the init line and names the disabled-path line, so no output stays diagnosable.

4. **Monitoring ordering is backwards.** Step 3 ("Enable monitoring") tells the reader to add values during the step 1 install, two sections after they installed, and the ServiceMonitor CRD requires the monitoring stack first. The README now states the ordering up front and gives the re-run path for readers who already installed.

5. **Duplicate metrics port fails the install.** `flow-control.values.yaml` declares a 9090 `extraServicePort` the chart already templates unconditionally on the EPP Service, so the install fails with `INSTALLATION FAILED ... duplicate entries for key [port=9090,protocol="TCP"]` regardless of whether `monitoring.values.yaml` is layered (that file never touches `extraServicePorts`). With the entry removed, the chart's own 9090 port remains, so the README's no-monitoring `curl :9090/metrics` check still works.

6. **Cleanup path drifted from the deploy path.** The deploy step renders `modelserver/gpu/vllm/${INFRA_PROVIDER}/`; cleanup rendered `modelserver/gpu/vllm/`, which no longer resolves the same manifests. Cleanup now matches, with a reminder to set `INFRA_PROVIDER` to the deploy-time value.

### DIFF
--- a/guides/env.sh
+++ b/guides/env.sh
@@ -7,6 +7,10 @@ export REPO_ROOT=${REPO_ROOT:-$(realpath $(git rev-parse --show-toplevel 2>/dev/
 export GAIE_VERSION=v1.5.0
 export ROUTER_CHART_VERSION=v0
 export ROUTER_EPP_VERSION=main
+# GitHub release tag for CRD manifests. Distinct from ROUTER_CHART_VERSION:
+# the chart channel (v0) floats on the OCI registry and has no matching
+# GitHub release, so release-asset URLs must use a real release tag.
+export ROUTER_RELEASE_VERSION=v0.9.0
 export ROUTER_STANDALONE_CHART=oci://ghcr.io/llm-d/charts/llm-d-router-standalone
 export ROUTER_GATEWAY_CHART=oci://ghcr.io/llm-d/charts/llm-d-router-gateway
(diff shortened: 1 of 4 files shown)


LukeAVanDrie:
cc @RishabhSaini

PR #2213, opened 2026-08-07
docs(flow-control): make use-case verification observable and fix guide friction

LukeAVanDrie:
The Use Case 2 load test could not demonstrate the queuing it claims to verify, and an audit of the rest of the guide surfaced more breakage in the same file. Every changed command was validated on a live cluster (GKE, Helm 4.2.0, one Qwen3-32B TP=2 replica).

## Use Case 2

- Replace the dead `hey` download (S3 returns 403; the debug image's busybox wget lacks TLS) with a curl loop using tools already in the debug image.
- Sustain the load so the queue is observable: `max_tokens: 500`, `ignore_eos: true`, ~2k-token prompt; burst in the background and poll while it runs. The old payload drained in under a second.
- Derive the test sizing from the values file instead of hardcoding it. `MAX_CONCURRENCY` is read out of `router/flow-control.values.yaml` with awk and passed into the debug pod; the burst is `MAX_CONCURRENCY + SURPLUS` and the expected queue peak is `SURPLUS` (20, kept below the best-effort band's `maxRequests`). A retuned values file changes the test with it, so the guide's numbers cannot drift out from under the config.
- Scale the pool to one replica for the burst, recording the prior replica count and restoring it afterward rather than assuming 8.
- Pre-flight a single request (expect 200) before the burst; sum all `priority="-10"` series in the poll; the peak-0 diagnosis covers poll timing and replica count.

## Rest of the guide

- Use Case 1: add a verification step; `llm_d_epp_flow_control_request_queue_duration_seconds_count` must show a `priority="100"` series (records on every outcome, including immediate dispatch).
- `helm install` → `helm upgrade --install` in both install commands; the re-run instructions already assumed it.
- Add `-c epp` to the verification grep as an in-block comment; kubectl defaults to the Envoy sidecar, whose log never matches.
- Pass `MODEL_NAME` into the debug pod; the UC1 request body and UC2 payload use it in-pod.
- Document the router pod's 8 vCPU / 16 GiB request at the install step, citing `base.values.yaml` as the source.
- Drop the stale `[Experimental]` title, a duplicate QoS caveat, and the hardcoded `--model` value.

## helpers/benchmark.md

Harness- and provider-specific material moved to the benchmarking helper (the declared single source of truth for benchmarking), which the guide now links instead of duplicating:

- New troubleshooting entry for `ProvisioningFailed ... multi writer with mount access type`: the RWX StorageClass requirement, the `storage.workloadPvc.storageClassName` scenario override, a GKE Filestore walkthrough, and stale-claim recovery (PVC StorageClass is immutable and the CLI reuses an existing claim).
- New troubleshooting entry for the harness pod stuck `Pending` on its 16 vCPU / 32 GiB request.
- New "Cleaning up harness resources" section; the guide's Cleanup links it. The workload PVC is backed by shared storage that bills until removed.
- The guide's benchmark section keeps a single NOTE pointing at both.

Follow-up: a `llmdbenchmark cleanup` subcommand upstream would replace the hand-maintained resource list in benchmark.md entirely; issue to be filed on llm-d/llm-d-benchmark.

## Observed on the live run

- UC1: `outcome="Dispatched", priority="100"` count 1 after one premium request.
- UC2 (pre-parameterization commands): peak queue depth 18 from a 150-request burst against `maxConcurrency` 132 — the same arithmetic the derived commands now encode with `SURPLUS=20` (expected peak 20). 
- Benchmark: the documented GKE failure mode reproduces; with the StorageClass override on a fresh claim, Filestore bound in ~2.5 min and `shared_prefix_synthetic.yaml` ran end to end.
- Not covered on a 2-GPU cluster: detector behavior while ready endpoints drain, 8-replica scale-back timing.

### DIFF
--- a/guides/flow-control/README.md
+++ b/guides/flow-control/README.md
@@ -223,6 +223,30 @@ export MAX_CONCURRENCY=$(awk '$1 == "maxConcurrency:" {print $2}' \
 echo "maxConcurrency: ${MAX_CONCURRENCY}"   # expect the integer set in the values file
 ```
 
+**Grant read access to the EPP metrics endpoint.** The EPP authenticates every metrics
+scrape against the Kubernetes API: a TokenReview on the caller's bearer token, then a
+SubjectAccessReview on the `/metrics` URL. Give the debug pod's service account
+permission to read `/metrics`, and give the EPP's service account the
+`system:auth-delegator` role it needs to run the reviews (the router chart does not
+ship this binding):
+
+```bash
+kubectl create clusterrole ${GUIDE_NAME}-metrics-reader \
+    --verb=get --non-resource-url=/metrics \
+    --dry-run=client -o yaml | kubectl apply -f -
+kubectl create clusterrolebinding ${GUIDE_NAME}-metrics-reader \
+    --clusterrole=${GUIDE_NAME}-metrics-reader \
+    --serviceaccount=${NAMESPACE}:default \
+    --dry-run=client -o yaml | kubectl apply -f -
+kubectl create clusterrolebinding ${GUIDE_NAME}-epp-auth-delegator \
+    --clusterrole=system:auth-delegator \
+    --serviceaccount=${NAMESPACE}:${GUIDE_NAME}-epp \
+    --dry-run=client -o yaml | kubectl apply -f -
+```
+
+Without these grants the metrics endpoint returns `401 Unauthorized`, and every
+metrics check in this guide reads as empty grep output.
+
 **Open a temporary interactive shell inside the cluster:**
 
 ```bash
@@ -237,14 +261,21 @@ kubectl run curl-debug --rm -it \
     -- /bin/bash
 ```
 
-**From inside the debug pod, check the metrics:**
+**From inside the debug pod, check the metrics.** The pod's automounted service account
+token authenticates the scrape:
 
 ```bash
-curl http://${GUIDE_NAME}-epp:9090/metrics | grep llm_d_epp_flow_control_queue_size
+TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+curl -s -o metrics.txt -w "%{http_code}\n" -H "Authorization: Bearer ${TOKEN}" \
+  http://${GUIDE_NAME}-epp:9090/metrics   # expect: 200
+grep llm_d_epp_flow_control_queue_size metrics.txt
 ```
 
 Expected: one series per active flow (tenant × priority), all `0` on an idle pool. A
 flow's series appears after its first request queues, so a quiet pool may show none.
+A `401` means the metrics-access grants above were skipped; a `500` means the EPP's
+service account lacks the auth-delegator binding (the EPP log shows a failed
+TokenReview).
 
 ## Use Cases
 
@@ -296,7 +327,8 @@ Still inside the `curl-debug` pod, confirm the request was accounted under the p
 band:
 
 ```bash
-curl -s http://${GUIDE_NAME}-epp:9090/metrics \
+TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+curl -s -H "Authorization: Bearer ${TOKEN}" http://${GUIDE_NAME}-epp:9090/metrics \
   | grep 'llm_d_epp_flow_control_request_queue_duration_seconds_count' \
   | grep 'priority="100"'
 ```
@@ -383,9 +415,10 @@ To verify backpressure management, you must overwhelm the pool's capacity. Becau
    and record the peak:
 
     ```bash
+    TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
     PEAK=0
     for i in $(seq 1 30); do
-      Q=$(curl -s http://${GUIDE_NAME}-epp:9090/metrics \
+      Q=$(curl -s -H "Authorization: Bearer ${TOKEN}" http://${GUIDE_NAME}-epp:9090/metrics \
         | awk '/llm_d_epp_flow_control_queue_size.*priority="-10"/ {s+=$2} END {print s+0}')
       [ "$Q" -gt "$PEAK" ] && PEAK=$Q
       sleep 1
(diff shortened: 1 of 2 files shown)


guides/flow-control/README.md
```diff
@@ -118,13 +118,17 @@ Flow Control is a software-level scheduling feature at the EPP layer and is enti
...
 ```bash
-helm install ${GUIDE_NAME} \
+helm upgrade --install ${GUIDE_NAME} \
     ${ROUTER_STANDALONE_CHART} \
     -f ${REPO_ROOT}/guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml \
 ```
 
+The router pod requests 8 vCPU and 16 GiB of memory (4 vCPU / 8 GiB each for the EPP and
```
alexagriffith: instead of hard coding this , is there a link or piece of code to point to that in case it changes, that link would show the update?

guides/flow-control/README.md
```diff
@@ -196,17 +200,52 @@ export IP=$(kubectl get gateway llm-d-inference-gateway -n ${NAMESPACE} -o jsonp
...
 
-To fully verify that queuing and backpressure are working, you must apply concurrent load. We will demonstrate this using a load generation tool in **Use Case 2** below. For now, set up your test environment.
+To fully verify that queuing and backpressure are working, you must apply concurrent load; [Use Case 2](#use-case-2-backpressure-management) does that with a burst of concurrent requests. For now, set up the test environment.
+
+**Read `maxConcurrency` from [router/flow-control.values.yaml](./router/flow-control.values.yaml).**
+The Use Case 2 load test sizes its burst from `MAX_CONCURRENCY`; a retuned values file
+changes the burst with it:
+
+```bash
+export MAX_CONCURRENCY=$(awk '$1 == "maxConcurrency:" {print $2}' \
```
alexagriffith: The `$1 == "maxConcurrency:"` match is fine today, but it isn't scoped to the `concurrency-detector` block — it matches any line whose first field is `maxConcurrency:` at any indentation. If a future values file gains a second `maxConcurrency:` (e.g. a per-band or per-profile setting), this prints two values and the `export` silently captures both, breaking the burst sizing with no error.

Cheap hardening — pin it to a single expected match and fail loudly otherwise:

```bash
export MAX_CONCURRENCY=$(awk '$1 == "maxConcurrency:" {print $2; found++} END {exit found!=1}' \
    ${REPO_ROOT}/guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml) \
  || { echo "expected exactly one maxConcurrency: in the values file" >&2; }
echo "maxConcurrency: ${MAX_CONCURRENCY}"
```

Or, if you'd rather stay a one-liner, `grep -m1` documents the single-match intent: `grep -m1 'maxConcurrency:' ... | awk '{print $2}'`. Non-blocking.

guides/flow-control/README.md
```diff
@@ -196,17 +200,52 @@ export IP=$(kubectl get gateway llm-d-inference-gateway -n ${NAMESPACE} -o jsonp
...
+    ${REPO_ROOT}/guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml)
+echo "maxConcurrency: ${MAX_CONCURRENCY}"   # expect the integer set in the values file
+```
+
+**Grant read access to the EPP metrics endpoint.** The EPP authenticates every metrics
+scrape against the Kubernetes API: a TokenReview on the caller's bearer token, then a
+SubjectAccessReview on the `/metrics` URL. Give the debug pod's service account
+permission to read `/metrics`, and give the EPP's service account the
+`system:auth-delegator` role it needs to run the reviews (the router chart does not
```
alexagriffith: Just FYI — the chart does actually ship this binding if you have the flag on. `routerlib`'s RBAC template grants `tokenreviews` / `subjectaccessreviews` / `get /metrics` to the EPP SA, but it's gated behind `router.monitoring.prometheus.enabled` (default `false`). So setting that flag in `flow-control.values.yaml` would give you the EPP-side `auth-delegator` grant and make the `${GUIDE_NAME}-epp-auth-delegator` binding here redundant.

Two things worth noting before changing the flag

1. Enabling it also pulls in a `ServiceMonitor` (the provider defaults to `prometheusoperator`), which needs the Prometheus Operator CRD, a dependency this guide currently avoids. 
2. And the debug-pod read grant (`metrics-reader` → `default` SA) would still be needed regardless, since the chart doesn't grant arbitrary callers.

So my suggestion would be one of:
- keep the inline RBAC but reword the "chart does not ship this" note (it's only true when `prometheus.enabled: false`), or
- enable the flag and accept the ServiceMonitor / CRD dependency.

Non-blocking.

helpers/benchmark.md
```diff
@@ -315,6 +331,30 @@ You passed a local filesystem path to `--output`. `--output` only accepts `local
...
+
+If a run already failed on the wrong class, delete the stale claim before retrying. A PVC's StorageClass is immutable, and the CLI reuses an existing claim:
+
+```bash
+kubectl delete pod/access-to-harness-data-workload-pvc pvc/workload-pvc -n <namespace>
+```
+
+### Harness pod stuck `Pending` with `Insufficient cpu`
+
+The harness pod requests 16 vCPU and 32 GiB of memory. It stays `Pending` until a node with that much allocatable capacity is available; check `kubectl describe pod -l app=llmdbench-harness-launcher -n <namespace>` for the scheduling events.
```
alexagriffith: Minor: the memory figure isn't asserted upstream, and hard-coding these numbers risks going stale. Suggest linking the CPU figure to the source of truth and dropping the specific memory number (your line otherwise unchanged):

```suggestion
The harness pod requests 16 vCPU ([resource requirements](https://github.com/llm-d/llm-d-benchmark/blob/main/docs/resource_requirements.md)). It stays `Pending` until a node with that much allocatable capacity is available; check `kubectl describe pod -l app=llmdbench-harness-launcher -n <namespace>` for the scheduling events.
```

Non-blocking.

guides/flow-control/README.md
LukeAVanDrie: Ah, I had filed https://github.com/llm-d/llm-d-router/issues/2370. Let me update that accordingly.

helpers/benchmark.md
```diff
@@ -293,6 +294,21 @@ llmdbenchmark \
 
 All three are also exposed as env vars: `LLMDBENCH_WAIT_TIMEOUT`, `LLMDBENCH_DATA_ACCESS_TIMEOUT`, `LLMDBENCH_PVC_BIND_TIMEOUT`.
 
+## Cleaning up harness resources
+
+The CLI tears down the harness launcher pod after each run but leaves the workload PVC and its supporting resources in the namespace, so workload data survives between runs. The PVC is backed by shared storage (on GKE, a Filestore instance) that bills until removed. When you are done benchmarking in a namespace, delete the leftovers:
+
+```bash
```
alexagriffith, line 301: was going to say we should codify this, I see you made an issue so 👍

alexagriffith (commented):
Reviewed the flow-control guide + benchmark helper changes. Verified the technical claims against llm-d-router `origin/main`: metric names/labels (`flow_control_queue_size`, `request_queue_duration_seconds`) check out, the metrics-auth story is accurate (`MetricsEndpointAuth` defaults true, `filters.WithAuthenticationAndAuthorization`), and the resource figures match `base.values.yaml`.

Left a handful some comments: awk-scoping fragility, a note that the chart can ship the metrics-auth RBAC when `prometheus.enabled: true`, and a suggestion to link the harness resource figure to its upstream source rather than hard-code it. None are blockers. Nice work, and thanks for filing llm-d-benchmark#1789 for the cleanup follow-up.

alexagriffith (approved):

guides/flow-control/README.md
```diff
@@ -398,15 +534,19 @@ llmdbenchmark \
...
     --gateway-class  "${GATEWAY_CLASS}" \
-    --model          "Qwen/Qwen3-32B" \
+    --model          "${MODEL_NAME}" \
     --namespace      "${NAMESPACE}" \
     --harness        inference-perf \
     --workload       random_concurrent.yaml \
     --analyze
 ```
 
 > [!NOTE]
```
alexagriffith, line 549: Same harness figure as the `helpers/benchmark.md` troubleshooting section. Upstream only asserts the **CPU** default (16 CPUs, tunable via `LLMDBENCH_HARNESS_CPU_NR`) in [resource_requirements.md](https://github.com/llm-d/llm-d-benchmark/blob/main/docs/resource_requirements.md); the `32 GiB` **memory** figure isn't stated anywhere upstream. Whatever you settle on for the benchmark.md copy, worth mirroring here so the two stay in sync. Non-blocking.

guides/flow-control/README.md
LukeAVanDrie: Took your version with one twea: `export VAR=$(...)` masks the substitution's exit status (export returns 0), so I split the assignment from the export to let awk's` exit n!=1` propagate.

guides/flow-control/README.md
LukeAVanDrie: Went with option 1, keeping the inline RBAC to avoid the ServiceMonitor/CRD dependency and reworded the note to say the chart ships the grant behind prometheus.enabled.

guides/flow-control/README.md
LukeAVanDrie: Applied your suggestion in both places: CPU figure linked to resource_requirements.md, memory figure dropped.

guides/flow-control/README.md
LukeAVanDrie: done, see related comment below

Gregory-Pereira (approved):
/lgtm

PR #2224, opened 2026-08-10
fix: Disable USE_SCCACHE when the sccache fallback kicks in

### LINKED ISSUE #2223: CUDA image build fails: sccache fallback in setup-sccache.sh does not disable USE_SCCACHE
When `setup-sccache.sh` cannot start sccache it degrades gracefully, but the degrade path is incomplete and every CUDA image build fails at the gdrcopy step.

### What happens

`docker/scripts/common/setup-sccache.sh` deletes the sccache binary and unsets the CMake launcher variables, then returns success:

```bash
if ! /usr/local/bin/sccache --start-server; then
    echo "Warning: sccache failed to start, continuing without cache" >&2
    rm -f /usr/local/bin/sccache || true
    unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
    return 0
fi
```

It does not clear `USE_SCCACHE`. Every consumer sources the script and then branches on that original input variable, so they still believe sccache is available:

```bash
. /usr/local/bin/setup-sccache
...
if [ "${USE_SCCACHE}" = "true" ]; then
    export CC="sccache gcc" CXX="sccache g++"   # binary was just deleted
fi
```

gdrcopy builds with a plain Makefile rather than CMake, so unsetting the `CMAKE_*_COMPILER_LAUNCHER` variables has no effect on it, and the hardcoded `sccache` in `CC` is what actually breaks:

```
/usr/local/bin/setup-sccache: line 36: /usr/local/bin/sccache: No such file or directory
Warning: sccache failed to start, continuing without cache
+ export CC='sccache gcc' CXX='sccache g++'
sccache gcc -O2 -fPIC -I ../include -I gdrdrv/ -D GDRAPI_ARCH=ARM64 -c -o gdrapi.o gdrapi.c
make[1]: sccache: No such file or directory
make[1]: *** [<builtin>: gdrapi.o] Error 127
make: *** [Makefile:63: lib] Error 2
```

### Impact

All CUDA-family jobs fail identically at `Dockerfile.cuda` (`build-gdrcopy.sh`), on both architectures and both base OSes:

- `cuda-build-llm-d (linux/amd64, rhel, ubi9)`
- `cuda-build-llm-d (linux/amd64, ubuntu, ubuntu24.04)`
- `cuda-build-llm-d (linux/arm64, rhel, ubi9)`
- `cuda-build-llm-d (linux/arm64, ubuntu, ubuntu24.04)`
- `cuda-build-llm-d-debug (linux/amd64, rhel, ubi9)`
- `gb200-build-llm-d`
- `aws-build-llm-d (linux/amd64, rhel, ubi9)`

Example run: https://github.com/llm-d/llm-d/actions/runs/31362149292

Two more scripts carry the same latent bug and would fail the same way if the build got that far:

- `docker/scripts/cuda/builder/build-ucx.sh:29` — `export CC="sccache gcc" CXX="sccache g++"`
- `docker/scripts/cuda/builder/build-nixl.sh:32` — `export CC="sccache gcc" CXX="sccache g++" NVCC="sccache nvcc"`

And four scripts call bare `sccache --show-stats` after the binary has been removed, which aborts under `set -e`:

- `build-gdrcopy.sh:61`, `build-ucx.sh:62`, `build-lmcache.sh:88`, `build-compiled-wheels.sh:106`

### Why it surfaced now

`detect-changes` lists `docker/common-versions` as a trigger for every platform. PRs that only touch a single platform Dockerfile skip the CUDA jobs entirely, so the bug stays hidden; any PR that edits the shared `common-versions` file turns all platform builds on and hits it immediately.

### Suggested fix

Clear `USE_SCCACHE` in both degrade branches of `setup-sccache.sh`. The script is sourced, so a single change propagates to all consumers, and `build-nvshmem.sh:32` already uses `export USE_SCCACHE="false"` for the same purpose.

yuanwu2017:
Fixes #2223

### Problem

`docker/scripts/common/setup-sccache.sh` degrades gracefully when sccache cannot start, but the degrade path leaves the build broken in two separate ways.

**1. `USE_SCCACHE` is never cleared.**

The script removes the binary, unsets the CMake launcher variables and returns success — but the input variable still says `true`. Every consumer sources the script and then branches on it, so they still believe sccache is available:

```bash
. /usr/local/bin/setup-sccache
...
if [ "${USE_SCCACHE}" = "true" ]; then
    export CC="sccache gcc" CXX="sccache g++"   # binary was just deleted
fi
```

gdrcopy builds with a plain Makefile rather than CMake, so unsetting the `CMAKE_*_COMPILER_LAUNCHER` variables has no effect on it. The hardcoded `sccache` in `CC` is what actually breaks the build:

```
make[1]: sccache: No such file or directory
make[1]: *** [<builtin>: gdrapi.o] Error 127
make: *** [Makefile:63: lib] Error 2
```

`build-ucx.sh:29` and `build-nixl.sh:32` set `CC` the same way and fail identically. `build-gdrcopy.sh:62`, `build-ucx.sh:62`, `build-nvshmem.sh:147`, `build-lmcache.sh:88` and `build-compiled-wheels.sh:106` call bare `sccache --show-stats` after the binary is gone, which aborts under `set -e`.

**2. Deleting the binary breaks the final image layer.**

`docker/Dockerfile.cuda:493-494` copies sccache out of the builder stage, and its comment states the invariant the degrade path violates:

```dockerfile
# sccache binary is always installed in builder; only configured/started when USE_SCCACHE=true
COPY --from=builder /usr/local/bin/sccache /usr/local/bin/sccache
```

Once the builder stage stops failing, the build gets one step further and dies here instead:

```
ERROR: failed to build: failed to solve: failed to compute cache key:
failed to calculate checksum of ref ...: "/usr/local/bin/sccache": not found
```

Together these fail every CUDA-family image build — both architectures, both base OSes, plus the debug, gb200 and aws variants.

**3. The install-time smoke test uses the binary as a success proxy.**

`install-sccache.sh:40` decides whether to run its verification compile by testing whether the binary still exists:

```bash
source /usr/local/bin/setup-sccache

# verify sccache works with a simple test (only if binary still exists after setup)
if [ -x /usr/local/bin/sccache ]; then
    echo "int main() { return 0; }" | sccache gcc -x c - -o /dev/null
```

That check is only meaningful because of the `rm -f` in (2), and it contradicts the file's own header comment: "Always install sccache binary (small download) so `COPY --from=builder` works in downstream stages regardless of `USE_SCCACHE` setting." With the binary preserved, the smoke test runs even though the cache is unreachable and aborts the builder stage under `set -Eeux`:

```
+ sccache gcc -x c - -o /dev/null
sccache: error: Server startup failed: ... InvalidAccessKeyId
#24 ERROR: process "/bin/sh -c /tmp/install-sccache.sh ..." did not complete successfully: exit code: 2
```

### Fix

Export `USE_SCCACHE=false` in both fallback branches, stop deleting the binary, and gate the smoke test on the flag rather than on the binary's existence.

The script is sourced, so the export covers every consumer in one place, and it matches the pattern `build-nvshmem.sh:32` already uses when it opts out of sccache. Every reference to the `sccache` binary in every consumer script is inside an `if [ "${USE_SCCACHE}" = "true" ]` guard, so the `rm -f` no longer protects against anything — its stated purpose ("so meson/cmake can't accidentally try to use it") is already covered by the unset launchers plus the disabled flag.

Only the degrade paths and the one guard are touched; the successful-sccache path is unchanged.

### Verification

Ran the real `build-gdrcopy.sh` in a container with no working sccache, which is exactly the CI condition:

| | exit code | output |
|---|---|---|
| before | `2` | `make[1]: sccache: No such file or directory` / `Error 127` |
| after | `0` | `cc -shared -Wl,-soname,libgdrapi.so.2 -o libgdrapi.so.2.5 ...` → installed |

And confirmed the resulting state after sourcing the script with a stub sccache that fails to start the way CI's does:

```
+ '[' true = true ']'
+ source /usr/local/bin/setup-sccache
++ /usr/local/bin/sccache --start-server
sccache: error: Server startup failed: InvalidAccessKeyId
Warning: sccache failed to start, continuing without cache
++ export USE_SCCACHE=false
+ '[' false = true ']'
---- results ----
exit code: 0
binary PRESENT
```

### Why this surfaced now

`detect-changes` lists `docker/common-versions` as a trigger for every platform. PRs touching only one platform's Dockerfile skip the CUDA jobs entirely, so the bug stays hidden; any PR that edits the shared `common-versions` file turns all platform builds on and hits it immediately. #1230 is the PR that ran into it.

The three failure modes masked each other: the builder stage died at gdrcopy before reaching `COPY --from=builder`, and the smoke test was skipped only because the binary had been deleted. The CI credentials for the `vllm-nightly-sccache` bucket currently return `InvalidAccessKeyId`, so the degrade path is taken on every run rather than occasionally.

### DIFF
--- a/docker/scripts/common/setup-sccache.sh
+++ b/docker/scripts/common/setup-sccache.sh
@@ -38,6 +38,8 @@ if [ "${USE_SCCACHE}" = "true" ]; then
         # Remove sccache binary so meson/cmake can't accidentally try to use it
         rm -f /usr/local/bin/sccache || true
         unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
+        # Sourcing scripts branch on this to decide whether to put sccache in CC/CXX
+        export USE_SCCACHE="false"
         return 0
     fi
 
@@ -47,6 +49,7 @@ if [ "${USE_SCCACHE}" = "true" ]; then
         # Remove sccache binary so meson/cmake can't accidentally try to use it
         rm -f /usr/local/bin/sccache || true
         unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
+        export USE_SCCACHE="false"
         return 0
     fi
 
--- a/docker/scripts/common/setup-sccache.sh
+++ b/docker/scripts/common/setup-sccache.sh
@@ -35,8 +35,6 @@ if [ "${USE_SCCACHE}" = "true" ]; then
 
     if ! /usr/local/bin/sccache --start-server; then
         echo "Warning: sccache failed to start, continuing without cache" >&2
-        # Remove sccache binary so meson/cmake can't accidentally try to use it
-        rm -f /usr/local/bin/sccache || true
         unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
         # Sourcing scripts branch on this to decide whether to put sccache in CC/CXX
         export USE_SCCACHE="false"
@@ -46,8 +44,6 @@ if [ "${USE_SCCACHE}" = "true" ]; then
     if ! /usr/local/bin/sccache --show-stats >/dev/null 2>&1; then
         echo "Warning: sccache not responding properly, disabling cache" >&2
         /usr/local/bin/sccache --stop-server 2>/dev/null || true
-        # Remove sccache binary so meson/cmake can't accidentally try to use it
-        rm -f /usr/local/bin/sccache || true
         unset CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER CMAKE_CUDA_COMPILER_LAUNCHER
         export USE_SCCACHE="false"
         return 0
--- a/docker/scripts/cuda/builder/install-sccache.sh
+++ b/docker/scripts/cuda/builder/install-sccache.sh
@@ -36,8 +36,8 @@ if [ "${USE_SCCACHE}" = "true" ]; then
     # shellcheck source=/dev/null
     source /usr/local/bin/setup-sccache
 
-    # verify sccache works with a simple test (only if binary still exists after setup)
-    if [ -x /usr/local/bin/sccache ]; then
+    # setup-sccache clears USE_SCCACHE when it could not reach the cache
+    if [ "${USE_SCCACHE}" = "true" ]; then
         echo "int main() { return 0; }" | sccache gcc -x c - -o /dev/null
         echo "sccache installation and S3 connectivity verified"
     fi


yuanwu2017:
The three commits here fix three symptoms of a single design problem, and each one only became visible after the previous was fixed. Recording the sequence since the intermediate CI runs on this PR look like the fix kept failing:

| commit | symptom | run |
|---|---|---|
| `37e7243` | `build-gdrcopy.sh` → `make[1]: sccache: No such file or directory` / `Error 127` | [31362149292](https://github.com/llm-d/llm-d/actions/runs/31362149292) |
| `207f340` | final layer → `COPY --from=builder /usr/local/bin/sccache: not found` | [31364574234](https://github.com/llm-d/llm-d/actions/runs/31364574234) |
| `aacbf08` | `install-sccache.sh` smoke test → `exit code: 2` | [31369626636](https://github.com/llm-d/llm-d/actions/runs/31369626636) |

The common cause is that the fallback signalled "sccache is unavailable" by mutating global state — deleting `/usr/local/bin/sccache` — rather than by clearing the flag that every consumer actually branches on. Two other places had then come to depend on that deletion as a side channel: `Dockerfile.cuda:494` expects the binary to survive, and `install-sccache.sh:40` used its absence as a proxy for "setup failed". Making the state explicit removes the need for all three.

I checked for further occurrences: `install-sccache.sh:40` was the only place probing for the binary's existence. `build-lmcache.sh:53-54` (`command -v sccache`) and `Dockerfile.cuda:548` (`sccache --show-stats`) are both already inside `USE_SCCACHE` guards, so they are unaffected.

Current status on [31370374603](https://github.com/llm-d/llm-d/actions/runs/31370374603):

```
success      cuda-build-llm-d-debug (amd64, rhel/ubi9)
success      cuda-build-llm-d       (arm64, ubuntu)
success      gb200-build-llm-d
success  cuda-build-llm-d       (amd64, rhel/ubi9)
success  cuda-build-llm-d       (amd64, ubuntu)
success  cuda-build-llm-d       (arm64, rhel/ubi9)
success  aws-build-llm-d        (amd64, rhel/ubi9)
```

Both OS variants and both architectures have now completed the full builder + final image path, which is what was failing before.

One thing worth separating out: this PR makes the fallback behave correctly, but the fallback should not be running at all. The sccache S3 credentials are rejected on every build (`InvalidAccessKeyId`), so the CUDA images compile with a completely cold cache — the builder stage takes around 30 minutes, of which NVSHMEM alone is ~15. That is filed separately as #2225. Fixing the credentials would hide this bug again rather than resolve it, so I would suggest not treating #2225 as a reason to close this.

Gregory-Pereira (approved):
great catch
/lgtm

PR #2208, opened 2026-08-07
fix(autoscaling): make the queue-based KEDA+EPP nightly deploy and validate reliably

mamy-CS:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it:**

The queue-based KEDA+EPP autoscaling nightly (added in #2192) fails on the shared
OCP GPU cluster. This fixes the deploy script and workflow so the path installs,
discovers the EPP, and validates cleanly. Four independent issues:

1. **Unique release name** (deploy script). Installing the router as
   `optimized-baseline` collides with existing optimized-baseline InferencePools on
   the shared cluster, sending the EPP into a hot reconcile loop (NOT_SERVING).
   Renamed the release to `keda-epp-queue`; the EPP Service is discovered at runtime
   and the ScaledObject query is templated to it.

2. **Pin the router chart to a released version** (deploy script). `guides/env.sh`
   pins every guide to the mutable `v0` tag, which picked up llm-d-router#1681 and
   stopped creating the EPP `-epp` Service with flowControl/monitoring — `helm
   install` reports `deployed` but the Service never appears. Pinned to the released
   `v0.9.0` (predates #1681; verified it creates the Service). Repo-wide env.sh fix
   tracked in #2207; chart bug in llm-d-router#2323.

3. **Poll for the EPP Service** (deploy script). It isn't present the instant `helm
   install` returns; poll up to 5m and dump `all,inferencepool` on failure.

4. **Fix stale `gateway_host`** (workflow). e2e-validate.sh curls
   `http://<gateway_host>:80/v1/models` and exits 1 if unreachable. The rename makes
   the EPP Service `keda-epp-queue-epp`, so `gateway_host` now points there (was
   `optimized-baseline-epp`).

5. **Warm up the EPP before the gate** (deploy script). The v0.9.0 EPP creates the
   per-model `flow_control_queue_size` / `request_running` series lazily on the first
   request, not at idle — so the deploy sends a few `/v1/completions` after the decode
   is Ready, before the Fallback gate, or the trigger errors on an empty series.

6. **Fix the ClusterRoleBinding subject namespace** (deploy script). kustomize's
   `namespace` transformer doesn't rewrite a CRB subject's namespace when the inner
   `ocp/` overlay already set it, so the metrics-reader SA was bound in the wrong
   namespace → KEDA's Thanos queries 403 → `TriggerError`. Now rewritten explicitly.

**Which issue(s) this PR resolves:**

Related to #2207, llm-d-router#2323

[...]

PR #2192, opened 2026-08-06
feat(autoscaling): add queue-based KEDA+EPP autoscaling nightly (OCP GPU)

mamy-CS:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Adds a nightly test lane for the **queue-based KEDA + EPP** autoscaling guide (`guides/workload-autoscaling/README.hpa-epp.md`), which has no CI today. The existing `workload-autoscaling` nightly tests the legacy WVA path (`wva_desired_replicas`); this lane tests the EPP-direct path. KEDA scales the decode Deployment straight from `llm_d_epp_flow_control_queue_size`, with **no WVA controller**.

This is the "wiring / signal-liveness" gate: it validates that the guide deploys, the EPP emits the queue metric with the right labels, user-workload-monitoring scrapes it into Thanos, KEDA authenticates to Thanos, and the HPA is driven by the real signal (`Fallback=False`) the silent-failure class (401 / wrong CA / label drift / missing metric) that breaks autoscaling invisibly. Load-driven scale-up actuation is a planned follow-up as V2.

Files:

- **`scripts/nightly-deploy-ocp-keda-epp.sh`** — the EPP-direct sibling of `nightly-deploy-ocp.sh`. Installs the EPP router as helm release `optimized-baseline` (matching the guide → EPP Service `optimized-baseline-epp`) with the `monitoring` + `keda-epp-queue/router.values.yaml` values (the latter enables `featureGates: [flowControl]`, the source of the queue metric), applies `keda-epp-queue/ocp`, discovers the EPP Service name at runtime and templates both trigger queries' `namespace`/`service`/`model_name` to it, labels the namespace for UWM scraping, starts the decode Deployment at 1 replica, waits for it to be ready (the queue metric is per-model and only exists once the EPP has a ready backend endpoint), then gates on `Fallback=False`. `maxReplicaCount` is capped at the GPU budget (2).
- **`nightly-e2e-workload-autoscaling-keda-epp-ibm-acc-gpu-vllm-x.yaml`** — the nightly workflow, calling `reusable-nightly-e2e-openshift.yaml@main` with the deploy script as `custom_deploy_script` (no WVA image-build job). Distinct namespace (`llm-d-nightly-keda-epp-queue`), concurrency group, badge (`keda-epp-queue-ocp`), and schedule (`0 5 * * *`) from the WVA lane. Declares the inputs `/test-nightly` passes so it can be triggered by comment.
[...]

PR #2228, opened 2026-08-10
feat(autoscaling): add scale-event validation to the queue-based KEDA+EPP nightly

mamy-CS:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it:**

The queue-based KEDA+EPP autoscaling nightly (added in #2192, hardened in #2208) gates on `Fallback=False` - KEDA is reading a real EPP metric from Thanos, not `spec.fallback`. That proves the signal pipeline is live but not that load actually produces a scaling decision. This adds the Phase-2 scale-event proof.

- **New `e2e-validate-scale-event.sh`** - wired via `e2e-validate.sh --extra-validate scale-event` (the existing hand-off hook that `predicted-latency` / `flow-control` already use; no llm-d-infra change). It discovers the ScaledObject -> Deployment -> KEDA HPA, fires a concurrent burst at the single replica, and asserts `hpa.status.desiredReplicas` 1 -> 2 (scale up) then -> 1 (scale down). It targets the `request_running > 16` trigger, and asserts the autoscaler decision, not the 2nd pod becoming `Ready` - on a contended cluster that pod may stay `Pending`, and the test still passes, keeping it deterministic and gateable.

- **Deploy script** - the ScaledObject kustomize patch also lowers `scaleDown.stabilizationWindowSeconds` 300 -> 60 (nightly overlay only; the guide default stays 300, where the long window guards the vLLM-startup scale-down race) so scale-down is observable within the run.

- **Workflow** - `e2e_validate_args: '--extra-validate scale-event'`.

**Which issue(s) this PR resolves:**

Part of the queue-based KEDA+EPP autoscaling nightly (follows #2192, #2208).

- **Validation**

Calibrated live on OpenShift cluster: the burst drove `request_running` to 30 (threshold 16) -> `desiredReplicas -> 2` in a single KEDA poll -> `-> 1` after the 60s window; the 2nd pod stayed `Pending` throughout and the test passed by design. Adds ~2 min to the nightly. Burst tunables are overridable (`--concurrency` / `--total` / `--max-tokens`).

**Does this PR introduce a user-facing change?**

```release-note
NONE
```

### DIFF
--- a/.github/scripts/e2e/e2e-validate-scale-event.sh
+++ b/.github/scripts/e2e/e2e-validate-scale-event.sh
@@ -1,0 +1,229 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+# -----------------------------------------------------------------------------
+# e2e-validate-scale-event.sh
+# -----------------------------------------------------------------------------
+# Validates the queue-based KEDA + EPP autoscaling path beyond signal-liveness.
+#
+# The deploy script already gates on Fallback=False (KEDA reads a real metric,
+# not spec.fallback). That proves the *wiring*, not that load actually drives a
+# scaling DECISION. This script fires a concurrent burst and asserts the KEDA
+# scale event fired:
+#
+#   baseline desiredReplicas == min  ->  burst  ->  desiredReplicas > min  (scale UP)
+#                                     ->  drain  ->  desiredReplicas == min (scale DOWN)
+#
+# Key property: we assert the autoscaler DECISION (hpa .status.desiredReplicas),
+# NOT that the new pod becomes Ready. On a contended GPU cluster the 2nd replica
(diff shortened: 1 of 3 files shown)


mamy-CS:
/test-nightly workload-autoscaling-keda-epp-ibm-acc-gpu-vllm-x

lionelvillard:
/test-nightly workload-autoscaling-keda-epp-ibm-acc-gpu-vllm-x

clubanderson:
### Nightly E2E Result: `workload-autoscaling` (OpenShift)

| | |
|---|---|
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/31414427591) |
| Ref | `8b1abda` |

lionelvillard (approved):

PR #2229, opened 2026-08-10
fix(autoscaling): enable flowControl via helm in the saturation guide

mamy-CS:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it:**

The `[Experimental]` saturation-based autoscaling guide cannot enable EPP flow control as written, so `llm_d_epp_flow_control_pool_saturation` is never emitted and the guide is non-functional. `keda-epp-saturation/base/epp-endpoint-picker-config.yaml` is an `apiVersion: router.llm-d.ai/v1beta1, kind: EndpointPickerConfig` resource applied via `kubectl apply -k`, (the `EndpointPickerConfig` type is an `llm-d.ai/v1alpha1` config-file schema, not a cluster CRD). So the guide's `kubectl apply -k .../keda-epp-saturation/ocp` fails with `no matches for kind "EndpointPickerConfig" in version "router.llm-d.ai/v1beta1"`.

Every other guide (queue, optimized-baseline, flow-control, pd-disaggregation, wide-ep, predicted-latency, …) enables flow control via a helm `pluginsCustomConfig` value. This PR aligns the saturation guide with that pattern:

- Adds `keda-epp-saturation/router.values.yaml` (`featureGates: [flowControl]` + the four optimized-baseline plugins), layered onto the router with a `helm upgrade`.
- Removes the `base/epp-endpoint-picker-config.yaml` and drops it from `base/kustomization.yaml` (base is now ScaledObject + TriggerAuthentication only, like the queue path).
- Updates `README.epp-keda-saturation.md`: adds the `helm upgrade` step, fixes the prerequisite wording, and removes the stale instruction

**Which issue(s) this PR resolves:**

Fixes the experimental saturation-based autoscaling guide.

**Validation**

Validated end-to-end on OpenShift (pokprod): `kubectl apply -k .../ocp` now applies clean, the router emits `pool_saturation{inference_pool=<release>}`, and KEDA reads it from Thanos (ScaledObject `Ready`, `Fallback=False`). The `router.values.yaml` mirrors `keda-epp-queue/router.values.yaml` (flow control emits both `pool_saturation` and `queue_size`).

**Does this PR introduce a user-facing change?**

```release-note
Fix the experimental saturation-based autoscaling guide so EPP flow control is enabled via a helm values file instead of a non-existent EndpointPickerConfig, so pool_saturation is emitted and the guide works.
```


guides/workload-autoscaling/optimized-baseline-autoscaling/keda-epp-saturation/router.values.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+router:
+  epp:
+    pluginsConfigFile: "optimized-baseline-keda-epp-saturation-plugins.yaml"
+    pluginsCustomConfig:
+      optimized-baseline-keda-epp-saturation-plugins.yaml: |
+        apiVersion: llm-d.ai/v1alpha1
+        kind: EndpointPickerConfig
+        featureGates:
+        - flowControl
```
asm582, line 17: I dont think we need a new file, we need to use whatever plugins optimized-baseline-plugins use and just patch that enables flow-control

asm582 (changes requested):

guides/workload-autoscaling/optimized-baseline-autoscaling/keda-epp-saturation/router.values.yaml
mamy-CS: the EPP just takes `--config-file`, no separate flag. Updated to reuse optimized-baseline's actual plugins.

asm582: Hmm, don't think we should maintain another helm file; optimized plugins can change in the router. Why not just discover and patch with flow-control?

mamy-CS: I did try the patch, but it was unreliable. The running EPP came up with zero flow-control metrics even on a fresh pod. The helm `pluginsCustomConfig` path enables it reliably (validated: pool_saturation → ScaledObject Ready / Fallback=False). We are reusing optimized-baseline's own plugins unchanged and we only add `flowControl`, no new plugin set. And it matches the flow-control and keda-epp-queue guides (the only other flowControl guides). SInce flowcontrol is going to be enabled by default next release the worry with drift should be ok.

PR #2239, opened 2026-08-11
Support tiered prefix cache on Intel XPU with vllm native offloading connector with file system

XinyuYe-Intel:
Support tiered prefix cache on Intel XPU with vllm native offloading connector with file system.

### DIFF
--- a/guides/tiered-prefix-cache/modelserver/xpu/vllm/native/fs/base/patch-vllm.yaml
+++ b/guides/tiered-prefix-cache/modelserver/xpu/vllm/native/fs/base/patch-vllm.yaml
@@ -1,0 +1,42 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: decode
+spec:
+  template:
+    spec:
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          args:
+            - "Qwen/Qwen3-32B"
+            - "--dtype=bfloat16"
+            - "--block-size=64"
+            - "--tensor-parallel-size=4"
+            - "--gpu-memory-utilization=0.85"
+            - "--kv-transfer-config"
+            - '{"kv_connector":"OffloadingConnector","kv_role":"kv_both","kv_connector_extra_config":{"spec_name":"TieringOffloadingSpec","cpu_bytes_to_use":107374182400,"block_size":256,"secondary_tiers":[{"type":"fs","root_dir":"/mnt/files-storage","n_read_threads":16,"n_write_threads":16}]}}'
+            - "--max-num-seq=128"
+            - "--uvicorn-log-level=warning"
+          env:
+            - name: HF_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: llm-d-hf-token
+                  key: HF_TOKEN
+          resources:
+            limits:
+              cpu: '8'
+              memory: 140Gi
+            requests:
+              cpu: '8'
+              memory: 140Gi
+            claims:
+              - name: intel-decode-claim
+          volumeMounts:
+            - name: files-storage
+              mountPath: /mnt/files-storage
+      volumes:
+        - name: files-storage
+          persistentVolumeClaim:
+            claimName: llm-d-kv-cache-storage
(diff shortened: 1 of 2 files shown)


vMaroon:
cc @kfirtoledo @effi-ofer

guides/tiered-prefix-cache/modelserver/xpu/vllm/native/fs/base/patch-vllm.yaml
```diff
@@ -0,0 +1,29 @@
...
+          args:
+            - "Qwen/Qwen3-8B"
+            - "--dtype=bfloat16"
+            - "--block-size=64"
+            - "--tensor-parallel-size=1"
+            - "--max-num-seqs=128"
+            - "--max-model-len=20480"
+            - "--uvicorn-log-level=warning"
+            - "--kv-transfer-config"
+            - '{"kv_connector":"OffloadingConnector","kv_role":"kv_both","kv_connector_extra_config":{"spec_name":"TieringOffloadingSpec","cpu_bytes_to_use":10737418240,"block_size":256,"secondary_tiers":[{"type":"fs","root_dir":"/mnt/files-storage","n_read_threads":16,"n_write_threads":16}]}}'
```
Gregory-Pereira: I wonder if its possible / more efficient to see if we can paramaterize the value of `cpu_bytes_to_use` in an ENV var that we can call to make this more dynamic

guides/tiered-prefix-cache/modelserver/xpu/vllm/native/fs/base/patch-vllm.yaml
```diff
@@ -0,0 +1,29 @@
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: modelserver
+          args:
+            - "Qwen/Qwen3-8B"
+            - "--dtype=bfloat16"
+            - "--block-size=64"
+            - "--tensor-parallel-size=1"
```
Gregory-Pereira: default arg - drop

Gregory-Pereira (commented):
some thoughts

guides/tiered-prefix-cache/modelserver/xpu/vllm/native/fs/base/patch-vllm.yaml
XinyuYe-Intel: An env var value still lives in the YAML, so it doesn't reduce config. The value that's actually meaningful is override-ability — but since vLLM doesn't expand env vars inside the --kv-transfer-config JSON, making it truly dynamic requires wrapping the command in a shell that interpolates ${CPU_BYTES_TO_USE}. Happy to do that if we want overlay-level overrides; otherwise I'd suggest keeping the literal for readability and revisiting if/when we need per-deployment tuning.

guides/tiered-prefix-cache/modelserver/xpu/vllm/native/fs/base/patch-vllm.yaml
XinyuYe-Intel: I'll change the model to Qwen3-32B to align with the cpu case, and tensor-parallel-size will become 4.

XinyuYe-Intel:
Hi @vMaroon @Gregory-Pereira @kfirtoledo @effi-ofer , could you please help review this PR?

guides/tiered-prefix-cache/modelserver/xpu/vllm/native/fs/base/patch-vllm.yaml
```diff
@@ -0,0 +1,44 @@
...
+  name: decode
+spec:
+  template:
+    spec:
+      containers:
+        - name: modelserver
+          command: ["/bin/bash", "-c"]
+          args:
+            - |
+              exec vllm serve \
```
yuanwu2017: Please refer to https://github.com/llm-d/llm-d/blob/main/guides/pd-disaggregation/modelserver/gpu/vllm/base/patch-decode.yaml#L11
Don't use the command in args.

XinyuYe-Intel: Changed the style.

maugustosilva (approved):

PR #2250, opened 2026-08-11
[ci/test]: Update gke workflows to use the cluster in us east5 

rahulgurnani:
**Problem**
Presently, the nightly pipeline executes against the cluster in us south1. Due to stockouts in that cluster, the pipeline has been failing recently.

**Proposed Fix**
Start using cluster in us east5, since the cluster does have reservation. This should help avoid stockout issues in future.

This change should fix the workflows that run on H100s. P/D and wide ep can be fixed separately since they need H200s.


rlakhtakia:
lgtm

maugustosilva (approved):

ahg-g:
isn't this already the same default cluster we have in https://github.com/llm-d/llm-d-infra/blob/main/.github/workflows/reusable-ci-nightly-benchmark.yaml ?

rahulgurnani:
> isn't this already the same default cluster we have in https://github.com/llm-d/llm-d-infra/blob/main/.github/workflows/reusable-ci-nightly-benchmark.yaml ?

yeah, that's correct. the cluster in east5 is already being used in these workflows.  Logs from pods show CUDA out of memory (torch.OutOfMemoryError) during model warmup and KV-cache allocation. It may require us to update the deployment config yamls. Investigating the OOM further.

In the meantime, raising a revert for this change.

PR #2252, opened 2026-08-11
guides(workload-autoscaling): fix saturation guide EPP flow control enablement

### LINKED ISSUE #2248: saturation autoscaling guide: EPP flow control never enabled (kubectl apply fails)
The experimental saturation-based autoscaling guide (`guides/workload-autoscaling/README.epp-keda-saturation.md`) cannot enable EPP flow control, so `llm_d_epp_flow_control_pool_saturation` is never emitted and the guide is non-functional.

`kubectl apply -k .../keda-epp-saturation/ocp` fails:

```
no matches for kind "EndpointPickerConfig" in version "router.llm-d.ai/v1beta1"
```

**Fix direction:** enable the `flowControl` feature gate through the router (reusing the optimized-baseline EPP plugins) instead of applying a config document - enabling the gate loads default flow-control policies, and `pool_saturation` emits once the pool has a ready endpoint. 

/kind bug

cc: @LukeAVanDrie closed this pr https://github.com/llm-d/llm-d/pull/2229 so guides stay consistent. Thanks.

asm582:
## What

Fixes the experimental saturation-based autoscaling guide, which could not enable EPP flow control. `kubectl apply -k .../keda-epp-saturation/ocp` failed with:

```
error: resource mapping not found for name: "optimized-baseline-epp-config" ...
no matches for kind "EndpointPickerConfig" in version "router.llm-d.ai/v1beta1"
ensure CRDs are installed first
```

Fixes #2248

## Why it failed

`base/epp-endpoint-picker-config.yaml` tried to apply the EPP config as a cluster resource. `EndpointPickerConfig` is not a Kubernetes API type — per [`docs/api-reference/endpointpickerconfig.md`](https://github.com/llm-d/llm-d/blob/main/docs/api-reference/endpointpickerconfig.md), it is the EPP binary's own config schema, delivered as a key inside the `<release>-epp` ConfigMap and read via `--config-file`. There is no CRD to apply it against.

The removed file was wrong in three further ways:

| | File declared | Actual |
|---|---|---|
| Group/version | `router.llm-d.ai/v1beta1` | `llm-d.ai/v1alpha1` |
| Schema | `spec.featureGates`, `spec.plugins.scheduling[].name` | top-level `featureGates`, `plugins[].type`, `schedulingProfiles[].plugins[].pluginRef` |
| Plugin set | `queue-scorer` / `kv-cache-utilization-scorer` / `prefix-cache-scorer` / `no-hit-lru-scorer` | this guide's set: `approx-prefix-cache-producer` / `inflight-load-producer` / `prefix-cache-affinity-filter` / `token-load-scorer` |

## Approach

Per the fix direction in #2248, this **reuses the existing EPP ConfigMap created by the llm-d router guide** rather than applying a config document. The guide now appends the feature gate to that ConfigMap with the two commands documented in Configure step 2 — read the current plugins config, then merge-patch `featureGates: [flowControl]` into that one key and restart the EPP.

Deliberately no Helm step: the patch is a small, in-place edit against whatever the optimized-baseline guide already deployed, so it works regardless of which router chart or mode was installed. The guide notes that Helm owns the ConfigMap, so a later `helm upgrade` reverts the patch and the two commands should be re-run.

The file is removed from the kustomize base, leaving only the KEDA resources that are genuine API objects — matching the sibling `keda-epp-queue/base`.

## Two additional query-default fixes

Both surfaced while verifying the trigger path against Thanos:

- **`inference_pool` was `"default"`.** The label carries the InferencePool name (the EPP's `--pool-name`), so the saturation trigger matched no series. Because KEDA suppresses trigger errors and serves `fallback` replicas, this looked healthy while never scaling.
- **Neither query pinned `namespace`.** Against a cluster-wide store such as OpenShift's Thanos Querier, the unpinned `sum(llm_d_epp_request_running{model_name=...})` aggregates every EPP on the cluster — on a shared cluster the deployment would scale on other tenants' traffic. Both queries now pin `namespace`, and the guide flags it as a value to edit.

## Testing

Verified on an OpenShift cluster against a live `optimized-baseline` deployment. All cluster interactions were **dry-run or read-only** — no resource was mutated.

**1. The originally failing command** — `oc apply -k .../keda-epp-saturation/ocp --dry-run=server`:

```
serviceaccount/keda-epp-metrics-reader unchanged (server dry run)
clusterrolebinding.rbac.authorization.k8s.io/keda-epp-metrics-reader-monitoring-view unchanged (server dry run)
secret/keda-epp-metrics-reader-token unchanged (server dry run)
scaledobject.keda.sh/optimized-baseline-nvidia-gpu-vllm-decode-saturation configured (server dry run)
triggerauthentication.keda.sh/prometheus-auth unchanged (server dry run)
```

No mapping error. Both the `k8s` and `ocp` overlays build cleanly, and `router.llm-d.ai/v1beta1` no longer appears anywhere in the repo.

**2. The ConfigMap patch from Configure step 2** — `--dry-run=server` against the live ConfigMap:

- Server accepted the patch.
- `featureGates: [flowControl]` present exactly once; the document parses as valid YAML.
- The ConfigMap's other keys (`default-plugins.yaml`, `payload-agnostic.yaml`, `launch-flags`) were left untouched, confirming the merge patch only rewrites the one key.

**3. Both trigger queries against Thanos Querier**, using the guide's own `keda-epp-metrics-reader` ServiceAccount token — the same auth path KEDA uses:

| Query | Result |
|---|---|
| `inference_pool="default"` (old default) | `[]` — empty, trigger never fires |
| `inference_pool="<pool name>"` (new default) | resolves |

The running-requests query returns an empty result before any traffic, since `llm_d_epp_request_running` is only created once the EPP has handled a request. KEDA's default `ignoreNullValues=true` reads that as 0, holding at `minReplicaCount`; this is now documented inline.

## Note for reviewers

The merge patch replaces the plugins key wholesale, so the document a user sends must be their own config plus the gate. The guide carries a warning about this: the example shown is the plugin list a stock optimized-baseline install produces, and anything extra in a given deployment (a `metrics-data-source` plugin, `parameters` blocks, the TensorRT-LLM plugin set) must be carried over or the patch drops it. This was observed on the test cluster, whose EPP config had an extra `metrics-data-source` plugin.

/kind bug


LukeAVanDrie:
@asm582 sketched out some longer term improvements here, let me know what you think: https://github.com/llm-d/llm-d-router/issues/2358

asm582:
> @asm582 sketched out some longer term improvements here, let me know what you think: [llm-d/llm-d-router#2358](https://github.com/llm-d/llm-d-router/issues/2358)

Thanks, @LukeAVanDrie. Looks good.

mamy-CS (approved):
Tested end-to-end on OpenShift: starting from a plain optimized-baseline install with a ready endpoint, flow control is off (pool_saturation absent). After the guide's kubectl patch + rollout restart, `pool_saturation{inference_pool="optimized-baseline"}` is emitted and reaches Thanos with the expected inference_pool/namespace/service labels, exactly what the ScaledObject queries. Guide works as written.

One optional nit: might be worth a one-line note that the CM patch is out-of-band from Helm, so a later helm upgrade of the router reverts it (re-run the patch). Otherwise LGTM.

maugustosilva (dismissed):

maugustosilva (approved):

PR #2263, opened 2026-08-12
guides: pin router chart to released v0.9.0

### LINKED ISSUE #2207: guides: pin router chart to released v0.9.0 instead of the mutable v0 tag (env.sh)
## What / Why

[`guides/env.sh:8`](https://github.com/llm-d/llm-d/blob/main/guides/env.sh#L8) sets `ROUTER_CHART_VERSION=v0`, the **mutable rolling** chart tag. Every guide (and the workload-autoscaling nightly) inherits this single source of truth, so they all install bleeding-edge chart and pick up regressions the moment they merge upstream before any release, with no way to pin or reproduce.

This just bit us. [llm-d-router#1681](https://github.com/llm-d/llm-d-router/pull/1681) (merged 2026-08-06) shipped straight into `v0` and broke `llm-d-router-standalone`: the EPP `-epp` Service is never created when `flowControl`/`monitoring` are enabled (`helm template` renders it and `helm install` reports `deployed`, but it never appears at runtime). That makes EPP metrics unscrapeable and breaks the queue-based KEDA+EPP autoscaling guide and its new nightly. The same #1681 also caused the EPP `--config-file` CrashLoop (fixed by [llm-d-router#2322](https://github.com/llm-d/llm-d-router/pull/2322)) and overlaps with the shell-var expansion fix in #2177. The underlying chart bug is tracked upstream as [llm-d-router#2323](https://github.com/llm-d/llm-d-router/issues/2323).

## Evidence - the released artifact is fine, only the rolling tag is broken

Identical standalone install (base + optimized-baseline + monitoring + flowControl values):

| chart version | EPP `-epp` Service | pod |
| --- | --- | --- |
| `v0` (rolling) | **not created** | Ready 2/2, `deployed` |
| `v0.9.0` (release) | **created** | Ready 2/2, `deployed` |

`git compare v0.9.0...main` confirms #1681 is not in `v0.9.0`.

## Proposal

Pin to the latest released chart in `guides/env.sh`, and bump deliberately on each router release:

```diff
-export ROUTER_CHART_VERSION=v0
+export ROUTER_CHART_VERSION=v0.9.0
```

This makes guide walkthroughs and CI reproducible instead of silently tracking `main`, and aligns standalone with how the `llm-d-router-gateway` path is already pinned to `v0.9.0`.

## Safety check (no guide depends on an unreleased feature)

Verified before proposing repo-wide: only two values keys were added on `main` since `v0.9.0` `additional_addresses` (standalone) and `podAnnotations` (routerlib) and **no guide sets either**. The chart features guides actually rely on (`featureGates: [flowControl]`, monitoring) are present and working in `v0.9.0`.

/kind bug

yiwxng:
env.sh set ROUTER_CHART_VERSION=v0, a rolling tag, so guides installed
whatever landed upstream last. llm-d-router#1681 went into v0 and broke EPP

service creation with flowControl/monitoring enabled.
pinning to a released tag makes guides and CI reproducible. env.sh is the
only assignment; everything else reads from it.

note v0.10.0-rc.1 was cut yesterday (pre-release), so this will need a bump when it goes stable.

fixes #2207

### DIFF
--- a/guides/env.sh
+++ b/guides/env.sh
@@ -5,7 +5,9 @@
 
 export REPO_ROOT=${REPO_ROOT:-$(realpath "$(git rev-parse --show-toplevel 2>/dev/null)" 2>/dev/null)}
 export GAIE_VERSION=v1.5.0
-export ROUTER_CHART_VERSION=v0
+# Pinned to a released chart tag rather than the mutable `v0` channel so guide
+# walkthroughs and CI are reproducible. Bump deliberately on each router release.
+export ROUTER_CHART_VERSION=v0.9.0
 export ROUTER_EPP_VERSION=main
 # GitHub release tag for CRD manifests. Distinct from ROUTER_CHART_VERSION:
 # the chart channel (v0) floats on the OCI registry and has no matching


diegocastanibm:
Thanks for digging into this, and for the clear before/after evidence in #2207.

That said, I'm going to close this without merging, because of where the pinning belongs in our release process rather than anything wrong with the change itself.

`main` floats on the `v0` chart channel deliberately. `main` is the development branch and is expected to track the newest chart; version pinning happens at release time, on the release branch, where all the component references are moved to the tags that were actually shipped.

PR #2258, opened 2026-08-12
add routing name

diegocastanibm:
Rename

PR #2268, opened 2026-08-13
Stale workflow filename left behind in TPU consolidate-status after nightly rename

weizhoublue:
PR #2258 renamed  `nightly-e2e-precise-prefix-cache-gke-acc-tpu-vllm-x.yaml` to `nightly-e2e-precise-prefix-cache-routing-gke-acc-tpu-vllm-x.yaml`
but left two stale references to the old filename in its companion consolidate-status workflow. 

As a result, the daily scheduled status check queries a non-existent workflow and permanently fails to find any success records for that lane.


weizhoublue:
@diegocastanibm 
Hi , I submit this follow-up PR for your #2258
Please help review, thanks

maugustosilva (approved):

PR #2270, opened 2026-08-13
Fix oneCCL init failure for tiered-prefix-cache on Intel XPU

yuanwu2017:
The XPU decode pod crash-looped during standup with:
  oneCCL: atl_ofi_comm.cpp:232 init_transport: EXCEPTION: failed to initialize ATL
  CCL_ERROR| atl_ofi_helper.cpp:1118 fi_getinfo error: ret -61, providers 0

With tensor-parallel-size=4 the vLLM XPU worker initializes the xccl/oneCCL backend, which defaults to the OFI transport. The bundled libfabric in the container exposes no providers because FI_PROVIDER_PATH is unset, so ATL initialization fails and the engine core never starts.

Point libfabric at the Intel MPI provider directory shipped in the image and select the tcp provider. Verified on a kind cluster with Intel XPUs: the decode pod now reaches Ready with tensor-parallel-size=4 and the OffloadingConnector.


yuanwu2017:
### Validation

Re-ran the Intel XPU nightly against this PR's code and it now passes:

- ✅ **Nightly - Tiered Prefix Cache CPU Offloading E2E (Intel XPU)** (`native`, `tensor-parallel-size=4`, `OffloadingConnector`): https://github.com/llm-d/llm-d/actions/runs/31682027641

Before this fix the decode pod died in `CrashLoopBackOff` during engine-core init:

```
CCL_WARN| value of CCL_ATL_TRANSPORT changed to be ofi (default:mpi)
CCL_ERROR| atl_ofi_helper.cpp:1118 atl_ofi_get_prov_list: fi_getinfo error: ret -61, providers 0
RuntimeError: oneCCL: atl_ofi_comm.cpp:232 init_transport: EXCEPTION: failed to initialize ATL
  -> vllm/v1/worker/xpu_worker.py:105  torch.distributed.all_reduce(torch.zeros(1).xpu())
```

Root cause: with `--tensor-parallel-size > 1` the XPU worker initializes oneCCL, which defaults `CCL_ATL_TRANSPORT=ofi`. The bundled libfabric in the image exposes **zero** providers unless `FI_PROVIDER_PATH` points at the Intel MPI provider directory (`fi_info -l` only lists `ofi_hook_noop` / `off_coll` without it). Other Intel XPU guides are unaffected because they run `tensor-parallel-size=1`, so oneCCL never initializes multi-rank comms.

With the fix the pod reaches `1/1 Running`, all four ranks (`Worker_TP0`..`Worker_TP3`) come up, and the benchmark completes end to end.

### Note on the other two red checks

Both are pre-existing and unrelated to this PR:

1. **CI Guide Dry Run** — a repo-wide regression already present on `main`. An unmodified `main`-based run reproduces the identical 25 failures (`recipes/gateway/*` missing `AgentgatewayParameters`/`ClientTrafficPolicy` CRDs, `wide-ep-lws/*` missing `ComputeDomain`, `workload-autoscaling/*` KEDA, etc.). All three overlays touched by this PR pass.
2. **Nightly - ... LMCache E2E (Intel XPU)** — this PR gets it past the oneCCL crash, but it then fails the smoke test for a separate reason: `_select_modelserver_command()` picks the *first* README command matching `modelserver/<acceleratorBackend>` and ignores `CONNECTOR`, while the guide README hardcodes the Intel XPU overlay paths. So the LMCache nightly always deploys `native/cpu/base` (Qwen3-32B) while the smoke test waits for Qwen3-8B. Tracked separately.

Gregory-Pereira (approved):
/lgtm

PR #2271, opened 2026-08-13
add LMetric routing policy

Spycsh:
Add LMetric https://arxiv.org/html/2603.15202v1#S5.SS1 as an optional policy for precise-prefix-cache-routing and add a router values overlay that replaces the default scoring profile with lmetric-scorer.

Also update the XPU example to use Qwen3-32B with FP8 quantization and TP=2 so it matches the guide's 32B baseline model.

Depends on https://github.com/llm-d/llm-d-router/pull/2351.


Gregory-Pereira:
@xiaojun-zhang Can you review this? Is it ready?

PR #2272, opened 2026-08-13
docs(tiered-prefix-cache): make vLLM native offloading the recommended default

michalmalka:
## Summary
- Add a dedicated quickstart (`guides/tiered-prefix-cache/vllm-native-quickstart.md`) covering only the vLLM native `OffloadingConnector` path (CPU RAM, optionally + filesystem).
- Highlight vLLM native offloading as the recommended default in both the well-lit-path concepts doc and the full guide, linking out to the quickstart.
- Full guide (LMCache, MooncakeStore, SGLang HiCache, TPU, Intel XPU) is unchanged aside from pointing its vLLM-native subsections at the new quickstart to avoid duplicated install commands.

## Test plan
- [ ] Manually click through the new/updated relative links (well-lit-path → quickstart, quickstart → full guide, full guide → quickstart) to confirm they resolve.
- [ ] Confirm the LMCache section's existing anchor link to the vLLM native filesystem heading still resolves in the full guide.
- [ ] Spot-check that quickstart commands match the original guide's commands (no unintended changes).

### DIFF
--- a/docs/well-lit-paths/foundations/tiered-prefix-cache.md
+++ b/docs/well-lit-paths/foundations/tiered-prefix-cache.md
@@ -52,7 +52,7 @@ This increases the **KV-working set size**, growing the **receptive-field** (the
 > CPU KV Cache offloading is low overhead and introduces ~no additional complexity. It can be enabled in almost all deployments. Storage offloading requires additional consideration.
 
 > [!TIP]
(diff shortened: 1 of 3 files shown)


michalmalka:
@vMaroon

zdtsw:
maybe i missed something  importnat
but the current README is already covered  native support as vllm's offering
why we create a new "quick" with similar content? 
if README still refers to the old kv-cache fs_backend, we can mark that as "deprecated", right?

PR #2067, opened 2026-07-17
Well-lit path and guide: P2P KV cache sharing

nilig:
Part of #2066.

Adds the P2P KV cache sharing well-lit path:

- `guides/p2p-kv-cache-sharing/`: guide in the canonical structure -
  configuration, install, mechanism-engaged verification gates,
  `llmdbenchmark` benchmarking steps, and benchmark reports for three
  testbeds: gpt-oss-120b/H200 aggregated (pull-vs-recompute crossover,
  shared-prefix pools, document Q&A), Qwen3-30B P/D agentic (prefill
  pulling decode-generated history: 6.3x median TTFT), and GLM-5.2-FP8
  wide-EP (load-spill payoff: -67% mean TTFT / 2.7x throughput with the
  pull under load-first placement, measured twice on independent builds;
  the precise-affinity pair on the same stack is a mechanism-verified
  null, and the overlay-era grid is retained only as an index-sizing
  failure-mode reproduction).
- Deployment manifests: `modelserver/gpu/vllm/` (aggregated replicas,
  `OffloadingConnector` with a P2P secondary tier, routing sidecar),
  `render/` (standalone tokenizer Service), `router/` values. The guide
  ships `epp-affinity-p2p` as the default arm - the safer general-purpose
  choice across regimes - with `epp-load-p2p` for owner-contended
  workloads (document Q&A) and the load-first pair for the wide-EP
  payoff benchmark. Self-contained overlays; all pass `kubectl
  kustomize`.
- NIXL P/D variant instructions: MultiConnector engine config, sidecar
  `--kv-connector=nixlv2 --enable-p2p-pull`, prefill-leg-only injection
  semantics.
- Engine requirement is a floor, no source overlay: any vLLM nightly at
  or after `nightly-6f91edf9` (2026-07-29) carries the P2P tier
  (vllm#48021) and its robustness fixes (vllm#49671, #49823, #49877);
  the kustomization pins that nightly for reproducibility. Wide-EP
  (multi-pod DP groups) additionally requires per-pod port-base
  compensation (documented in the guide) and a router carrying
  llm-d-router#2233 (merged) and #2234 (in review).
- `docs/well-lit-paths/foundations/enable-p2p-reuse.md`: the
  foundations page - mechanism, regime rules, and the placement law
  (the pull pays on KV the placement layer did not create).
- Calibration recipe for `minCachedTokenDelta`
  (`guides/recipes/router/calibration/`), gated on pulled bytes and an
  uncontaminated control.

[...]

PR #2277, opened 2026-08-13
docs(p2p): apply #2067 review follow-ups

nilig:
Follow-ups from the #2067 review comments:

- Rename the foundations page to "Enable P2P Prefix Cache Sharing" and align every reference - "P2P reuse" was not specific enough as a feature name.
- Call the routing layer the llm-d router in both benchmark-results pages (applies the review suggestion; the gpt-oss page had the same phrase).

Not included: the GLM load-spill payoff refresh - it needs the re-measured numbers and lands separately.

```release-note
NONE
```


vMaroon (approved):

Gregory-Pereira (approved):
/lgtm

PR #2290, opened 2026-08-14
fix: wrong KEDA_VERSION format in download URL

weizhoublue:
If `KEDA_VERSION=v2.20.1`, the download URL becomes `.../download/v2.20.1/keda-v2.20.1.yaml`,
while the actual URL is `.../download/v2.20.1/keda-2.20.1.yaml`.

```
https://github.com/kedacore/keda/releases/download/v2.20.1/keda-2.20.1.yaml
```

fix: set `KEDA_VERSION=2.20.1` and add the `v` prefix only in the tag path


Gregory-Pereira (approved):
/lgtm

PR #2296, opened 2026-08-14
Harden P2P cache-sharing guide and update GLM results

nilig:
## Summary

- mark the P2P KV cache-sharing path experimental and align its source-selection and P/D descriptions with shipped behavior
- replace the recorded-fork GLM result with the repeated C64 complete-policy comparison and a 300-second four-arm observation
- ship the four canonical C64 router configurations and exact-window JSON summaries with campaign provenance and checksums
- document the benchmark vLLM endpoint, pull-attribution limits, decode-capacity validity gate, and well-lit-path evidence

## Measurement scope

The repeated C64 headline compares calibrated approximate routing without P2P against the complete DP-aware precise+P2P policy. Its 9.97% paired-median successful-throughput improvement is a complete-policy result, not a P2P-only effect.

The four-arm result is one 300-second fixed-window observation. Its within-routing comparisons isolate the configured P2P factor for that observation, but do not establish repeated single-factor effect sizes.

## Validation

- markdownlint passes on every changed Markdown file
- yamllint passes on all four C64 EndpointPickerConfig files
- all C64 YAML files parse successfully
- all JSON summaries parse successfully and their documented SHA-256 checksums match
- reported four-arm and repeated-comparison percentage calculations recompute from the committed data
- local Markdown links resolve and git diff --check passes


vMaroon:
@copilot resolve the merge conflicts in this pull request

Gregory-Pereira (approved):
/lgtm

PR #2307, opened 2026-08-16
fix: grant FMA Launcher Pods RBAC Permissions to Advertise Serving State

aavarghese:
1. Similar to https://github.com/llm-d/llm-d-benchmark/pull/1799
2. Enhance verrify.test.sleep_wake section a bit to catch any FMA actuation issues (Today, only for manual deploying users since `kustomize` standup does not parse VERIFY and run it)

### DIFF
--- a/guides/fast-model-actuation/guide.yaml
+++ b/guides/fast-model-actuation/guide.yaml
@@ -124,8 +124,12 @@ verify:
             echo "== sleep/wake cycle ${i} =="
             kubectl scale deployment fma-requester -n ${NAMESPACE} --replicas=0
             kubectl wait --for=delete pod -l app=fma-requester -n ${NAMESPACE} --timeout=120s
+            # Confirm the launcher actually went to sleep before we wake it.
+            kubectl wait pod -l app.kubernetes.io/component=launcher -n ${NAMESPACE} \
+              --for='jsonpath={.metadata.labels.dual-pods\.llm-d\.ai/sleeping}=true' \
+              --timeout=120s
             kubectl scale deployment fma-requester -n ${NAMESPACE} --replicas=2
-            kubectl rollout status deployment/fma-requester -n ${NAMESPACE} --timeout=120s
+            kubectl rollout status deployment/fma-requester -n ${NAMESPACE} --timeout=300s
             kubectl get pods -l app.kubernetes.io/component=launcher -n ${NAMESPACE} \
               -o 'jsonpath={range .items[*]}{.metadata.name}{" sleeping="}{.metadata.labels.dual-pods\.llm-d\.ai/sleeping}{"\n"}{end}' || true
             kubectl run curl-wake-${i} --rm -i --restart=Never --attach \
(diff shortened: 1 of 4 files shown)


diegocastanibm:
/test-nightly fast-model-actuation-ibm-acc-gpu-vllm-x workload=nop.yaml harness=nop

guides/fast-model-actuation/guide.yaml
```diff
@@ -109,9 +112,29 @@ verify:
...
+      # to sleep), wait for the requester pods
+      # to terminate, scale up to 2, wait for rollout, then send a real
+      # inference request through the EPP.
+      - run: |
+          set -eu
+          IP=$(kubectl get service ${GUIDE_NAME}-epp -n ${NAMESPACE} -o jsonpath='{.spec.clusterIP}')
+          for i in 1 2; do
+            echo "== sleep/wake cycle ${i} =="
+            kubectl scale deployment fma-requester -n ${NAMESPACE} --replicas=0
+            kubectl wait --for=delete pod -l app=fma-requester -n ${NAMESPACE} --timeout=120s
```
diegocastanibm, line 126: Claude is suggesting to check for `sleeping=true` after this line. I like the idea. It is a way to be 100% that the instances are sleeping

guides/fast-model-actuation/rbac/role.yaml
```diff
@@ -1,8 +1,13 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: fma-launcher
+---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: fma-launcher-pod-reader
+  name: fma-launcher-pod-state-writer
```
diegocastanibm, line 9: Because the Role is renamed, kubectl apply creates the new objects alongside the old ones rather than replacing them (since the namespace is not deleted during the CI bcasue @maugustosilva added `kubectl delete namespace` with `skip_in: [ci]`). RBAC is purely additive, so in any namespace that has already run this guide, the old grant survives. 

I agree that `pod-state-writer` is better than `pod-reader`  based on the new verbs that you are setting, so I do not recommend to use the old name for the role. However we need to delete the old role, and rolebinding (that used default as SA as you can see in the `guide.yaml`.

I have asked Claude and the best way to do so is to add this code in the cleanup block in the guide:
```
     # Namespaced RBAC for this guide (Role + ServiceAccount from rbac/role.yaml,
     # plus the imperatively-created RoleBinding).
     - run: kubectl delete -n ${NAMESPACE} -f ${REPO_ROOT}/guides/${GUIDE_NAME}/rbac/role.yaml --ignore-not-found=true
     - run: kubectl delete rolebinding fma-launcher-pod-state-writer -n ${NAMESPACE} --ignore-not-found=true
     # Legacy RBAC from earlier revisions of this guide, which granted
     # pods get/list/watch to the "default" ServiceAccount. Deleted explicitly
     # because the rename means `apply` no longer supersedes it.
     - run: kubectl delete rolebinding fma-launcher-pod-reader -n ${NAMESPACE} --ignore-not-found=true
     - run: kubectl delete role fma-launcher-pod-reader -n ${NAMESPACE} --ignore-not-found=true
```
`--ignore-not-found=true keeps` this safe in fresh namespaces where the legacy objects were never created

aavarghese: I think we should add the first half for deleting the new RBAC. Keeping the legacy deletion in cleanup is confusing for a new FMA user...

aavarghese: https://github.com/llm-d/llm-d/pull/2307/changes/639d92a6708a857159f6ef826593c4436a6db2b9

diegocastanibm: I agree, but then the old RBAC is going to stay in the cluster forever. If you do not want to add it, then we should delete it manually

aavarghese: sure. Thank you! Deleted them manually. We're good to go now!
```
ansuvarghese$ oc get rolebindings -n llm-d-nightly-fast-model-actuation-ibm-gpu
NAME                            ROLE                                 AGE
fma-launcher-pod-state-writer   Role/fma-launcher-pod-state-writer   24h
inference-perf-runner-binding   Role/inference-perf-runner-role      4d4h
system:deployers                ClusterRole/system:deployer          6d13h
system:image-builders           ClusterRole/system:image-builder     6d13h
ansuvarghese$ oc get roles -n llm-d-nightly-fast-model-actuation-ibm-gpu
NAME                            CREATED AT
fma-launcher-pod-state-writer   2026-08-17T16:46:23Z
inference-perf-runner-role      2026-08-14T13:20:00Z
```

diegocastanibm: Thanks!

diegocastanibm (approved):
LGTM

PR #2308, opened 2026-08-17
proposal: add MetaX accelerator support

FouoF:
## Summary

- propose a phased MetaX C-series accelerator integration for llm-d
- define the optimized-baseline file layout and minimum validation contract
- record the current candidate vLLM-MetaX, MACA, and mcoplib release set
- add advanced P/D well-lit pace guide


Gregory-Pereira:
Hi @FouoF I am totally in favour of this I just want to point out that we are moving to entirely upstream images. We have agreed to host some `llm-d` inference server image variants as a stop gap measure while features do not yet exist upstream in sglang or vLLM. I see no issue with including metax hardware in our guide overlays provided you can use upstream images, or will move towards using them. Will need official endorsement from the maintainers though, cc @robertgshaw2-redhat @chcost @ahg-g

FouoF:
@Gregory-Pereira Is there related issue or docs about 

> we are moving to entirely upstream images

I'm not sure if the "upstream images" means the official vllm repo. If it is, I agree that it's best to integrate all hardware support to one inference engine but the fact is that most hardware vendors maintain their own vllm images. I wonder if you have some idea to address this problem.

archlitchi:
@Gregory-Pereira ptal

guides/optimized-baseline/modelserver/metax/vllm/kustomization.yaml
Gregory-Pereira, line 1: Delete this directory - you can still use our existing bases `guides/recipes/modelserver/base/single-host` and just remap the labels

guides/optimized-baseline/modelserver/metax/vllm/patch-vllm.yaml
Gregory-Pereira, line 1: Again delete this - we dont need 2 overlay directories. You should consume the shared base and then patch everything in a single overlay. llm-d is a production project - single GPU modelservers are not really a production example

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
Gregory-Pereira, line 1: Rename: to just `vllm`. 8 gpus is the default for optimized baseline anyway

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: optimized-baseline-metax-vllm-decode
```
Gregory-Pereira: rename to just `deocde` the rest will come from the generate name prefixing in the kustomization

guides/pd-disaggregation/modelserver/metax/vllm/base/kustomization.yaml
Gregory-Pereira, line 1: Same as above - delete file, lets only have the single node example

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/kustomization.yaml
Gregory-Pereira, line 1: rename to just `guides/optimized-baseline/modelserver/metax/vllm/kustomization.yaml` we are getting rid of the single GPU overlay

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
...
+resources:
+  - ../vllm
+namePrefix: eight-gpu-
+labels:
+  - pairs:
+      llm-d.ai/model: DeepSeek-R1-Distill-Llama-70B
+    includeSelectors: true
+    includeTemplates: true
+patches:
+  - path: patch-profile.yaml
```
Gregory-Pereira:
```suggestion
  - path: patch-vllm.yaml 
```

To be consistent with other guide examples

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
```diff
@@ -0,0 +1,42 @@
...
+spec:
+  replicas: 1
```
Gregory-Pereira: We should do 2 replicas here. If you dont have 2 nodes on hand I would suggest doing 2 replicas of TP=4 - the whole point of llm-d is balancing traffic across modelservers. If we only have 1 you get no performance bonuses unless you do something like offloading

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
```diff
@@ -0,0 +1,42 @@
...
+            - deepseek-ai/DeepSeek-R1-Distill-Llama-70B
+            - --trust-remote-code
+            - --tensor-parallel-size=8
+            - --pipeline-parallel-size=1
+            - --data-parallel-size=1
+            - --gpu-memory-utilization=0.85
+            - --max-model-len=4096
+            - --max-num-batched-tokens=4095
+            - --distributed-executor-backend=mp
+            - --disable-access-log-for-endpoints=/health,/metrics,/v1/models
```
Gregory-Pereira: Remove any of these that are default arguments:

```suggestion
            - serve
            - deepseek-ai/DeepSeek-R1-Distill-Llama-70B
            - --port
            - "8000"
            - --trust-remote-code
            - --tensor-parallel-size=8
            - --gpu-memory-utilization=0.85
            - --max-model-len=4096
            - --max-num-batched-tokens=4095
            - --distributed-executor-backend=mp
            - --disable-access-log-for-endpoints=/health,/metrics,/v1/models
```

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
```diff
@@ -0,0 +1,42 @@
...
+            - "8000"
+            - --served-model-name
+            - deepseek-ai/DeepSeek-R1-Distill-Llama-70B
+            - --trust-remote-code
+            - --tensor-parallel-size=8
+            - --pipeline-parallel-size=1
+            - --data-parallel-size=1
+            - --gpu-memory-utilization=0.85
+            - --max-model-len=4096
+            - --max-num-batched-tokens=4095
```
Gregory-Pereira: Is this intentionally 4095 and not 4096?

Gregory-Pereira: These im guessing come from some sort of rightsizing calculation? Others might benefit from sharing those as comments - optional

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
```diff
@@ -0,0 +1,89 @@
...
+  name: decode
+spec:
+  replicas: 1
+  template:
+    spec:
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          args:
+            - "Qwen/Qwen3-14B"
```
Gregory-Pereira: Can we choose a more production sized model? Why don't we stick with qwen/qwen3-32b?

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
```diff
@@ -0,0 +1,89 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: decode
+spec:
+  replicas: 1
```
Gregory-Pereira: Single replica again - we should use prefill 4 replicas TP=2, decode 2 replicas TP=4

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
```diff
@@ -0,0 +1,89 @@
...
+            - "--port"
+            - "8200"
+            - "--served-model-name"
+            - "Qwen/Qwen3-14B"
+            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
+            - "--tensor-parallel-size=1"
+            - "--max-num-batched-tokens=8192"
+            - "--gpu-memory-utilization=0.9"
+            - "--kv-transfer-config"
+            - '{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_load_failure_policy":"fail"}'
```
Gregory-Pereira: Drop default args
```suggestion
            - "Qwen/Qwen3-14B"
            - "--port"
            - "8200"
            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
            - "--max-num-batched-tokens=8192"
            - "--kv-transfer-config"
            - '{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_load_failure_policy":"fail"}'
```

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
```diff
@@ -0,0 +1,89 @@
...
+                  name: llm-d-hf-token
+                  key: HF_TOKEN
+            - name: VLLM_NIXL_SIDE_CHANNEL_HOST
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+            - name: VLLM_NIXL_SIDE_CHANNEL_PORT
+              value: "5600"
+            - name: VLLM_HTTP_TIMEOUT_KEEP_ALIVE
+              value: "120"
```
Gregory-Pereira: Only set this on prefill:

```suggestion
```

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
```diff
@@ -0,0 +1,89 @@
...
+          securityContext:
+            capabilities:
+              add:
+                - IPC_LOCK
+            seccompProfile:
+              type: Unconfined
+          resources:
+            limits:
+              memory: 64Gi
+              metax-tech.com/gpu: "1"
```
Gregory-Pereira: Lets make this a 16 GPU deployment as stated above:

4 Replicas = TP 2 prefill
2 Replicas = TP 4 decode

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
```diff
@@ -0,0 +1,89 @@
...
+            failureThreshold: 6
+          readinessProbe:
+            periodSeconds: 10
+            timeoutSeconds: 5
+            failureThreshold: 6
+      volumes:
+        - name: shm
+          emptyDir:
+            medium: Memory
+            sizeLimit: 100Gi
```
Gregory-Pereira: Was this chosen intentionally? 100Gi of `/dev/shm` here seems insane. I would be looking in the 1-2 Gi range for comparable deployments on nvidia hardware

guides/pd-disaggregation/modelserver/metax/vllm/base/patch-prefill.yaml
Gregory-Pereira, line 1: Pretty much same comments as above but keep `VLLM_HTTP_TIMEOUT_KEEP_ALIVE`.

Gregory-Pereira (changes requested):
These are my question and comments. To make this process easier I have drafted a PR suggesting my changes here: https://github.com/FouoF/llm-d/pull/1, but you do not have to accept all of them.

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
```diff
@@ -0,0 +1,42 @@
...
+            periodSeconds: 10
+            timeoutSeconds: 5
+          resources:
+            requests:
+              cpu: "16"
+              memory: 64Gi
+              metax-tech.com/gpu: "8"
+            limits:
+              memory: 256Gi
+              metax-tech.com/gpu: "8"
```
Gregory-Pereira: We recommend specialized networking for PD - which is usually considered table stakes making it worth while. Usually this is `rdma/ib` or `rdma/roce_gdr`. Is there a cluster with this accelerated networking and MetaX GPUs this could be tested on?

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
archlitchi: Qwen3-14B is what we used for kubeCon keynote demo, because we have limited time so we can only pick small model to minimize warm up time, it works fine with 32B

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
archlitchi: adopted

guides/pd-disaggregation/modelserver/metax/vllm/base/kustomization.yaml
archlitchi: adopted

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
archlitchi: adopted

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/kustomization.yaml
archlitchi: adopted

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/kustomization.yaml
archlitchi: adopted

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
archlitchi: adopted

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
archlitchi: no, but once i have access to their hypernode, i'll be able to do testing on top of that, for now, i have only a 8xGPU node

archlitchi: > We recommend specialized networking for PD - which is usually considered table stakes making it worth while. Usually this is `rdma/ib` or `rdma/roce_gdr`. Is there a cluster with this accelerated networking and MetaX GPUs this could be tested on?

I'll follow-up and submit another PR once i have validated that

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
archlitchi: adopted

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
archlitchi: adopted

guides/pd-disaggregation/modelserver/metax/vllm/patch-decode.yaml
archlitchi: after coordinating with them, the best practice here is 32G, i'll update the part here

guides/optimized-baseline/modelserver/metax/vllm-eight-gpu/patch-profile.yaml
archlitchi: yeah, i got the parameter from their initial docker run script, i've updated their latest parameter.

Gregory-Pereira (approved):
/lgtm

PR #2314, opened 2026-08-18
fix CPU builds for NIXL 

Gregory-Pereira:
TSIA

### DIFF
--- a/docker/scripts/cpu/install_nixl.py
+++ b/docker/scripts/cpu/install_nixl.py
@@ -89,8 +89,8 @@ def build_and_install_prerequisites(args):
         return
 
     print("\n--> No installed package or cached wheel found. Starting full build process...", flush=True)
-    print("\n--> Installing auditwheel...", flush=True)
-    run_command([sys.executable, '-m', 'pip', 'install', 'auditwheel'])
+    print("\n--> Installing auditwheel and patchelf...", flush=True)
+    run_command([sys.executable, '-m', 'pip', 'install', 'auditwheel', 'patchelf'])
     install_system_dependencies()
     ucx_install_path = os.path.abspath(UCX_INSTALL_DIR)
     print(f"--> Using wheel cache directory: {WHEELS_CACHE_HOME}", flush=True)


docker/scripts/cpu/install_nixl_from_source_ubuntu.py
```diff
@@ -0,0 +1,254 @@
...
+    try:
+        nixl_release_url = "https://api.github.com/repos/ai-dynamo/nixl/releases/latest"
+        with urllib.request.urlopen(nixl_release_url) as response:
+            data = json.load(response)
+            return data.get("tag_name", "0.7.0")
+    except Exception:
+        return "0.7.0"
+
+
+NIXL_VERSION = os.environ.get("NIXL_VERSION", get_latest_nixl_version())
```
diegocastanibm: move this line with the rest of the constants? IMO, move it to line 20 would be better

docker/scripts/cpu/install_nixl_from_source_ubuntu.py
```diff
@@ -0,0 +1,254 @@
...
+        flush=True,
+    )
+    print("\n--> Installing auditwheel...", flush=True)
+    run_command([sys.executable, "-m", "pip", "install", "auditwheel"])
+    install_system_dependencies()
+    ucx_install_path = os.path.abspath(UCX_INSTALL_DIR)
+    print(f"--> Using wheel cache directory: {WHEELS_CACHE_HOME}", flush=True)
+    os.makedirs(WHEELS_CACHE_HOME, exist_ok=True)
+
+    # -- Step 1: Build UCX from source --
```
diegocastanibm: I think it is better to have each step in different functions. It would be easier to follow

diegocastanibm (dismissed):
just some small suggestions

docker/scripts/cpu/install_nixl_from_source_ubuntu.py
Gregory-Pereira: This script was pulled from https://github.com/vllm-project/vllm/blob/v0.26.0/tools/install_nixl_from_source_ubuntu.py exactly as is in an effort to reduce drift between upstream and downstream. Honestly maybe we just call the script directly from the checked out repo

docker/scripts/cpu/install_nixl_from_source_ubuntu.py
Gregory-Pereira: Same as above - I am open to putting in changes we just need to decide if we want to do that as a whole and break alignment with upstream

docker/scripts/cpu/install_nixl_from_source_ubuntu.py
maugustosilva: While I understand the goal of making the code "cleaner" (and thus easier to follow), I vote for just call the script from the repo. We have **plenty** to maintain already!

docker/scripts/cpu/install_nixl_from_source_ubuntu.py
maugustosilva: While I understand the goal of making the code "cleaner" (and thus easier to follow), I vote for just call the script from the repo. We have **plenty** to maintain already!

Gregory-Pereira:
Resolves: https://github.com/llm-d/llm-d/actions/runs/32081267870/job/95544569201

PR #2332, opened 2026-08-19
fix(guides): unify PROVIDER_NAME default to none

### LINKED ISSUE #2330: [Bug :bug:]: modelexpress-p2p and optimized-baseline default PROVIDER_NAME differently
### Contact Details

lukevandrie@google.com

### What happened?

`guides/optimized-baseline/guide.yaml` defaults `PROVIDER_NAME` to `none`, matching the README default shipped in #2260. `guides/modelexpress-p2p/guide.yaml` defaults it to `gke`. Both guides pass the value to the same `--set provider.name=` flag in the router gateway install, so the two guides document different defaults for the same knob.

The defaults have opposite failure modes. With `none`, the chart renders no provider-specific resources; on GKE the `HealthCheckPolicy` is omitted, the default HTTP health check fails against vLLM, and requests return 503. With `gke`, non-GKE clusters receive GKE-only resources that fail to apply.

Expected: one default for both guides, with the tradeoff documented where the variable is set.
### Version

main (2026-08-19), router chart channel v0

### Relevant log output

```shell

```

LukeAVanDrie:
`guides/optimized-baseline` and `guides/modelexpress-p2p` documented different defaults for `PROVIDER_NAME`, the value passed to `--set provider.name=` in the router gateway install. PR #2260 changed the optimized-baseline README default to `none` inside the rendered `guide:env.static` block. Both `guide.yaml` files, the render source of truth, still said `gke`, so `scripts/guide.py render` would have reverted the README.

Both guides default `PROVIDER_NAME` to `none`, matching the router chart's default for `provider.name`. GKE users must set `PROVIDER_NAME=gke`: with `none`, the chart omits the `HealthCheckPolicy`, the load balancer health check fails against vLLM, and requests return 503. Each `guide.yaml` carries a comment above the variable stating the requirement, and each README states it in prose above the env block.

Related fixes in the same files:

- optimized-baseline `guide.yaml` declares `ROUTER_CHART_VERSION`. The variable existed only in the README's rendered block, so `guide.py render` deleted it and `render --check` failed on `main`.
- `guides/recipes/router/README.md` documented `provider.name=<gke|istio|none>`. The placeholder lists all four accepted values, and a paragraph describes what each value renders.

Seven other guide READMEs export `PROVIDER_NAME=gke` without a warning; a follow-up issue should track that sweep.

**Verification:**

- `scripts/guide.py check` and `scripts/guide.py render --check` pass for both guides.
- The scheduled GKE nightlies for both guides use `gateway_class: epponly` (standalone mode), which does not read `PROVIDER_NAME`, so the default change does not affect CI.

Fixes #2330

### DIFF
--- a/guides/modelexpress-p2p/README.md
+++ b/guides/modelexpress-p2p/README.md
@@ -73,6 +73,8 @@ git clone https://github.com/llm-d/llm-d.git && cd llm-d && git checkout ${BRANC
 
 * Set the guide specific environment variables:
 
+`PROVIDER_NAME` selects gateway-provider-specific resources in the router chart. The default `none` renders no provider resources. On GKE, set `PROVIDER_NAME=gke`, or the load balancer health check fails against vLLM and requests return 503. `istio` adds a DestinationRule; `agentgateway` requires no provider-specific resources.
+
 <!-- guide:env.static start -->
 ```bash
 export BRANCH=main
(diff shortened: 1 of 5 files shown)


Gregory-Pereira (approved):
/lgtm

LukeAVanDrie:
/assing @ahg-g 

can you do a quick scan to make sure I am not missing anything for GKE guidance?

ahg-g:
@LukeAVanDrie sorry I missed this, pls resolve the conflicts

PR #2333, opened 2026-08-19
docs: describe flow control for optimized baseline

nt591:
In Slack, @vMaroon and I discussed the benefits observed by setting a max concurrency to maintain a healthy throughput and limit KV cache thrashing under load. This PR adds a blurb under the optimized baseline as an optional configuration, and why. I limited the detail and preferred to link out where appropriate to minimize the risk of documentation drifting.

### DIFF
--- a/guides/optimized-baseline/README.md
+++ b/guides/optimized-baseline/README.md
@@ -279,6 +279,36 @@ The recipe (`calibrate.sh`) runs a short Kubernetes Job that measures true prefi
 
 For reference values across the (model, accelerator) combinations shipped under `guides/` — and which ones still need a calibration run — see the [**configuration matrix**](../recipes/router/calibration/configuration-matrix.md).
 
+## Optional: Enable flow control
+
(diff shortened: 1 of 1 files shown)


guides/optimized-baseline/README.md
```diff
@@ -329,6 +279,36 @@ The recipe (`calibrate.sh`) runs a short Kubernetes Job that measures true prefi
...
+featureGates:
+- flowControl
+
+plugins:
+# Keep the existing optimized-baseline plugins here.
+- type: concurrency-detector
+  parameters:
+    maxConcurrency: 8 # Example only; tune this for each endpoint's performance and latency constraints.
+    concurrencyMode: requests
+    headroom: 0.0
```
vMaroon, line 300: This technically does not need flow-control enabled - it can be a soft-gate. E.g., once no longer applicable, it will be broken.

Wonder if we should make the concurrency guard a part of the main configuration, and have this section explain flow-control on and off.

cc @ahg-g @kaushikmitr

vMaroon (approved):

PR #2320, opened 2026-08-19
docs: migrate deprecated plugin and config

zdtsw:
# Notes
- from `deciderPluginName` to `deciders.prefill`
- remove `disagg-header-handler `as it is already native support in `disagg-profile-handler`

cc @roytman

PR #2336, opened 2026-08-20
feat(agentic-serving): add eval-containers benchmark + nightly for the guide

natan-ergas:
Adds the agentic `eval-containers` benchmark to the **agentic-serving guide**, in
two pieces:

1. **A benchmark template** —
   `guides/agentic-serving/benchmark-templates/agentic-serving-eval-containers.yaml`,
   a third sibling beside the two existing `inference-perf` templates, plus the
   README section documenting it. Makes the eval a documented, runnable benchmark
   of the guide today, by hand, against any model the guide serves.
2. **A nightly** — `nightly-e2e-eval-containers-ibm-acc-gpu-vllm-x.yaml`, which
   `uses:` llm-d-infra's `reusable-ci-nightly-benchmark.yaml` and passes the eval
   as the **workload** (`harness: eval-containers`), the same way
   `inference-perf` is the workload for the perf nightlies. Runs at `0 2 * * *`,
   clear of the 03:00 slot the other eight OCP nightlies share.

### Why the eval complements inference-perf rather than duplicating it

The guide's existing templates report throughput, TTFT, and ITL — how fast the
stack serves a synthetic agentic load. This one drives a **real coding agent**
(aider-polyglot) and reports **reward**: the fraction of tasks actually solved.
The guide's own Benchmarking section already named that as a gap — *"Replaying
complete agent dependency graphs with tool timing and sub-agent fan-out remains a
separate evaluation mode."*

The two modes catch different failures. A broken tool-call parser, reasoning
traces leaking into content, or truncation mid-solution each leave TTFT and ITL
perfectly healthy while the agent silently fails every task. Reward sees them: a
uniform 0.0 across tasks points at a transport or parser fault, a spread reflects
genuine task difficulty.

Unlike its siblings the template takes `${MODEL}` rather than pinning a handle,
because the harness resolves the served model from the deployment itself
(`EVAL_MODEL="${LLMDBENCH_DEPLOY_CURRENT_MODEL:?}"`). One template therefore
covers Nemotron-3-Ultra, GLM-5.2, Qwen3-Coder-480B, and smaller models where a
cheap repeatable signal matters more than frontier scale.

### Revised from the dispatch approach (thanks @maugustosilva)

The first version of this PR used a cross-repo `workflow_dispatch` into
llm-d-benchmark, on the grounds that the shared engine couldn't drive the eval.
That's no longer true, and the review was right to push back:

- `--cluster-config` and `-j` were missing from the engine. **llm-d-infra#244
  has since merged** and adds both (`cluster_config`, `tasks`).
- `eval-containers` is already a declared value of the engine's `harness` input,
  and `workload/harnesses/eval-containers-llm-d-benchmark.sh` is in
  llm-d-benchmark on `main`.

With those in place a separate dispatch path buys nothing, so this revision
**deletes** `EVAL_DISPATCH_TOKEN`, the `github-script` dispatch, and the
run-URL polling — about 85 lines. A reusable-workflow call needs no PAT, and the
run now shows up as a normal job in this repo instead of a link to another one.

### The anyuid SCC is a prerequisite, not a CI step

My earlier argument was that the eval needs the **anyuid SCC** bound before
standup (the eval image and vLLM both run as root), and that putting an
OpenShift-specific SCC grant into the engine all nightlies share was too wide a
blast radius. The second half stands; the first half was wrong about where the
grant belongs.

`config/cluster-configs/examples/openshift-pokprod.yaml` already supplies the
`serviceAccount` and `runAsUser` to the cluster-generic scenario, and its own
header documents the binding as a **one-time-per-namespace prerequisite**:

```
oc create serviceaccount eval-runner -n llm-d-nightly-eval-containers-ibm-gpu
oc adm policy add-scc-to-user anyuid -z eval-runner -n llm-d-nightly-eval-containers-ibm-gpu
```

That's cluster state, not per-run work. It's recorded in the workflow header,
and the shared engine carries no OpenShift privilege logic.

### On the nightly's model

@maugustosilva — agreed on the direction, and this now uses the mechanism you
described, with the eval as the guide's workload.

The nightly serves **Qwen/Qwen3.6-27B on a single H100**, via
`config/scenarios/examples/eval-containers-aider-polyglot-gpu.yaml` — already on
llm-d-benchmark `main`, so there's no cross-repo dependency here. That model is
deliberate rather than a stand-in: a nightly wants a fast, cheap, every-night
signal, and this configuration scores non-trivially on aider-polyglot, so a score
regression is meaningful. The guide's three documented deployments are 64-GPU
configurations — right as reference deployments, but a nightly on that footprint
is a capacity event, not a regression test.

Worth stating plainly, since my first version of this PR got it wrong: nothing in
the guide scopes it by model size. There is no `gpu.product` selector or H200
constraint in the manifests, and the guide's substance — the routing and
scheduling profile in `router/agentic-serving-gpu.values.yaml` (prefix-cache
affinity, token-load scoring, prefill/decode filters) — binds to model servers by
label:

```yaml
  modelServers:
    matchLabels:
      llm-d.ai/guide: "agentic-serving-gpu"
```

`llm-d.ai/guide` and `llm-d.ai/model` are independent labels, so the guide's
identity is decoupled from the model by design.

The nightly drives the eval through the shared engine via `scenario_dir: examples`
against the eval's own scenario, which is what makes it self-contained: the
scenario is on llm-d-benchmark `main`, so nothing else has to land first. If this
guide later migrates to the `guide.yaml` method, the same job becomes
`scenario_dir: guides` + `standup_scenario: agentic-serving` by changing two
inputs, and the served model then comes from the guide's own manifests via the
engine's `--model` auto-detection.

### Cluster selection is by workflow name

Worth flagging for reviewers, since it's non-obvious: the engine picks the
cluster by grepping `github.workflow` for a provider substring
(`reusable-ci-nightly-benchmark.yaml:289-303`). `"OCP"` in the title selects the
self-hosted pok-prod runner pool and `KUBECONFIG_DATA_OCP`. So the
`(OCP GPU)` suffix and the `-ibm-` filename segment are load-bearing.

### Testing

- `actionlint` clean, and the full `pre-commit` suite passes on the new file
  (including `sync-nightly-matrix --check` — `eval-containers` is a workload,
  not one of the ten release-gating guides in `matrix_common.GUIDES`, so it's
  correctly absent from the matrix table).
- Every value in the `with:` block diffed against the engine's declared inputs.
  `scenario_dir: examples` and `workload: aider-polyglot.yaml` both resolve
  (`config/scenarios/examples/eval-containers-aider-polyglot-gpu.yaml`,
  `workload/profiles/eval-containers/aider-polyglot.yaml.in`).
- `decode_pods` is deliberately **not** exposed: the engine only applies it to
  kustomize `patch-*.yaml` manifests (step gated
  `if: env.LLMDBENCH_CICD_METHOD == 'kustomize'`), and this scenario stands up
  via `modelservice`, so the input would silently do nothing. Replica count
  lives in the scenario.
- Unlike the siblings, every input referenced in `with:` is declared or literal,
  so `actionlint` reports no `property "..." is not defined` warnings. The rest
  of the family has 535 of these; I didn't touch those files.

Not exercised: a live nightly, which needs the one-time SCC binding above on
pok-prod.

### Note on the DCO bot comment

The "Unsigned commits detected!" comment above is stale — it fired at 07:32:10
against an earlier push, 11 minutes before the signed commit landed. Both
commits on this branch are SSH-signed and report `verified: true` /
`reason: valid` from the commits API.


maugustosilva:
The plan is to have a the actual guide [agentic-serving](https://github.com/llm-d/llm-d/tree/main/guides/agentic-serving) tested via `reusable-ci-nightly-benchmark.yaml` (which already used `llm-d-benchmark`) and then use the `agentic-eval` as the workload to benchmark it.

natan-ergas:
Rebased onto `main` (`b6e9da2`) — clean, no conflicts. The only upstream commit touching `guides/agentic-serving/` was #2320, which edits `router/*.values.yaml`; this PR touches the README and `benchmark-templates/`, so there was no overlap. All three files are byte-identical to their pre-rebase versions (verified by checksum).

### State

Two pieces, both aimed at the **agentic-serving guide**:

| File | |
|---|---|
| `guides/agentic-serving/benchmark-templates/agentic-serving-eval-containers.yaml` | +104 |
| `guides/agentic-serving/README.md` | +33 |
| `.github/workflows/nightly-e2e-eval-containers-ibm-acc-gpu-vllm-x.yaml` | +122 |

1. **Benchmark template** — a third sibling beside the two `inference-perf` templates, declaring `harness: eval-containers` on the aider-polyglot task set, plus the README section documenting it. Reports **reward** (fraction of tasks solved) rather than latency, covering the mode the guide's own Benchmarking section calls out as missing. Model-agnostic via `${MODEL}`, since the harness resolves the served handle from the deployment.
2. **Nightly** — `uses:` llm-d-infra's `reusable-ci-nightly-benchmark.yaml` with the eval as the workload, on OCP/pok-prod at 02:00 UTC. Serves Qwen/Qwen3.6-27B on one H100 via `eval-containers-aider-polyglot-gpu.yaml`, which is already on llm-d-benchmark `main` — no cross-repo dependency.

### How it was tested

Everything below was re-run **after** the rebase, against the new base:

- **`CI Guides Check`** (added upstream in #2384/#2320 and triggered by `guides/**`), run exactly as the workflow does:
  - `python scripts/guide.py render guides/*/ --check` → exit 0. `agentic-serving` has no root `guide.yaml`, so it isn't rendered and the README edit can't stale a generated file.
  - `python -m pytest scripts/tests/ -v` → **28 passed**.
- **`pre-commit`**, all three files → every hook passes, including `yamllint`, `markdownlint`, `actionlint`, and `sync nightly testing matrix`. `eval-containers` is a workload rather than one of the ten release-gating guides in `matrix_common.GUIDES`, so it's correctly absent from the badge matrix and `sync-nightly-matrix.py --check` stays green.
- **`actionlint`** on the workflow → clean, no findings. Every input in `with:` is declared or literal, so unlike the sibling nightlies it emits no `property "..." is not defined` warnings.
- **Input contract** — all 21 inputs passed to the engine, and every input passed to `reusable-update-badge.yaml`, diffed against those workflows' current declared inputs: exact match (an unknown input would 422). `failure_category` is still an engine output.
- **Scenario/workload resolution** — `scenario_dir: examples` + `standup_scenario: eval-containers-aider-polyglot-gpu` and `workload: aider-polyglot.yaml` both confirmed present on llm-d-benchmark `main`.
- **Signing** — all three commits SSH-signed and `verified: true`/`valid`; `signed-commits` and `DCO` both pass on this push. (The earlier "Unsigned commits detected!" comment was stale — it fired ~11 minutes before the signed commit landed.)

### Cluster prerequisites are in place

The one-time per-namespace setup noted in the workflow header has been applied on pok-prod, and verified rather than assumed:

- Namespace `llm-d-nightly-eval-containers-ibm-gpu` and ServiceAccount `eval-runner` created; `RoleBinding/system:openshift:scc:anyuid` bound to it.
- **Server-side admission check**: a `runAsUser: 0` pod as `eval-runner` is admitted (`oc apply --dry-run=server`), and the assigned SCC comes back as `anyuid`. That exercises the real admission controller, which is what standup hits at step 05.
- Default StorageClass is `ibm-spectrum-scale-fileset` (RWX), which the model PVC and the shared results PVC need at `-j > 1`.
- 14 nodes carry `nvidia.com/gpu.product=NVIDIA-H100-80GB-HBM3`, the label the scenario's `nodeSelector` matches, with **40 of 96 allocatable GPUs free** — the scenario's 10 decode replicas bind with headroom.

`-j` is set to **20**: 20 agent pods over 10 decode replicas is 2 concurrent requests per H100, about what a 32768-token KV budget sustains.

**Not exercised:** a live end-to-end run, which happens once this is merged and the workflow runs on its own schedule. The remaining CI jobs here show `action_required` pending a maintainer's approval to run.

Issue #2341, opened 2026-08-20
[Feature]: Migrate flow-control guide + nightly to the guide.yaml method


LukeAVanDrie:
### Feature Area

CI/CD

### Problem Statement

The flow-control guide runs its nightly on a bespoke lane. `nightly-e2e-flow-control-gke-acc-gpu-vllm-x.yaml` calls `reusable-nightly-e2e-gke.yaml` (llm-d-infra) with `custom_deploy_script: guides/flow-control/scripts/nightly-deploy-gke.sh` and a bespoke validate script at `.github/scripts/e2e/e2e-validate-flow-control.sh`.

There is no `guides/flow-control/guide.yaml`, so `scripts/guide.py` neither renders nor checks the README, and the deploy commands are maintained in three places: the README, the deploy script, and the validate script. These drift. Nightly run #76 (2026-08-20) failed with a 404 on the GAIE CRD manifest because #2299 fixed the CRD URLs in `guides/env.sh` and the README while the deploy script, restored in #1918 before #2299 merged, kept the old hand-built URL.

### Proposed Solution

Adopt the guide.yaml method used by optimized-baseline and fast-model-actuation, in two phases.

**Phase 1 (llm-d repo only): guide.yaml as the single source of truth**

- Author `guides/flow-control/guide.yaml` and re-render the README from it with `scripts/guide.py render`.
- Rewrite `nightly-deploy-gke.sh` to consume the guide-defined commands, so a fix to the guide reaches the nightly without a second edit.
- Hardening that can ride along: pin `GAIE_VERSION` and `ROUTER_RELEASE_VERSION` in the workflow env or log the resolved releases before applying, and add retries to the CRD fetches. Both address nightly failures that have no llm-d commit in the blame window.
- Optionally wire `scripts/guide.py render --check` into CI. `guides/templates/README.md` recommends it and no workflow runs it today.

**Phase 2 (cross-repo): move the nightly to the standardized lane**

The workflow header documents the constraints the bespoke lane exists to satisfy. Phase 2 must resolve each of them:

1. Model detection. The harness's kustomize standup detects the model from `guides/<scenario>/modelserver/<accel>/<backend>/<infra>/`. flow-control keeps no modelserver overlay and borrows optimized-baseline's via `kubectl kustomize | sed`. The flow-control special case in `reusable-ci-nightly-benchmark.yaml` (llm-d-infra) runs after the detection step and never executes. The ordering or the detection fallback needs to change.
[...]

PR #2299, opened 2026-08-14
Add variables `GATEWAY_API_URL`, `GAIE_URL` and `ROUTER_RELEASE_URL` automatically calculated from `guides/env.sh`

maugustosilva:
Additionally, set as many images `ghcr.io/llm-d/llm-d-*` to `main` as possible

PR #1918, opened 2026-06-24
ci: restore bespoke flow-control nightly e2e

LukeAVanDrie:
**Problem**

The flow-control nightly e2e has failed every scheduled run since #1728 merged. Two hours after #1728 landed, a fleet-wide normalization to the benchmark harness (#1766) moved flow-control into `reusable-ci-nightly-benchmark.yaml` and deleted the bespoke `custom_deploy_script` and `--extra-validate` handoff. The bespoke e2e never reached a scheduled run.

The benchmark harness cannot stand the guide up. Its kustomize standup detects the model from `guides/<scenario>/modelserver/<accel>/<backend>/<infra>/`. The flow-control guide keeps no `modelserver/` overlay and borrows optimized-baseline's, which is what the guide README tells users to do. Runs fail at standup with `unable to detect model`. `reusable-ci-nightly-benchmark.yaml` carries a flow-control special case that borrows the optimized-baseline overlay, but it runs after the model-detection step that needs the missing directory, so it never executes. Filing that against the harness separately.

A benchmark run would also leave the feature untested. Flow control is work-conserving, so `sanity_random` never produces queue contention. Backpressure, per-band isolation, and QoS ordering go unexercised.

`workload-autoscaling` runs a bespoke lane for the same missing-overlay reason.

**Change**

Point the `nightly` job at `reusable-nightly-e2e-gke.yaml` with `e2e_validate_args: '--extra-validate flow-control'`. Move the deploy logic into `guides/flow-control/scripts/nightly-deploy-gke.sh`, following the workload-autoscaling precedent, which also picks up shellcheck via `ci-pr-checks`.

The script follows the guide README's install steps. It carries three CI-only deviations, each marked inline: a low `maxConcurrency` so the saturation gate closes under the validate script's burst, `prometheus.auth.enabled=false` so the validate pod can scrape `/metrics`, and the optimized-baseline modelserver relabel the README already documents.

`reusable-nightly-e2e-gke.yaml` declares no outputs, so `failure_category` resolves empty and the badge falls back to plain pass/fail. The workload-autoscaling lanes behave the same way.

**Fixes since the last review**

[...]

PR #2345, opened 2026-08-20
feat(guides): migrate flow-control to the guide.yaml method

LukeAVanDrie:
# What this PR does

Phase 1 of #2341: `guides/flow-control/guide.yaml` becomes the single source of truth for the guide's commands. The README's bash fences render from it, and the nightly deploy script consumes it, so a guide fix reaches the nightly without a second edit.

The flow-control nightly has broken a few times by missing fixes that landed in the README and `guides/env.sh`. Run #/76 (2026-08-20) failed with a 404 on the GAIE CRD manifest because #2299 fixed the CRD URLs while the deploy script, restored in #1918 before #2299 merged, kept a hand-built URL.

## Changes

- **`guides/flow-control/guide.yaml`** (new): env, prerequisites, deploy, verify, benchmark, and cleanup sections, following the optimized-baseline conventions. The CRD steps use `${GAIE_URL}`/`${ROUTER_RELEASE_URL}` from `guides/env.sh`, which fixes the run #/76 404 and supersedes the standalone hotfix #2342. Two automation hooks, `ROUTER_VALUES` and `EXTRA_HELM_ARGS`, let CI inject a values file and helm flags without the guide documenting anything CI-specific.
- **`scripts/guide.py`**: new `emit` subcommand. `guide.py emit <guide> env deploy.standalone --context ci --var NAME=VALUE` prints an executable script assembled from guide.yaml sections. `--context` drops `skip_in:` steps; `--var` overrides drive `when:` resolution, are validated against declared `values:` lists, and are shell-quoted (defaults emit verbatim so `$(…)` still runs). A sensitive variable without an override is omitted, and `when:` on one is an error. Validation now rejects YAML null/boolean env values (previously rendered as `export VAR=None`) and non-boolean `sensitive:` flags.
- **`guides/flow-control/README.md`**: 19 marker pairs inserted and fences re-rendered. The command diff is behavior-identical: fences de-indented out of bullets, the env block split into static/source fences, helm commands using `${ROUTER_VALUES}`/`${EXTRA_HELM_ARGS}`, `INFRA_PROVIDER` moved to the env block, cleanup steps separated by blank lines. The stateful multi-terminal use-case fences stay unmarked.
- **`guides/flow-control/scripts/nightly-deploy-gke.sh`**: rewritten as a wrapper over `guide.py emit`. All CI-only overrides survive with their fail-loud guards: maxConcurrency lowered to 4, EPP metrics auth disabled via `EXTRA_HELM_ARGS`, and a single decode replica. The replica patch is the one step that stays hand-mirrored, because a sed cannot interpose in the guide's `kustomize | sed | kubectl apply` pipe; the comment cross-references `deploy.modelserver`. The CRD installs retry 3× and run under `bash -x` so the log records the resolved release URLs, which matters when a failure has no llm-d commit in the blame window.
- **`.github/workflows/ci-guides-check.yaml`** (new): PR gate running `scripts/guide.py render guides/*/ --check` plus the guide.py test suite, path-filtered to `guides/**` and `scripts/guide.py`. All four converted guides pass today. `guides/templates/README.md` recommended the check; nothing ran it before.
- **`scripts/tests/test_guide.py`** (new): 26 tests covering emit semantics, per-guide render regressions, and a nightly-contract test asserting the emitted scripts contain the CI values path, the auth `--set`, and the env.sh URL forms.
- **`guides/templates/README.md`**: documents `emit` and the live CI gate. Emitting a parent section such as `deploy` concatenates every named sub-group, so automation must pass specific sub-paths; only `when:` encodes exclusivity.

## Verification

- `scripts/guide.py render guides/*/ --check` passes for all four converted guides, and rendering is idempotent.
- `python -m pytest scripts/tests/` — 26 passed.
- The emitted CRD/deploy/objectives scripts with the nightly's exact `--var` set pass `bash -n`, and their commands match the old deploy script except for the intended deltas (URL form, `ROUTER_VALUES`, `EXTRA_HELM_ARGS`).
- Markdownlint findings on the touched READMEs are byte-identical to main; none are inside rendered marker regions.
- Remaining: dispatch the nightly on the PR branch (`/test-nightly`) to confirm the validate script still sees unauthenticated `:9090/metrics`, one replica, applied objectives, and `maxConcurrency: 4`.

## Notes for reviewers

- Phase 2 of #2341 (moving the nightly to the standardized benchmark lane) needs coordinated llm-d-infra and llm-d-benchmark changes and is out of scope.

Part of #2341.

### DIFF
--- a/.github/workflows/ci-guides-check.yaml
+++ b/.github/workflows/ci-guides-check.yaml
@@ -1,0 +1,49 @@
+name: CI Guides Check
+
+# Guides on the guide.yaml method render their README bash fences from the
+# YAML (see guides/templates/README.md). The job fails when a README is stale
+# relative to its guide.yaml (fix: `scripts/guide.py render <guide-dir>`) and
+# runs the guide.py unit tests.
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - .github/workflows/ci-guides-check.yaml
+      - guides/**
+      - scripts/guide.py
+      - scripts/tests/**
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-guides-check-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  guides-check:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@v7
+        with:
+          python-version: '3.x'
+          cache: pip
+          # No requirements file — the workflow itself pins the installs.
+          cache-dependency-path: .github/workflows/ci-guides-check.yaml
+
+      - name: Install dependencies
+        run: python -m pip install --quiet 'pyyaml==6.*' pytest
+
+      - name: Check rendered READMEs are current
+        run: python scripts/guide.py render guides/*/ --check
+
+      - name: Run guide.py tests
+        run: python -m pytest scripts/tests/ -v
--- a/guides/flow-control/README.md
+++ b/guides/flow-control/README.md
@@ -70,47 +70,70 @@ Flow Control is a software-level scheduling feature at the EPP layer and is enti
 * Have the [proper client tools installed on your local system](../../helpers/client-setup/README.md) to use this guide.
 * Checkout llm-d repo:
 
-  ```bash
-  export branch="main" # branch, tag, or commit hash
-  git clone https://github.com/llm-d/llm-d.git && cd llm-d && git checkout ${branch}
-  ```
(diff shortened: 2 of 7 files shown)


clubanderson:
### Nightly E2E Result: `flow-control` (GKE)

| | |
|---|---|
| Guide | `flow-control` |
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/32426968376) |
| Ref | `b4b9527` |

Vezio:
This looks pretty good. I'm not sure if the `test` is needed there, since it seems to be hardcoding values exclusively for flow-control. The rendering suite has a bunch of checks today there that serve as tests. So it might be worthwhile using that as it's intended purpose. But otherwise, looks pretty good.

maugustosilva (approved):

PR #2347, opened 2026-08-21
docs(operations): add OrcaRouter as an external LLM provider in the LiteLLM guide

XiaoHuo888-hue:
## Summary

Add a named **OrcaRouter** external LLM provider to the LiteLLM "Serve External APIs" guide, mirroring the existing Google Gemini external provider pattern.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible gateway that smart-routes `orcarouter/auto` to an appropriate upstream model. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

## What changed

- `docs/operations/serve-external-apis/litellm.md`
  - **Step 1 (Secrets)**: added `ORCAROUTER_API_KEY` export and `litellm-env` secret literal.
  - **Step 3 (`model_list`)**: added a named `orcarouter-auto` entry:
    ```yaml
    - model_name: orcarouter-auto
      litellm_params:
        model: openai/orcarouter/auto
        api_base: https://api.orcarouter.ai/v1
        api_key: os.environ/ORCAROUTER_API_KEY
    ```
  - **Step 4 (Verify)**: updated `/v1/models`, `/key/generate`, the inference example (new subsection C), and `/key/info` to cover `orcarouter-auto`.
- `docs/operations/serve-external-apis/README.md`
  - Added OrcaRouter to the section intro and the architecture diagram.

## Validation

- `markdownlint` (repo `.markdownlint.json`) — clean.
- `check-yaml` / `check-json` / `trailing-whitespace` / EOF / mixed-line-ending (pre-commit hooks) — clean.
- YAML `model_list` block parses with all four entries intact.
- Live connectivity: `POST https://api.orcarouter.ai/v1/chat/completions` with model `orcarouter/auto` returned **HTTP 200** (`ORCA-LIVE-OK`, finish `stop`).

No code or cluster behavior is changed; this is a docs-only addition.

Disclosure: I'm an engineer on the OrcaRouter team.


nicolexin:
Hi @XiaoHuo888-hue Thanks for putting this together! I have two questions

1. How does cost tracking work?

The note added says:
```
The orcarouter-auto model is routed through the OrcaRouter gateway, so model_info pricing and budgets are handled by LiteLLM based on token usage, exactly as with gemini-flash.
```

I am not sure how that works. gemini-flash is priced automatically because it resolves to a first-party gemini/ model in LiteLLM's built-in price map. `orcarouter/auto` picks the best model for you and `orcarouter/auto` itself is not in LiteLLM's map, so my understanding is that LiteLLM records cost 0.0, spend never increments, and the budget capping won't apply to this route. 

2. Does this fit the section structurally?

Please correct me if I am wrong here. OrcaRouter is a model aggregator — its catalog is largely OpenAI, Gemini, Anthropic models etc, which are the same providers this guide already documents connecting to directly. So adding it routes traffic through an intermediary to reach providers already covered, and it changes the data path to client → LiteLLM → OrcaRouter → provider.

IIUC OrcaRouter's key feature is adaptive routing, and that's what LiteLLM is already doing in this guide — routing and fallbacks in the model list, and virtual keys, budgets, spend tracking, rate limiting. So this adds a second routing-and-governance layer behind the first one. Feature-wise, I think OrcaRouter should sit next to Kong and LiteLLM, however, it's not something we can deploy in the cluster and so it also does not compose with the llm-d stack.

Gregory-Pereira:
bump on this whats the status

PR #2350, opened 2026-08-21
feat(guides): add diffusion-serving image-to-image guide

zetxqx:
## What

Adds `guides/diffusion-serving/image-to-image`: deploy `Qwen/Qwen-Image-Edit-2511` behind the llm-d router on vLLM-Omni or SGLang, served over the OpenAI-compatible `POST /v1/images/edits` endpoint, which carries `multipart/form-data`.

## Why

The default parser chain covers the endpoint: `openai-parser` parses the multipart form and extracts the model and the scalar form fields (llm-d/llm-d-router#2562), so the router values only configure load-based scoring with `active-request-scorer`, the same signal as the text-to-image guide. Both engines run the same checkpoint, so they are comparable under one router configuration.

## Verified on GKE

- A multipart edit through the router on SGLang v0.5.17 returns a valid edited image in about 15 s on one H100.
- All kustomize overlays build; the values file renders through `helm template`.
- The overlays use the release component image defaults (SGLang v0.5.16, vLLM-Omni v0.26.0); those exact tags have not been exercised end to end, so this PR stays draft until they are.

```release-note
Added a diffusion-serving image-to-image guide: serve Qwen/Qwen-Image-Edit-2511 on vLLM-Omni or SGLang behind the llm-d router.
```

### DIFF
--- a/guides/diffusion-serving/image-to-image/README.md
+++ b/guides/diffusion-serving/image-to-image/README.md
@@ -1,0 +1,248 @@
+# [Experimental] Diffusion Image-to-Image Guide
+
+This guide deploys an image editing model behind the llm-d router and serves it over the OpenAI-compatible `POST /v1/images/edits` endpoint, which carries `multipart/form-data` because the OpenAI API accepts image uploads on it.
+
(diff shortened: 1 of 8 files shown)


ahg-g:
pls rebase

zetxqx:
> pls rebase

done!

Issue #1983, opened 2026-07-01
[Observability][Feature]: Consolidate observability assets for llm-d


gyliu513:
### Feature Area

Optimized baseline

### Problem Statement

Here are some tasks we need follow up for each repo in llm-d org.

* **Move component user-facing assets into `llm-d/llm-d`.** Migrate user-deployable scrape manifests, `PrometheusRule` alerts, and dashboards currently in component `deploy/` folders (e.g. [llm-d-router#1668](https://github.com/llm-d/llm-d-router/pull/1668)) into llm-d recipes; leave only definitional/dev-CI artifacts behind.
* **Render metric references on llm-d.ai.** Ensure each component's metric reference is present under `docs/operations/observability/` and kept in sync with the component's `docs/metrics.md` (per [llm-d-router#1636](https://github.com/llm-d/llm-d-router/pull/1636#issuecomment-4696434898)).

Check https://github.com/llm-d/llm-d/blob/main/proposals/observability-integration.md for more detail.

Please feel free to open sub issues for separate repos and link it back to this parent issue.

One example issue: https://github.com/llm-d/llm-d-router/issues/1855

### Proposed Solution

see above

### Willingness to Contribute

Yes, I can submit a PR

gyliu513:
## Sub tasks
- [x] Router https://github.com/llm-d/llm-d-router/issues/1855
- [ ] WVA https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1466
- [ ] kv cache https://github.com/llm-d/llm-d-kv-cache/issues/708
- [ ] batch gateway https://github.com/llm-d/llm-d-batch-gateway/issues/619
- [ ] IPP Grafana dashboard
- [ ] TBA

PR #2363, opened 2026-08-24
docs(observability): add WVA dashboards and alerts to llm-d

SrikarChittemsetty:
## What this PR does

Moves the Workload-Variant-Autoscaler's user-facing observability assets into `llm-d/llm-d`, per the [observability-integration proposal](https://github.com/llm-d/llm-d/blob/main/proposals/observability-integration.md) and #1983. This is the llm-d half of [wva#1466](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1466); the cleanup PR in the component repo follows once this lands, as that issue's acceptance criteria require.

It mirrors the layout the EPP migration established rather than inventing one — dashboards alongside the existing seven, the `PrometheusRule` as a sibling of `epp-alerting-rules.yaml`.

**Added:**

| File | Source in wva |
|---|---|
| `guides/recipes/observability/grafana/dashboards/llm-d-wva-operational.json` | `deploy/grafana/operational-dashboard.json` (16 panels) |
| `guides/recipes/observability/grafana/dashboards/llm-d-wva-benchmark-scaleup.json` | `deploy/grafana/benchmark-dashboard.json` (5 panels) |
| `guides/recipes/observability/alerts/wva-alerting-rules.yaml` | `config/components/prometheus-alerts/prometheusrule.yaml` (5 alerts) |

## The dashboards are not copied verbatim

This is the part worth reviewing closely. The two dashboards did not use this repo's datasource convention, and copying them across unchanged would have produced dashboards that load but don't bind:

- The operational dashboard bound every panel to a `$datasource_uid` template variable — a name that exists only in the wva repo.
- The benchmark dashboard had **no datasource variable at all** and no `uid` on its datasource references, relying on whatever Grafana had set as default.

Every dashboard already here uses `${DS_PROMETHEUS}` with a standard `datasource`-type template variable. Both files were retargeted — **47 datasource references in total** — and given that variable, so they behave like their neighbours. `load-llm-d-dashboards.sh` globs `*.json`, so no registration was needed.

I left the panel content, queries, and dashboard `uid`s untouched; the `uid`s are stable so existing links keep working.

## Alerts

Renamed to `wva-alerting-rules` and relabelled `app: wva-metrics`, matching `epp-alerting-rules.yaml`. The component repo used `release: kube-prometheus-stack`, which is a kube-prometheus-stack discovery convention rather than this repo's.

Worth noting: the `ruleSelector` caveat already documented in `alerting.md` Step 1 applies here identically — in central mode `ruleSelector: {}` picks up any rule, and in scoped mode neither this file nor the EPP one carries the `monitoring-ns` label. I have not changed that behaviour, only matched the existing precedent.

## Docs

- `alerting.md` — intro now covers both components, plus a WVA section with apply/verify steps and a reference table for the five `wva.rules` alerts. These fire on autoscaler health rather than request-path health, which the section calls out.
- `metrics.md` — the expected `kubectl get configmaps` output now lists the two new dashboards, so the sample stays accurate.

## What is deliberately not here

The ServiceMonitor and the `docs/user-guide/monitoring.md` runbook. I asked on [wva#1466](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1466) whether the scrape manifest belongs under `guides/workload-autoscaling/.../monitoring/` or in the shared recipe directory — the issue says "or equivalent recipe path" and I would rather match @gyliu513's intent than guess and have it moved twice. Happy to add both here once that's settled, or in a follow-up.

## Testing

Both dashboards parse as valid JSON and retain their panel counts; no `$datasource_uid` references remain. The `PrometheusRule` retains all five alerts across the single `wva.rules` group.

I have not applied these to a live cluster — I don't have a Grafana/Prometheus stack with WVA running to hand, so the datasource rebinding is verified by structure rather than by rendering. Worth a look from someone who can load them.

---

*Disclosure: prepared with AI assistance. The asset inventory, the datasource-convention mismatch, and the alert reference were checked against both repositories before writing.*

### DIFF
--- a/docs/operations/observability/alerting.md
+++ b/docs/operations/observability/alerting.md
@@ -1,6 +1,6 @@
 # Alerting
 
-This page covers a default set of Prometheus alerting rules for the EPP (Endpoint Picker). For Prometheus and Grafana installation, see [Observability Setup](./setup.md) first, and for the metrics these alerts are built on, see [Metrics](./metrics.md).
+This page covers the default sets of Prometheus alerting rules shipped for the EPP (Endpoint Picker) and the Workload-Variant-Autoscaler (WVA). For Prometheus and Grafana installation, see [Observability Setup](./setup.md) first, and for the metrics these alerts are built on, see [Metrics](./metrics.md).
 
 The rules ship as a [`PrometheusRule`](https://prometheus-operator.dev/docs/getting-started/design/#prometheusrule) custom resource, so they require the Prometheus Operator (bundled with the kube-prometheus-stack installed by the [setup guide](./setup.md)).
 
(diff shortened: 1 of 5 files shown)


SrikarChittemsetty:
Thanks @gyliu513, and sorry for the slow reply on this.

I went and checked the datasource binding across all seven, because I wanted to know whether mine were actually the odd ones out. The picture is more mixed than either of us described:

| Dashboard | `__inputs` | `templating.list` | datasource var |
|---|---|---|---|
| llm-d-diagnostic-drilldown | yes | 3 | `DS_PROMETHEUS` |
| llm-d-failure-saturation | yes | 3 | `DS_PROMETHEUS` |
| llm-d-inference-gateway | **no** | 1 | `DS_PROMETHEUS` |
| llm-d-pd-coordinator-metrics | yes | 0 | none |
| llm-d-performance-kv-cache | yes | 3 | `DS_PROMETHEUS` |
| llm-d-sglang-overview | yes | 3 | `DS_PROMETHEUS` |
| llm-d-vllm-overview | yes | 3 | `DS_PROMETHEUS` |
| **llm-d-wva-operational** (this PR) | no | 4 | `DS_PROMETHEUS` |
| **llm-d-wva-benchmark-scaleup** (this PR) | no | 1 | `DS_PROMETHEUS` |

So five of the seven carry `__inputs` *and* a live `type: datasource` variable, which is the combination you get from a Grafana export. `llm-d-inference-gateway` has no `__inputs` and just the variable, which is exactly the shape mine are in. Only `llm-d-pd-coordinator-metrics` has `__inputs` with an empty `templating.list`.

My reading is that `__inputs` is the dashboard-import mechanism, and `load-llm-d-dashboards.sh` does not use the import API. It writes ConfigMaps labelled `grafana_dashboard=1` for the sidecar to provision, and on that path the `${DS_PROMETHEUS}` variable is what does the binding. That is consistent with `llm-d-inference-gateway` working without `__inputs`.

That said, you are right that this is the part worth checking rather than assuming, and I want to be straight that I have not verified it: I said in the PR description that I have not loaded these into a real Grafana, and that is still true. I do not have a cluster with Grafana and WVA to hand.

Two ways forward, your call:

1. Someone loads them once and confirms the panels bind. That settles it either way.
2. I add `__inputs` to both so they match the five-dashboard majority rather than the one-dashboard minority. It costs nothing and removes the question, even if my reading above is right.

Happy to do 2 now if you would rather not spend a cluster on it. Just say which.

On the other three points, thanks, that unblocks the WVA-side cleanup. ServiceMonitor under `guides/workload-autoscaling/`, relabel dev/CI in place rather than moving anything `install.sh` or e2e depends on, and leave new rules to #919. I will open that PR against the component repo and link it here.

PR #2366, opened 2026-08-24
Slack notification workflow

### LINKED ISSUE #1548: [CI/CD] Add daily Slack report for nightly test status
## Summary

Add a daily Slack report summarizing the status of all nightly CI workflows. This improves operational visibility into CI health and helps teams quickly identify failing tests requiring attention.

## Proposed Solution

Create `.github/workflows/nightly-status-report.yaml` that runs daily at 12:00 UTC (after all nightlies complete) and posts a Slack message summarizing results.

**Implementation:**
1. Create Slack incoming webhook for `#llm-d-ci` (or similar channel)
2. Add `SLACK_NIGHTLY_WEBHOOK` to GitHub secrets
3. Create workflow that:
   - Fetches last run status for each nightly workflow
   - Aggregates results (passed/failed counts)
   - Formats Slack message with links to failed runs
   - Posts via webhook

**Future enhancements:**
- Track consecutive failures (add emoji for chronic issues)
- Compare today vs yesterday (trend indicators)
- Alert on-call when critical tests fail
- Weekly summary with pass rate graphs

## Deliverables

- [ ] `.github/workflows/nightly-status-report.yaml` - Daily summary workflow
- [ ] Documentation in CONTRIBUTING.md about Slack bot
- [ ] Slack webhook configured (org admin task)
- [ ] Test run and validate message format

## Related Issues

- #1518 - Document CI system (this adds operational visibility)
- #1520 - v0.8 Release Goals (sig-cicd: solidify CI coverage and operations)

## Labels

`release/v0.8`, `sig-cicd`, `enhancement`, `good first issue`

## DRI

TBD (volunteer welcome!)

diegocastanibm:
## What

Posts the result of every **scheduled** nightly to the Slack channel of the SIG that owns it, from within the repo instead of via the GitHub app for Slack.

Closes #1548.

## Why

The GitHub app for Slack (`/github subscribe`) hit two limits we could not work around: it only supports **one workflow subscription per channel** (so `#sig-router`, which owns 16 nightlies, could never be covered), and it **also fired on manual `workflow_dispatch` runs**.

## How

```
.github/slack-channels.yaml          <- source of truth (keyed by workflow FILE name)
        |  scripts/sync-slack-channels.py --fix  (generates the `workflows:` list)
        v
.github/workflows/notify-slack-nightly.yaml   <- GENERATED list, do not hand-edit
        |  at runtime
        v
.github/scripts/notify-slack-nightly.py       <- routes on workflow_run.path
```

Two identifiers, deliberately: the **trigger** must use display names because `workflow_run` matches on nothing else, which makes it the part that breaks *silently* — rename a nightly and notifications just stop, indistinguishable from a quiet night. Generating that list turns the rename red in the same PR. **Runtime routing** uses the workflow *file* name, which is stable, so a rename can only stop notifications, never misroute them.

A nightly in neither `channels` nor `skip` fails the new `sync-slack-channels` pre-commit hook — the only place the mistake can be caught, since an unrouted nightly is also absent from the trigger and therefore never fires at all.

**Routing**: 40 nightlies across 7 channels, 4 explicitly skipped with a reason (= all 44 `nightly-*.yaml`). `#sig-router` 16, `#sig-kv-disaggregation` 7, `#llm-d-intel` 6, `#sig-pd-disaggregation` 5, `#sig-autoscaling` 3, `#sglang-collab` 2, `#fast-model-actuation` 1. Note SGLang and Intel XPU route by engine/accelerator rather than by guide, because that is who acts on a failure. Run `python scripts/sync-slack-channels.py --audit` for the full table.

## Decisions worth a look in review

- **Only `schedule` runs notify** — the whole point of the change. The filter must live in the job's `if:`; `workflow_run` cannot filter on the upstream run's own event.
- **Re-runs of a scheduled run DO notify**, marked `(attempt N)`: when a SIG re-runs a nightly that failed on a transient cluster error, that result is the current truth about the guide.
- **`cancelled` / `skipped` do not notify.** Every nightly sets `cancel-in-progress`, so dispatching one mid-flight cancels the previous run — not news for a channel.
- **The guide's result comes from the `nightly` jobs, not the run conclusion.** `update-badge` runs with `if: always()`, so a green test whose badge push failed surfaces as a failed *run*; reporting that verbatim would tell a channel its guide is broken when it is not. That case gets its own `test passed, update-badge failed` message.
- **Unrouted at runtime → `#llm-d-ci-alerts` + warning, not a failed job.** A red notifier is a second alert channel nobody watches.
- Slack action **pinned by SHA**; `permissions` limited to `contents: read` + `actions: read`, which matters because a `workflow_run` workflow gets secrets and a privileged token even when the upstream run had neither.

## Files

New: `.github/slack-channels.yaml`, `scripts/sync-slack-channels.py` (`--check`/`--fix`/`--audit`, modelled on `sync-nightly-matrix.py` and reusing `matrix_common.py`), `.github/workflows/notify-slack-nightly.yaml`, `.github/scripts/notify-slack-nightly.py`.

Modified: `.pre-commit-config.yaml` (new hook), `.github/workflows/README.md` (new "Slack notifications" section + a third step in "Adding a new guide").

No nightly workflow is touched, and nothing is needed in `llm-d-infra`.

## Testing

Validated against real runs (success, failure, cancelled), including `32211600170`, which renders identically to what the GitHub app posted for it:

```bash
GH_TOKEN=$(gh auth token) python .github/scripts/notify-slack-nightly.py \
    --repo llm-d/llm-d --run-id <run-id> --dry-run
```
A real test can be seen in `#llm-d-ci-alerts` where the app has posted a filing workflow using:
```
  GH_TOKEN=$(gh auth token) .venv/bin/python .github/scripts/notify-slack-nightly.py \
    --repo llm-d/llm-d --run-id 32737554939 --channel '#llm-d-ci-alerts' --json \
  | curl -s -X POST https://slack.com/api/chat.postMessage \
      -H "Authorization: Bearer $SLACK_BOT_TOKEN" \
      -H 'Content-type: application/json; charset=utf-8' \
      --data-binary @- | jq '{ok, error, channel}'
```
Plus 25 unit checks over the pure functions, all five guardrail failure modes verified to fail actionably and be fixed by `--fix`.

## Required before this takes effect

Needs Slack + repo admin; full checklist with an app manifest is in `.github/workflows/README.md`.

- [x] Create and install the Slack app with the **`chat:write`** bot scope
- [x] Create `#llm-d-ci-alerts` (fallback channel)
- [x] `/invite @llm-d-ci` in all 8 channels — skipping this is the likely first-deploy failure (`not_in_channel`)
- [x] Add the `SLACK_BOT_TOKEN` repository secret

## Follow-ups

The 3 multimodal-serving nightlies need a SIG owner and a channel; they sit in `skip` until then. `nightly-build-image` is unrouted by design — happy to point it at a CI channel if reviewers prefer.

@llm-d/release-crew

### DIFF
--- a/.github/scripts/notify-slack-nightly.py
+++ b/.github/scripts/notify-slack-nightly.py
@@ -1,0 +1,363 @@
+#!/usr/bin/env python3
+"""Build the Slack message for a finished nightly run.
+
+Usage:
+  # Render without posting -- works locally against any real run.
+  GH_TOKEN=$(gh auth token) python .github/scripts/notify-slack-nightly.py \
+      --repo llm-d/llm-d --run-id 32211600170 --dry-run
+
+  # In CI: write `channel` and `payload` to $GITHUB_OUTPUT for the posting step.
+  python .github/scripts/notify-slack-nightly.py \
+      --repo "$GITHUB_REPOSITORY" --run-id "$RUN_ID" --github-output
+
+The run is fetched from the API by id rather than read out of the workflow_run
+event payload. That keeps one code path for the real trigger and for a manual
+replay, and -- more importantly -- the event payload cannot answer the question
+this script has to answer: see nightly_outcome() below.
+
+Routing is keyed off the workflow's file name, taken from the run's `path`,
+because file names are stable while display names are not. See
+.github/slack-channels.yaml.
+"""
+
+import argparse
+import json
+import os
+import sys
+import urllib.error
+import urllib.request
+from datetime import datetime
+from pathlib import Path
+
+import yaml
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+MAPPING_PATH = REPO_ROOT / ".github" / "slack-channels.yaml"
+
+API_ROOT = "https://api.github.com"
+
+# Jobs that publish the gh-pages badge rather than exercise the guide. They run
+# with `if: always()`, so they are present even when the test failed.
+BADGE_JOB_PREFIX = "update-badge"
+
+FAILED_CONCLUSIONS = ("failure", "timed_out")
+
+# Outcomes worth a message. The notify workflow also filters on the run-level
+# conclusion so it can skip spinning up a runner, but this is the authoritative
+# check: a manual replay bypasses that filter, and "nothing to say about this
+# run" is a property of the run, not of how the script was invoked.
+#
+# "cancelled" is the common exclusion: every nightly sets cancel-in-progress, so
+# dispatching one while a run is in flight cancels that run, which is not news.
+# "skipped" means no job actually exercised the guide.
+NOTIFIABLE_OUTCOMES = ("success", "failure", "timed_out")
+
+
+# ---------------------------------------------------------------------------
+# GitHub API
+# ---------------------------------------------------------------------------
+
+
+def api_get(path: str, token: str) -> dict:
+    request = urllib.request.Request(
+        f"{API_ROOT}{path}",
+        headers={
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"Bearer {token}",
+            "X-GitHub-Api-Version": "2022-11-28",
+            "User-Agent": "llm-d-notify-slack-nightly",
+        },
+    )
+    try:
+        with urllib.request.urlopen(request, timeout=30) as response:
+            return json.load(response)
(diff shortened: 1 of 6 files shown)


Gregory-Pereira (approved):
/lgtm

PR #2372, opened 2026-08-25
 Fix stale forward-looking version claims in WVA guides for 0.9.0

### LINKED ISSUE #2357: guides(wva): README shipped in v0.9.0 describes 0.9.0 as a future release (CRD removal, saturation engine v2 default)
**Area:** guides / workload-autoscaling (WVA)
**Path:** Docs
**Severity:** medium — the guide describes the release it ships in as a future event, so a reader on v0.9.0 cannot tell whether the CRD they are about to use still exists or whether the engine they are enabling manually is already the default.
**Build under test:** tag `v0.9.0` (commit `3291bca445be5bd309387fa78cc24f487f07003d`)

## Description

`guides/workload-autoscaling/wva/README.md` talks about 0.9.0 in the future tense, inside the 0.9.0 tag:

- line 5: "The VariantAutoscaling CRD **will be removed in 0.9.0**."
- line 129: "Saturation engine v2 **will be the default in the next release (0.9.0)**, but for now it must be enabled manually. The v1 saturation engine will be deprecated in 0.9.0 and removed in 0.10.0."

A reader on this tag is left with two unanswered questions: is `VariantAutoscaling` still available in the version they are running, and does enabling saturation engine v2 by hand still apply?

## Repro

1. `git checkout v0.9.0`
2. `grep -n '0\.9\.0' guides/workload-autoscaling/wva/README.md`
3. Both statements refer to 0.9.0 as upcoming.

## Expected

At the 0.9.0 tag, the guide states the current state: whether the CRD was removed, and whether v2 is now the default.

## Actual

Statements written for a pre-0.9.0 reader, shipped unchanged in 0.9.0.

## Probable root cause

Release-cycle text written during 0.8.x and not revisited at the 0.9.0 cut.

## Suggested fix

Reword to the state of this release, and prefer explicit version comparisons ("as of 0.9.0, X") over "the next release", which goes stale silently at every cut.

## Scope

Related drift in the same area: `docs/architecture/advanced/autoscaling/hpa-wva.md` opens with a deprecation notice for the VariantAutoscaling approach, so a reader gets the deprecation from one page and the removal timeline from another, neither anchored to the version they are on.

---
Found while building a documentation-derived knowledge base from the v0.9.0 tree, by reading every version claim against the tag it ships in. No secrets in this report.

Neha-dot-Yadav:
Fixes #2357

`guides/workload-autoscaling/wva/README.md` and
`docs/architecture/advanced/autoscaling/hpa-wva.md` described the 0.9.0 release in future tense even though they ship inside the 0.9.0 tag itself:

- "The VariantAutoscaling CRD will be removed in 0.9.0",  but the CRD was already removed as of 0.9.0 (confirmed against
  llm-d-workload-variant-autoscaler release-0.9, which no longer ships
  a CRD manifest).
- "Saturation engine v2 will be the default in the next release (0.9.0), but for now it must be enabled manually", but v2 is
  already the default as of 0.9.0 (confirmed via the shipped saturation ConfigMap, which already includes the `analyzers:`
  section that selects v2).
  
A reader on the v0.9.0 tag had no way to tell from the docs whether these "future" events had already happened.

### DIFF
--- a/docs/architecture/advanced/autoscaling/hpa-wva.md
+++ b/docs/architecture/advanced/autoscaling/hpa-wva.md
@@ -1,7 +1,7 @@
 # HPA/KEDA with WVA Metrics
 
 > [!WARNING]
-> **Deprecation Notice:** The `VariantAutoscaling` (VA) CRD-based approach described in this document is deprecated. The recommended path is the [HPA + WVA guide](../../../../guides/workload-autoscaling/wva/README.md), which configures HPA directly with WVA-published metrics without requiring the `VariantAutoscaling` CRD. If you are currently using VA objects, see the [Migration from VA to HPA + WVA](#migration-from-va-to-hpa--wva) section below.
+> **Deprecation Notice:** The `VariantAutoscaling` (VA) CRD-based approach described in this document was deprecated in llm-d 0.8.0, and the CRD was removed in 0.9.0. The recommended path is the [HPA + WVA guide](../../../../guides/workload-autoscaling/wva/README.md), which configures HPA directly with WVA-published metrics without requiring the `VariantAutoscaling` CRD. If you are currently using VA objects, see the [Migration from VA to HPA + WVA](#migration-from-va-to-hpa--wva) section below.
 
 ## Functionality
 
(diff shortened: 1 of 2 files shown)


Neha-dot-Yadav:
Hi @paulohenriquevn,
This same stale text exists unchanged on release-0.9. 
Should this doc fix also be backported there, or is release-0.9 docs-frozen and only main/future releases get corrected?

lionelvillard:
we may not need this anymore. @Neha-dot-Yadav can you rebase and check whether this PR is still needed?

PR #2379, opened 2026-08-25
Make DisaggregatedSet the base of the wide-ep guide

BenjaminBraunDev:
## What

Makes `DisaggregatedSet` the way the wide-EP guide deploys its NVIDIA GPU model servers.

- The `DisaggregatedSet` manifests replace the LeaderWorkerSet prefill/decode manifests as `modelserver/gpu/vllm-deepseek-r1-0528/base`.
- The provider overlays (`gke`, `coreweave`, `dgx-cloud-gb200`) compose off the DS base. The separate `disaggregatedset/` tree and the `legacy-lws` overlay are removed.
- The `topology-aware/` overlays are dropped, to be re-added once topology aware placement is validated with `DisaggregatedSet`.
- The generator script is removed. The DS manifest is the source of truth.
- The gke overlay inlines the NCCL tuner/net plugin env vars because `disable-gke-nccl-tuner-patch` targets Deployment/LWS and does not match the DisaggregatedSet kind.
- LWS prerequisite is now v0.10.0+ installed with `--set enableDisaggregatedSet=true`.
- The Intel XPU path is unchanged (plain LWS).
- Pins each role's `rolloutStrategy` to maxSurge 0, maxUnavailable 1 to match previous LWS defaults.
- Adds the `disaggregatedset-rollout-screener` to the EPP config so prefill and decode never pair across revisions during a rollout. The XPU overrides keep the screener-free config since XPU stays on plain LWS.
- Migrates the gke overlay to DRA with GKE managed DRANET via the shared `gke-rdma` component.
- Adds cluster prerequisites to the GKE overlay README. Repo-wide prereq work: #2429.
- Fixes a broken link in the pd-disaggregation README.

Every overlay except gke renders byte-identical to its pre-restructure `DisaggregatedSet` render, which matches the LWS pod specs it replaces (env sorted by name, ServiceAccount name differs by design).

> [!NOTE]
> Before this merges, nightly wide-EP e2e needs LWS v0.10.0 with `enableDisaggregatedSet=true`. That change goes in llm-d/llm-d-benchmark, whose standup installs LWS. Standup skips the install when the LWS CRD already exists, so already-provisioned e2e clusters need a one-time helm upgrade.


guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/kustomization.yaml
```diff
@@ -2,18 +2,21 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
+# DisaggregatedSet role pod templates: roles/0 is prefill, roles/1 is decode.
+# The disable-gke-nccl-tuner-patch component targets Deployment/LeaderWorkerSet
+# and does not match the DisaggregatedSet, so the equivalent env vars are set
+# in the patches below.
 patches:
   - target:
-      kind: LeaderWorkerSet
-      name: wide-ep-llm-d-prefill
+      kind: DisaggregatedSet
+      name: wide-ep-lws-nvidia-gpu-vllm
```
ahg-g, line 14: as a followup, lets please drop the lws substring, also this is a long name, I think we better shorten it because we may hit limits on the names of the objects that gets created by disaggset.

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/kustomization.yaml
```diff
@@ -145,31 +194,93 @@ patches:
               {"interfaceName":"eth9","network":"rdma-7"}
             ]
       - op: add
-        path: /spec/leaderWorkerTemplate/workerTemplate/spec/containers/0/env/-
+        path: /spec/roles/1/spec/leaderWorkerTemplate/workerTemplate/metadata/annotations
```
ahg-g, line 197: why are we adding these?

@huaxig can you please review

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/README.md
```diff
@@ -12,6 +12,64 @@ This overlay configures GKE-specific settings for DP-aware WideEP scheduling on
 | RDMA network annotations | Configures multi-NIC RDMA interfaces (eth2-eth9 → rdma-0 through rdma-7). |
 | `DEEP_EP_DEVICE_TO_HCA_MAPPING` | Maps GPUs to NICs for efficient NVSHMEM NIC selection. |
 | `NVSHMEM_DISABLED_GDRCOPY` | Recommended on GKE. |
-| `HF_HUB_DISABLE_XET` | Disables HF XET when loading model from host storage. |
 | Host volumes | GKE-specific hostPath for model and JIT caches. |
-| `disable-gke-nccl-tuner-patch` | Disables GKE's built-in NCCL tuner. |
+| `NCCL_TUNER_PLUGIN` / `NCCL_NET_PLUGIN` | Disables GKE's built-in NCCL tuner and net plugin. |
+
+## Cluster Prerequisites
```
ahg-g, line 18: @huaxig pls review

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/topology-aware/gke-a4/kustomization.yaml
```diff
@@ -1,18 +1,6 @@
+# Same as the topology-aware gke overlay; kept as a separate overlay to
```
ahg-g: I suggest to completely drop the topology-aware pkg and we add it later as a followup. Topology-aware placement with kueue needs to be revisited and tested throughly with disaggregatedSet

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/kustomization.yaml
BenjaminBraunDev: Agreed, will handle in a follow-up. I can shorten it along with the child object names in mind, something like `wide-ep-vllm`

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/kustomization.yaml
BenjaminBraunDev: These are carried over from the current overlay. It used to target kind: LeaderWorkerSet so it applied to both prefill and decode. With DisaggregatedSet the patch needs to name each role path so the block appears once on each.

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/topology-aware/gke-a4/kustomization.yaml
BenjaminBraunDev: sounds good. It renders against the DS base but haven't tested topology aware placement w/ DS

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/README.md
BenjaminBraunDev: For context: this section is what a fresh cluster (w/ a3-ultra nodepool) needed before the guide worked. I found that
1. node pools created with --accelerator-network-profile=auto never finish deleting, so I document the manual VPC route
2. the default R580 driver crashes DeepEP high-throughput kernels with cudaErrorIllegalAddress

guides/wide-ep-lws/router/wide-ep-lws.values.yaml
hasB4K, line 1: are we not planning to have 2 EPP in the guide? 1 for prefill and 1 for decode?

guides/wide-ep-lws/router/wide-ep-lws.values.yaml
```diff
@@ -12,6 +12,15 @@ router:
...
         plugins:
+        # Pin each request to a single DisaggregatedSet revision during
+        # rollouts so prefill and decode never pair across revisions.
+        - type: disaggregatedset-rollout-screener
+          parameters:
+            scope:
+              labelSelector: "disaggregatedset.x-k8s.io/name=wide-ep-lws-nvidia-gpu-vllm"
+            revisionGating:
+              mode: max-role
+              requiredRoles: [prefill, decode]
```
yankay, line 23: Could we change this to:

requireRoles:
  values: [prefill, decode]

I rendered the Helm values and passed the resulting EPP config through the current router loader. The existing shape fails with `json: unknown field "requiredRoles"`, while the corrected shape loads successfully.

guides/wide-ep-lws/router/wide-ep-lws.values.yaml
ahg-g: I think starting with a single one is better since we are not yet deploying the coordinator by default

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/README.md
ahg-g: @BenjaminBraunDev as a followup, is there a way to define an lws spec that gets patched for both prefill and decode roles on the disaggset (just like what we have now) instead of patching individual parameters twice?

huaxig: @BenjaminBraunDev. 
for
>1. node pools created with --accelerator-network-profile=auto never finish deleting, so I document the manual VPC route

I strongly recommend we keep using `--accelerator-network-profile=auto` and avoid falling back to the **manual VPC route**.

Here is the architectural context for why:
1. The auto flag is what enables Managed DRANET, delegating the creation, configuration, and management of the subnets directly to the GKE Network controller. Using explicit `--additional-node-network` flags forces a fallback to the legacy DRANET approach. Especially now that Managed DRANET is released, we should not recommend the manual fallback in our official guides. Manual management requires extra configuration to maintain optimal performance and is prone to causing resource leaks.
2. Regarding the node pool deletion issue: I have tested the auto profile previously and it works correctly. The deletion process does introduce additional latency because GKE has to clean up all the auto-provisioned subnets in the background. If the deletion is genuinely wedging indefinitely for you, this is a bug rather than expected behavior. If you can confirm it is permanently stuck, we should cut an issue to the Network team to fix the root cause rather than changing our guides to recommend the legacy manual approach.

Let's stick to the auto profile here and route any hanging deletion issues to the network infrastructure team.

huaxig: >2. the default R580 driver crashes DeepEP high-throughput kernels with cudaErrorIllegalAddress

I don't have too many concerns with documenting the driver pin for now—using a specific, stable driver version is a acceptable mitigation plan to unblock users.

However, I am much more concerned about the underlying root cause, especially since this is triggered under high-throughput conditions. We should try to establish a stable reproduction path so we can deeply investigate why it's crashing. My strong suspicion is that a `cudaErrorIllegalAddress` occurring specifically under heavy load is likely an HBM management issue—highly likely a specific edge-case OOM scenario disguised as an illegal memory access, rather than a pure driver bug.

If we can stably reproduce and confirm this hypothesis, we might be able to avoid the crash entirely by tuning the appropriate memory/concurrency parameters instead of being locked out of the newer driver lines. Furthermore, isolating the root cause here gives us a great opportunity to contribute the fix or  guidance upstream to the community.

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/coreweave/kustomization.yaml
```diff
@@ -3,39 +3,72 @@ kind: Kustomization
...
 patches:
   - target:
-      kind: LeaderWorkerSet
+      kind: DisaggregatedSet
+      name: wide-ep-lws-nvidia-gpu-vllm
     patch: |-
       # We add the volume for different providers since the hostPath can be different.
       - op: replace
-        path: /spec/leaderWorkerTemplate/workerTemplate/spec/volumes/1
+        path: /spec/roles/0/spec/leaderWorkerTemplate/workerTemplate/spec/volumes/1
```
huaxig, line 15: Could we refactor this to use a structural patch instead? Matching explicitly by name: hf-cache and name: jit-cache ensures safe merging regardless of array order

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/kustomization.yaml
huaxig: Good catch! @ahg-g. 

@BenjaminBraunDev, we actually have a critical issue here. These annotations are a very legacy workaround.

With the GKE Managed DRANET setup, we absolutely do not need to manually patch any of the `networking.gke.io/interfaces` or other annotations. Please refer to our latest P/D  overlay recipes, where we have already entirely eliminated the need for these manual network patches. 

If you need any help updating this section to align with the DRANET standards, just let me know and I’d be happy to help push a fix for this part!

huaxig: To solve this cleanly, I have already standardized the GKE RDMA overlay into `/guides/recipes/modelserver/components/gke-rdma`, which can be directly reused here. For reference, the **modelexpress p2p** well-lit path is also reusing this exact component.

guides/wide-ep-lws/router/wide-ep-lws.values.yaml
BenjaminBraunDev: What EPP image did you test with? The field was renamed in llm-d-router#2571 on september 8th. `requireRoles.values` is the old shape and `requiredRoles` is current. I believe `requiredRoles` is correct going forward

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/kustomization.yaml
BenjaminBraunDev: Okay I will switch this PR to the gke-rdma component with Managed DRANET.

Though I want to note that while PD and modelexpress use use NIXL for transfers, wide-EP uses DeepEP and the GPU drives the NIC directly. I cannot get H200 capacity I need to test this on DRANET at the moment, so I have not verified.

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/README.md
BenjaminBraunDev: @huaxig 
1. I switched it to the auto profile and made the README point to the DRA/DRANET setup docs.

2. Oh it's not that it happened at high throughput, "High throughput" is the DeepEP kernel variant name. For me, the crash happened every boot during vllm profile_run for that specific kernel, and not say, the low-latency variant. Setup: GKE 1.36 RAPID default driver, A3 Ultra H200, and ran through using this guide's prefill manifest

Also I can confirm it was permanently stuck for 12+ days in STOPPING and it blocked any new auto-profile pools on the cluster, which is why I kept a warning in the readme. I can follow up with more info if need be. (this has also happened on a few other clusters I made to test RDMA)

BenjaminBraunDev: @ahg-g 
Added an openapi schema in the coreweave overlay to allow roles/volumes to match by name. I still don't think you can have a single spec apply to both roles? not without adding support for it in the DisaggregatedSet API itself.

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/coreweave/kustomization.yaml
BenjaminBraunDev: Done. Kustomize needed a schema for the DisaggregatedSet merge keys so I added in the overlay, then made it a strategic merge patch with name matching.

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/README.md
BenjaminBraunDev: Oh actually we can make this work with anchors and <<, since *common lets a role add its own keys on top for the parts that differ. Added this to the PR for gke and coreweave.

guides/wide-ep-lws/modelserver/gpu/vllm-deepseek-r1-0528/gke/disaggregatedset-schema.json
```diff
@@ -0,0 +1,100 @@
+{
```
ahg-g, line 1: any chance we can unify this across providers? or does it have to be in the same directory as the patch?

ahg-g:
/approve

Great! pls lets followup with the other guides.

guides/wide-ep-lws/router/wide-ep-lws.values.yaml
yankay: Thanks for catching this. I checked the older schema used in `v0.10.0`; `requiredRoles` is indeed correct on `main`.

PR #2382, opened 2026-08-26
docs(api): document requestHandler.parsers as a list

### LINKED ISSUE #2355: docs(api): `requestHandler.parser` (object) in the reference vs `requestHandler.parsers` (list) in the gRPC guide
**Area:** docs / API reference (EndpointPickerConfig)
**Path:** Docs
**Severity:** medium — the two pages describe incompatible YAML shapes for the same field, and a reader has no way to tell which one the EPP accepts. One of them makes the config fail to load.
**Build under test:** tag `v0.9.0` (commit `3291bca445be5bd309387fa78cc24f487f07003d`)

## Description

`requestHandler` is documented with two different shapes in the same API reference:

- `docs/api-reference/endpointpickerconfig.md:77`, under `## RequestHandlerConfig`:

  | Field | Description |
  | --- | --- |
  | `parser` | [ParserConfig](#parserconfig) <br/> Specifies the parsing logic for protocol messages. |

  Singular, a single object, with `ParserConfig.pluginRef` **Required**.

- `docs/api-reference/epp-grpc-apis.md:29-32`, in a complete working example:

  ```yaml
  requestHandler:
    parsers:
    - pluginRef: vllmgrpcParser
  ```

  Plural, a list.

A user configuring a gRPC model server follows the second page; a user reading the reference table writes the first. Only one can be right.

## Repro

1. `git checkout v0.9.0`
2. `grep -n 'parser' docs/api-reference/endpointpickerconfig.md` → `parser` (object) at line 77.
3. `grep -n 'parsers:' docs/api-reference/epp-grpc-apis.md` → `parsers:` (list) at line 31.

## Expected

One shape, documented consistently. If both are accepted (for example a deprecated singular form and a current plural one), the reference says so and states the precedence.

## Actual

Two shapes, each presented as correct, with no cross-reference between them.

## Probable root cause

The field appears to have moved from a single parser to a list, and the reference table was not updated — the same pattern as the `saturationDetector` move tracked in llm-d-router#1308.

## Suggested fix

Confirm the current shape against `apix/config/v1alpha1` in `llm-d-router` at the pinned EPP version (`v0.10.0`), then update `docs/api-reference/endpointpickerconfig.md` — including the `ParserConfig` section and its `Required` marker — to match, keeping a deprecation note if the singular form is still accepted.

## Related

Same class of drift as the `saturationDetector` documentation issue filed alongside this one.

---
Found while building a documentation-derived knowledge base from the v0.9.0 tree; the two API-reference pages were read against each other. No secrets in this report.

rishabhsinha17:
## Summary
`docs/api-reference/endpointpickerconfig.md` documented `requestHandler.parser` as a single object while `docs/api-reference/epp-grpc-apis.md` shows `requestHandler.parsers` as a list. The schema defines only the list form inside `requestHandler`. The configuration guide also presented the deprecated top-level `parser` field as current.

Fixes #2355

## Changes
- `docs/api-reference/endpointpickerconfig.md`: `RequestHandlerConfig` table row `parser` ([ParserConfig]) replaced with `parsers` ([][ParserConfig]).
- `docs/architecture/core/router/epp/configuration.md`: top-level structure sketch lists `requestHandler` instead of the deprecated `parser`; the Parsers section example now uses the `requestHandler.parsers` list form; added a note that the top-level `parser` field is deprecated, with the precedence rule.
- `docs/api-reference/epp-grpc-apis.md` already uses the list form and is unchanged.

## Validation
- Schema checked in llm-d/llm-d-router at tag `v0.10.0` (commit `0b43c187`), file `apix/config/v1alpha1/endpointpickerconfig_types.go`:
  - lines 320-325: `RequestHandlerConfig` has a single field `Parsers []ParserConfig` with JSON tag `parsers,omitempty`.
  - lines 72-78: top-level `Parser *ParserConfig` (JSON tag `parser,omitempty`) is marked Deprecated, the new field wins when both are set, tracked in llm-d-router#1308.
  - Same shape at llm-d-router `main` (commit `d3248497`, lines 361-366).
- `markdownlint-cli@0.47.0` (version pinned in `.pre-commit-config.yaml`) with the repo `.markdownlint.json` over both changed files: no new findings. The single reported error (MD051 at `endpointpickerconfig.md:90`, broken `#datalayerextractor` fragment) exists on `main` before this change.

### DIFF
--- a/docs/architecture/core/router/epp/configuration.md
+++ b/docs/architecture/core/router/epp/configuration.md
@@ -12,7 +12,7 @@ plugins:
 - ....
 featureGates:
   ...
-parser:
+requestHandler:
   ...
 flowControl:
   ...
@@ -169,18 +169,22 @@ This section covers components that process requests and responses before they r
 
 #### Parsers
 
-The `parser` section configures how the EPP understands protocol messages (e.g., OpenAI or vLLM payloads). To use a non-default parser, you must first instantiate it in the `plugins` section and then reference its name in the `parser` field:
+The `requestHandler.parsers` section configures how the EPP understands protocol messages (e.g., OpenAI or vLLM payloads). By default, when `requestHandler.parsers` is unspecified, the EPP configures three built-in parsers, `openai-parser`, `anthropic-parser`, and `vllmhttp-parser`, instantiated automatically without `plugins` entries. A request is handled by the first configured parser that claims its path (for example `/v1/chat/completions` by `openai-parser`, `/v1/messages` by `anthropic-parser`, `/inference/v1/generate` by `vllmhttp-parser`).
+
+Setting `parsers` replaces this default list. To use a non-default parser, you must first instantiate it in the `plugins` section and then reference its name in an entry of the `parsers` list:
 
 ```yaml
 plugins:
 - name: myParser
   type: vllmgrpc-parser
 # ...
-parser:
-  pluginRef: myParser
+requestHandler:
+  parsers:
+  - pluginRef: myParser
 ```
 
-If unspecified, `openai-parser` is used by default.
+> [!NOTE]
+> The top-level `parser` field (a single object) is deprecated in favor of the `requestHandler.parsers` list. If both are set, `requestHandler.parsers` is used. See [llm-d-router#1308](https://github.com/llm-d/llm-d-router/issues/1308).
 
 #### Admitters & Data Producers
 
--- a/docs/api-reference/endpointpickerconfig.md
+++ b/docs/api-reference/endpointpickerconfig.md
@@ -74,7 +74,7 @@ Configures request handling behavior.
 
 | Field | Description |
 | --- | --- |
-| `parser` | [ParserConfig](#parserconfig) <br/> Specifies the parsing logic for protocol messages. |
+| `parsers` | [][ParserConfig](#parserconfig) <br/> List of parsing plugins used to process protocol messages. If unspecified, `openai-parser`, `anthropic-parser`, and `vllmhttp-parser` are configured by default. |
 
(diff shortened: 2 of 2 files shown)


docs/architecture/core/router/epp/configuration.md
```diff
@@ -169,19 +169,23 @@ This section covers components that process requests and responses before they r
 
 #### Parsers
 
-The `parser` section configures how the EPP understands protocol messages (e.g., OpenAI or vLLM payloads). To use a non-default parser, you must first instantiate it in the `plugins` section and then reference its name in the `parser` field:
+The `requestHandler.parsers` section configures how the EPP understands protocol messages (e.g., OpenAI or vLLM payloads). To use a non-default parser, you must first instantiate it in the `plugins` section and then reference its name in an entry of the `parsers` list:
```
ahg-g: can we clarify what is configured by default pls?

rishabhsinha17: Done in 77f9e5c: the Parsers section now opens with the default (three built-in parsers, `openai-parser`, `anthropic-parser`, `vllmhttp-parser`, instantiated automatically; a request goes to the first configured parser claiming its path; setting `parsers` replaces the list) and the api-reference `parsers` row says the same, with the stale `(default: openai-parser)` on `pluginRef` dropped. Source: `ensureParsers` in pkg/epp/config/loader/defaults.go at v0.10.0 (llm-d-router#1488).

rishabhsinha17:
Addressed the default parsers comment in 77f9e5c and CI is green. @ahg-g whenever you have a chance to take another look.

PR #2385, opened 2026-08-26
[Helper] Creates a Helper File for llm-d-planner

Vezio:
# Summary

- Creates initial documentation for `llm-d-planner` for interested users to begin _planning_ workloads for their desired use case using `llm-d`

- Provides an overview, general installation, quick start, and capabilities that `llm-d-planner` will bring to the `llm-d` ecosystem

### DIFF
--- a/helpers/planner.md
+++ b/helpers/planner.md
@@ -1,0 +1,218 @@
+# Planning an llm-d Deployment
+
+[llm-d-planner](https://github.com/llm-d-incubation/llm-d-planner) helps you narrow down a particular `llm-d` configuration needed to support your usage requirements by using a data driven, AI infused methodology.
+
+You provide the `planner` with key details for your workload and it searches benchmark data for the model and accelerator combinations that can meet them. The `planner` will then generate deployment artifacts and manifests for the option you choose.
+
+There are a variety of ways to interact with the `planner`, some feature direct UI experience and others feature embedding it's API within your existing library (`llm-d-benchmark` leverages the `Capacity Planner` option today!).
+
+## Contents
+
+1. [What you get](#what-you-get)
+2. [Installation](#installation)
+3. [Quick start](#quick-start)
+4. [Use cases, traffic profiles, and SLOs](#use-cases-traffic-profiles-and-slos)
+5. [Where the numbers come from](#where-the-numbers-come-from)
+6. [Deploying onto llm-d](#deploying-onto-llm-d)
+7. [Bringing your own benchmark data](#bringing-your-own-benchmark-data)
+8. [Good to know](#good-to-know)
+
+## What you get
+
+The project bundles three tools. The `Planner` is the main workflow and calls the other 2 workflows if benchmark data runs scarce, but both of those are useful in their own right, and each one has a UI page, a CLI command, and REST endpoints.
+
+| Tool | Answers | Interface |
+| --- | --- | --- |
+| **Planner** | Given this workload and these SLOs, what should I deploy, and what will it cost? | UI main page, REST, Python library |
+| **Capacity Planner** | Does this model fit? Memory for weights, KV cache (MHA, GQA, MQA, MLA), activations, and overhead, along with minimum accelerator count, maximum context length, and concurrent request capacity. | UI page, `planner plan`, REST |
+| **GPU Recommender** | Which accelerator should I ask for? Estimated TTFT, ITL, and throughput per accelerator type, with a cost comparison, without running a benchmark. | UI page, `planner estimate`, REST |
+
+## Installation
+
+How much you install depends on how much of the tool you need. If you only want the Capacity Planner, the GPU Recommender, or the recommendation pipeline as a library, the wheel is enough:
+
+```bash
+pip install llm-d-planner
+```
+
+That core install runs without a server, an LLM, or a cluster behind it.
+
+Optional extras cover the rest: `[estimation]` adds HuggingFace model config lookups and roofline estimates, `[llm]`, `[openai]`, and `[vertex]` each add natural-language intent extraction through that provider, `[server]` adds the REST API, `[kubernetes]` lets the library apply a generated bundle, and `[ui]` adds the Streamlit interface.
+
+> [!NOTE]
+> `planner estimate` also needs BentoML's roofline model, which is not published to PyPI:
+>
+> ```bash
+> pip install "llm-optimizer @ git+https://github.com/bentoml/llm-optimizer.git"
+> ```
+
+The upstream [README prerequisites](https://github.com/llm-d-incubation/llm-d-planner/blob/main/README.md#prerequisites) and [Developer Guide](https://github.com/llm-d-incubation/llm-d-planner/blob/main/docs/DEVELOPER_GUIDE.md) carry the current list and the platform notes that go with it. Also see [Deployment Guide](https://github.com/llm-d-incubation/llm-d-planner/blob/main/docs/DEPLOYMENT_GUIDE.md) covers hosting a single shared instance for a team.
+
+## Quick start
+
+First clone the repo:
+
+```bash
+git clone git@github.com:llm-d-incubation/llm-d-planner.git
+```
+
+## UI Mode
+
+```bash
+make setup          # install dependencies, pull the Ollama model
+make start          # start the database, Ollama, backend, and UI
+make db-load-blis   # load the bundled BLIS benchmark data
+```
+
+Open `http://localhost:8501` and the main page walks you through the workflow. Most of it is clicking forward; steps 4 and 6 are where your own judgement comes in.
+
+1. **Describe your use case.** For example: "I need a chatbot for 5000 users with low latency" is enough, and a local LLM reads structured intent out of it. If you already know exactly what you want, a form mode lets you skip the LLM and fill the fields in directly.
+2. **Analyze use case.** This produces the intent: use case, user count, priorities, and any accelerator or model preferences you mentioned.
+3. **Generate specification.** The intent becomes a traffic profile and a set of SLO targets.
+4. **Review the specification.** Everything the previous step derived is editable and shows its recommended range beside it, including the TTFT, ITL, and E2E targets, the percentile they apply at, the expected RPS, and the weights used for scoring.
+5. **Generate recommendations.** The Planner searches its benchmark data for configurations that meet those targets.
+6. **Select a recommendation.** You get four ranked views of the same search, sorted for balance, quality, cost, and latency, and you choose which trade-off you want.
+7. **Deploy.** The Deployment tab lets you review, copy, or download the generated manifests, or apply them to a connected cluster.
+
+When you are done, `make stop` leaves Ollama and the database running, and `make stop-all` shuts everything down.
+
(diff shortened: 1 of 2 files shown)


helpers/planner.md
```diff
@@ -0,0 +1,218 @@
...
+print(f"KV cache left: {kv_cache_gb:.1f} GB across {TP} GPUs")
+print(f"per request:   {per_request_gb:.2f} GB at {MAX_MODEL_LEN} tokens")
+print(f"headroom for {concurrent} concurrent requests")
+```
+
+These functions need the `[estimation]` extra, since they read the model config from HuggingFace. `allocatable_kv_cache_memory()` returns a negative number when the weights and activation memory do not leave room for any KV cache at all, which is how `llmdbenchmark` catches a deployment that would fail before it launches it. The full recommendation pipeline is a library too, through the `Planner` class, and both it and the matching REST endpoints are documented in the upstream [Programmatic API User Guide](https://github.com/llm-d-incubation/llm-d-planner/blob/main/docs/PROGRAMMATIC_API_USER_GUIDE.md).
+
+## Use cases, traffic profiles, and SLOs
+
+The Planner does not reason about an arbitrary workload description from scratch. It recognizes nine use cases, and each one carries a traffic profile describing the token shape of the workload and an experience class describing how fast the interaction has to feel. Those two together produce the SLO targets you are asked to approve.
```
jgchn: We can add that the workload shape matches with the workload catalog use cases used in Prism as well.

Vezio: Ah good point, we should talk about Prism here too. Thanks @jgchn

anfredette:
Thanks for putting this together, Tyler! It's great to have an independent look at the Planner project, and you did a really thorough job writing this up — the three-tool structure, the CLI examples, the library code sample, the llm-d deployment section, and the installation extras are all spot-on.

There are a few inaccuracies I've noted in inline comments below. Most of them trace back to issues on our side: `docs/traffic_and_slos.md` was an early design doc that had drifted from the actual code, and `slo_templates.json` had become mostly obsolete with the real data living in `usecase_slo_workload.json`. I've opened [llm-d-incubation/llm-d-planner#355](https://github.com/llm-d-incubation/llm-d-planner/pull/355) to clean that up so the Planner docs match what the code actually does.

I've left inline comments on the specific items to address. Thanks again!

— Andre

helpers/planner.md
```diff
@@ -0,0 +1,218 @@
+# Planning an llm-d Deployment
+
+[llm-d-planner](https://github.com/llm-d-incubation/llm-d-planner) helps you narrow down a particular `llm-d` configuration needed to support your usage requirements by using a data driven, AI infused methodology.
+
+You provide the `planner` with key details for your workload and it searches benchmark data for the model and accelerator combinations that can meet them. The `planner` will then generate deployment artifacts and manifests for the option you choose.
+
+There are a variety of ways to interact with the `planner`, some feature direct UI experience and others feature embedding it's API within your existing library (`llm-d-benchmark` leverages the `Capacity Planner` option today!).
```
anfredette: "it's API" → "its API" (possessive, not contraction)

helpers/planner.md
```diff
@@ -0,0 +1,218 @@
...
+    MODEL, config, GPU_MEMORY, tp=TP, max_model_len=MAX_MODEL_LEN
+)
+per_request_gb = KVCacheDetail(MODEL, config, MAX_MODEL_LEN).per_request_kv_cache_gb
+concurrent = max_concurrent_requests(MODEL, config, MAX_MODEL_LEN, GPU_MEMORY, tp=TP)
+
+print(f"weights:       {weights_gb:.1f} GB")
+print(f"KV cache left: {kv_cache_gb:.1f} GB across {TP} GPUs")
+print(f"per request:   {per_request_gb:.2f} GB at {MAX_MODEL_LEN} tokens")
+print(f"headroom for {concurrent} concurrent requests")
+```
```
anfredette: The SLO values and some traffic profiles in this table don't match what the tool actually produces. This was our fault — they came from `docs/traffic_and_slos.md`, which was an early design doc that had drifted from the code. The upstream fix is in [llm-d-incubation/llm-d-planner#355](https://github.com/llm-d-incubation/llm-d-planner/pull/355).

Specific issues:
- **Traffic profiles**: `translation` and `content_generation` both use **512 → 256** in the code, not 1024 → 1024.
- **SLO values**: The system presents these as **ranges** (e.g., chatbot TTFT: 100–500 ms), not single fixed values. The default within the range depends on the user's latency priority setting (high → 25th percentile, medium → 50th, low → 75th). Since this is a helper doc and the user will see the actual values in the tool, I'd suggest either showing the ranges or just describing the concept without specific numbers and pointing to the upstream doc.
- **Experience class column**: This concept exists in our design rationale doc but is not surfaced anywhere in the tool's UI, CLI, or API. Users won't find it when using the Planner. Consider removing this column or noting it's a design concept.

helpers/planner.md
```diff
@@ -0,0 +1,218 @@
...
+
+```python
+from planner.capacity_planner import (
+    KVCacheDetail,
+    allocatable_kv_cache_memory,
+    get_model_config_from_hf,
+    max_concurrent_requests,
+    model_memory_req,
+)
+
```
anfredette: This function actually returns **0** (not a negative number) when the model doesn't fit — the implementation is `return max(0, available_memory - total_consumed)`. The check for "doesn't fit" is `<= 0`, not `< 0`.

helpers/planner.md
```diff
@@ -0,0 +1,218 @@
...
+Optional extras cover the rest: `[estimation]` adds HuggingFace model config lookups and roofline estimates, `[llm]`, `[openai]`, and `[vertex]` each add natural-language intent extraction through that provider, `[server]` adds the REST API, `[kubernetes]` lets the library apply a generated bundle, and `[ui]` adds the Streamlit interface.
+
+> [!NOTE]
+> `planner estimate` also needs BentoML's roofline model, which is not published to PyPI:
+>
+> ```bash
+> pip install "llm-optimizer @ git+https://github.com/bentoml/llm-optimizer.git"
+> ```
+
+The upstream [README prerequisites](https://github.com/llm-d-incubation/llm-d-planner/blob/main/README.md#prerequisites) and [Developer Guide](https://github.com/llm-d-incubation/llm-d-planner/blob/main/docs/DEVELOPER_GUIDE.md) carry the current list and the platform notes that go with it. Also see [Deployment Guide](https://github.com/llm-d-incubation/llm-d-planner/blob/main/docs/DEPLOYMENT_GUIDE.md) covers hosting a single shared instance for a team.
```
anfredette: Suggest: "The Deployment Guide covers hosting a single shared instance for a team."

helpers/planner.md
```diff
@@ -0,0 +1,218 @@
...
+per_request_gb = KVCacheDetail(MODEL, config, MAX_MODEL_LEN).per_request_kv_cache_gb
+concurrent = max_concurrent_requests(MODEL, config, MAX_MODEL_LEN, GPU_MEMORY, tp=TP)
+
+print(f"weights:       {weights_gb:.1f} GB")
+print(f"KV cache left: {kv_cache_gb:.1f} GB across {TP} GPUs")
+print(f"per request:   {per_request_gb:.2f} GB at {MAX_MODEL_LEN} tokens")
+print(f"headroom for {concurrent} concurrent requests")
+```
+
+These functions need the `[estimation]` extra, since they read the model config from HuggingFace. `allocatable_kv_cache_memory()` returns a negative number when the weights and activation memory do not leave room for any KV cache at all, which is how `llmdbenchmark` catches a deployment that would fail before it launches it. The full recommendation pipeline is a library too, through the `Planner` class, and both it and the matching REST endpoints are documented in the upstream [Programmatic API User Guide](https://github.com/llm-d-incubation/llm-d-planner/blob/main/docs/PROGRAMMATIC_API_USER_GUIDE.md).
```
anfredette: `allocatable_kv_cache_memory()` actually returns 0 (not a negative number) when the model doesn't fit — the implementation is return max(0, available_memory - total_consumed). The check for "doesn't fit" is <= 0, not < 0.

helpers/planner.md
```diff
@@ -0,0 +1,218 @@
...
+| `chatbot_conversational` | 512 → 256 | Conversational | ≤150 ms | ≤25 ms | ≤7 s |
+| `code_completion` | 512 → 256 | Instant | ≤100 ms | ≤20 ms | ≤5 s |
+| `code_generation_detailed` | 1024 → 1024 | Interactive | ≤300 ms | ≤30 ms | ≤25 s |
+| `translation` | 1024 → 1024 | Deferred | ≤400 ms | ≤35 ms | ≤35 s |
+| `content_generation` | 1024 → 1024 | Deferred / batch | ≤500 ms | ≤35 ms | ≤40 s |
+| `summarization_short` | 4096 → 512 | Interactive / deferred | ≤600 ms | ≤30 ms | ≤15–20 s |
+| `document_analysis_rag` | 4096 → 512 | Interactive | ≤600 ms | ≤30 ms | ≤18 s |
+| `long_document_summarization` | 10240 → 1536 | Deferred | ≤1 s | ≤40 ms | ≤60 s |
+| `research_legal_analysis` | 10240 → 1536 | Batch / offline | ≤2 s | ≤45 ms | ≤75–90 s |
+
```
anfredette, line 148: The SLO values and some traffic profiles in this table don't match what the tool actually produces. This was our fault — they came from `docs/traffic_and_slos.md`, which was an early design doc that had drifted from the code. The upstream fix is in [llm-d-incubation/llm-d-planner#355](https://github.com/llm-d-incubation/llm-d-planner/pull/355).

Specific issues:
- **Traffic profiles**: `translation` and `content_generation` both use **512 → 256** in the code, not 1024 → 1024.
- **SLO values**: The system presents these as **ranges** (e.g., chatbot TTFT: 100–500 ms), not single fixed values. The default within the range depends on the user's latency priority setting (high → 25th percentile, medium → 50th, low → 75th). Since this is a helper doc and the user will see the actual values in the tool, I'd suggest either showing the ranges or just describing the concept without specific numbers and pointing to the upstream doc.
- **Experience class column**: This concept exists in our design rationale doc but is not surfaced anywhere in the tool's UI, CLI, or API. Users won't find it when using the Planner. Consider removing this column or noting it's a design concept.

maugustosilva:
@ahg-g @chcost

Vezio:
@jgchn @anfredette thanks for your comments! I will go ahead and make those changes shortly. Very much appreciate the second and third set of eyes, especially on documentation.

anfredette (approved):
@Vezio thanks for addressing my comments!  This looks good to me!

Vezio:
@ahg-g - can I have your review here? Incubation team has signed off above.

CC: @maugustosilva @chcost

jgchn (approved):

helpers/planner.md
Gregory-Pereira: Lets move this from helpers to `docs/operations` because its a not helper which is traditionally manifests, or scripts

Gregory-Pereira (changes requested):
One small thing but we can merge after / the content looks good

PR #2388, opened 2026-08-27
proposal: selective tracing for llm-d

gyliu513:
Adds a design proposal for **selective tracing**: a second, orthogonal control on top of the existing sampling-ratio knob (`OTEL_TRACES_SAMPLER_ARG`) from [Distributed Tracing for llm-d](https://github.com/llm-d/llm-d/blob/main/proposals/distributed-tracing.md). Where sampling decides *which requests* get traced, this proposal decides *which components* actually record and export spans for a request that's already been sampled in. That original work is implemented and documented in [Distributed Tracing](https://github.com/llm-d/llm-d/blob/main/docs/operations/observability/tracing.md).

/cc @ahg-g @robertgshaw2-redhat @chcost


gyliu513:
@rameshdoddaiah I think you have some comments for depth-based selection in our weekly call, after more thinking, I want to take this as low priority but focus on process level and scope level for now.

The proposal's existing two layers, process-level on/off and EPP's scope-level filtering by `TracerScope` name, both map directly to concrete operator scenarios (e.g., "I only want KV-cache spans," "I only want the router + kv-cache hop"). Depth doesn't map to any of the user stories the same way; nobody asks for "the first N levels of the trace tree," they ask for spans from a specific subsystem or component.

It's also harder to implement consistently: depth is a property of the whole trace tree, but each component's `SpanProcessor` only sees its own local spans at `OnEnd`. It has no notion of its position in the global tree unless we propagate a depth counter through baggage and have every hop increment and check it, which is materially more machinery than filtering on a scope name already attached to the span. Depth is also fragile: refactor the code to add one more layer of nesting, and "the useful spans happen to be at level 3" silently breaks, whereas scope names are a stable, self-describing selection key.

Thoughts? Thanks!

rdoddaia:
@gyliu513 
I agree that scope-level filtering is more useful than depth for the primary operator use cases.
I do think depth could still be useful as a complementary mechanism for controlling trace volume/noise in deeply nested or agentic workflows. In that case, the intent is less “I want spans at level N” and more “I only want the top N levels of the distributed execution.”

So I'm fine with keeping depth low priority for now, but I would suggest leaving it as a possible future extension .

Thanks,
-Ramesh

rdoddaia (approved):

nicolexin:
The proposal LGTM! I just have one comment for the "scope-level filtering".

Would it be too late to filter when all the spans are generated? If we are dropping some scope in the middle, would the child get disconnected and point to a non-existent parent and have it's own tree?

I haven't verified that but I think we can check scope within the Tracer() initialization here: https://github.com/llm-d/llm-d-router/blob/ebc59e514b58eed1e6af796225f5786a3700a715/pkg/common/observability/tracing/telemetry.go#L282. If the scope is not included, we can simply return a noop tracer which carries the parent context forward.

gyliu513:
Thanks @nicolexin , good catch!

I think you're right that OnEnd-based filtering has this hole. Since tracer.Start() already mints a real span ID before the filter runs, a disabled-scope span that has enabled-scope children would leave those children pointing at a parent that never gets exported, orphaned in the backend.

Filtering at Tracer() acquisition and handing back a no-op tracer for disabled scopes is the right fix, it's the same mechanism the process-level OTEL_TRACES_EXPORTER=none path already relies on (no-op Start() returns the parent context unchanged, no new span ID minted), so context still passes through correctly to any children. I'll update the `Scope-level filtering inside EPP` section to filter at tracer-acquisition time instead of OnEnd, and adjust the scopeFilterProcessor design accordingly.

gyliu513:
under review

PR #2390, opened 2026-08-27
docs(observability): add batch-gateway dashboards and alerts to llm-d

UgaTheDev:
## What this PR does

Moves the Batch Gateway's user-facing observability assets into `llm-d/llm-d`, per the [observability-integration proposal](https://github.com/llm-d/llm-d/blob/main/proposals/observability-integration.md) and #1983. This is the llm-d half of [batch-gateway#619](https://github.com/llm-d/llm-d-batch-gateway/issues/619); the cleanup PR in the component repo follows once this lands.

It follows the pattern of #2363 (the WVA slice) rather than inventing a layout — dashboards alongside the existing seven, the `PrometheusRule` as a sibling of `epp-alerting-rules.yaml`.

**Added:**

| File | Source in batch-gateway |
|---|---|
| `guides/recipes/observability/grafana/dashboards/llm-d-batch-gateway-apiserver.json` | `charts/batch-gateway/dashboards/apiserver.json` (4 panels) |
| `guides/recipes/observability/grafana/dashboards/llm-d-batch-gateway-processor.json` | `charts/batch-gateway/dashboards/processor.json` (17 panels) |
| `guides/recipes/observability/grafana/dashboards/llm-d-batch-gateway-gc.json` | `charts/batch-gateway/dashboards/gc.json` (5 panels) |
| `guides/recipes/observability/alerts/batch-gateway-alerting-rules.yaml` | `charts/batch-gateway/templates/prometheus-rule.yaml` (5 alerts) |

## The dashboards are not copied verbatim

Same class of problem #2363 hit, more severe here. None of the three dashboards had a datasource template variable, and none of their datasource references carried a `uid` — each file had exactly one `{"type": "prometheus"}` reference, on the `namespace` query variable, with the panels carrying no datasource at all. Copied across unchanged they would load and render empty, binding to whatever Grafana had set as default.

Every dashboard already here uses `${DS_PROMETHEUS}` with a standard `datasource`-type template variable. All three were given that variable and had every panel and target bound to it — **71 datasource references in total** (11 / 46 / 14). `load-llm-d-dashboards.sh` globs `*.json`, so no registration was needed.

Panel content, queries, and dashboard `uid`s are untouched; the `uid`s are stable so existing links keep working.

## Alerts

The chart's `PrometheusRule` is a Helm template, so it was rendered to static YAML using the chart's own `values.yaml` defaults (300s queue wait, 10% failure ratio, 3% expired ratio, 90% worker saturation, 65m reconciler window). Renamed to `batch-gateway-alerting-rules` and relabelled `app: batch-gateway-metrics`, matching `epp-alerting-rules.yaml`.

**One deliberate deviation from `epp-alerting-rules.yaml`, worth a maintainer's eye:** every expression keeps a `namespace="batch-gateway"` matcher. The EPP rules carry no namespace matcher because `llm_d_epp_*` names are unambiguous; the batch-gateway names are not (`jobs_processed_total`, `active_workers`, `total_workers`), and an unscoped rule would silently aggregate unrelated workloads. `batch-gateway` is the namespace the guide's `export NAMESPACE=batch-gateway` creates. Applying the `PrometheusRule` into a namespace does not scope its queries, so this can't be left to `kubectl apply -n`. Called out in both a file comment and an `IMPORTANT` note in `alerting.md`. Happy to drop the matcher if you'd rather match the EPP file exactly.

The `ruleSelector` caveat already documented in `alerting.md` Step 1 applies here identically — I have not changed that behaviour.

## Docs

- `alerting.md` — intro now covers both components, plus a Batch Gateway section with apply/verify steps and a reference table for the five `batch-gateway.rules` alerts. These fire on batch job throughput and GC reconciler health rather than request-path health, which the section calls out.
- `metrics.md` — a "Key Batch Gateway Metrics" section (11 operationally useful series), the two new dashboard entries in the `kubectl get configmaps` sample output and the manual-import table.
- `guides/batch-serving/batch-gateway/README.md` — the 4-step "Enable Monitoring" section #619 asks for, which the guide was missing entirely.

On the metric reference: rather than copy the component's 126-line `docs/guides/metrics.md` across (which would drift), the new section is a table of the operationally relevant series in the same style as the vLLM/SGLang/EPP tables, linking the component doc as the definitional source of truth for names, types, and labels. Say the word if you want the full doc mirrored instead.

## What is deliberately not here

The `PodMonitor`s and `ServiceMonitor`. #619 targets them at `guides/recipes/batch-gateway/monitoring/` "(or equivalent)", and #2363 raised the same open question for the WVA scrape manifest — whether these belong under the per-guide directory or the shared recipe directory — which is still unanswered. I would rather not guess and have it moved twice. Step 1 of the new guide section therefore still uses the chart's `podMonitor.enabled` / `serviceMonitor.enabled` flags, with an in-tree note saying so and pointing at #1983. Happy to add the standalone manifests here in a follow-up once the path is settled.

**Path note:** #619 refers to `llm-d/guides/batch-gateway/README.md`, which no longer exists — the guide has since moved to `guides/batch-serving/batch-gateway/README.md`. That is the file this PR edits.

## Testing

- All three dashboards parse as valid JSON and retain their panel counts (4 / 17 / 5) and `uid`s; no unbound datasource references remain.
- The `PrometheusRule` parses as valid YAML, retains all five alerts in the single `batch-gateway.rules` group, and contains no unrendered Helm template expressions.
- `python scripts/guide.py render guides/*/ --check` passes (the batch-gateway guide has no `guide.yaml`, so its README is hand-maintained — this edit does not desync a rendered file).
- `markdownlint-cli` 0.47.0 and `yamllint` at the repo's pinned versions/config report only findings that are already present on `main` — MD060 fires 70 times on `metrics.md` before this change, and the existing `epp-alerting-rules.yaml` produces the identical yamllint indentation output, so the new YAML matches established style.

I have not applied these to a live cluster — I don't have a Prometheus/Grafana stack with the Batch Gateway running to hand, so the datasource rebinding is verified structurally rather than by rendering. Worth a look from someone who can load them.

---

*Disclosure: prepared with AI assistance. The asset inventory, the datasource-convention mismatch, the chart's default thresholds, and the guide's namespace were checked against both repositories before writing.*

### DIFF
--- a/docs/operations/observability/alerting.md
+++ b/docs/operations/observability/alerting.md
@@ -1,6 +1,6 @@
 # Alerting
 
-This page covers a default set of Prometheus alerting rules for the EPP (Endpoint Picker). For Prometheus and Grafana installation, see [Observability Setup](./setup.md) first, and for the metrics these alerts are built on, see [Metrics](./metrics.md).
+This page covers default sets of Prometheus alerting rules for the EPP (Endpoint Picker) and for the Batch Gateway. For Prometheus and Grafana installation, see [Observability Setup](./setup.md) first, and for the metrics these alerts are built on, see [Metrics](./metrics.md).
 
 The rules ship as a [`PrometheusRule`](https://prometheus-operator.dev/docs/getting-started/design/#prometheusrule) custom resource, so they require the Prometheus Operator (bundled with the kube-prometheus-stack installed by the [setup guide](./setup.md)).
 
@@ -80,6 +80,32 @@ The error-ratio alerts are `0/0`-safe: with no traffic the expression yields no 
 > [!NOTE]
 > `EPPExtProcStreamErrors` relies on opt-in metrics enabled by the EPP `--enable-grpc-stream-metrics` flag. When the flag is unset, the series are absent and the alert never fires. The matcher excludes `OK` and `Canceled` (a normal client disconnect) — adjust it for your environment.
 
+### Batch Gateway (`batch-gateway.rules`)
+
+These alerts cover the Batch Gateway processor and GC reconciler rather than the request path, so they are shipped as a separate `PrometheusRule` and are only relevant if you deployed the [Batch Gateway guide](../../../guides/batch-serving/batch-gateway/README.md).
+
+```bash
+kubectl apply -n ${NAMESPACE} -f guides/recipes/observability/alerts/batch-gateway-alerting-rules.yaml
+kubectl get prometheusrules -n ${NAMESPACE}
+```
+
+You should then see the `batch-gateway.rules` group under **Status → Rule Health** in the Prometheus UI.
+
+> [!IMPORTANT]
+> Every expression in this file is scoped with `namespace="batch-gateway"`, the namespace the Batch Gateway guide deploys into. The Batch Gateway metric names are unprefixed and generic (`jobs_processed_total`, `active_workers`), so an unscoped rule can pick up unrelated workloads. If you deployed into a different namespace, edit the matchers to match — applying the `PrometheusRule` into your namespace does **not** scope its queries.
+
+| Alert | Severity | Fires when | Why it matters |
+|-------|----------|-----------|----------------|
+| `BatchGatewayHighQueueWait` | warning | p95 `job_queue_wait_duration_seconds` > 300s for 15m | Jobs are backing up in the priority queue faster than workers drain it |
+| `BatchGatewayHighJobFailureRate` | warning | Failed share of `jobs_processed_total` > 10% for 10m | Job execution is failing at a rate users will notice |
+| `BatchGatewayExpiredJobsDetected` | warning | Expired share of `jobs_processed_total` > 3% for 10m | Jobs are aging out before execution — capacity or completion-window problem, not a code failure |
+| `BatchGatewayWorkersSaturated` | warning | `active_workers / total_workers` > 90% for 15m | The worker pool is the bottleneck; raise `NumWorkers` or add replicas |
+| `BatchReconcilerErrors` | warning | `batch_reconciler_errors_total` increased over 65m | The GC orphan reconciler is failing, so orphaned jobs are not being recovered |
+
+The two ratio alerts use `clamp_min` on the denominator, so they stay silent when no jobs are being processed rather than dividing by zero.
+
+`BatchReconcilerErrors` uses `increase()` over a 65m window rather than `rate()`: the reconciler runs on a 60m interval by default and increments the counter at most once per cycle, which is too sparse for a `rate()` threshold. If you shorten `reconciler.interval` in the GC config, shorten the window to match — a window shorter than the interval can miss a once-per-cycle failure, and a much longer one keeps the alert firing after the error has aged out.
+
 ## Customization
 
 These rules are a starting point, not a tuned policy. Common adjustments:
@@ -94,6 +120,7 @@ See the [PromQL Reference](./promql.md) for more queries you can promote into al
 
 ```bash
 kubectl delete -n ${NAMESPACE} -f guides/recipes/observability/alerts/epp-alerting-rules.yaml
+kubectl delete -n ${NAMESPACE} -f guides/recipes/observability/alerts/batch-gateway-alerting-rules.yaml
 ```
 
 ## Troubleshooting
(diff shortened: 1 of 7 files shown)


guides/recipes/observability/grafana/dashboards/llm-d-batch-gateway-gc.json
```diff
@@ -0,0 +1,225 @@
...
+        "regex": "",
+        "skipUrlSync": false,
+        "type": "datasource"
+      },
+      {
+        "current": {
+          "selected": false,
+          "text": "default",
+          "value": "default"
+        },
```
gyliu513: The `namespace` template variable's `current` is hardcoded to `"default"` here, unlike the sibling dashboards added in this same PR (`llm-d-batch-gateway-apiserver.json` and `llm-d-batch-gateway-processor.json`) which both use `"current": {}`. Combined with `refresh: 2` below instead of `refresh: 1`, this dashboard will default to `namespace="default"` on load, following the guide's `NAMESPACE=batch-gateway` convention, all 5 panels will show "No data" until the user manually changes the variable. Looks like this file was copy-pasted from a different template and not reset. Suggest matching the other two new dashboards: `"current": {}` and `"refresh": 1`.

docs/operations/observability/alerting.md
```diff
@@ -80,6 +80,32 @@ The error-ratio alerts are `0/0`-safe: with no traffic the expression yields no
...
+
+```bash
+kubectl apply -n ${NAMESPACE} -f guides/recipes/observability/alerts/batch-gateway-alerting-rules.yaml
+kubectl get prometheusrules -n ${NAMESPACE}
+```
+
+You should then see the `batch-gateway.rules` group under **Status → Rule Health** in the Prometheus UI.
+
+> [!IMPORTANT]
+> Every expression in this file is scoped with `namespace="batch-gateway"`, the namespace the Batch Gateway guide deploys into. The Batch Gateway metric names are unprefixed and generic (`jobs_processed_total`, `active_workers`), so an unscoped rule can pick up unrelated workloads. If you deployed into a different namespace, edit the matchers to match — applying the `PrometheusRule` into your namespace does **not** scope its queries.
```
gyliu513: This section documents the `namespace="batch-gateway"` matcher tradeoff but never states the `PrometheusRule`'s label (`app: batch-gateway-metrics`). Elsewhere in this doc, self-hosted-Prometheus users are told to match `ruleSelector` against `app: epp-metrics` — but that's specific to the EPP rule. A reader who follows that instruction for a scoped/custom Prometheus setup will apply this rule successfully with no error, but Prometheus will never discover it, so no Batch Gateway alert will ever fire. Suggest adding a note about the `app: batch-gateway-metrics` label so custom `ruleSelector` configs can be updated to match both rule sets.

guides/recipes/observability/grafana/dashboards/llm-d-batch-gateway-gc.json
```diff
@@ -0,0 +1,225 @@
...
+        "datasource": {
+          "type": "prometheus",
+          "uid": "${DS_PROMETHEUS}"
+        },
+        "definition": "label_values(batch_reconciler_cycle_duration_seconds_count, namespace)",
+        "includeAll": false,
+        "multi": false,
+        "name": "namespace",
+        "refresh": 2,
```
gyliu513:
```suggestion
        "refresh": 1,
```

guides/recipes/observability/grafana/dashboards/llm-d-batch-gateway-gc.json
UgaTheDev: Good catch. Switched to `"current": {}` to match the apiserver and processor dashboards, in b8744a6.

guides/recipes/observability/grafana/dashboards/llm-d-batch-gateway-gc.json
UgaTheDev: Applied, `refresh: 1` in b8744a6.

docs/operations/observability/alerting.md
UgaTheDev: Added a sentence to the IMPORTANT callout naming the rule's own label, `app: batch-gateway-metrics`, and noting that `app: epp-metrics` is specific to the EPP rule. In b8744a6.

Issue #2225, opened 2026-08-10
sccache S3 credentials are invalid: every CUDA image build runs with no cache


yuanwu2017:
### Summary

The AWS credentials used for the sccache S3 backend are rejected by S3, so **every** CUDA image build compiles from scratch with no cache. This is not intermittent — the credentials are unconditionally invalid, so the fallback path in `docker/scripts/common/setup-sccache.sh` is taken on every run.

### Evidence

Every CUDA builder job logs this while starting the sccache server:

```
sccache: error: Server startup failed: cache storage failed to read:
  PermissionDenied (permanent) at read => S3Error {
    code: "InvalidAccessKeyId",
    message: "The AWS Access Key Id you provided does not exist in our records." }

Context:
   uri: https://s3.us-west-2.amazonaws.com/vllm-nightly-sccache/llm-d-cache/.sccache_check
   response: Parts { status: 403, ... }
```

Observed in every CUDA job I have looked at, across different days and different PRs:

| run | job | platform |
|---|---|---|
| [31362149292](https://github.com/llm-d/llm-d/actions/runs/31362149292) | 93373145809 | linux/amd64, ubuntu |
| [31364574234](https://github.com/llm-d/llm-d/actions/runs/31364574234) | 93380605080 | linux/amd64, ubuntu |
| [31369626636](https://github.com/llm-d/llm-d/actions/runs/31369626636) | 93395676034 | linux/amd64, rhel/ubi9 |

The secret is present and non-empty — buildx mounts it and the log shows `AWS_ACCESS_KEY_ID=***`, and `[ -f /run/secrets/aws_access_key_id ]` passes. S3 returns `InvalidAccessKeyId`, which means the key ID itself no longer exists in IAM, i.e. it was deleted or rotated rather than merely lacking permissions (that would be `AccessDenied`).

`build-image.yaml` passes these at lines 270, 424, 798 and 1115:

```yaml
secrets: |
  aws_access_key_id=${{ secrets.AWS_ACCESS_KEY_ID_S3_BUCKET_ONLY }}
  aws_secret_access_key=${{ secrets.AWS_SECRET_ACCESS_KEY_S3_BUCKET_ONLY }}
```

Both of those repository secrets were created on 2025-08-10 and have never been updated since:

```
AWS_ACCESS_KEY_ID_S3_BUCKET_ONLY      created 2025-08-10  updated 2025-08-10
[...]

PR #2412, opened 2026-08-31
fix: correct always-true AWS credential check in setup-sccache.sh

### LINKED ISSUE #2411: setup-sccache.sh always logs "AWS credentials: set" regardless of whether they are
## Summary

`docker/scripts/common/setup-sccache.sh` prints a diagnostic line reporting whether AWS credentials were detected for the sccache S3 backend, but the check is broken and always reports `set`, even when `AWS_ACCESS_KEY_ID` is empty/unset.

## Where

`docker/scripts/common/setup-sccache.sh:57`:

```bash
echo "  - AWS credentials: $([ -n \"${AWS_ACCESS_KEY_ID:-}\" ] && echo 'set' || echo 'NOT SET')"
```

## Why it is wrong

The `\"` sequences are inside a double-quoted string that also contains a `$(...)` command substitution. The outer double quotes already unescape `\"` to a literal `"` character before the substitution's command is parsed, so by the time `[ -n ... ]` actually runs, its argument is the literal string `"${AWS_ACCESS_KEY_ID:-}"` *including the quote characters* — which is always non-empty, so `-n` is always true.

`shellcheck` flags this directly:

```
docker/scripts/common/setup-sccache.sh:57:39: error: -n doesn't work with unquoted arguments. Quote or use [[ ]]. [SC2070]
docker/scripts/common/setup-sccache.sh:57:39: error: Argument to -n is always true due to literal strings. [SC2157]
```

Reproduction:

```bash
$ unset AWS_ACCESS_KEY_ID
$ echo "  - AWS credentials: $([ -n \"${AWS_ACCESS_KEY_ID:-}\" ] && echo 'set' || echo 'NOT SET')"
  - AWS credentials: set   # wrong — the variable is unset
```

## Impact

This is a small piece of the puzzle behind #2225 (sccache S3 credentials being invalid on every CUDA build): the build log's own self-reported credential status is unconditionally wrong, so this diagnostic line gives false confidence that credentials were present even when they were missing entirely, making the real breakage harder to notice from the logs.

## Suggested fix

Drop the unnecessary backslash-escaping — the `$(...)` already opens a fresh quoting context, so a plain `"..."` inside it works correctly:

```diff
-    echo "  - AWS credentials: $([ -n \"${AWS_ACCESS_KEY_ID:-}\" ] && echo 'set' || echo 'NOT SET')"
+    echo "  - AWS credentials: $([ -n "${AWS_ACCESS_KEY_ID:-}" ] && echo 'set' || echo 'NOT SET')"
```

Verified this reports correctly in both states after the fix, and `shellcheck` passes clean.

AlSh007:
## Summary

`docker/scripts/common/setup-sccache.sh` logs a diagnostic line reporting whether AWS credentials were detected for the sccache S3 backend. The check was broken and always printed `set`, regardless of whether `AWS_ACCESS_KEY_ID` was actually present.

The backslash-escaped quotes inside the `$(...)` command substitution were unescaped by the *outer* double-quoted string before the substitution's command was parsed, so `[ -n ... ]` ended up testing a literal, always-non-empty string (`"${AWS_ACCESS_KEY_ID:-}"` including the quote characters) instead of the variable's actual value. `shellcheck` flags this directly (SC2070, SC2157).

This is a small piece of the puzzle behind #2225 (sccache S3 credentials being invalid on every CUDA build): the build log's own self-reported credential status was unconditionally wrong, giving false confidence that credentials were present even when missing entirely.

## Changes

- `docker/scripts/common/setup-sccache.sh`: drop the unnecessary backslash-escaping so the nested `[ -n "..." ]` check quotes the variable correctly.

Fixes #2411

## Test plan

- [x] Reproduced the bug locally: with `AWS_ACCESS_KEY_ID` unset, the old line printed `AWS credentials: set`; the fixed line correctly prints `NOT SET`
- [x] `shellcheck docker/scripts/common/setup-sccache.sh` passes clean (previously reported SC2070/SC2157)
- [x] `bash -n docker/scripts/common/setup-sccache.sh` — syntax OK

### DIFF
--- a/docker/scripts/common/setup-sccache.sh
+++ b/docker/scripts/common/setup-sccache.sh
@@ -54,6 +54,6 @@ if [ "${USE_SCCACHE}" = "true" ]; then
     echo "  - Region: ${SCCACHE_REGION}"
     echo "  - Key prefix: ${SCCACHE_S3_KEY_PREFIX}"
     echo "  - Socket: ${SCCACHE_SERVER_UDS}"
-    echo "  - AWS credentials: $([ -n \"${AWS_ACCESS_KEY_ID:-}\" ] && echo 'set' || echo 'NOT SET')"
+    echo "  - AWS credentials: $([ -n "${AWS_ACCESS_KEY_ID:-}" ] && echo 'set' || echo 'NOT SET')"
     /usr/local/bin/sccache --show-stats 2>&1 | head -5
 fi


AlSh007:
Closing — #2413 fixed this same line and merged a couple of hours after I opened this, so this is redundant. Its approach is also the better one: computing the status into a variable avoids the command substitution entirely and clears shellcheck, whereas I only dropped the stray escaping.

Note that #2411 is still open — #2413 did not reference it, so it did not auto-close. It is fixed on main and can be closed.

PR #2415, opened 2026-08-31
fix(guides): merge duplicate 'env' key in tiered-prefix-cache LMCache patch

jcayab:
## Summary

`guides/tiered-prefix-cache/modelserver/gpu/vllm/lmcache-connector/cpu/base/patch-vllm.yaml` declared two sibling `env:` keys on the same container. YAML mappings can't have duplicate keys, so the parser keeps only the last one it sees — silently dropping the first `env:` block (which set `HF_TOKEN`) before kustomize's strategic-merge patch logic ever runs.

## Detail

Before:
```yaml
env:
  - name: HF_TOKEN
    ...
resources:
  ...
env:
  - name: LMCACHE_MAX_LOCAL_CPU_SIZE
    ...
```

I confirmed the duplicate key with a small PyYAML-based scan across every YAML file in the repo — this was the only hit. The sibling `fs/base` patch for the same LMCache connector already merges everything into a single `env:` list, so this brings the `cpu/base` variant in line with that pattern.

## Fix

Merged both blocks into one `env:` list on the container, keeping `HF_TOKEN` alongside the LMCache-specific vars.

## Testing
- Verified the resulting file has no duplicate mapping keys.
- Diffed against the sibling `fs/base/patch-vllm.yaml` to confirm the merged shape matches the established pattern.


maugustosilva (dismissed):

maugustosilva:
@jcayab please sign your commits and fix the conflicts and it should be ready to merge. Thank you.

PR #2416, opened 2026-08-31
fix(scripts): lint-envvars.py always exited 0, silently disabling the pre-commit/CI gate

jcayab:
## Summary

`scripts/lint-envvars.py` — the env-var declaration linter documented in `scripts/ENVVARS.md` and enforced via `.pre-commit-config.yaml` / the CI `pre-commit` job — has never actually failed a build. `main()` prints its findings but then unconditionally calls `sys.exit(0)`:

```python
if all_errors:
    for error in all_errors:
        print(error, file=sys.stderr)
    print(f"\n⚠ {len(all_errors)} warning(s) found", file=sys.stderr)
    sys.exit(0)          # <-- always succeeds, even with errors
```

Its sibling `lint-dockerfile-envvars.py` already exits 1 on failure, and `ENVVARS.md` describes this as an enforced check ("The `lint-envvars.py` tool enforces this at lint time"), not an opt-in warning — so this is a bug, not a deliberate "warn only" design.

## What flipping it on surfaces

Turning the exit code into a real gate immediately exposed two more issues, both fixed in this PR so the newly-real gate is green from the start:

1. **A detection false-negative in the linter itself**: `find_locally_defined_vars` only matched assignments anchored at the very start of a line, so a `case` arm one-liner like
   ```bash
   linux/arm64) export SCCACHE_SERVER_UDS="..." ;;
   ```
   was never recognized as a local definition, producing a false "undeclared variable" report on `setup-sccache.sh`. Fixed by also matching assignments right after a `)`.

2. **Real, pre-existing undeclared vars** across `docker/scripts/**` that had been accumulating silently since the check never actually failed anything:
   - `build-uccl.sh`: `TARGETOS`, `VIRTUAL_ENV` — used, never declared
   - `build-nixl.sh`, `build-nvshmem.sh`, `build-ucx.sh`: `MAX_JOBS` — used to parallelize the build, never declared
   - `install-vllm.sh`: `CUDA_MINOR`, `FLASHINFER_VERSION`, `NVSHMEM_BUILD_FROM_SOURCE`, `VLLM_CHERRYPICK_1`/`_REMOTE`, `VLLM_CHERRYPICK_2`/`_REMOTE`, `VLLM_SETUPTOOLS_SCM_PRETEND_VERSION` (the last is only referenced by a currently commented-out `export`; documented rather than removed since it looks intended for future re-enablement)

These are documentation-only additions to each script's existing "Required/Optional environment variables" header — no runtime behavior changes.

## Note for maintainers

Fixing the exit code also exposed that `docker/Dockerfile.rocm` currently fails the *sibling* `lint-dockerfile-envvars.py` check (missing `UCCL_DEVICE`/`UCCL_PREFIX` ARG/ENV declarations for `build-uccl.sh`). That's pre-existing and reproducible on `main` today, independent of this PR — I left it out of scope since it's a GPU-specific Dockerfile I can't build/test locally. Happy to file a follow-up issue if useful.

## Testing
- `python3 scripts/lint-envvars.py $(git ls-files 'docker/scripts/**/*.sh')` now exits 0 across the whole tree (previously always exited 0 regardless of content, which is exactly the bug).
- Manually reverted the exit-code fix locally to confirm the same command now correctly exits 1 when a script has an undeclared var.


docker/scripts/cuda/runtime/install-vllm.sh
```diff
@@ -3,17 +3,28 @@ set -Eeu
...
 # - VLLM_PREBUILT: whether to use prebuilt wheel (1/0)
 # - VLLM_USE_PRECOMPILED: whether to use precompiled binaries (1/0)
 # - VLLM_PRECOMPILED_WHEEL_COMMIT: commit SHA for precompiled wheel lookup (defaults to VLLM_COMMIT_SHA)
 # - CUDA_MAJOR: The major CUDA version
+# - CUDA_MINOR: The minor CUDA version
 # - BUILD_NIXL_FROM_SOURCE: if nixl should be installed by vLLM or has been built from source in the builder stages
+# - FLASHINFER_VERSION: flashinfer package version to install
+#
+# Optional environment variables:
+# - SUPPRESS_PYTHON_OUTPUT: If we should suppress vLLM installation logs
```
diegocastanibm: Move this line to line 26 so it stays right before `: "${SUPPRESS_PYTHON_OUTPUT:=}"`

diegocastanibm (changes requested):
Thanks for this. The `sys.exit(1)` fix is clearly right, and I verified that
`python3 scripts/lint-envvars.py $(git ls-files 'docker/scripts/**/*.sh')` exits 0
across all 13 in-scope scripts with the new docs in place.

One thing needs addressing before merge. The header comment block is read by **two** linters with opposite directions:

- `scripts/lint-envvars.py` (this PR): every `${VAR}` in the script body must appear in
  the comment block.
- `scripts/lint-dockerfile-envvars.py`: every var in the comment block must be declared
  as `ARG`/`ENV` in the Dockerfile that runs the script.

So documenting a var to satisfy the first linter creates a new obligation for the second.
And `lint-dockerfile-envvars.py:22-33` ends the required block at the first **non-comment** line (rule 4 in `scripts/ENVVARS.md`). `# Optional environment variables:` is a comment, so everything under that heading is parsed as **required** since the linter is not understanding the `Optional` comment.

Verified with Claude:

**`docker/scripts/common/build-uccl.sh`** — verified against `docker/Dockerfile.rocm`:

```
=== main ===
docker/Dockerfile.rocm:119: Script 'build-uccl.sh' in stage 'build' requires:
  - UCCL_DEVICE (not declared as ARG or ENV)

=== with this PR ===
docker/Dockerfile.rocm:119: Script 'build-uccl.sh' in stage 'build' requires:
  - TARGETOS (not declared as ARG or ENV)
  - UCCL_DEVICE (not declared as ARG or ENV)
  - VIRTUAL_ENV (not declared as ARG or ENV)
```

`UCCL_DEVICE` shows the bug already exists on `main`; the PR grows that failure from 1 var to 3. Both new ones are awkward to "fix" by declaring them:

- `TARGETOS` is already set inline on the invoking `RUN` (`Dockerfile.rocm:119`:
  `RUN TARGETOS=ubuntu UCCL_DEVICE=rocm /tmp/build-uccl.sh`) — the linter only sees
  stage-level `ARG`/`ENV`, so satisfying it means a redundant declaration.
- `VIRTUAL_ENV` is documented on the same line as "set by the Python environment, not by callers of this script" — the linter is now demanding a Dockerfile declaration for a var the docs say callers must not set.

CI is green only because the hook is scoped `files: ^docker/Dockerfile\.`
(`.pre-commit-config.yaml:52`) and this PR **DO NOT** touches Dockerfile. The errors surface on whatever unrelated PR next modifies `Dockerfile.rocm`.

**`docker/scripts/cuda/runtime/install-vllm.sh`** — same mechanism, and there's an extra wrinkle. Moving `: "${SUPPRESS_PYTHON_OUTPUT:=}"` from *above* the Required block to *below* it changes that var from excluded to required. Along with
`NVSHMEM_BUILD_FROM_SOURCE`, `VLLM_CHERRYPICK_1/2(_REMOTE)` and
`VLLM_SETUPTOOLS_SCM_PRETEND_VERSION`, none of which is declared in
`docker/Dockerfile.cuda`. This one is silent for a different reason: `find_script_runs`
(`lint-dockerfile-envvars.py:132-138`) doesn't join `RUN ... \` continuations, so it never
sees `/tmp/install-vllm.sh` at `Dockerfile.cuda:471` — confirmed by injecting a bogus
`# - ZZZ_BOGUS_VAR:` requirement, which went unreported. Fixing that parser or reflowing the `RUN` would immediately fail the runtime stage on genuinely optional vars.

Suggested fix — make `Optional` mean something, in `lint-dockerfile-envvars.py`:

```python
if '# Optional environment variables:' in line:
    in_block = False
    continue
```

and have `lint-envvars.py` accept declarations from either heading. That resolves both
files, clears the pre-existing `UCCL_DEVICE` failure, and keeps the docs honest about
`VIRTUAL_ENV`. Worth a matching note in `scripts/ENVVARS.md` so rule 4 reflects the new terminator.

The lower-effort alternative is to move the optional docs *below* a non-comment line (what `install-vllm.sh` did before this PR), but that relies on rule 4 as a trick rather than
expressing intent, and leaves the next person to rediscover it.

jcayab:
Thanks for the thorough review, @diegocastanibm — you're right, and the repro made it easy to verify.

Pushed a fix in `290aa280`: `# Optional environment variables:` now ends the required-var collection in `lint-dockerfile-envvars.py`, so vars listed under `Optional` no longer get swept into the set checked against the Dockerfile's `ARG`/`ENV` declarations. `lint-envvars.py` is untouched — it never distinguished the two headings, and shouldn't, since a var satisfies that check regardless of which section documents it.

Also updated `scripts/ENVVARS.md` rule 4 to describe the new terminator, per your suggestion.

Re-ran your repro after the fix:

```
$ python3 scripts/lint-dockerfile-envvars.py docker/scripts docker/Dockerfile.rocm
docker/Dockerfile.rocm:119: Script 'build-uccl.sh' in stage 'build' requires:
  - UCCL_PREFIX (not declared as ARG or ENV)
```

`TARGETOS`, `VIRTUAL_ENV`, and `UCCL_DEVICE` are all gone from the failure now (the first two are what this PR's docs commit added; `UCCL_DEVICE` is the pre-existing one you flagged — this fix resolves it as a side effect since it's also under `Optional`).

`UCCL_PREFIX` is still flagged, but it's a **different, separate bug**, unrelated to Required/Optional: it's declared via `ENV UCCL_PREFIX=...` in `Dockerfile.rocm`'s parent `base` stage and inherited into the `build` stage by Docker's normal `FROM base AS build` semantics, but `DockerfileParser` tracks `ARG`/`ENV` per-stage with no inheritance modeling, so it doesn't see it. I confirmed this is pre-existing on `main` independent of this PR (checked out `main` directly and reran the same command — `UCCL_PREFIX` fails there too). Fixing it needs a stage-inheritance-aware rewrite of `DockerfileParser`, and I can't build/test the ROCm image locally to validate that safely, so I've left it out of scope here — same caveat as in the original PR description, just narrowed down to its actual root cause now that the Required/Optional issue no longer masks it.

Also spot-checked `install-vllm.sh` against `docker/Dockerfile.cuda` per your second point — still passes silently for the reason you identified (`find_script_runs` doesn't join `RUN ... \` continuations, so it never sees that invocation at all). Left that alone too; agreed that fixing it would immediately surface `CUDA_MINOR`/`FLASHINFER_VERSION` and needs its own careful pass.

`python3 scripts/lint-envvars.py $(git ls-files 'docker/scripts/**/*.sh')` still exits 0 across all 13 scripts, and `cpu`/`cuda`/`rdma-tools`/`xpu` Dockerfiles all validate cleanly — only `rocm` has the one remaining (pre-existing, separate) finding.

diegocastanibm (commented):
Thanks @jcayab . May I ask you to pass the linters over all the files? I'm sure we will have lint errors on some of the Docker files

jcayab:
Good call, @diegocastanibm — there was one more, and it's fixable without a Docker build.

Ran both linters over every script and every Dockerfile in the repo (not just the ones this PR touches):

```
$ python3 scripts/lint-envvars.py $(git ls-files 'docker/scripts/**/*.sh')
✓ Checked 13 script(s) - all environment variables properly declared

$ python3 scripts/lint-dockerfile-envvars.py docker/scripts docker/Dockerfile.{cpu,cuda,rdma-tools,rocm,xpu}
```

Only `docker/Dockerfile.rocm`'s `UCCL_PREFIX` was still failing — the one I'd flagged as a separate, pre-existing bug in my last comment (`DockerfileParser` tracks `ARG`/`ENV` strictly per-stage, with no inheritance, so it missed `UCCL_PREFIX` being set via `ENV` in `Dockerfile.rocm`'s `base` stage and inherited into `build` via `FROM base AS build`).

I'd left that out of scope thinking it needed image-build validation, but it doesn't — it's a static-analysis gap in the parser itself, testable the same way as everything else here. Fixed in `46874f9c`: a stage now inherits `ENV` (only `ENV`, not `ARG` — per [Docker's docs](https://docs.docker.com/reference/dockerfile/#arg) an `ARG` goes out of scope at the end of its own stage and doesn't cross a `FROM` boundary) from whatever stage its `FROM` line names, when that name matches an already-parsed stage.

To make sure this was a precise fix and not just a new blind spot, I wrote two throwaway two-stage Dockerfiles: one where the required vars are genuinely undeclared anywhere (still correctly fails, all three flagged), and one where the var is only set via `ARG` in the parent stage rather than `ENV` (still correctly fails — confirms `ARG` isn't wrongly inherited too).

Current state, every Dockerfile in the repo:

```
docker/Dockerfile.cpu                        ✓
docker/Dockerfile.cuda                       ✓
docker/Dockerfile.rdma-tools                 ✓
docker/Dockerfile.rocm                       ✓  (was the one failure)
docker/Dockerfile.xpu                        ✓
guides/modelexpress-p2p/image/Dockerfile     ✓
helpers/interactive-pod/build/Dockerfile     ✓
```

The `install-vllm.sh` / `RUN ... \` continuation-parsing gap against `docker/Dockerfile.cuda` from your original review is still there and still silent (confirmed again just now) — that one's a different function (`find_script_runs`) and a different class of fix (joining line continuations before scanning for scripts), so I've left it alone rather than bundle it in here. Happy to open a follow-up for it if you'd rather track it separately.

jcayab:
Heads up on what happened between the last two pushes, since the diff might otherwise look confusing:

Someone else merged [#2446](https://github.com/llm-d/llm-d/pull/2446) into `main` while this PR was in review — same `DockerfileParser` stage-inheritance bug I'd just fixed in `46874f9c`, fixed independently over there. When `main` got merged into this branch (via GitHub's "Update branch"), the conflict in `scripts/lint-dockerfile-envvars.py` got resolved by concatenating both implementations rather than picking one — left two blocks both assigning `self.stages[self.current_stage]`, with the second silently clobbering the first (so my version was already dead code, just still sitting there).

The two versions actually disagreed on something real: mine only inherited `ENV` across a `FROM stage AS newstage` boundary, deliberately excluding `ARG` (going on the usual "ARG doesn't cross stages" advice). #2446's inherits both. Rather than assume my recollection of Docker's docs was right, I checked with an actual `docker build`:

```dockerfile
ARG GLOBAL_ARG=hello-global      # before the first FROM
FROM alpine AS base
ARG STAGE_ARG=hello-stage-base   # inside a stage
FROM base AS mid
FROM mid AS child
RUN echo "${GLOBAL_ARG}"         # -> empty
RUN echo "${STAGE_ARG}"          # -> hello-stage-base (2 stages down, no re-declare)
```

Turns out the "must re-declare ARG per stage" rule only applies to a *global* ARG declared before the first `FROM` — an ARG declared inside a stage carries forward through descendant stages exactly like `ENV` does. So #2446's ARG+ENV inheritance is the correct one; mine was subtly wrong (too conservative, could false-fail on a var that's only ever an inherited stage-scoped ARG).

Pushed `ecda720c`: removed my superseded block, kept #2446's, and folded in a comment with the empirical finding above so the next person doesn't have to re-derive it. `scripts/tests/test_lint_dockerfile_envvars.py` (#2446's own regression test) still passes, the full `scripts/tests/` suite is green (37 passed / 1 skipped — a macOS-only `realpath` gap, unrelated), and all 7 Dockerfiles in the repo still validate cleanly.

diegocastanibm (commented):
LGTM

diegocastanibm (approved):
LGTM

PR #2420, opened 2026-08-31
fix: request 1 DRA claim for SGLang GKE prefill to match TP=1

zetxqx:
The SGLang GKE prefill patch attaches 4 DRA claims per pod, but prefill runs with `--tensor-parallel-size=1`, so each pod allocates 4 GPUs and uses only 1. With 8 replicas the guide asks for 40 GPUs instead of 16, which is part of why the GKE GPU SGLang nightly cannot schedule.

Reduce prefill to 1 claim, matching the vLLM GKE overlays (`gpu/vllm/gke/base`, `gpu/vllm-ds/gke`). Decode keeps 4 claims for its TP=4.


zetxqx:
@rahulgurnani @huaxig Could you help take a look? This should make the nightly scheduling easier for sglagn nightly

rahulgurnani:
Looks good to me. Thanks for the change!

PR #2424, opened 2026-09-01
fix: add shared HF cache and disable Xet for vLLM GKE PD overlay

zetxqx:
Mirror the SGLang GKE overlay: use a node-local hostPath `HF_HOME` cache shared across pods and restarts, and set `HF_HUB_DISABLE_XET=1`

to fix nightly benchmark GKE vllm PD, e.g: https://github.com/llm-d/llm-d/actions/runs/33466012376


guides/pd-disaggregation/modelserver/gpu/vllm/gke/base/kustomization.yaml
```diff
@@ -32,9 +32,25 @@ patches:
                 env:
                   - name: UCX_IB_ROCE_REACHABILITY_MODE
                     value: "all"
+                  - name: HF_HOME
```
ahg-g, line 35: why not apply this across all providers?

PR #2425, opened 2026-09-01
feat(guides): add diffusion-serving text-to-speech guide

zetxqx:
Adds a diffusion-serving text-to-speech guide: Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice on vLLM-Omni behind the llm-d router, served over the OpenAI-compatible POST /v1/audio/speech endpoint.

- Routes on load via active-request-scorer; prefix-cache-aware routing is follow-up work tracked in llm-d/llm-d-router#2483.
- Requires an EPP with /v1/audio/speech request parsing from llm-d/llm-d-router#2484 (landed after router v0.10.0).
- Verified end to end on GKE (1x H100 per replica): non-streaming WAV, SSE stream (usage in speech.audio.done), raw audio stream, and even load spread across replicas under a concurrent burst.


zetxqx:
@ahg-g  this should be ready for review now

PR #2436, opened 2026-09-04
fix(ci): separate GKE native offloading badge

git-jxj:
The GKE GPU native-offloading and LMCache nightlies write to the same badge file, so whichever finishes last sets the status shown for both tests.

Use `tiered-prefix-cache-gke-cpu-native` for the native workflow and update its link in the release testing matrix.

Verified that the two workflows produce distinct badge URLs and that the README matches the generated matrix. Pre-commit still reports Markdown table formatting and undeclared workflow inputs; these also fail on unmodified main.


diegocastanibm (approved):
@maugustosilva I might need to update the `sync-nightly-matrix.py` script, but this modification makes sense to me

PR #2437, opened 2026-09-04
fix(ci): reject HTTP errors in gateway smoke tests

git-jxj:
The gateway smoke validator currently reports success when inference endpoints return HTTP errors with JSON bodies. For example, twenty HTTP 400 responses can still produce `All 10 iterations succeeded` because curl normally exits zero for HTTP errors.

Add `--fail-with-body` to chat and completion requests, retain the response body, and print the HTTP status. Failure messages distinguish HTTP 4xx/5xx, curl/kubectl execution errors, and responses containing no JSON object. The existing nightly smoke validator exercises these diagnostics.

Validation:
- Local HTTP server and real curl: HTTP 200 completes all ten iterations; independent HTTP 400/500 errors on both endpoints fail and retain response-body diagnostics and pod cleanup.
- Direct shell checks cover HTTP error, execution error, missing JSON object, and success diagnostics.
- `bash -n`, `shellcheck -x --severity=warning`, applicable changed-file pre-commit hooks, and `git diff --check` pass.

Local validation used a temporary kubectl adapter outside the repository. No live Kubernetes E2E run was performed.


diegocastanibm (commented):
I wonder why do we need that test in python to test a shell script that exercise an e2e test.
98 lines of Python fixture to guard a 6-line bash change. Python mocking kubectl to test bash is an awkward layer. Finally this test it not wired into any pipeline, so right now it has zero value and nonzero maintenance surface.
The root cause of the lost time was that a harness bug produced an error message pointing at cluster config. It was not "no unit test".

My recommendation is to fix the diagnostics, and drop the Python test. The nightly then catches the bug and names it correctly.

maugustosilva (approved):

PR #2438, opened 2026-09-04
fix(ci): wait for flow-control request bursts

git-jxj:
The flow-control validator launches background request bursts inside command substitutions, so their PIDs cannot reliably be waited on by the parent shell. The final metrics scrape can run before requests finish and report misleading flow-control configuration failures.

Start each band in the parent shell and capture its PID with `$!`. Wait for every band, including when an earlier band fails. Before evaluating metrics, check each band's process exit status and require the expected number of HTTP 2xx responses. Failed or incomplete bursts print their logs and stop validation before configuration assertions. These diagnostics run in the existing nightly validator.

Validation:
- Six temporary local shell scenarios cover concurrent success, failures in the premium and standard processes, HTTP 500, HTTP 302, and missing responses.
- Every scenario waits for all three bands before cleanup; request failures stop before flow-control metric assertions.
- `bash -n`, `shellcheck`, applicable changed-file pre-commit hooks, and `git diff --check` pass.

Local validation used temporary fixtures outside the repository. No live Kubernetes E2E run was performed.


diegocastanibm (commented):
Comment in the same line as this one:
https://github.com/llm-d/llm-d/pull/2437#pullrequestreview-5114509852

maugustosilva (approved):

PR #2443, opened 2026-09-06
feat(guides): serve encode disaggregation on the vLLM nightly image

### LINKED ISSUE #2696: token-producer: address render across discovered engine endpoints instead of one URL
`token-producer` takes a single `vllm.url`, so render capacity is bounded by what one Service can expose, which is one endpoint per pod regardless of how many render-capable processes the fleet actually runs.

The wide-EP precise cell makes the gap concrete. Two prefill pods and one decode pod, DP8 per role, so 24 vLLM API servers, every one of them able to serve `/v1/*/render`. The guide's render Service selects the prefill pods and forwards to the rank-0 port, so it reaches 2 of those 24. A Service cannot do better: it has one `targetPort`, and a named port resolves per pod, so each pod contributes exactly one endpoint. Selecting decode as well would make it 3. Ranks 1 through 7 are unreachable through any single-URL Service.

This is the ceiling #2025 describes from the other side. That issue asks for observability and sizing guidance because "whatever the replica count, the ceiling exists" - a fixed pool moves the cliff without removing it. Addressing more endpoints is the other half: capacity that scales with the fleet rather than with a replica count someone has to predict.

Wonder if we could reuse the discovery the precise producer already runs. `precise-prefix-cache-producer` builds a `kvevents.SubscriberManager` whose `podDiscoveryConfig` watches pods by label selector and connects to a per-rank port derived by convention. Render needs the same two things: the set of engine pods, and a port per rank. Pointing the tokenizer at discovered endpoints instead of a Service name would take the reference cell from 2 render processes to 24 without deploying anything extra, and would keep working as the fleet scales.

Design points worth settling before code:

- **Which ports.** ~~For decode the engine ports are not the pool's `targetPorts`.~~ **Corrected:** they are usable. The pool's decode endpoint is the routing sidecar (`--port=8000 --vllm-port=8200` in the wide-EP guide), whose catch-all (`mux.Handle("/", s.decoderProxy)` in `pkg/sidecar/proxy`) reverse-proxies unregistered paths to the local engine, so a render sent to the pool port is served. I originally cited `pkg/coordinator/server/passthrough.go`, which forwards to the gateway instead, but that belongs to `cmd/coordinator`, a different binary. So the inference port is a workable default even on a disaggregated pool. An explicit port mapping is still needed where render is on a separate listener, where the sidecar terminates TLS since discovery addresses endpoints over HTTP, or where a sidecar does not forward render at all.
- **Readiness.** A Service only sends to Ready endpoints. Discovery-based addressing has to reproduce that, or render calls land on pods still loading weights.
- **Balancing.** Render cost scales with prompt length, so least-outstanding beats round-robin on mixed traffic. With per-connection balancing through a Service today this is implicit; addressing endpoints directly makes it a choice we own.
- **Failure handling.** One slow or dead endpoint should cost one retry elsewhere, not the full `vllm.timeout`. This is where the fail-fast idea in #2025 fits naturally.
- **Keep the single-URL mode.** The sidecar, a dedicated `vllm launch render` pool, and SGLang backends that do not serve the render endpoints all still need it, so this should be an additional addressing mode rather than a replacement.
- **Connection pooling.** The render transport is HTTP/1.1 with `MaxIdleConnsPerHost` 16. Fanning out across many endpoints changes what that bound means per endpoint.

Related: llm-d/llm-d#2439 and llm-d/llm-d#2440, where the same guide's render Service was returning 404 because vLLM now gates its scale-out endpoints, which silently disabled precise routing with requests still returning 200. Same class of failure, in that the tokenization path degrades without saying so.

cc @vMaroon

revit13:
## Summary

Move the Coordinator Disaggregation guide and the multimodal Encode Disaggregation
guide (E/PD and E/P/D) from pinned development vLLM images to the upstream vLLM
nightly image. Configure the
[ECCPU Connector](https://docs.vllm.ai/en/latest/features/ec_cpu_connector/) in its
P2P NIXL mode, with the V2 model runner that it requires. In the Coordinator guide,
replace the Coordinator's CPU render sidecar with a Service that fronts the prefill
model servers. Add CoreWeave overlays for E/PD and E/P/D. Make the vLLM model
servers and the Coordinator pull their moving image tags (`:nightly`, `:main`) on
every start.

The nightly pin is temporary: we expect to move to the vLLM 0.29.0 release image
once it ships, since encoder-cache transfer is not in any released version yet.

### Both guides: vLLM nightly image and ECCPU Connector

- **Image:** the `gpu-vllm/nightly` component (`docker.io/vllm/vllm-openai:nightly`)
  replaces `ghcr.io/revit13/vllm-openai:61e53561a` (Coordinator guide) and the
  `gpu-vllm/ec-connector` component (`ghcr.io/llm-d/vllm-openai:ec-cpu-connector-support`,
  Encode Disaggregation guide). The component already adds the `USER=llm-d`
  workaround for vllm-project/vllm#44548, so the inline copies are removed.
- **`imagePullPolicy: Always`:** the `gpu-vllm/nightly` component now sets it on the
  vLLM container. `:nightly` moves every day, and with `IfNotPresent` a node that
  cached an older nightly runs a different vLLM build. In a live E/PD test, one encode
  pod ran vLLM 0.28.1 from a cached image while the other pods ran 0.29.1. The EC
  connector compatibility check includes the vLLM version.
- **`VLLM_USE_V2_MODEL_RUNNER=1`:** set only on instances that have
  `--ec-transfer-config`, because `ECCPUConnector` raises `ValueError` without the V2
  model runner:
  - Coordinator guide: encode, prefill
  - E/P/D: encode, prefill
  - E/PD: encode, decode
- **EC config:** P2P NIXL mode (`"ec_enable_nixl": true`). `ec_cpu_bytes`, which is
  required and has no default, replaces `num_ec_blocks`. The sizes keep the earlier
  capacity (Qwen3-VL-32B uses 40,960 bytes per encoder token):

  | Instances | Before | `ec_cpu_bytes` | Blocks | `/dev/shm` |
  |---|---|---|---|---|
  | E/P/D encode, E/P/D prefill, E/PD encode | `num_ec_blocks: 1000000` | 40 GiB | 1,048,576 | 90Gi |
  | E/PD decode | default (100,000) | 4 GiB | 104,857 | 20Gi |
  | Coordinator encode, prefill | default (100,000) | 4 GiB | 104,857 | 8Gi |

- **Docs:** each `--ec-transfer-config` has a comment that links to
  https://docs.vllm.ai/en/latest/features/ec_cpu_connector/.

### Coordinator Disaggregation guide

- **`--enable-scale-out` on all model servers:** `vllm serve` serves `/render`,
  `/derender` and `/inference/v1/generate` only with this flag
  (vllm-project/vllm#55176 replaced `VLLM_ENABLE_SCALE_OUT_ENDPOINTS` with it).
  Prefill needs it for render. All roles need it because the Coordinator sends
  `/inference/v1/generate` to every phase for `/inference/v1/generate`, Responses
  and Messages requests.
- **Render Service (`render/`, new):** render (`/v1/*/render`) is CPU-only work in
  the vLLM API server, so any vLLM instance that serves the model can handle it.
  Before this change, all render calls went to one `vllm launch render` sidecar in
  the Coordinator pod. llm-d/llm-d-router#2701 (fixes llm-d/llm-d-router#2696) adds
  render balancing across discovered endpoints, but only for the EPP
  `token-producer`. The Coordinator's `render` step still takes one `address`. So
  this guide spreads the render load with a Service: `coord-disaggregation-render`
  fronts the prefill pods on their `modelserver` port, and render capacity grows
  with the prefill replicas. The render step address is
  `http://coord-disaggregation-render.${NAMESPACE}.svc:8000`. The `.svc` suffix
  matches `NO_PROXY` in `multimedia-downloader/patch-coordinator-ca.yaml`, so render
  calls do not go through the media proxy.
  - Prefill only, not all vLLM pods: render on decode slows token streaming.
    Encode sets no `--max-model-len`, so render there accepts prompts that prefill
    and decode reject. Prefill returns one token per request, so its API server has
    spare CPU.
  - The `vllm-render` sidecar (`vllm-openai-cpu:v0.26.0`) and its volumes are
    removed. Render now runs on the same vLLM version as the model servers, and no
    extra copy of the model is downloaded.
- **Coordinator image:** `ghcr.io/llm-d/llm-d-router-coordinator:main` now uses
  `imagePullPolicy: Always`, because `:main` moves with every merge to llm-d-router
  main.
- **README:** step 2 applies `render/`, Cleanup deletes it, and the image note
  describes the nightly image and the EC settings.

### Encode Disaggregation guide

- **CoreWeave overlays (new):** `e-pd/coreweave` and `e-p-d/coreweave` add
  `rdma/ib: "1"` to each role. `rdma/ib` is removed from the E/P/D base decode and
  prefill patches, so the base and GKE overlays no longer request it. The README
  lists `coreweave` as a provider.
- **README:** describes the nightly image, the V2 model runner and the P2P NIXL
  settings. The GKE bullet no longer claims DRA and DRANet (RoCE), because the GKE
  overlays do not configure RDMA yet.

## Defaults

- `--enable-scale-out`: off in `vllm serve` when not set. Set only in the
  Coordinator guide. The multimodal guides do not use the scale-out endpoints and
  keep the default. vLLM builds from before 2026-09-14 do not know the flag.
- `VLLM_USE_V2_MODEL_RUNNER`: not set on decode in the Coordinator guide and in
  E/P/D. vLLM then uses V2 by default (vllm-project/vllm#53183) and falls back to V1
  when a feature needs it.
- `ec_cpu_bytes`: required, with no default. vLLM does not check it against the
  `/dev/shm` size, so each value is below its pod's `/dev/shm` limit.
- `imagePullPolicy`: Kubernetes uses `IfNotPresent` when it is not set and the tag is
  not `:latest`. It is now `Always` on the vLLM container in every guide that uses
  `gpu-vllm/nightly` (E/PD decode and wide-EP GLM-5.2 already set it) and on the
  Coordinator.
- Render step `address`: no default in the Coordinator. It is now the Service URL.

## Testing

- [x] `kustomize build` passes for the Coordinator model servers (`base`,
  `coreweave`, `gke`), `coordinator/` and `render/`, and for E/PD and E/P/D
  (`base`, `coreweave`, `gke`). All 38 kustomizations that use `gpu-vllm/nightly`
  build, and the only change there is `imagePullPolicy: Always`.
- [x] Checked the built manifests. Every Deployment uses
  `docker.io/vllm/vllm-openai:nightly`. No env var name is duplicated. The V2 runner
  is set only on EC instances. The render Service selects only prefill pods. The
  render address host matches `NO_PROXY`.
- [x] Deployed E/P/D on CoreWeave and sent an image request. The encode pod logged
  `EC save`, `serving NIXL READ request` and `NIXL READ completed`, and prefill
  answered. A text-only request skipped encode.
- [x] Deployed E/PD on CoreWeave and sent an image request. The encode pod logged
  `NIXL READ completed`, and decode answered.
- [x] Deployed the Coordinator guide on CoreWeave (3 EPPs) and sent a text and an
  image request through the Gateway. Render went through `coord-disaggregation-render`
  to a prefill pod (`/v1/chat/completions/render` 200, 83 tokens, 1 image), and the
  encode pod logged `NIXL READ completed`. Both requests returned 200.

## Notes

- The Service balances connections, not requests. For more render capacity, add
  prefill replicas or CPU. llm-d/llm-d-router#2701 balances render per request, but
  only in the EPP. If the Coordinator's `render` step gets a similar option, it can
  replace this Service.
- The `ec_cpu_bytes` sizes depend on the model. For another model, calculate
  `out_hidden_size * (1 + deepstack layers) * dtype bytes` per token.
- GKE RDMA: the E/PD, E/P/D and Coordinator GKE overlays do not configure RDMA yet,
  so the NIXL KV-cache and encoder-cache transfers use TCP on GKE. We will open a
  follow-up issue to add the `gke-rdma` DRA component (one GPU+NIC claim per TP rank),
  as in the P/D Disaggregation guide, once it can be tested on a GKE A3/A4 cluster
  with DRA and DRANet.


omerpaz95:
LGTM.

roytman (commented):
This PR removes the last two consumers of `gpu-vllm/ec-connector` (`e-pd/base` and `e-p-d/base` switch their `components:` entry to `gpu-vllm/nightly`). After this change, `guides/recipes/modelserver/components/images/gpu-vllm/ec-connector/` has zero references anywhere in the repo.

That component was already a near-duplicate of `gpu-vllm/nightly` (identical `USER=llm-d` workaround patches for `Deployment`/`LeaderWorkerSet`/`DisaggregatedSet`; the only difference was the pinned image tag `ec-cpu-connector-support` vs `nightly`), so nothing is lost by the switch, and reusing the already-shared `nightly` component (used by `tiered-prefix-cache` and `wide-ep-lws`) instead of updating `ec-connector` in place avoids ending up with two identical components under different names.

Two loose ends this PR leaves behind:
- `guides/recipes/modelserver/components/images/gpu-vllm/ec-connector/kustomization.yaml` is now dead and could be deleted.
- `guides/recipes/modelserver/components/images/README.md` still lists `ec-connector` in its directory tree and describes it in prose as the current example of a pinned dev-build stop-gap, which is no longer accurate once this merges.

Not a blocker, worth a follow-up cleanup (or a one-line mention in the PR description) so a future reader has a signal it's unused.

PR #2447, opened 2026-09-07
fix(multimodal): avoid duplicate USER environment

git-jxj:
The E/P/D encode deployment includes `USER=llm-d` twice. Remove the duplicate while retaining the existing workaround value. Regression tests render both base and GKE overlays, check every container for duplicate environment names, and confirm the encode deployment still has exactly one `USER=llm-d` entry.

Validation: `python -m pytest -q scripts/tests`: 30 passed, including both kustomize renders.
Changed-file pre-commit checks passed.


git-jxj:
The current head is `70959ec758519a5bc01461d69e824d575be64e55`. Signed-commits and DCO are successful.

The current head's pull-request workflows still need GitHub Actions approval: [CI - PR Checks](https://github.com/llm-d/llm-d/actions/runs/34109617982), `CI Guides Check`, `CI Guide Dry Run`, `Non-Main Gatekeeper`, `Prow Remove LGTM`. They report `action_required`, so these checks have not executed. Could a maintainer approve the pending workflow runs for `70959ec75`?

scripts/tests/test_epd_user_env.py
```diff
@@ -0,0 +1,42 @@
...
+
+ROOT = Path(__file__).resolve().parents[2]
+OVERLAY_ROOT = (
+    ROOT
+    / "guides/multimodal-serving/e-disaggregation/modelserver/gpu/vllm/e-p-d"
+)
+
+
+@pytest.mark.parametrize("variant", ["base", "gke"])
+def test_rendered_containers_have_unique_environment_names(variant: str) -> None:
```
ahg-g: this is overkill, I wouldn't add this test file.

git-jxj: Removed `scripts/tests/test_epd_user_env.py` in abf0381b, as requested. The PR now contains only the three-line removal of the duplicate `USER` entry. I also checked both the base and GKE overlays with `kustomize build`: all rendered container environment names are unique, and the encode container retains `USER=llm-d`.

PR #2458, opened 2026-09-08
feat(guides): add Kueue-based replica rebalancing to workload-autoscaling

asm582:
## What

Adds a second replica-rebalancing path to the workload-autoscaling guide that
enforces a shared GPU budget with [Kueue](https://kueue.sigs.k8s.io/) instead of
a control loop over HPAs.

Each model gets a `ClusterQueue` with a guaranteed GPU floor, all of them in one
cohort, and with Kueue's `deployment` integration every replica pod becomes a
`Workload` that is admitted only when quota is free. Enforcement therefore
happens *below* the HPA, on the pods it creates:

- the HPA needs no annotation, and KEDA keeps sole ownership of the HPA it
  generates — so this works with KEDA-generated HPAs as they are
- an idle model's GPUs are lent to a busy one, and a model reclaims its floor by
  preemption (`preemption.reclaimWithinCohort: Any`)
- over-budget demand becomes gated `Pending` pods rather than a lowered
  `maxReplicas`, so "desired" and "ready" replicas legitimately differ

This does not change the existing
[experimental replica rebalancer](https://github.com/llm-d/llm-d/tree/main/guides/workload-autoscaling/replica-rebalancing);
it sits alongside it as a second Features entry. The two enforce the same budget
from opposite sides of the HPA and must never run at the same time, so the new
guide assumes the rebalancer is uninstalled and that the hard
`requests.nvidia.com/gpu` `ResourceQuota` it reads is gone (a gated pod still
counts against a `ResourceQuota`, which would stop the ReplicaSet from creating
the very pod Kueue is meant to queue).

## Files changed

| File | Change |
|---|---|
| `guides/workload-autoscaling/README.md` | new `### Kueue-Based Replica Rebalancing (Experimental)` entry under Features, next to the existing rebalancer entry |
| `guides/workload-autoscaling/kueue-rebalancing/guide.yaml` | new — machine-readable guide source: env, Kueue install, quota apply, opt-in, rollout strategy, verification and cleanup steps |
| `guides/workload-autoscaling/kueue-rebalancing/README.md` | new — the guide itself, with every bash block rendered from `guide.yaml` by `scripts/guide.py` |
| `guides/workload-autoscaling/kueue-rebalancing/kueue.values.yaml` | new — Kueue chart overrides that confine the controller and its webhooks to the model namespace |
| `guides/workload-autoscaling/kueue-rebalancing/optimized-baseline/base/kustomization.yaml` | new — quota overlay |
| `.../base/resourceflavor.yaml` | new — one `ResourceFlavor` for the accelerator pool being rationed |
| `.../base/clusterqueues.yaml` | new — two `ClusterQueue`s (floors of 5 + 5 across 10 GPUs) sharing cohort `llm-d-gpu`, with `reclaimWithinCohort: Any` |
| `.../base/localqueues.yaml` | new — one `LocalQueue` per model in the workload namespace |
| `.github/scripts/install-kueue-crds.sh` | new — installs the Kueue CRDs for the dry-run job |
| `.github/workflows/ci-kustomize-dry-run.yaml` | one step added to run the above |

All objects use `kueue.x-k8s.io/v1beta2`, the storage version since Kueue 0.16.
The guide pins the chart at 0.19.1.

## CI

The `dry-run` job auto-discovers every `kustomization.yaml` under `guides/*/`, so
it picks up the new quota overlay with no workflow change — but it would have
failed without the Kueue CRDs, since the objects are all custom resources. Kueue
publishes no CRD-only release asset and the chart's CRD templates contain Helm
directives, so `install-kueue-crds.sh` renders them from the chart the same way
the other per-operator installers work:

```bash
helm template kueue oci://registry.k8s.io/kueue/charts/kueue \
  --version "${KUEUE_VERSION}" -n kueue-system \
  -s templates/crd/kueue.x-k8s.io_clusterqueues.yaml \
  -s templates/crd/kueue.x-k8s.io_localqueues.yaml \
  -s templates/crd/kueue.x-k8s.io_resourceflavors.yaml \
  -s templates/crd/kueue.x-k8s.io_workloads.yaml \
  | kubectl apply --server-side -f -
```

The step is placed alongside the other CRD installers in
`ci-kustomize-dry-run.yaml`, so the new overlay is `kustomize build`-ed and
server dry-run like every other overlay in `guides/`.

## Validation

Static checks:

- `kustomize build` on the quota overlay — 5 objects
- every Kueue object checked field-by-field against the `v1beta2` CRD OpenAPI
  schemas rendered from the 0.19.1 chart (field names, enums, required fields)
- `scripts/guide.py render --recursive --check guides` passes, so `README.md` and
  `guide.yaml` are in sync as CI requires
- yamllint and markdownlint with the repo's configs; relative links and anchors
  resolve; `bash -n` on the new script

End to end on a kind cluster (Kueue 0.19.1, k8s 1.32, one node advertising 10
emulated `nvidia.com/gpu`, the shipped floors of 5 + 5), running every bash block
rendered in `README.md` in order and unedited — install, quota apply, rollout
strategy, opt-in, all four verification tests, cleanup. All exited 0, and:

- installing with `kueue.values.yaml` left `mpod.kb.io` selecting only the model
  namespace; a pod labelled `kueue.x-k8s.io/queue-name` in an unrelated namespace
  ran ungated with no `Workload` created
- the pod-template opt-in produced one `Workload` per replica, and cleanup removed
  the label and rolled both Deployments back with no quota objects or `Workload`s
  left behind
- model A scaled to 6 against a floor of 5 → 6 admitted, `borrowed=1`
- model B then scaled to 5 → A's borrowed replica preempted (`due to reclamation
  within the cohort`), A's replacement pod `Pending` behind
  `kueue.x-k8s.io/admission,kueue.x-k8s.io/topology`, `borrowed` back to 0; both
  queues settled on their floors when scaled back down
- with the chart-default `maxSurge: 25%` at full quota the opt-in rollout stalls on
  a gated surge pod and `kubectl rollout status` exits 1; with `maxSurge: 0`
  applied first the same rollout completes — which is why the guide sets the
  strategy before opting in
- the new `verify` assertion exits 0 on a healthy cluster and 1 both when the
  queues cannot be reconciled and when `cohortName` is removed from one
  `ClusterQueue`

## Not tested on OpenShift

This has only been validated on upstream Kubernetes (kind). It has **not** been
tested on OpenShift, and specifically not with the Red Hat build of Kueue
operator, which the guide mentions as an alternative to the upstream chart. The
quota objects should be identical either way, but that path needs confirmation
before anyone relies on it.

The OpenShift path also differs materially: the Red Hat build of Kueue is
installed as an operator with a cluster-scoped `Kueue` CR rather than this Helm
chart, so `kueue.values.yaml` does not apply there, `integrations.frameworks` must
be set explicitly on the CR, and namespaces need `kueue.openshift.io/managed=true`.
The README documents that route, but it needs verification by someone with a
cluster.

### DIFF
--- a/.github/scripts/install-kueue-crds.sh
+++ b/.github/scripts/install-kueue-crds.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+# Install Kueue CRDs for CI validation of the Kueue-based rebalancing guide.
+#
+# On real clusters Kueue is installed as an operator (or via its Helm chart).
+# This script installs only the CRDs so kubectl dry-run can validate
+# ResourceFlavor, ClusterQueue and LocalQueue resources without deploying the
+# controller and its webhooks.
+#
+# Kueue publishes no CRD-only release asset, so the CRDs are rendered from the
+# chart's own templates with --show-only.
+
+set -euo pipefail
+
+KUEUE_VERSION=${KUEUE_VERSION:-"0.19.1"}
+KUEUE_CHART=${KUEUE_CHART:-"oci://registry.k8s.io/kueue/charts/kueue"}
+
+# Only the kinds the guide's overlays declare, plus Workload, which is what the
+# guide's verification commands read.
+CRDS=(
+  clusterqueues
+  localqueues
+  resourceflavors
+  workloads
+)
+
+show_only=()
+for crd in "${CRDS[@]}"; do
+  show_only+=(-s "templates/crd/kueue.x-k8s.io_${crd}.yaml")
+done
+
+echo "Installing Kueue CRDs (v${KUEUE_VERSION})..."
+helm template kueue "${KUEUE_CHART}" \
+  --version "${KUEUE_VERSION}" \
+  -n kueue-system \
+  "${show_only[@]}" \
+  | kubectl apply --server-side -f -
+
+echo "Kueue CRDs installed."
--- a/.github/workflows/ci-kustomize-dry-run.yaml
+++ b/.github/workflows/ci-kustomize-dry-run.yaml
@@ -54,6 +54,9 @@ jobs:
       - name: Install KEDA CRDs
         run: .github/scripts/install-keda-crds.sh
 
+      - name: Install Kueue CRDs
+        run: .github/scripts/install-kueue-crds.sh
+
       - name: Install Envoy Gateway CRDs
         run: .github/scripts/install-envoy-gateway-crds.sh
 
--- a/guides/workload-autoscaling/README.md
+++ b/guides/workload-autoscaling/README.md
@@ -109,3 +109,15 @@ annotated HPAs when multiple models share a GPU budget. Integration with
 KEDA-generated HPAs requires annotation-propagation support and is not covered
 by the KEDA+EPP guide. The feature is a proof of concept and is not
 production-ready; use it only in test or development environments.
+
+### Kueue-Based Replica Rebalancing (Experimental)
+
+The [Kueue-Based Replica Rebalancing guide](./kueue-rebalancing/README.md)
+enforces the same shared GPU budget with [Kueue](https://kueue.sigs.k8s.io/)
+instead of a control loop over HPAs. Each model gets a `ClusterQueue` with a
+guaranteed GPU floor in a shared cohort, and replica pods are admitted only when
+quota is free — so an idle model's GPUs are lent to a busy one and reclaimed by
+preemption. Enforcement happens below the HPA, so KEDA-generated HPAs work
+unmodified, but over-budget demand shows up as `Pending` pods rather than a
+lowered `maxReplicas`. Do not run this alongside the experimental replica
+rebalancer.
--- a/guides/workload-autoscaling/kueue-rebalancing/README.md
+++ b/guides/workload-autoscaling/kueue-rebalancing/README.md
@@ -33,23 +33,6 @@ when demand returns.
          admitted pods → scheduled    over-budget pods → gated (Pending)
 ```
 
-## How It Compares to the Replica Rebalancer
-
-| | Replica rebalancer | Kueue |
-|---|---|---|
-| Where the budget is enforced | above the HPA, by patching `maxReplicas` | below the HPA, by gating pods |
-| GPU budget defined as | `ResourceQuota` `requests.nvidia.com/gpu` | `ClusterQueue` `nominalQuota`, shared through a cohort |
-| Reaction time | the `COORDINATOR_INTERVAL` loop (default 15s) | event-driven, on pod creation and deletion |
-| Opt-in | `llm-d.ai/epp-inference-pool` annotation on the HPA | `kueue.x-k8s.io/queue-name` label on the Deployment |
-| GPUs per replica | computed by the controller | read from the pod's `resources.requests` |
-| KEDA-generated HPAs | needs annotation propagation, not covered | works unchanged — Kueue never reads the HPA |
-| Reclaiming lent capacity | not supported | `preemption.reclaimWithinCohort: Any` |
-| Over-budget demand | never created, because the ceiling drops | created and queued as gated pods |
-| Extra components | WVA controller with the coordinator enabled | Kueue |
-
-Both keep total GPU usage inside the budget. They must not run at the same time:
-each would react to the capacity the other freed.
-
 ## Prerequisites
 
 1. Multiple inference pools in one namespace, from the
--- a/guides/workload-autoscaling/kueue-rebalancing/README.md
+++ b/guides/workload-autoscaling/kueue-rebalancing/README.md
@@ -1,0 +1,358 @@
+# Kueue-Based Replica Rebalancing
+
+When several model deployments share one GPU budget, their HPAs scale
+independently and none of them knows what the others are consuming. Two models
+scaling up at the same time collectively ask for more GPUs than the budget
+holds.
+
+The [experimental replica rebalancer](../replica-rebalancing/README.md) solves
+this *above* the HPA: a control loop reads a `ResourceQuota` and patches
+`spec.maxReplicas` on annotated HPAs so the ceilings always sum to the budget.
+
+This guide solves it *below* the HPA with [Kueue](https://kueue.sigs.k8s.io/).
+With the `deployment` integration, every replica pod becomes its own Kueue
+`Workload`, and Kueue holds a pod at a scheduling gate until GPU quota is free
+for it. The HPA is never touched, so it needs no annotation and KEDA keeps sole
+ownership of the HPA it generates. Each model gets a guaranteed floor of GPUs,
+lends what it is not using to the others, and preempts to take its floor back
+when demand returns.
+
+```text
+      KEDA ScaledObject            KEDA ScaledObject
+              │                            │
+        HPA (model-a)                HPA (model-b)     ← untouched, unannotated
+              │  replicas                  │  replicas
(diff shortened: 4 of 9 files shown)


asm582:
Pushed a fix commit (f1fec37) addressing the review feedback, and validated the guide end to end on a local kind cluster (Kueue 0.19.1, k8s 1.32, one node advertising 10 fake `nvidia.com/gpu`) before pushing.

### What changed

**Kueue is now confined to the model namespace.** With chart defaults the `pod` integration is on and every webhook inherits `managedJobsNamespaceSelector`, so `mpod.kb.io` intercepts pod CREATE in every namespace except `kube-system` and `kueue-system`, with `failurePolicy: Fail`. New `kueue.values.yaml` names the model namespace in that selector and the install step passes `-f`. The knob lives inside `controllerManagerConfigYaml`, a single string Helm replaces rather than merges, so the file repeats the 0.19.1 defaults verbatim, with a note to re-diff against `helm show values` when bumping the version.

**Opt-in moved to the pod template.** Kueue's `vdeployment` webhook treats `metadata.labels[kueue.x-k8s.io/queue-name]` on a Deployment as immutable once `status.readyReplicas > 0`, and rejects removing it at any replica count. The original Step 3 patch therefore failed on any serving model server:

```text
Error from server (Forbidden): admission webhook "vdeployment.kb.io" denied the request:
metadata.labels[kueue.x-k8s.io/queue-name]: Invalid value: "model-a-lq": field is immutable
```

Labelling `spec.template.metadata.labels` works instead — each replica pod becomes its own Workload through the `pod` integration — and it can be patched back out at cleanup. The Deployment-level label is still the right choice at install time, so Step 3 now shows that as a kustomize patch on the decode Deployment.

**Deploy step order fixed.** The opt-in patch changes the pod template, and at full quota the default `maxSurge: 25%` surge pod has no GPU to claim: Kueue gates it and `kubectl rollout status` never returns. `spec.strategy` is now set first (that patch does not touch the pod template, so it rolls nothing), then opt-in.

Smaller items: cluster-scoped queue names carry an `llm-d-` prefix and a common label so verification can select them out of a shared cluster; the LocalQueue namespace moved into `kustomization.yaml`; `verify` now asserts each ClusterQueue is `Active=True` and still in its cohort and exits non-zero otherwise, instead of printing status that passed either way; the OpenShift note documents the Red Hat build of Kueue (`openshift-kueue-operator`, cluster-scoped `Kueue` CR, required capitalised `integrations.frameworks`, `kueue.openshift.io/managed=true`); the one-way Deployment label is recorded under Limitations.

### Files changed

| File | Change |
| --- | --- |
| `guides/workload-autoscaling/kueue-rebalancing/README.md` | rewritten Steps 1–4, Limitations and Cleanup |
| `guides/workload-autoscaling/kueue-rebalancing/guide.yaml` | values-file install, `GUIDE_ROOT`/`QUOTA_ROOT`, strategy-before-opt-in, pod-template opt-in, `admission` assertion |
| `guides/workload-autoscaling/kueue-rebalancing/kueue.values.yaml` | new — namespace-scoped `managedJobsNamespaceSelector` |
| `.../optimized-baseline/base/clusterqueues.yaml` | `llm-d-` prefixes, comments on nominal-vs-physical quota |
| `.../optimized-baseline/base/localqueues.yaml` | namespace removed, taken from kustomization |
| `.../optimized-baseline/base/resourceflavor.yaml` | `llm-d-gpu-default`, note on heterogeneous hardware |
| `.../optimized-baseline/base/kustomization.yaml` | `namespace:` + common label with `includeSelectors: false` |

### What was validated locally

Every rendered bash block from the README was executed in order, unedited, against the kind cluster: `prerequisites.kueue` → `prerequisites.integrations` → `deploy.quota` → `deploy.rollout_strategy` → `deploy.optin` → all four `verify` tests → `cleanup`. All exited 0.

- **Webhook scope.** After installing with `kueue.values.yaml`, `mpod.kb.io` reports `namespaceSelector` matching only the model namespace. A pod carrying `kueue.x-k8s.io/queue-name` created in an unrelated namespace ran ungated with no Workload created.
- **Opt-in and cleanup are reversible.** The pod-template patch rolled both Deployments, produced one Workload per replica, and the cleanup patch removed the label and rolled them back with no Workloads, ClusterQueues, ResourceFlavor or LocalQueues left behind.
- **Borrowing.** Model A scaled to 6 against a floor of 5 → 6 admitted, `borrowed=1` on its ClusterQueue.
- **Reclaim by preemption.** Model B then scaled to 5 (its full floor) → A's borrowed replica was preempted (`due to reclamation within the cohort; preemptor path: /llm-d-gpu/llm-d-model-b-cq; preemptee path: /llm-d-gpu/llm-d-model-a-cq`), A's replacement pod sat Pending behind `kueue.x-k8s.io/admission,kueue.x-k8s.io/topology`, and `borrowed` returned to 0. Scaling both back to 1 settled both queues on their floors.
- **Rollout ordering.** With chart-default `maxSurge: 25%` at full quota, the opt-in rollout stalls on a gated surge pod and `kubectl rollout status` exits 1; with `maxSurge: 0` applied first the same rollout completes.
- **`verify` assertion.** Confirmed it exits 0 on a healthy cluster, and exits 1 both when the queues cannot be reconciled (controller scaled to 0) and when `cohortName` is removed from one ClusterQueue.

### CI

`python3 scripts/guide.py render guides/workload-autoscaling/kueue-rebalancing --check` reports up to date, and markdownlint (pinned `markdownlint-cli` v0.47.0 with the repo's `.markdownlint.json`) reports no new findings. All commits are DCO signed off.

### Not tested on OpenShift

Everything above was run on kind with fake GPU capacity. This has **not** been tested on OpenShift, and the OpenShift path differs materially: the Red Hat build of Kueue is installed as an operator with a cluster-scoped `Kueue` CR rather than this Helm chart, so `kueue.values.yaml` does not apply there, `integrations.frameworks` must be set explicitly on the CR, and namespaces need `kueue.openshift.io/managed=true`. The README documents that route but it needs verification by someone with a cluster.

asm582:
Addressed the `for` loop feedback in 75e913a — none of those steps actually needed iteration.

- **Same patch for both model servers** (rollout strategy, and the label removal at cleanup): `kubectl patch` takes both names in one call.
- **Different patch per model** (the opt-in): written out as one patch plus one `rollout status` per model, so adding a third model is a copy of the pair rather than another entry in a colon-delimited list.
- **Patch bodies** moved from JSON to the flow-YAML form `kubectl patch` also accepts, which removes the backslash-escaped quotes and lets `${QUEUE_A}` interpolate directly:

```bash
kubectl patch deployment ${DEPLOYMENT_A} -n ${NAMESPACE} --type=merge \
  -p "spec: {template: {metadata: {labels: {kueue.x-k8s.io/queue-name: ${QUEUE_A}}}}}"
```

- **The two verification loops** became single expressions over the whole collection: the ClusterQueue check prints an `ACTIVE`/`COHORT` table via `custom-columns` and asserts with one `jq -e`; the gated-workload report reads every Workload in one `kubectl get` instead of one call per name.

Re-verified on kind after the rewrite — the rendered blocks take both Deployments from unlabelled to admitted, borrow (`borrowed=1`), reclaim by preemption, report the gated Workload's `insufficient unused quota for nvidia.com/gpu in flavor llm-d-gpu-default` reason, and roll the label back out with nothing left behind. The `admission` assertion still exits 0 on a healthy cluster and 1 both when `cohortName` is removed from a ClusterQueue and when the queues are absent entirely.

lionelvillard (approved):

PR #2466, opened 2026-09-09
docs: add token-aware autoscaling guide (KEDA + EPP token backlog)

asm582:
## Summary

Adds a new autoscaling path under `guides/workload-autoscaling/`: **token-aware autoscaling**, where KEDA scales model servers on *tokens of outstanding work* rather than a count of requests.

The existing KEDA + EPP paths (`keda-epp-queue`, `keda-epp-saturation`) scale on request-shaped signals — queue depth, running requests, pool saturation. Those hold well when prompt sizes are homogeneous and drift when they are not: an 8192-token prompt is 16× the prefill work of a 512-token one, and a request counter rates them the same, so the same request rate can represent a third of a replica or three replicas of prefill work. This path counts the tokens themselves.

It needs no new components — KEDA and Prometheus only, same as the sibling guides — and no EPP config change, because both base guides already register the plugins it reads from.

## How it works

Each phase is measured in its own unit, because they are bound by different resources.

**Prefill is a rate problem.** `llm_d_epp_inflight_tokens` is a token backlog; dividing it by the calibrated `peakPrefillThroughput` converts tokens into **seconds of queue wait**:

```promql
(
  sum(llm_d_epp_inflight_tokens{namespace="llm-d-optimized-baseline",
                                producer_name="inflight-load-producer"}) / 15928
) or vector(0)
```

The useful consequence is that the threshold becomes **derivable from a latency target** instead of tuned by trial:

```text
threshold ≈ TTFT_SLO − (ISL / V_P)
            └ product ┘  └ idle floor ┘
```

`ISL / V_P` is the irreducible idle TTFT for a prompt. Autoscaling only ever removes *queueing*, so the threshold is whatever the SLO has left after that floor. The guide works this through for three workload shapes and shows the arithmetic. It also flags the case where `TTFT_SLO ≤ floor`, in which no threshold can meet the SLO and scaling is the wrong lever — you need a higher `V_P` or a smaller uncached ISL.

**Decode is an occupancy problem.** No measured rate exists for it. `vllm:kv_cache_usage_perc` is already a fraction of capacity, so the threshold is compared to it directly and `0.8` reads as "act at 80 % full":

```promql
sum(vllm:kv_cache_usage_perc{namespace="llm-d-pd-disaggregation", pod=~".*decode.*"}) or vector(0)
```

This is a vLLM metric rather than an EPP one deliberately: the autoscaler needs a KV reading **per pod** so it can filter by role and aggregate. The EPP's only KV metric is a single pool-wide average covering both roles, which cannot be split.

## Layout

Follows the `<strategy>/<topology>` layout of `keda-epp-saturation`, with `base` / `k8s` / `ocp` under each topology:

```
keda-epp-token-aware/
├── README.md
├── guide.yaml
├── optimized-baseline/      # P+D co-located (kv_both)
│   └── base/ k8s/ ocp/
└── pd-disaggregation/       # prefill and decode split across pods
    └── base/ k8s/ ocp/
```

| | P+D co-located | P/D-disaggregated |
|---|---|---|
| Objects | one `ScaledObject`, two triggers | one `ScaledObject` per role |
| Scaling rule | `max(ceil(backlog_s / 1.5), ceil(kv / 0.8))` — the current bottleneck wins | prefill and decode scale independently |
| Trade-off | one calibration, one object; a prompt-heavy and a generation-heavy burst both scale the whole pod | tracks lopsided load between phases, at the cost of two objects |

The token math is identical in both. Only the query shape (role selectors) and the number of `ScaledObject`s differ.

The `ocp` overlays reuse `recipes/autoscaling/metrics-reader` for Thanos bearer auth, matching the sibling guides, and give the cluster-scoped `ClusterRoleBinding` a namespace-unique name so deployments do not collide across namespaces or across the guides sharing that recipe.

## Points worth reviewer attention

**Both triggers use `sum()`, never `avg()` or `max()`.** With `metricType: AverageValue` the HPA computes `ceil(value / threshold)`, so the numerator must grow with the pool. A per-pod average cannot exceed ~1.0 regardless of fleet size, so `ceil(avg / 0.8)` pins a Deployment at 2 replicas — and worse, the average *drops* each time a replica is added, so it oscillates rather than holding. `sum()` grows with the pool and `AverageValue` divides it back out per replica, which is what makes `0.8` mean "80 % full per pod".

**The plain `sum()` prefill query depends on llm-d-router#2577.** Before that fix, `inflight_tokens` series were not pruned when an endpoint was removed (llm-d-router#2529), so a scaled-down pod stranded a frozen series and the sum could never fall back — scale-up worked, scale-down did not. #2577 is on `main`, which is what `ROUTER_EPP_VERSION` defaults to, but it merged after `v0.10.0`, so it is not in a tagged release yet. The guide therefore keeps a short subsection documenting the liveness-intersection workaround for anyone pinned to `v0.10.0` or earlier. **That subsection can be dropped if a release carrying #2577 ships before this merges** — happy to do so on request.

**`peakPrefillThroughput` appears in two places** and is read by different components: on `prefix-cache-affinity-filter` (the EPP router, for prefix-cache-aware routing) and as the divisor inside the trigger PromQL (KEDA, for the unit conversion). The guide is explicit that these are separate and that kustomize cannot reach inside a PromQL string, so the divisor is a documented hand edit — same as the label selectors in `keda-epp-saturation`.

## Validation

Done locally against this branch:

- `scripts/guide.py check` passes; `scripts/guide.py render --check` reports the README up to date, so the marker rendering is idempotent
- `kubectl kustomize` builds all six overlay directories; the rendered `ocp` output was inspected to confirm the PromQL backreferences survive, the namespace transformer rewrites the `ClusterRoleBinding` subject, and both triggers are repointed at Thanos with `authModes: bearer`
- `pre-commit` passes on every changed file, including `markdownlint` and `yamllint`
- all relative links and internal anchors in the new README resolve
- every numeric cell in the threshold table was re-derived from `ISL`, `V_P` and the SLO and checked against the rendered text

**Not yet done: an end-to-end run on a cluster.** The queries and the design are ported from a working deployment, but this specific overlay set has not been applied to a live cluster and driven through a scaling event. Treating this as a draft for that reason. Marking the guide `[Experimental]` in the title, consistent with `keda-epp-saturation`.

The guide is not added to the CI dry-run exclusions, so the kustomize dry-run job will cover all six overlays.

## Also in this PR

`guides/workload-autoscaling/README.md` gains the new path in the Paths section and a column in the Choosing a Scaling Signal table.

That file also carried three pre-existing `markdownlint` violations (two `MD013` long lines, one `MD001` heading-increment). Since CI lints changed files, touching the file made them blocking, so they are fixed here: two paragraphs split at sentence boundaries with no wording changed, and two `####` headings under an `##` corrected to `###`. Heading anchors are text-derived, so no inbound links change. Called out to explain the otherwise-unrelated moved lines.

## Reference

The token-velocity framing — sizing each role by dividing a token rate by that role's token throughput, so both phases share a denominator in tokens/s — follows *"TokenScale: Timely and Accurate Autoscaling for Disaggregated LLM Serving with Token Velocity"* ([arXiv:2512.03416](https://arxiv.org/abs/2512.03416)), cited in the guide.

### DIFF
--- a/guides/workload-autoscaling/README.md
+++ b/guides/workload-autoscaling/README.md
@@ -56,9 +56,17 @@ Adapter.
 
 The [Saturation-based Autoscaling](./keda-epp-saturation/README.md) path has KEDA query Prometheus directly for InferencePool-scoped saturation and running-request metrics, then scale the model server Deployment. KEDA consumes two EPP metrics — pool saturation level (0.0–1.0+, normalized measure of how loaded the pool is) and active in-flight request count — and drives scaling decisions. Well-suited for simple homogeneous deployments where each model scales independently and you want minimal operational overhead.
 
+#### Token-Aware Autoscaling
+
+The [Token-Aware Autoscaling](./keda-epp-token-aware/README.md) path scales on **tokens of outstanding work** rather than a count of requests. An 8192-token prompt is 16x the prefill work of a 512-token one, and a request or queue-depth counter rates them the same, so the same request rate can mean a third of a replica or three replicas of prefill work.
+
+This path measures each phase in its own unit: prefill divides the EPP's in-flight token backlog by a calibrated `peakPrefillThroughput` to get **seconds of queue wait**, comparable directly against a share of your TTFT SLO, while decode reads per-pod KV cache occupancy. Overlays are provided for both topologies: P+D co-located (both phases in one pod) and P/D-disaggregated (split across pods). Best when prompt sizes vary widely, or when you want the prefill threshold derived from a latency target rather than tuned by trial. Requires one calibration run per hardware/model combination.
+
 #### SLO-Aware Autoscaling (KEDA + estimated latency)
 
-A specialization of this path drives the HPA from the pool's **latency** rather than its queue depth. The [SLO-Aware Autoscaling](./slo-aware/README.md) sub-path scales directly against latency SLOs, using the EPP's **estimated** TTFT/TPOT as the signal — either its ML-predicted latency (from the online-trained predictor) or the actual measured latency aggregated in real time when the predictor isn't enabled. A Prometheus recording rule turns that estimate into a single saturation ratio (latency ÷ SLO), and a KEDA `ScaledObject` with an [expr-lang](https://expr-lang.org/) formula computes the desired replica count and drives a standard HPA — no custom controller. With the ML predictor, capacity is added as pressure builds rather than after the queue has already formed. Best when clients express per-request latency SLOs and you want scaling driven by the objective itself rather than a proxy metric.
+A specialization of this path drives the HPA from the pool's **latency** rather than its queue depth. The [SLO-Aware Autoscaling](./slo-aware/README.md) sub-path scales directly against latency SLOs, using the EPP's **estimated** TTFT/TPOT as the signal — either its ML-predicted latency (from the online-trained predictor) or the actual measured latency aggregated in real time when the predictor isn't enabled.
+
+A Prometheus recording rule turns that estimate into a single saturation ratio (latency ÷ SLO), and a KEDA `ScaledObject` with an [expr-lang](https://expr-lang.org/) formula computes the desired replica count and drives a standard HPA — no custom controller. With the ML predictor, capacity is added as pressure builds rather than after the queue has already formed. Best when clients express per-request latency SLOs and you want scaling driven by the objective itself rather than a proxy metric.
 
 ### KEDA + WVA Metrics (Legacy)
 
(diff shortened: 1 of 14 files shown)


asm582:
Pushed `be80a09b`, addressing the three defects and two gaps from the Kind-cluster walkthrough. Each was reproduced against a live KEDA 2.20.2 apiserver before fixing, and re-verified after.

**Verify step two was broken.** `kubectl describe hpa | grep -A2 "External metric"` matched nothing and exited 1 — confirmed. kubectl labels those rows `(target average value)` and writes `external metric` only in lower case, inside the `FailedGetExternalMetric` message.

One correction to the suggested fix: `grep -i` is not the right repair. On my reproduction it *did* match — but only because that HPA was broken, so the string appeared in the failure message. On a healthy HPA it is absent entirely, so `grep -i` prints output on failure and nothing on success. Inverted, and silently so. I dropped the grep instead:

```bash
kubectl get hpa -n ${NAMESPACE} -o jsonpath='…ScalingActive…'
kubectl get hpa -n ${NAMESPACE}
```

Neither command can produce empty output, which felt like the right property for a step whose whole purpose is catching silent success.

That run also handed over a better diagnostic than the guide had, now documented: `0/1.500 (avg)` means the query resolved and the pool is idle, while `<unknown>/1.500 (avg)` means the trigger never evaluated — the state KEDA otherwise reports only as an HPA condition while serving `fallback` replicas.

**`pollingInterval: 15` removed from all three ScaledObjects.** I tested four states: as-shipped warns; removed is silent; `useCachedMetrics: true` is silent; `minReplicaCount: 0` is silent — which independently confirms your scale-to-zero point. I chose removal over `useCachedMetrics` because it documents the behaviour that already exists instead of changing it, and added a comment recording that the HPA sync period governs and when the field would start mattering.

**Expected-output block replaced** with output captured from a live cluster — twelve columns, and `0/1.500 (avg), 0/800m (avg)`.

**Both gaps documented.** The closed-loop caveat now sits with the replica-count table, including the evenly-spread per-endpoint check that separates it from a stuck metric — that observation was the most useful thing in the report, since the table silently assumed open-loop arrivals. And the generic-Kubernetes-without-Prometheus-auth case is now covered; I tested that patch snippet rather than just writing it, since shipping another unrun command seemed like a poor response to this review.

**Left alone deliberately:** the parent README's "Supported (Needs Validation)" scale-to-zero cell, which is still accurate — we ship `minReplicaCount: 1` and nobody has validated zero. The guide now says what scale-to-zero would require.

Still outstanding, unchanged by this: no GPU run, so nothing here speaks to whether `V_P` is stable under real batching — which is the assumption the threshold derivation rests on. The P/D topology remains validated only to `--dry-run=server`, and the `ocp` overlays build but have never been applied.

lionelvillard (approved):

PR #2467, opened 2026-09-09
feat(workload-autoscaling): add AMD ROCm support for the WVA nightly

PrateekKumar1709:
## Summary

Adds nightly coverage for the `workload-autoscaling` guide on the AMD ROCm CI cluster, alongside the existing OCP and CKS lanes. Split per review into #2469, which carried the workflows and has merged, and this PR.

This half is what the lane deploys: a kustomize component that retargets the KEDA ScaledObject at the AMD model server and MI355X, the deploy script the nightly invokes, and a pointer to the component from the wva guide.

The ROCm cell in the nightly matrix was originally part of this PR. #2471 regenerated the matrix after #2469 merged and picked the cell up already, so it is dropped here. It reports `resource not found` until the lane's first run writes `badges/wva-rocm.json` to gh-pages.

## Schedules

The workflows merged dispatch-only, so that main never referenced a deploy script that was not there yet. With the script landing here, the 13:00 UTC nightly and the 15:00 UTC status roll-up go back on their crons, clear of the 03:00, 05:00, 07:00 and 10:00 slots the other four ROCm nightlies hold. That is the three line workflow change in this PR.

## Design choices

**Standup runs through the deploy script.** The guide composes onto `optimized-baseline` rather than shipping a model server of its own, so the lane stands the composed stack up through the reusable's `custom_deploy_script` hook and keeps every resource in one namespace.

**Prometheus wiring is discovered, not pinned.** The script reads the monitoring namespace from the Prometheus CR and configures the controller and the KEDA trigger for whichever scheme the stack serves, creating the `prometheus-tls-cert` Secret the controller mounts when it is TLS. One script therefore stays correct across clusters instead of encoding a single cluster's monitoring setup. If the expected Prometheus service is absent it fails immediately with a clear message rather than timing out later against a missing metric source.

**vLLM metrics are guaranteed before autoscaling is asserted.** The lane layers the model server monitoring component on so vLLM is scraped. Without it WVA's saturation engine has no input and publishes a desired count equal to the current one, which satisfies every health check while exercising no autoscaling.

**Sized and scheduled to share the cluster.** Two replicas at TP2, matching the four GPUs the lane reserves, with no PriorityClass and no preemption. ClusterRoleBindings carry a namespace hash and label, so concurrent runs cannot collide and teardown cannot strand cluster-scoped RBAC.

## Test plan

Ran the deploy script and `e2e-validate.sh` on MI355X exactly as the workflow invokes them, in the nightly namespace.

WVA produces real decisions rather than echoing the current replica count:

```
analyzer-result   {"modelID": "Qwen/Qwen3-32B", "analyzer": "saturation", "supply": 3043200, "demand": 0, ...}
scaling-decision  {"decisions": [{"name":"optimized-baseline-rocm-vllm-decode-scaler","curr":2,"tgt":1,"action":"scale-down"}]}
EmitReplicaMetrics completed  {"currentReplicas": 2, "desiredReplicas": 1, "accelerator": "MI355X"}
```

KEDA consumes that metric rather than its fallback:

```
NAME                                        READY  ACTIVE  FALLBACK  TRIGGERS
optimized-baseline-rocm-vllm-decode-scaler  True   True    False     prometheus
```

`e2e-validate.sh` passed 10/10 through the EPP, and teardown left no stranded ClusterRoleBindings and released every GPU.

KEDA is a cluster prerequisite rather than something the lane installs, and is installed and healthy on the CI cluster.

## On the /test-nightly run

The run triggered from this PR failed in `build-wva-image`, before it reached a cluster. The WVA Dockerfile pins `quay.io/projectquay/golang:1.26` by digest `4142f907`, and quay re-pushed that tag on 2 September under digest `74560d1d`, so the pinned digest no longer resolves. This is not specific to the AMD lane: the CKS WVA nightly has failed at the same job every night going back to mid August. llm-d/llm-d-workload-variant-autoscaler#1541, the bump from golang 1.26 to 1.27, re-pins to a live digest and would unblock every WVA nightly.

## Related

- CI half, merged: #2469
- Reusable workflow the lane calls: llm-d/llm-d-infra#243

### DIFF
--- a/.github/workflows/consolidate-status-workload-autoscaling-amd-acc-rocm-vllm-x.yaml
+++ b/.github/workflows/consolidate-status-workload-autoscaling-amd-acc-rocm-vllm-x.yaml
@@ -19,6 +19,9 @@ on:
         type: string
         default: 'main'
 
+  schedule:
+    - cron: '0 15 * * *'  # 15:00 UTC daily (after the 13:00 UTC nightly)
+
 jobs:
   check_for_success:
     uses: llm-d/llm-d-infra/.github/workflows/reusable-query-success-past-runs.yaml@main
--- a/.github/workflows/nightly-e2e-workload-autoscaling-amd-acc-rocm-vllm-x.yaml
+++ b/.github/workflows/nightly-e2e-workload-autoscaling-amd-acc-rocm-vllm-x.yaml
@@ -8,14 +8,12 @@ name: Nightly - Workload Autoscaling E2E (AMD ROCm)
 # chart on main may require features not yet in a tagged release image.
 #
 # The ROCm cluster is shared with four other nightlies, so this run does not
-# preempt: allow_gpu_preemption stays false and the deploy script sets no
-# PriorityClass.
-#
-# Dispatch-only for now. The 13:00 UTC schedule, which is clear of the
-# 03:00/05:00/07:00/10:00 slots the other four ROCm nightlies hold, is added
-# alongside the deploy script this lane calls.
+# preempt: allow_gpu_preemption stays false, the deploy script sets no
+# PriorityClass, and the schedule is clear of the 03:00/05:00/07:00/10:00 slots.
 
 on:
+  schedule:
+    - cron: '0 13 * * *'  # 13:00 UTC daily (clear of the other four ROCm nightlies)
   workflow_dispatch:
     inputs:
       image_tag:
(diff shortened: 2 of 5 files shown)


clubanderson:
### Nightly E2E Result: `workload-autoscaling` (AMD ROCm)

| | |
|---|---|
| Result | :x: **Failed** |
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/34511310302) |
| Ref | `50234fb` |

clubanderson:
### Nightly E2E Result: `workload-autoscaling` (AMD ROCm)

| | |
|---|---|
| Result | :x: **Failed** |
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/34511310302) |
| Ref | `50234fb` |

clubanderson:
### Nightly E2E Result: `workload-autoscaling` (AMD ROCm)

| | |
|---|---|
| Run | [View logs](https://github.com/llm-d/llm-d/actions/runs/34511310302) |
| Ref | `50234fb` |

lionelvillard (approved):

PR #2468, opened 2026-09-09
Update async guides + observability script

jtechapps:
- Update multitenant guide to integrate Flow Control router InferenceObjectives, tier-priority merge policy, and per-model worker pools
- Unify Cloud Monitoring dashboard across Redis and Pub/Sub backends using pure llm_d_async_* Prometheus metrics:
  - Add separate tiles for broker backlog and in-process queue depth
  - Add tiles for exceeded deadline request rate and token processing throughput
  - Add documentation note that deadline proximity is specific to Redis Sorted Sets
- Add multi-tenant end-to-end stress testing and heavy benchmark scripts
- Add standalone manifests for vLLM, Prometheus, Grafana, and InferenceObjectives
- Update Helm value overlays to modern unified transport configuration

### DIFF
--- a/guides/batch-serving/asynchronous-processing/multitenant/README.md
+++ b/guides/batch-serving/asynchronous-processing/multitenant/README.md
@@ -34,32 +34,72 @@ model — and within each pool three teams contend by **tier** and **quota**. Th
 
 The three dimensions:
 
-- **Model → pool isolation.** The publisher sends a request to the `(team, model)` queue and sets
-  `payload.model`; the gateway routes it to that model's `InferencePool`. Each pool has its own workers
-  and its own saturation gate, so one model saturating does not park the other.
+- **Model → pool isolation.** The publisher sets `payload.model`, which the gateway routes to that model's
+  `InferencePool`. Each model gets its own worker pool (`model-a`, `model-b`) and independent saturation gate,
+  so one model saturating does not park the other.
+- **Queue as the serving dimension (tier × model).** A queue represents a **serving SLA tier** for a target model
+  (e.g., interactive latency-sensitive, standard async, or batch throughput), **not a rigid single-tenant silo**.
+  In production, multiple distinct teams can publish into the **same queue** concurrently. The request payload
+  identifies the team via `metadata.team`.
 - **Team → reservation classification.** The per-team **`redis-quota`** gate runs in **`classifying`**
-  mode (keyed on `metadata.team`, with a **per-model prefix** so each `(team, model)` has its own
-  counter): within quota → `reserved` (org-guaranteed), over quota → `overflow` (admitted and
-  deprioritized, **not** nacked).
-- **Tier → priority.** A per-queue `tier` label: `interactive` (premium) > `async` (standard) > `batch`.
+  mode (keyed dynamically on `metadata.team`, with a **per-model prefix** such as `quota:a:team:<team>`):
+  each team maintains its own independent concurrency counter in Redis. When multiple teams share a serving queue,
+  one team exceeding its quota does not exhaust another team's budget: within quota → `reserved` (org-guaranteed),
+  over quota → `overflow` (admitted and deprioritized, **not** nacked).
+- **Tier → priority.** A per-queue `tier` label: `interactive` (premium SLA) > `async` (standard SLA) > `batch`.
 
 The [**tier-priority merge policy**](https://github.com/llm-d/llm-d-async/pull/294) runs
 **per pool independently**: within each model it buckets requests into **6 strict lanes** by
-`(classification, tier)`, dispatches them in order, and stamps **`x-gateway-priority`** (0 = highest …
-5 = lowest):
+`(classification, tier)`, dispatches them in order, and stamps both **`x-gateway-priority`** (0 = highest …
+5 = lowest) and **`x-llm-d-inference-objective`** via `lane_objectives`.
 
-| Lane | `x-gateway-priority` | Who (within one model) |
-| :-- | :-- | :-- |
-| reserved + interactive | 0 | premium within quota |
-| reserved + async | 1 | standard within quota |
-| reserved + batch | 2 | batch within quota |
-| overflow + interactive | 3 | premium over quota |
-| overflow + async | 4 | standard over quota |
-| overflow + batch | 5 | batch over quota |
+By defining matching [`InferenceObjective`](#1-apply-inferenceobjectives-and-deploy-flow-control-router)
+resources in the cluster, `llm-d-async` and `llm-d-router` Flow Control speak the exact same language.
+Requests carry the authoritative objective and tenant identity (`x-llm-d-inference-fairness-id`), allowing
+`llm-d-router` to enforce multi-tenant fairness and priority band admission at the gateway:
+
+| Lane | Objective (`lane_objectives`) | Header `x-llm-d-inference-objective` | Router Band Priority | Who (within one model) |
+| :-- | :-- | :-- | :-- | :-- |
+| reserved + interactive | `reserved-interactive` | `reserved-interactive` | 100 | premium within quota |
+| reserved + async | `reserved-async` | `reserved-async` | 60 | standard within quota |
+| reserved + batch | `reserved-batch` | `reserved-batch` | 30 | batch within quota |
+| overflow + interactive | `overflow-interactive` | `overflow-interactive` | 10 | premium over quota |
+| overflow + async | `overflow-async` | `overflow-async` | 0 | standard over quota |
+| overflow + batch | `overflow-batch` | `overflow-batch` | -10 | batch over quota |
 
 So within each model **all reserved traffic drains before any overflow** (org priority), tier-ordered
 within each class; and the two models are fully independent. On Redis SortedSet, within a lane dispatch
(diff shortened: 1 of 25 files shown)


guides/batch-serving/asynchronous-processing/multitenant/dashboards/cloud-monitoring.json
```diff
@@ -1,5 +1,5 @@
 {
-  "displayName": "Async Processor \u2014 Multi-tenant Pub/Sub Demo",
+  "displayName": "LLM-D Async Processor",
```
ahg-g: llm-d is always small letters

shimib:
Scenario D's gate will never close as configured
  The prometheus-saturation gate queries Prometheus for the metric:
  1 - inference_extension_flow_control_pool_saturation{inference_pool="<pool>"}

  However, the guide never sets up scraping for the router (EPP), so this metric is completely missing from Prometheus. When a metric query returns empty data, the dispatch gate silently falls back to its fallback value
  (which defaults to 0.0 saturation / 1.0 budget).

  As a result, the gate stays 100% open, never saturates, and Scenario D runs without actually testing backpressure.
   * Fixes needed: Add a ServiceMonitor/PodMonitor for the router/EPP, and update the verification snippet to assert that async_gate_metric_source_available == 1.
   * Typo/Correction: The gate doesn't query "directly from llm-d-router's metrics endpoint" (it queries the Prometheus Server URL), and the metric name is inference_extension_flow_control_pool_saturation (not
     llm_d_epp_flow_control_pool_saturation).

  benchmark-heavy-redis.py queries a Prometheus that doesn't exist
  The python helper query_prom attempts to run kubectl exec deploy/prometheus in ${NAMESPACE}. Under a standard kube-prometheus-stack setup, Prometheus is deployed as a StatefulSet (prometheus-k8s) in llm-d-monitoring.

  Because of the bare except: block in the Python script, this failure is silently swallowed, and the In-Flight and In-Process columns simply display 0 for the entire run.

  Duplicate Heading
  The heading #### With Flow Control OFF (Baseline Router with Saturation Detection) appears twice. The second section should be renamed to reflect that saturation detection is disabled.

  InferenceObjectives only render for pool A
  The render() function maps POOL_NAME to ${POOL_A}. In a multi-pool isolation setup, this leaves Pool B's lanes without mapped objectives. Since the new default path is single-pool, we should update the introductory text
  to match, or re-introduce the "Single-model variant" note as an inverted guide.

guides/batch-serving/asynchronous-processing/multitenant/README.md
```diff
@@ -34,33 +34,74 @@ model — and within each pool three teams contend by **tier** and **quota**. Th
 
 The three dimensions:
 
-- **Model → pool isolation.** The publisher sends a request to the `(team, model)` queue and sets
-  `payload.model`; the gateway routes it to that model's `InferencePool`. Each pool has its own workers
-  and its own saturation gate, so one model saturating does not park the other.
+- **Model → pool isolation.** The publisher sets `payload.model`, which the gateway routes to that model's
```
ahg-g: does the async processor add something to the request to ensure that it matches the httproute fronting the specific infpool serving the model?

guides/batch-serving/asynchronous-processing/multitenant/README.md
```diff
@@ -34,33 +34,74 @@ model — and within each pool three teams contend by **tier** and **quota**. Th
...
+| overflow + batch | `overflow-batch` | `overflow-batch` | -10 | batch over quota |
 
 So within each model **all reserved traffic drains before any overflow** (org priority), tier-ordered
 within each class; and the two models are fully independent. On Redis SortedSet, within a lane dispatch
 is earliest-deadline-first (the deadline is the sorted-set score).
 
+### Priority Values: Flow Control ON vs. Flow Control OFF
+
+The system establishes two complementary priority representations:
+- **`x-gateway-priority` (0 to 5):** A scalar integer priority where **0 is highest and 5 is lowest**. This is standard for generic reverse proxies, HTTP load balancers, or Envoy ingress filters that order dispatch based on ascending integer rank.
```
ahg-g: What acts on this and why is it needed given that we already have inferenceObjectives? also, if this is an llm-d defined header? it should be prefixed with `x-llm-d`

guides/batch-serving/asynchronous-processing/multitenant/README.md
```diff
@@ -34,33 +34,74 @@ model — and within each pool three teams contend by **tier** and **quota**. Th
...
+- **`x-llm-d-inference-objective` & Router Band Priority (100 to -10):** The Kubernetes [`InferenceObjective`](#1-apply-inferenceobjectives-and-deploy-flow-control-router) specification, where **higher numerical values represent higher scheduling priority**.
+
+#### With Flow Control ON (`llm-d-router`)
+When `llm-d-router` is deployed with Flow Control enabled (`featureGates: [flowControl]` in `flow-control.yaml`):
+- **Centralized Priority Bands:** When model server capacity saturates (detected in real time via `concurrency-detector` or `utilization-detector`), requests are held in memory across priority bands matching the `InferenceObjective` priority (100, 60, 30, 10, 0, -10).
+- **Strict Band Dispatch:** The gateway scheduler drains highest-priority bands first: all `reserved` bands (100, 60, 30) dispatch before any `overflow` band (10, 0, -10) is admitted.
```
ahg-g: pls don't use the term gateway and use `llm-d router` when referring to functionality done by EPP

ahg-g (commented):
A couple of comments:

1) I think we need to clear up the terminology of "gateway" vs "llm-d router"; I think we should only use gateway if we are actually referring to k8s gateway and using httpRoutes etc., if we are referring to inference routing and functionality done by EPP, we should use the term "llm-d Router".

2) It seems we have two types of priorities, that part can be a confusing and I hope we can converge on one priority concept.

guides/batch-serving/asynchronous-processing/multitenant/README.md
ahg-g: is this actually referring to a k8s gateway with httpRoutes?

jtechapps: dispatching request means sending it to the inference endpoint which uses epp ext-proc. 

>does the async processor add something to the request to ensure that it matches the httproute fronting the specific infpool serving the model?

It only dispatches the request to the llm-d Router endpoint and sets extra headers such as inference objective.

I will reword this to make it more clear. Every model would have a separate endpoint anyway.

guides/batch-serving/asynchronous-processing/multitenant/README.md
jtechapps: >What acts on this and why is it needed given that we already have inferenceObjectives?

It's not needed with the inference objective stamping. I will remove. x-gateway-priority was tell downstream router of priority but llm-d-router uses inferenceObjectives mappings so it's not needed. It was still left in when we added InferenceObjective mapping.

guides/batch-serving/asynchronous-processing/multitenant/dashboards/cloud-monitoring.json
jtechapps: Fixed

guides/batch-serving/asynchronous-processing/multitenant/README.md
jtechapps: Reworded this and gateway is only mentioned in context of when HttpRoutes are used.

guides/batch-serving/asynchronous-processing/multitenant/README.md
jtechapps: Removed this section and only mention inference objective stamping. Created https://github.com/llm-d/llm-d-async/issues/441 in llm-d-async to change default stamp behavior.

guides/batch-serving/asynchronous-processing/multitenant/README.md
jtechapps: Changed terminology here and elsewhere.

jtechapps:
Also addressed Shimi's comments:

1. Changed the router values to render servicemonitor so that a separate podmonitor isn't needed.

> "the metric name is inference_extension_flow_control_pool_saturation (not
> llm_d_epp_flow_control_pool_saturation)" 

`llm_d_epp_flow_control_pool_saturation` is the new metric name and `inference_extension_flow_control_pool_saturation` is deprecated. llm-d-async uses the old metric and this needs to be fixed. I created https://github.com/llm-d/llm-d-async/issues/440. I have the deprecated metric name used in this PR. I can either update the README to use the new metric name now or change it after there is an llm-d-async release with the metric rename.

2. benchmark-heavy-redis.py queries a Prometheus that doesn't exist
Fixed dead Prometheus queries in benchmark-heavy-redis.py: Replace failing kubectl exec deploy/prometheus with direct HTTP PromQL queries supporting --prom-url and background port-forwarding.

3. Fixed the heading it now says "without Saturation Detection"

4. InferenceObjectives only render for pool A

The demo uses the same InferencePool and router deployment. I clarified InferenceObjectives pool binding: Document single-router demo binding via POOL_NAME and explain namespace isolation for multi-pool environments.

PR #2469, opened 2026-09-10
ci(workload-autoscaling): add AMD ROCm WVA nightly workflows

PrateekKumar1709:
## What this adds

The CI half of the Workload Variant Autoscaler nightly lane for the AMD ROCm cluster, split out of #2467 so it can be merged and dispatched ahead of the code:

- `.github/workflows/nightly-e2e-workload-autoscaling-amd-acc-rocm-vllm-x.yaml`, the nightly caller
- `.github/workflows/consolidate-status-workload-autoscaling-amd-acc-rocm-vllm-x.yaml`, its past-status companion

The caller builds the WVA controller image from the autoscaler repo before testing, since the chart on main can require flags that are not present in the last tagged release image, then hands off to `reusable-nightly-e2e-amd.yaml` in llm-d-infra with the guide's deploy script, a 4 GPU request, and preemption disabled so the run coexists with the four other nightlies on the shared cluster.

## Dispatch-only for now

Both workflows omit their schedules here. The deploy script the caller invokes, `guides/workload-autoscaling/scripts/nightly-deploy-amd.sh`, lands in #2467, so a scheduled run against main before that merges could only fail. The 13:00 and 15:00 UTC crons are added in #2467 alongside the script, which keeps main free of a lane that cannot pass at any point.

Dispatch covers what is needed in the meantime. The caller takes an `llm_d_ref` input that overrides the ref the reusable checks out, so this lane can be run against the unmerged code by dispatching from main with `llm_d_ref` set to `refs/pull/2467/merge`. The same input also carries the release branch through when `release-e2e.yaml` dispatches the lane with `matrix_type`.

A dispatch holds 4 GPUs on the shared ROCm cluster for roughly 25 minutes.

## Validation

The lane has been run end to end on the ROCm CI cluster with the #2467 code in place: the autoscaler made real scaling decisions off live vLLM metrics, KEDA consumed the published desired replica count, `e2e-validate.sh` passed 10/10, and teardown left no residue.

## Related

- Code half: #2467
- Reusable workflow this calls: llm-d/llm-d-infra#243

### DIFF
--- a/.github/workflows/consolidate-status-workload-autoscaling-amd-acc-rocm-vllm-x.yaml
+++ b/.github/workflows/consolidate-status-workload-autoscaling-amd-acc-rocm-vllm-x.yaml
@@ -1,0 +1,29 @@
+name: Past Status - Workload Autoscaling E2E (AMD ROCm)
+
+on:
+  workflow_dispatch:
+    inputs:
+      workflow_name_query:
+        description: 'Name of the workflow to be queried'
+        required: false
+        type: 'string'
+        default: 'nightly-e2e-workload-autoscaling-amd-acc-rocm-vllm-x.yaml'
+      time_window_query:
+        description: 'How many days in the past to query'
+        required: true
+        type: 'string'
+        default: '5'
+      branch_query:
+        description: 'Branch to be queried'
+        required: false
+        type: string
+        default: 'main'
+
+jobs:
+  check_for_success:
+    uses: llm-d/llm-d-infra/.github/workflows/reusable-query-success-past-runs.yaml@main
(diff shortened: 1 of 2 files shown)


maugustosilva (approved):

maugustosilva:
@PrateekKumar1709 please sign your commits so we can merge

PrateekKumar1709:
> @PrateekKumar1709 please sign your commits so we can merge

Hi @maugustosilva, they are signed. That bot comment is stale and it was posted before the force-push, and GitHub leaves it in the timeline rather than updating it.

PR #2474, opened 2026-09-11
feat(observability): wire metric-to-trace navigation for EPP exemplars

### LINKED ISSUE #2462: [Feature]: [Observability] Grafana panel for metric-to-trace navigation via exemplars
### Feature Area

Observability / Monitoring

### Problem Statement

metrics tell you the fleet is slow, they don't tell you which request was slow.

an operator sees p99 jump on a dashboard and the only way to find an actual slow request is to guess a time range and scroll through traces by hand. the trace exists, there's just no path from the number to it.

llm-d/llm-d-router#2774 fixes the data side - the EPP now attaches the trace ID to llm_d_epp_request_duration_seconds as a prometheus exemplar and serves /metrics in openmetrics format so it survives the scrape. but nothing displays it yet, so the link is there and unusable.

### Proposed Solution

a panel under guides/recipes/observability/grafana/dashboards/ that shows the jump working.

what it needs:
- a latency panel (heatmap or p99 timeseries) over llm_d_epp_request_duration_seconds
- exemplars enabled on the prometheus datasource
- a datasource link from the exemplar's trace_id label to the traces datasource

end result: you see a spike, click the dot on it, and land on the trace for that exact request.

### Alternatives Considered

leaving it to operators to wire up themselves. the exemplars are exposed either way so this works, but the datasource link config isn't obvious and it's the kind of thing that's easier to ship once than to document.

adding it to an existing dashboard instead of a new panel - probably fine, whoever picks it up can decide. the diagnostic drilldown one might be the natural home.

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

follow up to llm-d/llm-d-router#2637, implemented in llm-d/llm-d-router#2774.

splitting it out because verifying it needs the whole stack running - an EPP serving real traffic, prometheus scraping it, a traces backend receiving spans, and grafana wired to both. i can't run that locally so i can't confirm the click-through actually works, only that the exemplar reaches the wire.

happy to write the panel json if someone with an environment can verify it, or hand it over entirely to whoever has the stack up.

one thing for whoever picks it up: prometheus needs exemplar storage on (--enable-feature=exemplar-storage) and has to request openmetrics on the scrape. without both, the exemplars get dropped in transit even though the EPP is emitting them correctly - and nothing errors, they just aren't there.

relates to SIG Observability.

sudoalok:
Closes #2462.

The EPP attaches trace IDs to `llm_d_epp_request_duration_seconds` as Prometheus exemplars (llm-d/llm-d-router#2774), but nothing in the recipe made them usable. Three things were missing, and none of them error when absent - the exemplar just quietly isn't there:

- `enableFeatures: [exemplar-storage]` - without it Prometheus parses the exemplar off the scrape and drops it
- a Jaeger datasource in Grafana - the tracing script installs Jaeger, the prometheus script only registered Prometheus, so the exemplar had nowhere to point
- `exemplarTraceIdDestinations` on the Prometheus datasource - without it `trace_id` renders as plain text instead of a link

The values file is written twice, once per mode, so all three go into both branches. The TLS branch already wrote its own `jsonData`; both settings are now built into a single block, because two `jsonData` keys under one datasource is valid YAML in which one silently wins.

Jaeger is optional and its namespace is chosen at install time, so the link is wired only when a `jaeger-collector` service is found, with `TRACING_NAMESPACE` to override.

The only dashboard change is `"exemplar": true` on the P99 query of the existing latency panel. All three quantiles read the same buckets, so enabling them all would draw the same points three times.

### Verifying

`scripts/test-prometheus-exemplar-config.sh` runs the real installer against stubbed `kubectl`/`helm` - no cluster needed - and asserts on the values file helm would actually receive. 43 assertions over six configurations: no tracing installed, central mode, central + TLS, individual mode, explicit `TRACING_NAMESPACE`, and a `TRACING_NAMESPACE` holding no Jaeger.

### DIFF
--- a/guides/recipes/observability/grafana/dashboards/llm-d-failure-saturation-dashboard.json
+++ b/guides/recipes/observability/grafana/dashboards/llm-d-failure-saturation-dashboard.json
@@ -446,6 +446,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
+          "exemplar": true,
           "expr": "histogram_quantile(0.99, sum by(le) (rate(llm_d_epp_request_duration_seconds_bucket{namespace=~\"$namespace\",model_name=~\"$model_name\"}[5m])))",
           "hide": false,
           "instant": false,
--- a/guides/recipes/observability/install-prometheus-grafana.sh
+++ b/guides/recipes/observability/install-prometheus-grafana.sh
@@ -12,6 +12,9 @@ KUBERNETES_CONTEXT=""
 DEBUG=""
 CENTRAL_MODE=true
 ENABLE_TLS=false
+# Namespace holding the tracing stack (installed separately by
+# install-otel-collector-jaeger.sh). Auto-detected when unset.
+TRACING_NAMESPACE="${TRACING_NAMESPACE:-}"
 
 ### HELP & LOGGING ###
 print_help() {
@@ -32,6 +35,8 @@ Options:
 
 Environment Variables:
   MONITORING_NAMESPACE        Override default monitoring namespace
+  TRACING_NAMESPACE           Namespace of the Jaeger install to link exemplars to
+                              (auto-detected when unset)
 
 Examples:
   $(basename "$0")                              # Install central monitoring in llm-d-monitoring (watches all namespaces)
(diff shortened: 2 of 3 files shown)


sudoalok:
I checked the assertions actually bite by breaking each fix in turn - dropping `enableFeatures` from one branch only, making the TLS branch emit its own `jsonData`, removing the link, and removing `exemplar: true` from the panel. Every mutation was caught.

 Grafana painting a clickable dot and Jaeger resolving the ID needs a running stack, which I can't run locally. If anyone has the recipe up, that's the one thing I'd like confirmed: click a dot on the latency panel and see whether it lands on the trace.
 
 cc @gyliu513 @PlateauGao

PR #2476, opened 2026-09-11
Hardware support: Rebellions NPU (optimized-baseline)

rebel-minhopark:
## What this adds

An `optimized-baseline` model server overlay for Rebellions NPUs, plus the image component
and accelerator docs it needs. Modeled on the `xpu` and `amd` overlays.

- `guides/optimized-baseline/modelserver/npu/vllm/` — two replicas of `openai/gpt-oss-120b`,
  one NPU each, consuming `recipes/modelserver/base/single-host`
- `guides/recipes/modelserver/components/images/npu-vllm/release/` — the image component
- `docs/getting-started/accelerators.md` — a Rebellions NPU section and a support-table row
- `guides/optimized-baseline/router/npu.rbln.values.yaml` — a load-only scheduling profile
- `guides/recipes/router/calibration/configuration-matrix.md` — the measured
  `peakPrefillThroughput` for this configuration

`ACCELERATOR_TYPE=npu`. Rebellions' operator advertises one resource name and one DRA
DeviceClass across its product families, so the accelerator is spelled `npu` rather than a
family name.

## Prefix caching, and what it does to routing

The runtime disables prefix caching for sliding-window models and `gpt-oss-120b` is one, so
the overlay passes `--no-enable-prefix-caching`. Enabling it anyway is not a workaround:
the server starts and counts queries but returns no hits, measured at
`vllm:prefix_cache_queries_total` 112640 against `vllm:prefix_cache_hits_total` 0, with
warm and cold TTFT within 2 ms of each other at both 2k and 8k prompts.

That matters for this guide because its default scheduling profile is prefix-cache aware.
Left alone, the `prefix-cache-affinity-filter` would pin requests to endpoints it believes
are cache-warm while the model servers reuse nothing, trading load balance for an affinity
that does not exist. `router/npu.rbln.values.yaml` replaces the profile with the
`token-load-scorer` alone. Rebellions expects prefix caching for these models in the GA
runtime; the flag and that file both go away then.

## Device allocation

NPUs come through DRA, not an extended resource. A single `npu.rebellions.ai` DeviceClass
covers every Rebellions product, so the claim narrows to the REBEL family with a CEL
selector on `device.attributes["npu.rebellions.ai"].productName`. This needs
`resource.k8s.io/v1`, which is Kubernetes 1.34 or later — `kubeconform -strict` finds no
schema for the claim template at `--kubernetes-version 1.33.0` and validates all three
resources at 1.34.0. The accelerator docs state that requirement.

This overlay claims one NPU and no NIC, so it sets no NUMA constraint.

## Verification

Run on one bare-metal node, Kubernetes 1.36.3, image `ghcr.io/rbln-sw/rbln-serve:0.11.3a1`
pulled anonymously with no pull secret.

| | |
| :-- | :-- |
| Both replicas ready | 458 s, on separate NPUs |
| `/v1/completions` | returns, `vllm 0.24.0` on `torch 2.11.0`, `vllm_rbln 0.11.3a7` |
| Load, per replica | `vllm bench serve`, random 1k in / 1k out, 64 requests, concurrency 8 |
| Replica A | 64/64 succeeded, 201.8 output tok/s, TPOT p99 4.95 ms |
| Replica B | 64/64 succeeded, 201.7 output tok/s, TPOT p99 4.94 ms |
| Anonymous memory | 50.7 GiB against the 157Gi limit |
| Restarts | 0 |

The calibration recipe defines peak prefill throughput as `CHUNK_SIZE / median(TTFT)` with
`CHUNK_SIZE` matching `--max-num-batched-tokens`, which is 512 here. Five runs of 20
requests spread from 12467 to 13467; the matrix row carries their median, 12582. The
overlay does not set the value on the filter — the guide's "Adapting to other hardware"
section directs readers to the matrix instead.

## Three deliberate deviations from the base

The compile runs inside the pod on first start, so the startup budget is 7215 s and
`progressDeadlineSeconds` matches it rather than tripping at the 600 s default. A rolling
update would leave a replica's old and new pod both holding a claim with no spare
accelerator, hence `Recreate`. And `--num-gpu-blocks-override` carries a measured count
rather than a profiled one, because the KV cache lives in NPU memory that vLLM's profiling
pass cannot size; that also makes `--gpu-memory-utilization` redundant, so the overlay does
not set it.

## On the container image

`ghcr.io/rbln-sw/rbln-serve` is published by Rebellions. Most accelerators here take their
image from the engine project (`docker.io/vllm/*`, `docker.io/lmsysorg/sglang`) or from
llm-d itself (`ghcr.io/llm-d/llm-d-xpu` and friends, thin layers in `docker/`), with MetaX
the one vendor-published exception so far. @Gregory-Pereira raised the policy on #2308 and
we would rather answer it up front.

`vllm-rbln` is not a fork. It is an out-of-tree platform plugin that registers on the
`vllm.platform_plugins` entry point, and the image carries unmodified vLLM 0.24.0. So the
llm-d-built path is open in principle: a `docker/Dockerfile.rbln` layered on an upstream
vLLM release, the way `docker/Dockerfile.xpu` is.

What blocks it today is that the RBLN SDK is pre-GA. `vllm-rbln`, `optimum-rbln` and
`torch-rbln` are open source and on public PyPI, but the image also carries
`rebel-compiler`, `lmcache-rbln` and `nixl-rbln`, which are still internal projects, and
every RBLN package in it is a development build rather than a release. All of them become
public releases at GA. Until then llm-d CI cannot build the image, so we are asking for the
vendor image as a stop gap and will contribute a `docker/Dockerfile.rbln` once the packages
are published. The image component carries a comment saying so, and we are happy to track
it in an issue.

## Follow-up

A `pd-disaggregation` overlay is written and has run on eight NPUs with RoCE over DRA and
NIXL KV transfer, including NUMA-aligned NPU and VF placement, though that run predates the
image and naming this PR settles and will be repeated. Keeping it out of this PR per the
maintainers' preference for separate PRs; it will follow once this lands.

A nightly e2e lane needs a cluster with an Actions Runner Controller, which we do not have
registered yet. That is tracked separately and is not part of this PR.

### DIFF
--- a/guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
+++ b/guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
@@ -76,15 +76,16 @@ spec:
             - name: VLLM_RBLN_USE_VLLM_MODEL
               value: "1"
           resources:
-            # Anonymous memory measured 51Gi both idle and under load; the rest of
-            # this limit is headroom for the page cache the model read charges to
-            # the cgroup. Keep cpu even — a node with an SMT-aligned static CPU
-            # manager rejects odd counts.
+            # cpu is the host's core count divided by its NPU count, so a pod gets the
+            # share that goes with the one NPU it claims. Keep it even — a node with an
+            # SMT-aligned static CPU manager rejects odd counts. Anonymous memory
+            # measured 51Gi both idle and under load; the rest of that limit is headroom
+            # for the page cache the model read charges to the cgroup.
             limits:
-              cpu: "10"
+              cpu: "16"
               memory: 157Gi
             requests:
-              cpu: "10"
+              cpu: "16"
               memory: 157Gi
             claims:
               - name: npu-decode-claim
--- a/guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
+++ b/guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
@@ -1,0 +1,123 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: decode
+spec:
+  replicas: 2
+  # A rolling update keeps the old and the new pod of a replica both holding a
+  # claim, so neither becomes ready where there is no spare accelerator.
+  strategy:
+    type: Recreate
+  # Matches the startupProbe budget below. Left at the 600s default, a compile
+  # still running inside its budget trips ProgressDeadlineExceeded at ten minutes
+  # and anything reading Deployment conditions calls the rollout degraded.
+  progressDeadlineSeconds: 7215
+  template:
+    spec:
+      securityContext:
+        runAsUser: 2000
+        runAsGroup: 0
+        fsGroup: 1600
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          args:
+            - "openai/gpt-oss-120b"
+            - "--served-model-name=openai/gpt-oss-120b"
+            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
+            - "--max-model-len=131072"
+            - "--block-size=4096"
+            - "--enable-chunked-prefill"
+            - "--max-num-batched-tokens=512"
+            # The runtime turns prefix caching off for sliding-window models, so leaving
+            # it on reports 0 hits over every query. Drop this once that support lands.
+            - "--no-enable-prefix-caching"
+            # Both measured on one Rebellions NPU at this data-parallel size. The block floor
+            # is ceil(max-model-len / block-size) = 32, so 34 leaves one spare.
+            - "--num-gpu-blocks-override=34"
+            - "--max-num-seqs=1"
+          env:
+            - name: HF_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: llm-d-hf-token
+                  key: HF_TOKEN
+            - name: VLLM_CACHE_ROOT
+              value: /workspace/.cache
+            - name: RBLN_COMP_DTYPE
+              value: dlfloat
+            - name: RBLN_LOCAL_IP
+              value: 127.0.0.1
+            - name: RBLN_ROOT_IP
+              value: 127.0.0.1
+            - name: RBLN_WEIGHT_FREE
+              value: "1"
+            # Unreachable in practice: the startupProbe budget below is 7215s.
+            - name: VLLM_ENGINE_READY_TIMEOUT_S
+              value: "9600"
+            - name: VLLM_RBLN_AUTO_PORT
+              value: "1"
+            - name: VLLM_RBLN_BATCH_ATTN_OPT
+              value: "1"
+            - name: VLLM_RBLN_COMPILE_STRICT_MODE
+              value: "1"
+            - name: VLLM_RBLN_METRICS
+              value: "1"
+            - name: VLLM_RBLN_MOE_REDUCE_SCATTER
+              value: "1"
+            - name: VLLM_RBLN_SAMPLER
+              value: "1"
+            - name: VLLM_RBLN_SORT_BATCH
+              value: "1"
+            - name: VLLM_RBLN_SUB_BLOCK_CACHE
+              value: "False"
+            - name: VLLM_RBLN_USE_DEVICE_TENSOR
+              value: "1"
+            - name: VLLM_RBLN_USE_VLLM_MODEL
+              value: "1"
+          resources:
+            # Anonymous memory measured 51Gi both idle and under load; the rest of
+            # this limit is headroom for the page cache the model read charges to
+            # the cgroup. Keep cpu even — a node with an SMT-aligned static CPU
+            # manager rejects odd counts.
+            limits:
+              cpu: "10"
+              memory: 157Gi
+            requests:
+              cpu: "10"
+              memory: 157Gi
+            claims:
+              - name: npu-decode-claim
+          volumeMounts:
+            - mountPath: /workspace/.cache
+              name: rbln-compile-cache
+            - mountPath: /dev/shm
+              name: shm
+          # 15 + 30x240 = 7215s, wider than the 3615s the other overlays in this
+          # guide use. A cold compile measured 314s on one Rebellions NPU and the cache
+          # is an emptyDir, so a restart pays that again — but the first start also
+          # pulls 65GiB of weights, which on its own ran past half the narrower
+          # budget. The vendor's own reference deployment allows the same 240 checks.
+          startupProbe:
+            initialDelaySeconds: 15
+            periodSeconds: 30
+            timeoutSeconds: 5
+            failureThreshold: 240
+          livenessProbe:
+            periodSeconds: 10
+            timeoutSeconds: 5
+            failureThreshold: 3
+          readinessProbe:
+            periodSeconds: 5
+            timeoutSeconds: 2
+            failureThreshold: 3
+      resourceClaims:
+        - name: npu-decode-claim
+          resourceClaimTemplateName: npu-claim-template-decode
+      volumes:
+        - name: rbln-compile-cache
+          emptyDir: {}
+        - name: shm
+          emptyDir:
+            medium: Memory
+            sizeLimit: 20Gi
--- a/guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
+++ b/guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
@@ -23,17 +23,15 @@ spec:
           command: ["vllm", "serve"]
           args:
             - "openai/gpt-oss-120b"
-            - "--served-model-name=openai/gpt-oss-120b"
             - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
-            - "--max-model-len=131072"
             - "--block-size=4096"
-            - "--enable-chunked-prefill"
             - "--max-num-batched-tokens=512"
             # The runtime turns prefix caching off for sliding-window models, so leaving
             # it on reports 0 hits over every query. Drop this once that support lands.
             - "--no-enable-prefix-caching"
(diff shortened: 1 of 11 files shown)


guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,121 @@
...
+            # Anonymous memory measured 51Gi both idle and under load; the rest of
+            # this limit is headroom for the page cache the model read charges to
+            # the cgroup. Keep cpu even — a node with an SMT-aligned static CPU
+            # manager rejects odd counts.
+            limits:
+              cpu: "10"
+              memory: 157Gi
+            requests:
+              cpu: "10"
+              memory: 157Gi
```
rebel-jinsolkim, line 89: When running the benchmark with the CPU set to 10, there was performance degradation due to throttling. How about increasing the resource allocation (around 20-30) or removing it entirely?

rebel-minhopark:
Thanks — pushed changes for two of the four, and here is the evidence for the other two.

**Prefix caching and the cache-aware chain.** The concern is right and the suggested
remedy does not work here: the runtime disables prefix caching for sliding-window models,
and `gpt-oss-120b` is one. Enabling it anyway starts fine and counts queries but returns no
hits — `vllm:prefix_cache_queries_total` 112640 against `vllm:prefix_cache_hits_total` 0,
with warm and cold TTFT within 2 ms at both 2k and 8k prompts. So the overlay keeps
`--no-enable-prefix-caching`, now with a comment saying why, and ships
`guides/optimized-baseline/router/npu.values.yaml`, which replaces the scheduling profile
with the `token-load-scorer` alone. Rebellions expects prefix-caching support for these
models in the GA runtime; the flag and that file both go away then.

**`MODEL` default.** Added a note in the accelerator docs telling readers to export
`MODEL=openai/gpt-oss-120b`. Worth noting the same gap exists for the `xpu` overlay
(`Qwen/Qwen3-0.6B`) and the `cpu` overlay (`meta-llama/Llama-3.2-3B-Instruct`), which also
pin a model the guide does not default to — happy to follow up on those separately if the
maintainers want `guide.yaml` to carry a per-accelerator default.

**Model identifier casing in the calibration matrix.** Leaving this as is. That column
carries short display names rather than HF identifiers — `Qwen3-32B`,
`Llama-3.2-3B-Instruct`, and `gpt-oss-120B` in the existing `gpu/vllm/gpt-oss` row, which
is the same model. Lowercasing only our row would make it the odd one out.

**`USER` in the image component.** Leaving this as is too. All 16 vLLM image components in
`guides/recipes/modelserver/components/images/` carry the same patch, and this image is
vLLM 0.24.0 on torch 2.11.0, inside the range the linked issue describes. `LOGNAME` is set
by two pod-level patches rather than by any image component, and this image does not need
it: both replicas compiled and served with 0 restarts without it.

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
rebel-minhopark: Hello @rebel-jinsolkim, Thank you for your review! But, here is llm-d upstream repository, please leave reviews in English. 

I will change a values as a 16 (128 cores/8 npus)

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,124 @@
...
+      securityContext:
+        runAsUser: 2000
+        runAsGroup: 0
+        fsGroup: 1600
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          args:
+            - "openai/gpt-oss-120b"
+            - "--served-model-name=openai/gpt-oss-120b"
```
Gregory-Pereira: Is there a reason to include this? This should be the default to serve it as the same model name

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,124 @@
...
+            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
+            - "--max-model-len=131072"
+            - "--block-size=4096"
+            - "--enable-chunked-prefill"
```
Gregory-Pereira: This should be a default argument in vLLM correct? If so we can remove the flag from the manifest:

```suggestion
```

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,124 @@
...
+        runAsGroup: 0
+        fsGroup: 1600
+      containers:
+        - name: modelserver
+          command: ["vllm", "serve"]
+          args:
+            - "openai/gpt-oss-120b"
+            - "--served-model-name=openai/gpt-oss-120b"
+            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
+            - "--max-model-len=131072"
```
Gregory-Pereira: Isn't this the models default max length? https://huggingface.co/openai/gpt-oss-120b/blob/main/config.json#L53. If so we can remove it:

```suggestion
```

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,124 @@
...
+              valueFrom:
+                secretKeyRef:
+                  name: llm-d-hf-token
+                  key: HF_TOKEN
+            - name: VLLM_CACHE_ROOT
+              value: /workspace/.cache
+            - name: RBLN_COMP_DTYPE
+              value: dlfloat
+            - name: RBLN_LOCAL_IP
+              value: 127.0.0.1
```
Gregory-Pereira: Are these addresses used only for communication within each model instance? Since this deployment uses TP=1 with two independent replicas, localhost seems reasonable if no cross-Pod communication is needed. Just want to verify this should not be `0.0.0.0` which is used with most vLLM serving exampls for host

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
```diff
@@ -0,0 +1,124 @@
...
+            - name: RBLN_ROOT_IP
+              value: 127.0.0.1
```
Gregory-Pereira: same as above

guides/optimized-baseline/router/npu.rbln.values.yaml
Gregory-Pereira, line 1: I think we should rename this to `npu.rbln.values.yaml` because its possible other NPU providers dont have this sliding window attention limitation at this point in time

Gregory-Pereira (commented):
All this looks great, just a few questions / comments

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
rebel-minhopark: Right — removed. vLLM defaults `served_model_name` to the model path, and with the flag gone the server resolves `served_model_name=openai/gpt-oss-120b` and `/v1/models` returns the same id.

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
rebel-minhopark: Correct, removed. `max_position_embeddings` is 131072 in the model config, and with the flag gone the engine resolves `max_seq_len=131072` and `/v1/models` reports the same.

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
rebel-minhopark: Yes, removed. V1 enables it by default and the engine still resolves `enable_chunked_prefill=True` without the flag. I had kept it because vLLM lists it under non-default args, but that only reflects the flag being passed, not the resolved value differing.

guides/optimized-baseline/modelserver/npu/vllm/patch-vllm.yaml
rebel-minhopark: Dropped both. They are the RBLN runtime's control-plane addresses for rendezvous between ranks, not the serving host, which is why `0.0.0.0` would not apply — `RBLN_ROOT_IP` is an address a rank connects *to*. The runtime already defaults both to `127.0.0.1` when unset, and at one NPU per replica `world_size` is 1 so nothing reads them, so spelling out a default only invited that reading. The pd-disaggregation overlay that follows keeps them, where four devices share a pod and the rendezvous is real.

guides/optimized-baseline/router/npu.rbln.values.yaml
rebel-minhopark: Agreed, renamed to `npu.rbln.values.yaml`. The prefix-caching gap it works around is the RBLN runtime's rather than something inherent to NPUs, so the file should not claim the generic name. Note the overlay directory and `ACCELERATOR_TYPE` stay `npu`: Rebellions' operator advertises one resource name and one DRA DeviceClass across its product families, so there is no vendor-specific accelerator type to key off there.

rebel-minhopark:
Thanks for the review @Gregory-Pereira — all five are in as `c304bd3f`, on top rather than as a rewrite so the history stays readable.

The three flags restated a default and are gone; I verified on hardware that the server resolves the same `served_model_name`, `max_seq_len` and `enable_chunked_prefill` without them. `RBLN_LOCAL_IP` and `RBLN_ROOT_IP` turned out to restate a default too, so they are gone as well rather than explained. The router values are now `npu.rbln.values.yaml`.

Also folded in a cpu fix from an earlier review (`48dbe2cd`): the pod now requests the host's core count divided by its NPU count, 16, instead of the 10 it carried, which removed a throttling asymmetry between the two replicas.

Gregory-Pereira (approved):
/lgtm

PR #2445, opened 2026-09-07
fix: recognize RUN command environment variables

git-jxj:
A Dockerfile command such as `RUN UCCL_DEVICE=rocm /tmp/build.sh` supplies the variable to the script, but the environment linter only checks stage ARG/ENV declarations and incorrectly reports it missing. Include environment assignments from the current RUN command when validating script requirements. Assignments remain local to that invocation.

Validation: `python -m pytest -q scripts/tests`: 29 passed.
Changed-file pre-commit checks passed.

PR #2482, opened 2026-09-12
fix(scripts): lint scripts on RUN continuation lines

### LINKED ISSUE #2481: [Bug :bug:]: lint-dockerfile-envvars.py never checks scripts invoked on RUN continuation lines
### What happened?

`scripts/lint-dockerfile-envvars.py` (pre-commit hook `lint-dockerfile-envvars`, run in the CI `pre-commit` job) is supposed to verify that every var in a script's `# Required environment variables:` header is declared as `ARG`/`ENV` in the Dockerfile stage that runs it.

`find_script_runs` only regex-scans the line that literally starts with `RUN`. `DockerfileParser.parse` in the same file joins `\` continuations; `find_script_runs` does not. Almost every script in `docker/Dockerfile.cuda` is invoked as

    RUN --mount=type=cache,target=/var/cache/dnf \
        /tmp/install-base-packages.sh && \
        rm -f /tmp/install-base-packages.sh

so the script path is on the second line and is never seen.

On `main` (f97c2c5b) the linter sees 2 of the 12 script invocations in `Dockerfile.cuda` (`install-runtime-packages.sh`, `package-utils.sh`). The other 10 are unchecked.

Expected: every script invoked by a `RUN`, including on continuation lines, is checked.

Once continuations are joined, the linter reports one real finding: `build-gdrcopy.sh` lists `GDRCOPY_PREFIX` as required, but the script has hardcoded `PREFIX=/usr/local` since #391 and no Dockerfile declares it. It also reports `UCCL_DEVICE` from `build-uccl.sh`, which is the Optional-block false positive already fixed in #2416.

This is the `find_script_runs` follow-up discussed in #2416.

Related: #2445 also edits `find_script_runs` (inline `VAR=x /script.sh` assignments) but still scans line by line, so it does not cover this. The two changes are independent; whichever lands second needs a trivial rebase.

### Version

main (f97c2c5b)

### Area

CI/CD

### Relevant log output

    $ python3 - <<'EOF'
    import importlib.util, pathlib
    spec = importlib.util.spec_from_file_location("l", "scripts/lint-dockerfile-envvars.py")
    m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m)
    print(sorted({r[1] for r in m.find_script_runs(pathlib.Path("docker/Dockerfile.cuda").read_text())}))
    EOF
    ['install-runtime-packages.sh', 'package-utils.sh']

malamsyah:
Fixes #2481

`find_script_runs` only scanned the line starting with `RUN`, so scripts invoked after `RUN --mount=... \` were never checked — 10 of 12 in `Dockerfile.cuda`. Follow-up noted in #2416.

- Join `\` continuations (as `DockerfileParser.parse` already does) and report each script once per `RUN`.
- Drop the `GDRCOPY_PREFIX` header line from `build-gdrcopy.sh`; the script never reads it and no Dockerfile declares it. Surfaced by the fixed check.
- Regression test in `scripts/tests/test_lint_dockerfile_envvars.py`.

`python3 -m pytest scripts/tests/`: 38 passed, 1 skipped; the new test fails on `main`.

Running the linter over all Dockerfiles still reports `UCCL_DEVICE` until #2416 lands (Optional-block false positive fixed there). Happy to rebase once it merges.

Related: #2445 also edits `find_script_runs` (inline `VAR=x /script.sh` assignments) but still scans line by line, so it does not cover this. The two changes are independent; whichever lands second needs a trivial rebase.

### DIFF
--- a/docker/scripts/cuda/builder/build-gdrcopy.sh
+++ b/docker/scripts/cuda/builder/build-gdrcopy.sh
@@ -7,7 +7,6 @@ set -Eeux
 # - USE_SCCACHE: whether to use sccache (true/false)
 # - GDRCOPY_REPO: git repo to build GDRCopy from
 # - GDRCOPY_VERSION: git ref to build GDRCopy from
-# - GDRCOPY_PREFIX: location to install GDR Copy to
 # Optional environment variables:
 # - TARGETPLATFORM: platform target (linux/arm64 or linux/amd64)
 # - TARGETOS: OS type (ubuntu or rhel)
--- a/scripts/lint-dockerfile-envvars.py
+++ b/scripts/lint-dockerfile-envvars.py
@@ -114,11 +114,16 @@ def find_script_runs(dockerfile_content: str) -> List[Tuple[str, str, int]]:
     """find all RUN commands that execute scripts, return (stage, script_path, line_num)"""
     runs = []
     current_stage = None
-    line_num = 0
+    lines = dockerfile_content.split('\n')
+    i = 0
 
-    for line in dockerfile_content.split('\n'):
-        line_num += 1
-        stripped = line.strip()
+    while i < len(lines):
+        line_num = i + 1
+        stripped = lines[i].strip()
+
+        while stripped.endswith('\\') and i + 1 < len(lines):
+            i += 1
+            stripped = stripped[:-1] + ' ' + lines[i].strip()
 
         # track stage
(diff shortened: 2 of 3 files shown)


Gregory-Pereira:
Hi @malamsyah, we are actually removing the container image builds and the linting scripts. Thank you for the contribution but I think were moving in a different direction here

Issue #2429, opened 2026-09-01
Guides don't say what the cluster needs before you deploy


BenjaminBraunDev:
Deploying the wide-ep guide from scratch on a new GKE cluster hit four failures before the first request worked, all cluster setup no doc covers.

Guides state GPU counts, but not the rest of what the cluster needs: networking objects, driver versions, controller versions, CPU node sizes. Setup docs exist (`docs/infrastructure/providers/`, `docs/infrastructure/rdma/`) but only pd-disaggregation links one.

Proposal:

1. **Required "cluster requirements" block in every guide README**: accelerator type/count, RDMA yes/no, controller versions, router CPU node sizing, provider setup doc link. Four guides already render their README from `guide.yaml` with a CI freshness check; add the block to that template and migrate the remaining guides onto it (or add the block by hand where migration isn't worth it).
2. **One preflight script in `helpers/`**, generalizing `docs/infrastructure/providers/digitalocean/verify-do-prerequisites.sh`. It checks: client tools, CRDs, LWS version, required Network objects Ready, GPU driver in a known-good range, free CPU/GPU for the guide's pods.

Also: link `docs/infrastructure/rdma/` and `multi-node.md` from guides that need them.

The four failures, for reference: missing `rdma-0..7` Network objects (pods rejected at admission), R580 driver crashing DeepEP high-throughput kernels, router pod too big for the default CPU node, decode OOM at default `--gpu-memory-utilization`.

alexeymoskalev-devops:
Hi @BenjaminBraunDev — I'd like to take this on, in two steps.

**Step 1, preflight.** A cluster-level check that runs before anything is
deployed, generalizing the DigitalOcean script
(`docs/infrastructure/providers/digitalocean/verify-do-prerequisites.sh`):
client tools, Gateway API / GAIE / LWS CRDs present and at the versions
`guides/env.sh` expects, the guide's required `Network` objects (RDMA guides),
GPU driver within a known-good range (from GPU feature discovery node labels),
and enough schedulable CPU/GPU for the guide's pods, including a CPU node large
enough for the router. Of your four failures that covers three; decode OOM at
the default `--gpu-memory-utilization` is a runtime setting and belongs in the
requirements block instead. This is complementary to the in-pod preflight in
llm-d-pd-utils, which gates `vllm serve` after scheduling — the failures here
happen before a pod runs.

Shape: a bash script under `helpers/preflight/` that reads a new
[...]

PR #2484, opened 2026-09-12
docs(wide-ep-lws): link networking prerequisites

k21993:
Part of #2429.

The wide-ep-lws guide requires All-to-All RDMA (InfiniBand or RoCE) on the NVIDIA DeepEP path and full-mesh pod connectivity on the Intel XPU path, but it did not point at the networking setup docs, so a reader deploying on a fresh cluster had no path from "this needs RDMA" to how to provision it.

Rebased onto main now that #2379 has merged. That PR added a prerequisite for RDMA-capable nodes with GKE-specific links, so this adds the provider-neutral docs to that bullet instead of adding a second one:

- `docs/infrastructure/rdma/README.md` for the networking stack (NIXL/UCX, InfiniBand/RoCE) and verification
- `docs/infrastructure/multi-node.md` for cross-node setup

Both are also linked from the RDMA note above the Prerequisites section.

Docs only, one file. Checked that the links resolve.


k21993:
Rebased onto main after #2379 merged. The RDMA-capable nodes bullet it added now carries the provider-neutral networking links instead of this PR adding a second bullet. Still one file, +8/-1.

PR #2489, opened 2026-09-14
docs(guides): add GKE Pod Snapshots single-GPU user guide

RyanRosario:
A new guide, guides/gke-pod-snapshots, showing how to use [GKE Pod Snapshots](https://cloud.google.com/kubernetes-engine/docs/how-to/pod-snapshots) with GKE Sandbox (gVisor) to checkpoint and restore a single-GPU vLLM model server, using Qwen/Qwen3-32B on an H100 as the worked example.

The problem it targets: a large model server spends minutes on every cold start downloading weights and initialising the engine. Pod Snapshots let GKE capture the running container once and restore subsequent pods from that image instead of repeating the work.


RyanRosario:
On shipping the snapshot runtime as a ConfigMap

This part is deliberately provisional and I'd like reviewer input before it hardens.

The guide runs `python3 -m docker.scripts.snapshot.launcher `inside the stock `vllm/vllm-openai` image, so the Python package has to reach the container somehow. Today that is a pair of generated `ConfigMap`s.

Mechanics, briefly:

Generation lives in `docker/scripts/snapshot/kustomization.yaml` rather than in the guide overlay, because Kustomize will not let `configMapGenerator.files:` read outside its own root, but will let a `resources:` entry point at a directory that has its own kustomization.

There is one generator per Python package level (`snapshot-scripts`, `snapshot-scripts-vllm`) so the keys stay plain basenames. The consumer mounts the second nested inside the first.

The generated content hash rolls the pods when the sources change, which is the one genuinely nice property of this approach.

Known downsides: it ships runtime code as configuration, the two-level nested mount is subtle, and adding a module means remembering to touch the generator in a different tree.

Alternatives:

1. Keep it, and document it properly.
2. Bake the snapshot runtime into a purpose-built image. This deletes the generator, both mounts and the `PYTHONPATH` handling — at the cost of an image to build, publish and version alongside the guide.

I lean toward (2) longer term, but it is a bigger change than a guide PR should carry, so I have left the `ConfigMap` approach in place here and marked it as provisional in the file. **The implemented approach does not use a script to populate `ConfigMap` yet.**

Basically, it feels like the `ConfigMap` method doesn't work cleanly here. The question is how we want to proceed in such a way that is clean to the user and also matches the repo's conventions.

RyanRosario:
Testing status: incomplete, keeping this as a draft

Flagging clearly that this has not been validated end to end.

Validated so far:

scripts/guide.py check and scripts/guide.py render --check pass.
kubectl kustomize renders the overlay cleanly.
The model server deploys, reaches Ready, and the snapshot runtime is loaded and hooked into vLLM's application lifespan.
engine.sleep(level=1) releases GPU memory as intended.
Not yet validated:

The checkpoint itself. In my environment the gVisor checkpoint fails and no snapshot object is written to GCS. That means the restore path, and the cold-start improvement this guide exists to demonstrate, are both currently unverified.
Current evidence points at something environment- or platform-side rather than guide configuration — it reproduces with a minimal pod that has no GPU and no vLLM — but I am still confirming that and do not want to assert it here until I can back it up.

RyanRosario:
**Update:** This is a permissions issue and needs to be added to the user guide. Will be updated shortly.

RyanRosario:
**Update: checkpoint and restore now verified end to end. Root cause of the earlier failure was an
IAM gap in this guide.**

## Root cause

`podsnapshotstorageconfig.yaml` sets `tokenSource: "podKSA"`. Under that setting the gVisor sandbox
writes snapshot objects to GCS as the **KSA principal itself** via Workload Identity Federation.
## Measured results

> [!NOTE]
> These numbers are from a **single-A100 / `Qwen3-8B`** run, not the H100 / `Qwen3-32B` config the
> guide targets — I don't have H100 quota. They validate the mechanism; treat the absolute figures
> as indicative only. I have deliberately kept them out of the guide text.

| | |
|---|---|
| Snapshot size | 18.44 GiB (`pages.img` 18.4 GiB, `checkpoint.img` 9.8 MB, `pages_meta.img` 852 KB) |
| Upload time | 4m 1s @ ~82 MB/s |
| Pod deleted → `Ready` | 31s |
| Engine wake-up | 1.89s for tags `{'kv_cache', 'weights'}` |
| Weight re-downloads on restore | none — zero safetensors loads in the restored pod's logs |

Inference against the restored pod returns valid completions.

Two things worth noting from the data:

1. **Upload is GCS-bandwidth-bound, not CPU or disk.** gVisor logged `queue full for 3m59.6s` out of
   the 4m1.3s total, i.e. it spent essentially the whole checkpoint waiting on GCS.
2. **The snapshot is roughly weights-sized even after the on-disk cache purge.** `engine.sleep(level=1)`
   frees 27.54 GiB of GPU memory, of which 15.43 GiB is backed up into host RAM and 12.12 GiB
   discarded; the host-RAM copy is what gets serialized. Purging `MODEL_CACHE_DIR` prevents the
   weights being captured *twice*, it doesn't make the snapshot small. Scaling that to ~64 GB of
   `Qwen3-32B` weights at the observed throughput is where the new 2400s timeout comes from — that
   figure is extrapolation, not measurement.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+            # purge". expanduser resolves this to /root/.cache/huggingface/hub,
+            # which is the container's writable layer -- not the /.cache volume
+            # mounted below, which vLLM uses for its compile cache.
+            - name: MODEL_CACHE_DIR
+              value: "~/.cache/huggingface/hub"
+            - name: HF_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: llm-d-hf-token
+                  key: HF_TOKEN
```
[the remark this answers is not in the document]
RyanRosario: To be reviewed.

RyanRosario:
**Known limitation in current guide draft: the rollout strategy currently assumes a single replica**

`patch-vllm.yaml` sets `strategy: Recreate`. That was necessary because at `replicas: 1` on a pool with exactly one GPU, the default `RollingUpdate` tries to schedule the replacement pod before the old one releases its GPU, so every rollout deadlocks. `Recreate` resolves the deadlock but bakes in a single-replica assumption — at N replicas it tears
down all N at once.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/kustomization.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - podsnapshotstorageconfig.yaml
+  - podsnapshotpolicy.yaml
+  # Generates the snapshot-scripts ConfigMaps from the runtime sources.
+  - ../../../../../../docker/scripts/snapshot
+
+namePrefix: gke-pod-snapshots-nvidia-gpu-vllm-
+
+components:
+  - ../../../../../recipes/modelserver/components/images/gpu-vllm
+  - ../../../../../recipes/modelserver/components/disable-gke-nccl-tuner-patch
```
nicolexin: Do you have any idea why are referring to this script? I suppose it is only used for RDMA networking between multi-host deployments and we are not testing that one here.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
+# The PodSnapshotStorageConfig uses tokenSource "podKSA", so the gVisor sandbox
+# authenticates to GCS as this KSA principal directly via Workload Identity
+# Federation. No `iam.gke.io/gcp-service-account` annotation is needed -- the
+# bucket IAM binding targets the `principal://.../sa/<ksa>` subject instead.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sa
```
nicolexin: Why do we have this block here? I thought the recipe already comes with the exact same service account: https://github.com/llm-d/llm-d/blob/a130c6249152de47d5e2447501d8f2d3ee06c7ae/guides/recipes/modelserver/common/sa.yaml

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+kind: Deployment
+metadata:
+  name: decode
+spec:
+  replicas: 1
+  # At replicas=1 on a single-GPU pool, the default RollingUpdate wants to
+  # schedule the new Pod before releasing the old one's GPU, so every rollout
+  # deadlocks. Recreate frees the GPU first.
+  strategy:
+    type: Recreate
```
nicolexin, line 16: I understand your comment here (for testing purposes), but we should not leave this config for production use cases here. Did any of the test step require this particular configuration to work? Can we work around the test steps instead of updating the config here?

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+  template:
+    metadata:
+      annotations:
+        # Opts into GKE Sandbox's fast-startup path, matching the reference Pod.
```
nicolexin: Do we need this annotation for anything? I am not sure what the reference pod means here.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+        # Note: /proc/gvisor/checkpoint was writable without it, so this is not
+        # what exposes the file. Kept for parity.
+        cloud.google.com/gke-fast-pod-startup: "true"
+    spec:
+      runtimeClassName: gvisor
+      # recipes/modelserver/common/sa.yaml disables token automounting, which is
+      # right for model servers that never call a GCP API. This guide does: the
+      # PodSnapshotStorageConfig uses tokenSource "podKSA", so GKE reaches GCS as
+      # this Pod's KSA and needs its projected token.
+      automountServiceAccountToken: true
```
nicolexin: I had no trouble uploading to GCS without this flag, is this some new requirements from pod snapshot feature?

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+        - name: modelserver
+          command: ["python3", "-m", "docker.scripts.snapshot.launcher"]
+          args:
+            - "--model"
+            - "Qwen/Qwen3-32B"
+            - "--safetensors-load-strategy"
+            - "eager"
+            # Required for engine.sleep(level=1) to reclaim VRAM at all: vLLM
+            # only routes weights through CuMemAllocator when sleep mode is on at
+            # engine construction. Without it, sleep succeeds but frees nothing.
```
nicolexin: The sleep only frees the KV cache and nothing else. The model weights are offloaded as a "side effect" for us. I'd say this is nice to have but not really mandatory.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+            - "--enable-sleep-mode"
+            - "--gpu-memory-utilization"
+            - "0.90"
+            # Disables CUDA graph capture, matching the reference Pod. Note:
+            # graphs were tested and are NOT the cause of checkpoint failures.
+            # Costs decode latency, so re-evaluate once snapshots succeed.
+            - "--enforce-eager"
+            - "--port"
+            - "8000"
+            - "--disable-access-log-for-endpoints=/health,/metrics,/v1/models"
```
nicolexin, line 55: Do we still need this flag?

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+            - "Qwen/Qwen3-32B"
+            - "--safetensors-load-strategy"
+            - "eager"
+            # Required for engine.sleep(level=1) to reclaim VRAM at all: vLLM
+            # only routes weights through CuMemAllocator when sleep mode is on at
+            # engine construction. Without it, sleep succeeds but frees nothing.
+            - "--enable-sleep-mode"
+            - "--gpu-memory-utilization"
+            - "0.90"
+            # Disables CUDA graph capture, matching the reference Pod. Note:
```
nicolexin: Why do you disabled CUDA graph capture in here? This is only needed now for multiGPU workloads I believe.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+            # the V1 path, so it has no effect today. Kept for parity.
+            - name: VLLM_ENABLE_V1_MULTIPROCESSING
+              value: "0"
+            # Without this, vLLM's stdout is block-buffered and a slow cold start
+            # is indistinguishable from a hang in kubectl logs.
+            - name: PYTHONUNBUFFERED
+              value: "1"
+            - name: SNAPSHOT_PROVIDER
+              value: "gke_gvisor"
+            # Opts into the pre-checkpoint disk purge; unset means "do not
```
nicolexin: We should make it clear that model weights are duplicated: one copy in GPU, another copy is on disk. Unset this when model weights are mounted instead of downloaded as a local copy.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+          env:
+            # Snapshot configuration
+            - name: PYTHONPATH
+              value: "/workspace"
+            # Asks vLLM to keep the API server and engine core in one process so
+            # cuda-checkpoint need not cross a process boundary. Note: vLLM 0.25
+            # recognises this but still forks an "(EngineCore pid=NN)" child on
+            # the V1 path, so it has no effect today. Kept for parity.
+            - name: VLLM_ENABLE_V1_MULTIPROCESSING
+              value: "0"
```
nicolexin: Just so I understand this, vllm 0.25 onward see this flag but will disregard it? Does checkpoint/restore fail without this flag today?

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
```diff
@@ -0,0 +1,168 @@
...
+            # which is the container's writable layer -- not the /.cache volume
+            # mounted below, which vLLM uses for its compile cache.
+            - name: MODEL_CACHE_DIR
+              value: "~/.cache/huggingface/hub"
+            - name: HF_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: llm-d-hf-token
+                  key: HF_TOKEN
+            # gVisor and NCCL compatibility settings, matching the reference Pod
```
nicolexin: Can we cleanup the comment a bit (e.g. "reference pod") and I believe the only flag you need is VLLM_HOST_IP but please verify before deleting the rest.

guides/gke-pod-snapshots/README.md
```diff
@@ -0,0 +1,399 @@
...
+> [!IMPORTANT]
+> **The snapshot is roughly the size of the model, not a few gigabytes.** `engine.sleep(level=1)` frees *device* memory by offloading weights to host RAM; it does not discard them. A whole-pod snapshot captures that host RAM, so the GCS artifact scales with the model's memory footprint rather than with the purged disk cache. Size the node's memory to hold the offloaded weights alongside the running process, and budget GCS storage and egress accordingly.
+
+---
+
+## Single-GPU Scope & Technical Caveats
+
+1. **Single-GPU / Single-Rank Scope:**
+   The snapshot launcher is designed for single-GPU / single-pod deployments (`nvidia.com/gpu: 1`). Multi-node / multi-rank distributed barrier coordination across pods during checkpointing is not supported.
+2. **Sleep Mode Must Be Enabled (`--enable-sleep-mode`):**
```
nicolexin: See my previous comment, I think this is nice to have but not required.

guides/gke-pod-snapshots/README.md
```diff
@@ -0,0 +1,399 @@
...
+
+```text
+(APIServer pid=1) [vllm.snapshot.wrapper] INFO: Process restored from snapshot checkpoint. Resuming engine...
+(APIServer pid=1) [vllm.snapshot.wrapper] INFO: Executing engine.wake_up() to restore VRAM...
+(EngineCore pid=NN) INFO: It took N seconds to wake up tags {'kv_cache', 'weights'}.
+(APIServer pid=1) INFO: Application startup complete.
+```
+
+> [!NOTE]
+> The `Process restored from snapshot checkpoint` line is emitted unconditionally by the wrapper, including when the checkpoint failed, so it is not on its own evidence of a successful restore. The `wake up tags` line and the absence of weight-download output are the signals that the pod came back from a snapshot rather than cold-starting.
```
nicolexin: First of all, we should look into if there is a way to make `Process restored from snapshot checkpoint` reflect successful restoration. If we cannot, I think the user guide should provide other ways a validating a successful restoration, for example the k8s events should show pod x is successfully restored from snapshot.

guides/gke-pod-snapshots/README.md
```diff
@@ -0,0 +1,399 @@
...
+| `MASTER_ADDR` / `MASTER_PORT` | `127.0.0.1` / `29500` | Torch distributed rendezvous address for the single rank. |
+| `TORCH_NCCL_ENABLE_MONITORING` | `0` | Asks PyTorch to skip its NCCL watchdog threads, which could otherwise fire while the sandbox is frozen. |
+| `TORCH_NCCL_ASYNC_ERROR_HANDLING` | `0` | Disables asynchronous error handling threads during checkpointing. |
+| `NCCL_TUNER_PLUGIN` / `NCCL_NET_PLUGIN` | `none` / `""` | Disables GKE's injected NCCL plugins, applied by the shared `disable-gke-nccl-tuner-patch` component as in the other GKE overlays. |
+
+> [!NOTE]
+> Several of these are inert in this single-GPU configuration and are kept only for parity with the known-good reference pod. At `world_size=1` vLLM selects its own rendezvous port rather than honoring `MASTER_PORT`, the `NCCL_P2P`/`SHM_DISABLE` flags gate transports a single rank never establishes, and the PyTorch NCCL watchdog threads were observed to start regardless of `TORCH_NCCL_ENABLE_MONITORING`. Revisit once a snapshot completes successfully and there is a passing baseline to compare against.
```
nicolexin: The above list needs to be revisited again, there should be nothing related to NCCL for single GPU case. If the list is small it might make sense to merge into the caveat.
Also I think the last paragraph here could be confusing to the user, let's remove it if it does not add any value.

guides/gke-pod-snapshots/README.md
```diff
@@ -0,0 +1,399 @@
...
+   gcloud container node-pools create "<NODE_POOL_NAME>" \
+     --cluster="<CLUSTER_NAME>" \
+     --region="<REGION>" \
+     --node-locations="<ZONE>" \
+     --machine-type="<GPU_MACHINE_TYPE>" \
+     --disk-size="<DISK_SIZE>" \
+     --image-type=cos_containerd \
+     --workload-metadata=GKE_METADATA \
+     --sandbox type=gvisor \
+     --accelerator=type=<GPU_ACCELERATOR>,count=1,gpu-driver-version=latest \
```
nicolexin, line 106: Let's make the count also configurable here, and enable auto scaling. The workload itself has one replica using one GPU to begin with, but we are actually targeting an autoscaling story here. Snapshot once, and scale out fast.

guides/gke-pod-snapshots/README.md
```diff
@@ -0,0 +1,399 @@
...
+<!-- guide:verify.tests.snapshot end -->
+
+Status progression:
+
+1. `AwaitingCheckpoint`: GKE signaled gVisor to freeze the container runtime.
+2. `AllSnapshotsAvailable`: Snapshot files have been uploaded to GCS and are ready for restoration.
+
+### 2. Test Pod Restoration & Verify Restore Logs
+
+Delete the running pod so the Deployment schedules a replacement replica that restores directly from the snapshot:
```
nicolexin, line 317: Can we tell a scale out story instead? Technically it's the same since we are trying to show snapshot/restore works, but ultimately, we should show the user that the scale out is fast.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: Completely agree. I did run into this deadlock, but this is one of those parts that resulted from my debugging. This needs to be removed anyway and replaced with a rollout for scaling.

guides/gke-pod-snapshots/README.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+1. **First Pod (Cold Start & Snapshot Creation):** The initial replica loads the model weights (~64 GB for `Qwen/Qwen3-32B`), calls `engine.sleep(level=1)` to release physical GPU memory, purges the on-disk weight cache so the same bytes are not also captured from the container filesystem, and triggers a GKE Pod Snapshot uploaded to Google Cloud Storage (GCS).
+2. **Subsequent Pods (Fast Restoration):** When pods scale out or restart, GKE automatically restores the container from the GCS snapshot in seconds—bypassing model weight downloads and engine initialization.
+3. **Serve Traffic:** The restored pod immediately wakes up GPU memory and begins serving inference requests.
+
+> [!IMPORTANT]
+> **The snapshot is roughly the size of the model, not a few gigabytes.** `engine.sleep(level=1)` frees *device* memory by offloading weights to host RAM; it does not discard them. A whole-pod snapshot captures that host RAM, so the GCS artifact scales with the model's memory footprint rather than with the purged disk cache. Size the node's memory to hold the offloaded weights alongside the running process, and budget GCS storage and egress accordingly.
+
+---
```
nicolexin, line 15: Just a TODO here: once we get H100 working, we should run some basic benchmarking on improvements of model server startup time and post the result here (like the one in PR comments).

And one nit to the PR comment, we should measure restoration time from container creation to pod restore (not from pod deletion) since we are targeting a scale out use case and we don't want node provisioning time to pollute the actual gain from pod snapshot.

guides/gke-pod-snapshots/README.md
RyanRosario: Yes. This will be in my next update. Will resolve once addressed.

guides/gke-pod-snapshots/README.md
RyanRosario: Yes great. See the following comment.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/kustomization.yaml
RyanRosario: This can safely be removed. This was convention-following rather than a deliberate choice. Every `gke/` overlay in the repo includes this component so it got copied in with the rest of the GKE overlay scaffolding.

It isn't needed here. The component's `README` scopes it to tensor parallelism on clusters with the `gIB NCCL RDMA` libraries installed as you said. This guide is single-GPU (nvidia.com/gpu: 1, TP defaults to 1), so vLLM never creates an NCCL communicator and the tuner plugin is never loaded. 

I've removed it and re-verified end to end — full cold start, checkpoint, and restore all succeed, with zero NCCL-related lines in the model server logs.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: Removed. The base already provides it via recipes/modelserver/common/sa.yaml, and since patch-vllm.yaml is applied under patches: rather than resources:, that block was a no-op strategic merge that only restated the name the SA already had. It's a leftover from before the guide moved to tokenSource: podKSA, when it carried the iam.gke.io/gcp-service-account annotation.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: I will leave this unresolved until a future commit for scale out.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: Reference pod is the configuration I developed a while back in `pod.yaml`. It's just what the new configuration was based on. The fast-restore path is driven entirely by the `runtimeClassName: gvisor` spec and the target `PodSnapshotPolicy`

I've cleaned up this annotation and its comment in the latest commit to prevent user confusion.

guides/gke-pod-snapshots/README.md
RyanRosario: I agree. I only included it because it was in our test deployments and didn't want to risk breaking something by removing them. While they could be useful, we don't need them. Removed.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: No, when trying to resolve a permissions issue I had switched to experiment with a different method that did require this parameter. This is removed.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: Yes, but it's not specific to pod snapshots. If we remove it, this implementation is the only one that logs access to these endpoints. Basically, this is a convention used throughout the repo.

The recipe ships `args: []`, so every guide spells out the full list — this flag is part of the standard set ~50 other sites use.

It also keeps probe traffic (`/v1/models` every 5s, `/health` every 10s) out of the logs, which matters here since the guide has users read those logs to watch the checkpoint and restore. I'll resolve once I push the updates and we can reopen if need be.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: The online server never reads it because `async_llm.py` has no references to it. Under `vllm serve` it's inert.

Checkpoint/restore doesn't need it. Tonight's H100 run had it set to 0 and vLLM forked `EngineCore` anyway — that pod still checkpointed and restored in 31s. Removing the env var and the comment.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: I tested without it and hit an OOM. For a model this large, CUDA graphs need more memory than we have left at 0.95 — capture takes ~1.22 GiB and the sampler's warmup allocation then fails. For a smaller model it would make sense to trade the space for the performance.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: Added both points — the comment now explains that the weights exist twice at checkpoint time and that both copies would otherwise land in the snapshot, plus the guidance to leave it unset for mounted weights since there's no duplicate to drop and the purge would delete shared data.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: Done, and you are correct -- I tested it when removing the NCCL parameters.

guides/gke-pod-snapshots/README.md
RyanRosario: Agreed — reworded from "Must Be Enabled" to describe the consequence: without it sleep() frees nothing, so the checkpoint captures the GPU fully resident and the snapshot is larger.

guides/gke-pod-snapshots/README.md
RyanRosario: Went with the event since it comes from GKE and names the exact snapshot. Added it to the restore verification step:

```
kubectl get events -n ${NAMESPACE} --field-selector reason=GKEPodSnapshotting
Normal  GKEPodSnapshotting  pod/...-687fcc865f-dzkhn  Successfully restored the pod from PodSnapshot llm-d-gke-pod-snapshots/b96c4058-...
```

guides/gke-pod-snapshots/README.md
RyanRosario: Benchmarks added and measured from container creation to pod restore.

guides/gke-pod-snapshots/guide.yaml
```diff
@@ -0,0 +1,128 @@
...
+    # checkpoint has been written and the process resumed, so this wait covers:
+    # weight download, model load, engine.sleep(), and the full snapshot upload
+    # to GCS. The upload is bandwidth-bound and the snapshot is roughly
+    # weights-sized, because engine.sleep(level=1) backs the weights up into host
+    # memory, which is then saved into pages.img. At an observed ~80 MB/s to GCS,
+    # budget several minutes per 10 GiB of model weights on top of the download.
+    snapshot:
+      - run: |
+          kubectl wait --for=condition=ready pod -l llm-d.ai/guide=${GUIDE_NAME} -n ${NAMESPACE} --timeout=2400s
+          kubectl get podsnapshots -n ${NAMESPACE}
```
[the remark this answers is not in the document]
RyanRosario: this was a real gap, and it's exactly the failure I hit earlier in this PR: an IAM misconfiguration killed the checkpoint, the wrapper caught the exception and woke the engine, the pod went Ready, and the step printed an unusable `PodSnapshot` and passed. Added an explicit assertion.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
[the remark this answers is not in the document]
RyanRosario: Added a warning that the token makes the round trip to GCS and back, but since bucket's already require authentication and authorization, a robust fix is out of scope for a user guide. For ungated models, the token is optional anyway.

guides/gke-pod-snapshots/guide.yaml
```diff
@@ -0,0 +1,128 @@
...
+          kubectl logs -l llm-d.ai/guide=${GUIDE_NAME} -n ${NAMESPACE} --tail=20
+
+    request:
+      - run: |
+          kubectl run curl-test --rm -i --restart=Never \
+            --image=${CURL_TEST_IMAGE} \
+            --namespace="${NAMESPACE}" \
+            --env="IP=${IP}" \
+            --env="MODEL=${MODEL}" \
+            -- /bin/sh -c 'curl -sS -X POST "http://${IP}/v1/completions" -H "Content-Type: application/json" -d "{\"model\": \"${MODEL}\", \"prompt\": \"How are you today?\"}"'
```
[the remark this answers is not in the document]
RyanRosario: Good catch.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: Reworded so it no longer claims sleep mode is required. Kept the flag and documented the measured effect on Qwen3-32B: `--enable-sleep-mode` flips `enable_cumem_allocator`, which lets `sleep(level=1)` move 61.68 GiB of weights to host RAM and discard 12.52 GiB of KV cache so the KV cache isn't captured in the snapshot.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
nicolexin: Can we revisit this one? I was able to run H100 snapshot restore with graph capture and no issues. We may need to update the gpu_utilization, but without graph capture, you pay extra performance overhead every time you generate a token.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/kustomization.yaml
nicolexin: Oh I didn't know every other gke guide has it. In that case we should add it back, sorry about that!

RyanRosario: The reason the other guides have it though is because they are working with multi-GPU. My preference would be leave it out because it adds a lot of unnecessary complexity and is confusing. Even if other guides have it, I don't think it's a strong enough reason to keep it here. Let me know your thoughts.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: Will do. Any changes will be included in the next commit that includes the scale out story. I admit this isn't a great reason to avoid using CUDA graphs, but it adds some complexity to the user guide because I feel that we need to provide guidance on how to choose the gpu_utilization parameter.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,96 @@
...
+GKE `v1.36.4-gke.1082000`, gVisor sandbox, Spot provisioning, 300 GB boot disk;
+`Qwen/Qwen3-32B` on vLLM `v0.26.0` at `--gpu-memory-utilization 0.95` and `--enforce-eager`.
+
+> [!NOTE]
+> The servers run with `--gpu-memory-utilization=0.95` because the default `0.90` leaves too little KV
+> headroom for Qwen3-32B's 40960-token context on an 80&nbsp;GB H100 — at `0.90` the engine cannot start
+> at all. If you swap models, check the `Available KV cache memory` line during startup before
+> assuming the shipped value fits.
+
+The single-GPU A3 shapes (`a3-highgpu-1g`, `-2g`, `-4g`) are only offered as Spot or Flex-start VMs —
```
nicolexin: nit: 2g and 4g is not single GPU. But it seems like all H100 machine type should work here right? I am not sure if we need this paragraph.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,96 @@
...
+> [`llmdbenchmark`](https://github.com/llm-d/llm-d-benchmark) / `inference-perf`. Steady-state serving
+> performance after a restore is unchanged from a normal vLLM deployment — the snapshot restores the
+> same process, not a degraded one.
+
+**Reference configuration:** `a3-highgpu-1g` (1× NVIDIA H100 80GB HBM3, driver `580.126.20`),
+GKE `v1.36.4-gke.1082000`, gVisor sandbox, Spot provisioning, 300 GB boot disk;
+`Qwen/Qwen3-32B` on vLLM `v0.26.0` at `--gpu-memory-utilization 0.95` and `--enforce-eager`.
+
+> [!NOTE]
+> The servers run with `--gpu-memory-utilization=0.95` because the default `0.90` leaves too little KV
```
nicolexin: Can we leave the gpu-memory-utilization at 0.90 so that we can capture the CUDA graph? We can tweak the --max-model-len instead to something like 8192.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,96 @@
...
+
+Treat these as one measured data point, not a specification. Phase times are measured from
+**container start** (the first vLLM log line), excluding image pull, scheduling, and node
+provisioning, so they are comparable across runs on warm and cold nodes and isolate the snapshot's
+own contribution. Download speed, model size, and GPU will move the figures.
+
+| Metric | Without snapshots | Restore from snapshot |
+| :------------------------------- | ----------------: | --------------------: |
+| Container start → serving engine  | 2m 56s            | **26.0s**             |
+| Weight loads from disk     | 1                 | **0**                 |
```
nicolexin: what does 1 and 0 mean here? Is it true and false or some time measurement?

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,96 @@
...
+| Speedup                          | —                 | **~6.8×**             |
+
+> [!IMPORTANT]
+> The first pod additionally pays a **one-time 16m 41s** to create the snapshot, most of it the frozen
```
nicolexin: This is way too long, something must be off. It took me 75 seconds for checkpointing Qwen3-32B on H100. Either the duplicate weights are not deleted, or the bucket settings are not right (e.g. is your bucket and cluster in the same region?).

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,96 @@
...
+
+| Phase | Measured |
+| :--- | ---: |
+| Image pull (8.91 GB) | 69s |
+| Model loading — download + load into VRAM (61.03 GiB) | 123s |
+| &nbsp;&nbsp;&nbsp;&nbsp;↳ of which, safetensors read into VRAM | 41s |
+| Engine init (profile, KV cache, warmup) | 19.9s |
+| **Container start → checkpoint triggered** | **3m 16s** |
+| `engine.sleep(level=1)` freed | 76.43 GiB (0.94 GiB still in use) |
+| Time to fall asleep | 16.4s |
```
nicolexin: Where is this number coming from? I thought the sleep usually take subseconds.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,96 @@
...
+### Cold start — first pod, creates the snapshot
+
+| Phase | Measured |
+| :--- | ---: |
+| Image pull (8.91 GB) | 69s |
+| Model loading — download + load into VRAM (61.03 GiB) | 123s |
+| &nbsp;&nbsp;&nbsp;&nbsp;↳ of which, safetensors read into VRAM | 41s |
+| Engine init (profile, KV cache, warmup) | 19.9s |
+| **Container start → checkpoint triggered** | **3m 16s** |
+| `engine.sleep(level=1)` freed | 76.43 GiB (0.94 GiB still in use) |
```
nicolexin: I think we should state how much kv cache is cleared. The 76.43 includes kv cache and model weights. Model weights are offloaded to CPU but they are still captured.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,96 @@
...
+| Snapshot size in GCS | **64.73 GiB** |
+
+The checkpoint upload dominates: roughly 83% of the cold start is spent writing pages to GCS, during which
+the pod is frozen and emits no logs.
+
+### Restore — every subsequent pod
+
+| Phase | Measured |
+| :--- | ---: |
+| Pod created → scheduled | 2s |
```
nicolexin: I assume this data is based on pod deletion?
We should ignore node provisioning time so probably no need to report pod created -> scheduled time (or the next line) since it also ties to availability. Or we could replace with "node provisioning time: varies"

But one thing I do think is missing is image pulling. If it is a new node, chances are it will take around 5 minutes to pull vllm. I am a little skeptical about the 69 seconds above.
We should enable image streaming to reduce the image pulling time.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,96 @@
...
+| Scheduled → container started | 4s |
+| **Container started → pod `Ready`** | **26.0s** |
```
nicolexin: This should be the main focus here. I would measure from container creation/start to pod being restored, and another separate item from pod being restored to pod ready. We should also define what pod 'Ready' means.

Also is the pod ready include the engine.wakeup()?

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,96 @@
...
+| `engine.wake_up()` | 2.68s |
+| Safetensors loads performed | **none** |
```
nicolexin: nit: we can remove this. It's not in the order anyways.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
nicolexin: Can we leave the gpu-memory-utilization at 0.90 so that we can capture the CUDA graph? We can tweak the --max-model-len instead to something like 8192.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Yes. Removed. The user can read about any limitations (spot) in the general GKE documentation anyway.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: 0.90 restored and `max-model-len` added.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: D'oh. That should be 1=Yes, 0=No. Fixed.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: This was resolved by modifying a bucket permission. Without the permission, the transfer fell back to serial rather than parallel. This setting is also in the user guide in case it is useful to users.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: This was added after the most recent run of the test pipeline. 

The numbers in my previous reply above are from a run before I set
`--max-model-len 8192`. The doc has the current ones:
- KV cache discarded (**not** captured in the snapshot): **8.52 GiB**
- Weights offloaded to host RAM (**is** captured): **61.68 GiB**
- `CuMemAllocator: sleep freed 70.20 GiB memory in total, of which 61.68 GiB
  is backed up in CPU and the rest 8.52 GiB is discarded directly`
- `Sleep mode freed 72.92 GiB memory, 2.9 GiB memory is still in use`
  (1.86 GiB of the residual is CUDA graphs)
Weights are unchanged at 61.68 GiB; only the KV cache moved, from 12.52 GiB
down to 8.52 GiB with the smaller context length. And to your point: yes, the
weights are still captured in the snapshot — sleep only keeps the KV cache
out of it.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Removed!

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: For some reason this comment is posted twice. 0.90 restored and max-model-len added.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Running level=1 with --enable-sleep-mode omitted, sleep() returns in 0.021s, because CuMemAllocator has nothing registered to release. That's almost certainly the subsecond timing you saw.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Indeed image streaming was not enabled. Image pulling added and removed pod creation time and replaced node provisioning with "varies."

Also, you were right to be skeptical of the 69s back in the earlier thread — it
came from a different run and I've removed it.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Done —I've updated the table to break out Container start → process restored (StartedAt → PodRestored condition) and Process restored → pod Ready separately for both new-node and warm-node restores, and explicitly defined both metrics right below the table

guides/gke-pod-snapshots/README.md
RyanRosario: Added GPU_COUNT and MAX_NODES as configuration variables. Updated the cluster and node pool creation steps to incorporate scale out.

guides/gke-pod-snapshots/README.md
RyanRosario: Added scale out verification explanation.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: Removed `strategy: type: Recreate`—this is no longer needed.

**Why it was there originally vs. now:**
- **Originally:** The node pool had a fixed `--num-nodes=1` with no autoscaling, so Kubernetes' default `RollingUpdate` couldn't schedule a surge pod on a 1-node pool during rollouts and caused a deadlock.
- **Now:** Because we enabled node pool autoscaling (`--enable-autoscaling --min-nodes=1 --max-nodes=<MAX_NODES>`) and updated the verification step to test horizontal scale-out (`kubectl scale deployment ... --replicas=2`), Kubernetes' default `RollingUpdate` works out of the box and is much better for production.

guides/gke-pod-snapshots/modelserver/gpu/vllm/gke/patch-vllm.yaml
RyanRosario: This was fixed.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: UPDAET: I did an experiment here.

The copy alone takes 2.3s, so sub-second sleep isn't achievable for `level=1` and 61 GiB. gVisor does not add overhead (and was faster in this run, cause not yet identified); the microbenchmark accounts for about 11.5s of the 17.4s, with the remainder attributed to vLLM's per-tensor sleep path.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Fixed in 2970b8d6.

RyanRosario:
@nicolexin All feedback has been addressed!

guides/gke-pod-snapshots/guide.yaml
```diff
@@ -0,0 +1,135 @@
...
+          kubectl get podsnapshots -n ${NAMESPACE}
+
+    # Scaling out after PodSnapshot is Ready schedules a new replica that GKE
+    # restores directly from GCS.
+    restore:
+      - run: |
+          kubectl scale deployment -l llm-d.ai/guide=${GUIDE_NAME} -n ${NAMESPACE} --replicas=2
+          kubectl wait --for=condition=ready pod -l llm-d.ai/guide=${GUIDE_NAME} -n ${NAMESPACE} --timeout=600s
+          kubectl get events -n ${NAMESPACE} --field-selector reason=GKEPodSnapshotting --sort-by=.lastTimestamp
+          kubectl logs -l llm-d.ai/guide=${GUIDE_NAME} -n ${NAMESPACE} --tail=20
```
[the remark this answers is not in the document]
RyanRosario: `kubectl wait --for=condition=PodRestored` on the deployment selector would fail here because replica 1 (the initial pod that cold-started to create the snapshot) does not have `PodRestored=True`—only scaled-out replica 2 does. The human-facing verification steps immediately below check the `GKEPodSnapshotting` event and `engine.wake_up()` log line.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -1,96 +1,68 @@
...
-the pod is frozen and emits no logs.
+| Phase | Measured | Source / Notes |
+| :--- | ---: | :--- |
+| Node provisioning (`Pod created → Pod scheduled`) | varies (excluded) | Excluded to isolate pod startup latency from cloud VM provisioning |
+| Image pull (new node, image streaming enabled, 8.91 GB) | 1.8s | Kubelet `Pulling` → `Pulled` event (`containerd` mounts rootfs over `gcfs`; `69s` without Image Streaming) |
+| Python & CUDA library imports (on-demand via `gcfs`) | 73s | Container `startedAt` → vLLM log `Loading model from scratch...` |
+| Model loading — download + load into VRAM (61.03 GiB) | 130s | `Loading model from scratch...` → `Model loading took 61.03 GiB and 130.20 seconds` |
+| &nbsp;&nbsp;&nbsp;&nbsp;↳ of which, safetensors read into VRAM | 43s | `Loading weights took 43.41 seconds` |
+| Engine init (profile, KV cache, CUDA graph capture [9s / 1.86 GiB], warmup) | 73s | `Model loading took...` → `Executing engine.sleep(level=1) to release physical VRAM...` |
+| **Pod scheduled → serving-ready (checkpoint triggered)** | **4m 38s** | `PodScheduled=True` → `Executing engine.sleep(level=1)...` (`1.8s + 73s + 130s + 73s = 277.8s`) |
```
nicolexin: Can we restructure this table a little bit? The breakdown should be time focused, if we need we can separate table for the size involved but I think a note in the table for the size should be good enough.

Also we should make sure it shows model loading & engine init is actually part of the "pod scheduled -> serving ready" otherwise it feels like we are double counting. (question: is sleep part of it too?)

Let's the keep the "Total Pod scheduled -> snapshot" as the last line, and put the snapshot size in the note.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -1,96 +1,68 @@
...
+| `engine.sleep(level=1)`: weights offloaded to host RAM | 61.68 GiB | `CuMemAllocator: sleep freed 70.20 GiB memory in total, of which 61.68 GiB is backed up in CPU` |
+| `engine.sleep(level=1)`: KV cache discarded | 8.52 GiB | `...and the rest 8.52 GiB is discarded directly` (not captured in snapshot) |
+| GPU memory still in use after sleep | 2.90 GiB | `Sleep mode freed 72.92 GiB memory, 2.9 GiB memory is still in use` (includes `1.86 GiB` CUDA graphs) |
+| Time to fall asleep | 17.4s | `Executing engine.sleep(level=1)...` → `It took 17.412627 seconds to fall asleep` |
+| Checkpoint + upload to GCS (`componentCount: 2104`) | **34.0s** | `Triggering snapshot checkpoint...` → `gVisor checkpoint completed successfully` / `PodSnapshot` `Ready=True` |
+| **Total Pod scheduled → snapshot `Ready`** | **5m 29s** | `PodScheduled=True` → `PodSnapshot` condition `Ready=True` (`277.8s + 17.4s + 34.0s = 329.2s`) |
+| Snapshot size in GCS | **65.73 GiB** | Sum of objects under `gs://<bucket>/<snapshot-uid>/` (`pages.img` [`65.72 GiB`] + metadata) |
+
+A standalone microbenchmark on the same H100 node isolates where the **17.4s** goes: allocating `61.68 GiB` of pinned host memory
```
nicolexin: nit: we can remove the section. I don't think users needs to know the details here.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -1,96 +1,68 @@
...
-Upload ran at 83.5 MB/s here, 82.9 MB/s on an earlier vLLM v0.25.0 run, and 82.0 MB/s on a smaller
-A100 / 8B run — within 2% across different GPUs, two vLLM releases, and a 3.5× difference in snapshot
-size, which suggests a fixed ceiling of the snapshot write path rather than a property of the
-workload.
-
-As a planning rule, **snapshot GB ÷ 5 ≈ minutes of checkpoint freeze**.
+| Phase | Measured | Source / Notes |
+| :--- | ---: | :--- |
+| Node provisioning (`Pod created → Pod scheduled`) | varies (excluded) | GCE VM provisioning time (excluded to isolate pod restore latency) |
+| Pod scheduled → Pod restored (image mount + GCS snapshot stream) | **14.0s** | `PodScheduled=True` (container `startedAt` at `+6.0s`) → Pod condition `PodRestored=True` |
```
nicolexin: Please double check but I think there is also image pulling here.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -1,96 +1,68 @@
...
+| Pod restored → Pod `Ready` | **4.0s** | `PodRestored=True` (`Process restored from snapshot checkpoint`) → Pod condition `Ready=True` (includes `2.67s` `engine.wake_up()`) |
+| **Pod scheduled → Pod `Ready` (serving-ready)** | **18.0s** | `PodScheduled=True` → Pod condition `Ready=True` (`14.0s + 4.0s = 18.0s`; **15.4×** speedup vs `4m 38s`) |
```
nicolexin: nit: Can we add a "Total" here?

RyanRosario:
After all changes were made and issues resolved, the test pods used for benchmarking were torn down and started fresh. All statistics are accurate.

**Remaining work**

- [x] ~~Double check if dry run is required or if we can exclude it.~~ The guide is now included in the Kustomize dry run (it was previously excluded). Required vendoring the GKE `PodSnapshot` CRDs into `.github/scripts/` so the overlay validates server-side in `kind`.
- [x] ~~Make sure skipped tests are actually supposed to be skipped.~~ Verified: the `build-images/*` jobs skip because no Dockerfile changed, the four `nightly / Skip TPU v7 Run` jobs are cron-only, and `auto-gate` is push-event-only (fork PRs use `pr-gate`).
- [x] Final LGTM — @nicolexin has LGTM’d the user guide; 
- [ ] Final merge approval by @ahg-g ; pending
- [ ] Nightly E2E — @nicolexin and I agreed to address this in a follow-up.

All CI checks pass.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,83 @@
...
+<details>
+<summary><b><i>Click</i></b> to view the per-phase breakdown</summary>
+
+### Cold start — first pod, creates the snapshot
+
+| Phase | Measured | Source / Notes |
+| :--- | ---: | :--- |
+| Node provisioning (`Pod created → Pod scheduled`) | varies (excluded) | Excluded to isolate pod startup latency from cloud VM provisioning |
+| Image pull (new node, image streaming enabled, 8.91 GB) | 1.8s | Kubelet `Pulling` → `Pulled` event (`containerd` mounts rootfs over `gcfs`; `69s` without Image Streaming) |
+| Python & CUDA library imports (on-demand via `gcfs`) | 73s | Container `startedAt` → vLLM log `Loading model from scratch...` |
```
nicolexin: I am not sure if we want to show "Python & CUDA library imports" this is a image streaming cost being optimized here. Otherwise the image pull would take ~5mins if we are doing a standard pull.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
nicolexin: We can remove the section here, the explanation looks good to me but user might not be interested.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,83 @@
...
+| `engine.sleep(level=1)`: weights offloaded to host RAM | 61.68 GiB | `CuMemAllocator: sleep freed 70.20 GiB memory in total, of which 61.68 GiB is backed up in CPU` |
+| `engine.sleep(level=1)`: KV cache discarded | 8.52 GiB | `...and the rest 8.52 GiB is discarded directly` (not captured in snapshot) |
+| GPU memory still in use after sleep | 2.90 GiB | `Sleep mode freed 72.92 GiB memory, 2.9 GiB memory is still in use` (includes `1.86 GiB` CUDA graphs) |
+| Time to fall asleep | 17.4s | `Executing engine.sleep(level=1)...` → `It took 17.412627 seconds to fall asleep` |
+| Checkpoint + upload to GCS (`componentCount: 2104`) | **34.0s** | `Triggering snapshot checkpoint...` → `gVisor checkpoint completed successfully` / `PodSnapshot` `Ready=True` |
+| **Total Pod scheduled → snapshot `Ready`** | **5m 29s** | `PodScheduled=True` → `PodSnapshot` condition `Ready=True` (`277.8s + 17.4s + 34.0s = 329.2s`) |
+| Snapshot size in GCS | **65.73 GiB** | Sum of objects under `gs://<bucket>/<snapshot-uid>/` (`pages.img` [`65.72 GiB`] + metadata) |
+
+A standalone microbenchmark on the same H100 node timed the two largest components of the **17.4s**: allocating `61.68 GiB` of pinned
```
nicolexin: We can remove the section here, the explanation looks good to me but users doesn't need to know!

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,83 @@
...
+`torch.cuda.empty_cache()` that `CuMemAllocator.sleep()` runs after the copy, which is the only other work in that code path.
+Host page-locking — not PCIe bandwidth — dominates, which is why `level=1` (offload weights to host RAM) costs seconds where `level=2`
+(discard weights) would not. gVisor's GCS client then streams `pages.img` directly from host RAM to GCS using parallel composite uploads
+(`componentCount: 2104` chunks of `32 MiB`), completing the `65.73 GiB` checkpoint in **34.0s**.
+
+### Restore — every subsequent pod
+
+| Phase | Measured | Source / Notes |
+| :--- | ---: | :--- |
+| Node provisioning (`Pod created → Pod scheduled`) | varies (excluded) | GCE VM provisioning time (excluded to isolate pod restore latency) |
```
nicolexin, line 50: Is there also a image pulling time?

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
```diff
@@ -0,0 +1,83 @@
...
+| Pod scheduled → container start (image mount, sandbox create) | 6.0s | `PodScheduled=True` → container `state.running.startedAt` |
+| Container start → process restored | **8.0s** | Container `startedAt` → Pod condition `PodRestored=True` |
+| Process restored → Pod `Ready` | **4.0s** | `PodRestored=True` → Pod condition `Ready=True` |
+| **Pod scheduled → Pod `Ready` (serving-ready)** | **18.0s** | `PodScheduled=True` → `Ready=True` (`6.0s + 8.0s + 4.0s = 18.0s`; **15.4×** speedup vs `4m 38s`) |
```
nicolexin: Can we add a "Total" here for the last line?

nicolexin:
@RyanRosario nice work, thanks! LGTM on the user guide, just a few nits from me on the benchmarking report.

@ahg-g Can you please also take a look?

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Sure thing. Removed.

Agreed, and merged. You're right that the split was an artifact of Image
Streaming rather than two independent costs. There is now one row.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Restructured. Fortunately I had all of the results.

- `Measured` is now durations only. Sizes moved into the notes column, so the four GiB-valued rows are gone.
- Sub-phases are prefixed `↳` and each bold total's note names the rows that
  compose it, with the arithmetic. There is a caption explaining to the user.
- `Total Pod scheduled → snapshot Ready` (`5m 29s`) is the last row; the
  snapshot size sits in the checkpoint-upload note just above it.

On your question — no, sleep is not part of serving-ready. Serving-ready is
the instant the engine could serve traffic (weights loaded, KV cache allocated,
CUDA graphs captured), which is immediately *before* `engine.sleep(level=1)` is
called. That's now explicitly stated.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: I agree this is too much for the user. I overdid it in areas and drastically cut back what was written in other areas. Thanks!

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Yes — it's inside the `6.0s` row, not missing. With Image Streaming that fetch is cheap — the cold-start run measured `Pulling` →
`Pulled` at `1.8s` for the `8.91 GB` image on a newly provisioned node — which is why
`6.0s` covers fetch, mount and sandbox creation together. I didn't record whether
these specific nodes already had the image cached, so I've reported it as one window
rather than splitting it and implying a precision I don't have.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Added! Done, and I did the same for the cold-start table. The last row of each is now
labelled `Total` with its technical definition inline

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Correction to my earlier reply here. I said the pull/imports split was an artifact of Image Streaming — that was wrong. I went back to Cloud Logging and pulled the events for the actual benchmark pod, and there was no image pull at all: the node already had the image cached and the kubelet logged `Container image ... already present on machine`. There is no `Pulling` event anywhere in that run. The `1.8s` I had quoted came from an earlier run on a genuinely fresh node, so it did not belong in this table.

Since I was in there, I re-derived every cold-start figure from the pod's own event and log timestamps rather than from my notes:

| | was | measured |
|---|---:|---:|
| Pod scheduled → model load begins | 74.8s | 75.1s |
| Model loading | 130s | 129.8s |
| Engine init | 73s | 72.4s |
| Subtotal → serving-ready | 4m 38s | 4m 37s |
| Checkpoint + upload | 34.0s | 33.2s |
| Total → snapshot `Ready` | 5m 29s | 5m 28s |

`75.1 + 129.8 + 72.4 = 277.4` and `277.4 + 17.4 + 33.2 = 328.0`, so both totals now reconcile exactly against `PodScheduled` and the `PodSnapshot` `Ready=True` transition. The speedup is unchanged at **15.4×**.

The row is relabelled `Pod scheduled → model load begins (container start, Python & CUDA imports)`, and the note under the table states that no image pull is included and that a first pull onto a node that has never seen the image is extra.

guides/gke-pod-snapshots/benchmark-results/vllm-qwen3-32b-h100.md
RyanRosario: Following up on this one — I told you I hadn't recorded whether those nodes already had the image. I went back and checked, and they did: the kubelet logged `Container image ... already present on machine` for the restore pod, and there is no `Pulling` event in the run.

So the direct answer to your question is that there is no image pull time in the `18.0s` at all. The `6.0s` row is rootfs mount and gVisor sandbox creation only. The table now says that outright instead of hedging.

The caveat I've kept is that the snapshot holds the process memory image, not the container image, so a restore onto a node that has never seen the image would pay the fetch inside that same `6.0s` window. That case I genuinely haven't measured end to end, and the note says so.

RyanRosario:
Pushed `ab7edac6`. The GitHub Actions workflows are sitting at `action_required` (fork PR), so
in the meantime I ran the CI checks locally against the exact configs in this repo. All green:

**Linters (`.pre-commit-config.yaml`)**

- [x] `pre-commit run --from-ref origin/main --to-ref HEAD` — **exit 0**, 9 hooks passed
  (trailing-whitespace, end-of-file-fixer, check-yaml, check-added-large-files,
  check-merge-conflict, mixed-line-ending, check-case-conflict, markdownlint, yamllint)
- [x] `shellcheck -x --severity=warning .github/scripts/install-gke-crds.sh` — exit 0, no findings
- [x] `yamllint` with this repo's exact rule overrides, across all 8 changed YAML files — 0 findings
- [x] `markdownlint` v0.47.0 with `.markdownlint.json` on all changed Markdown — 0 findings
- [x] `actionlint` — 0 findings in any file this PR touches

**Generated artifacts**

- [x] `make verify-guide-readmes` → `guides/gke-pod-snapshots/README.md: up to date`

**Kustomize dry-run (`ci-kustomize-dry-run.yaml`)**

- [x] **kind** — fresh cluster, vendored `.github/scripts/gke-podsnapshot-crds.yaml` applied via
      `kubectl apply --server-side --validate=false` exactly as `install-gke-crds.sh` does, then
      `kustomize build` + `kubectl apply --dry-run=server`. Exit 0, 6/6 objects validated.
- [x] **Live GKE** (`a3-highgpu-1g` cluster with the real GKE Pod Snapshot CRDs and admission
      webhooks) — same overlay, exit 0, 6/6 objects validated.

```
serviceaccount/gke-pod-snapshots-nvidia-gpu-vllm-sa created (server dry run)
configmap/gke-pod-snapshots-nvidia-gpu-vllm-snapshot-scripts-hhg4t6cm79 created (server dry run)
configmap/gke-pod-snapshots-nvidia-gpu-vllm-snapshot-scripts-vllm-fkbct97k6t created (server dry run)
deployment.apps/gke-pod-snapshots-nvidia-gpu-vllm-decode created (server dry run)
podsnapshotpolicy.podsnapshot.gke.io/gke-pod-snapshots-nvidia-gpu-vllm-pod-snapshot-policy created (server dry run)
podsnapshotstorageconfig.podsnapshot.gke.io/gke-pod-snapshots-nvidia-gpu-vllm-pod-snapshot-storage-config created (server dry run)
```

**Commit hygiene**

- [x] GPG signature verified (`G`) — `signed-commits` check passing on the PR
- [x] DCO `Signed-off-by` present — `DCO` check passing on the PR

**Not run locally:** `hadolint` — this PR changes no Dockerfiles, so it has no files in scope.

Could someone with write access click **Approve and run workflows**? Thanks!

RyanRosario:
@ahg-g All CI checks are green and Nicole has reviewed/approved/LGTMed. The PR is ready for your final /approve to merge for the 0.10 code freeze today.

guides/gke-pod-snapshots/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+1. **First Pod (Cold Start & Snapshot Creation):** The initial replica loads the model weights, captures CUDA graphs, and calls `engine.sleep(level=1)` to offload weights from GPU VRAM to host CPU RAM while discarding the KV cache. It then purges the on-disk weight cache so the downloaded files are not duplicated in the container filesystem snapshot, and triggers a GKE Pod Snapshot uploaded to Google Cloud Storage (GCS).
+2. **Subsequent Pods (Fast Restoration):** When pods scale out or restart, GKE automatically restores the container from the GCS snapshot in seconds—bypassing model weight downloads and engine initialization.
+3. **Serve Traffic:** The restored pod immediately wakes up GPU memory and begins serving inference requests.
+
+---
+
+## Single-GPU Scope & Technical Caveats
+
+1. **Single-GPU Scope & Snapshot Warmup (`replicas: 1` → `N`):**
```
ahg-g: this doesn't make it explicit whether or not TP>1 is supported, pls clarify

nicolexin:
@RyanRosario Can you please change the well lit path title to "pod snapshot" and explicitly say in the README that we currently support GKE but we are also working on a platform agnostic solution based on CRIU? (and also contributions are welcomes here).

RyanRosario:
All feedback addressed.

@ahg-g  the scope was buried mid-sentence in a bullet titled "Single-GPU Scope & Snapshot
Warmup," which mixed two unrelated topics. I've split it so the support boundary is the heading itself. This guide does not support multi-GPU per pod only single

@nicolexin The guide is now "Pod Snapshots for Single-GPU Model Servers" (dropped "GKE"), and the entry in `guides/README.md` is now "Pod Snapshots". The overview no longer leads with GKE either; it describes what the guide does and notes that GKE is the current implementation, an explanation of a vendor neutral method and welcomes contributions. I left the directory and filenames as is to not break review chains and maintain the same structure for CI.

CI will likely need to be re-run. It passes locally.

PR #2490, opened 2026-09-14
docs(autoscaling): deprecate WVA and pin its image to release-0.9

lionelvillard:
## What

WVA is deprecated and gets no further releases — `v0.9.0` is the final version (upstream also carries a `wva-engine-eol` tag). This pins the WVA guide assets to that release and marks the path deprecated.

### Image pin

The controller overlays already tracked `?ref=release-0.9`, but the image tag came from whatever the upstream overlay happened to set. Both platform overlays now pin it explicitly:

```yaml
images:
  - name: ghcr.io/llm-d/llm-d-workload-variant-autoscaler
    newTag: v0.9.0
```

`kustomize build` on `wva/controller/platform/{k8s,ocp}` renders `…llm-d-workload-variant-autoscaler:v0.9.0`.

### Deprecation notices

- **`guides/workload-autoscaling/README.md`** — warning at the top pointing new deployments at the KEDA + EPP paths, naming Saturation-based Autoscaling as the closest replacement for WVA's saturation signal. `KEDA + WVA Metrics (Legacy)` → `(Deprecated)`.
- **`wva/README.md`** — replaces the stale VariantAutoscaling warning (it said the CRD "will be removed in 0.9.0", already past) with the deprecation notice, keeping the CRD fact as context.
- **`replica-rebalancing/README.md`** — deprecation folded into the existing experimental-feature blockquote, since the rebalancer ships inside WVA.

### Workflows

The three WVA nightly lanes each built the controller **from `main`** and passed it via `WVA_TAG`, which the deploy scripts turn into an `images:` override — that would have defeated the pin. For `nightly-e2e-workload-autoscaling-{ibm-acc-gpu,cks-acc-gpu,amd-acc-rocm}-vllm-x.yaml`:

- Removed the `build-wva-image` job; `nightly` no longer `needs:` it.
- `WVA_TAG` now comes only from the optional `image_tag` dispatch input, so the overlay pin applies by default.
- `caller_ref` default `main` → `release-0.9`, so the e2e suite matches the image.
- Schedules disabled on cks (07:30 UTC) and amd (13:00 UTC); ibm's was already off and its reason is updated. All three keep `workflow_dispatch`.

`nightly-e2e-workload-autoscaling-keda-epp-ibm-acc-gpu-vllm-x.yaml` is EPP-only and untouched.

## Testing

- `kustomize build` on both platform overlays renders `v0.9.0`.
- `make verify-guide-readmes` passes.
- `markdownlint-cli2` error count on the three READMEs unchanged from base (all pre-existing).
- All three workflows parse and no longer contain `build-wva-image`; no dangling references remain.

## Notes for reviewers

- `permissions: packages: write` was left in place on the three workflows — now unused for pushing, but reusable workflows inherit permissions and I did not want to risk an image pull.
- The three `consolidate-status-workload-autoscaling-{amd,cks,ibm}` badge workflows and the WVA badge in the guide README still exist. With schedules off those badges will go stale, so retiring them may be worth a follow-up.


ahg-g:
@lionelvillard can you pls rebase?

lionelvillard:
@ahg-g done!

ahg-g:
@lionelvillard sorry, another conflict

PR #2492, opened 2026-09-14
Guarantee a scale-from-zero path in the fast-model-actuation-keda verification

dumb0002:
## Summary
The verification request could hit an already-warm pool. The guide now stands up at
the `ScaledObject` floor of 1, waits for KEDA to scale it to zero, then sends the test
request from there.

- **`keda/base/`**: `cooldownPeriod` and the HPA's `scaleDown.stabilizationWindowSeconds`
  300s → 30s each — they apply sequentially, so both must be cut; scale-to-zero now
  takes ~1 min instead of ~10. This guide's overlay only; the upstream base keeps its
  production values.
- **Step 6**: new wait for `spec.replicas == 0`. That is what proves the metric pipeline
  works — `Ready` only means KEDA accepted the trigger.
- **Step 2**: one request at `--max-time 600` instead of a 12×15s retry loop — flow
  control holds it through the whole 0→1 transition. Now also checks HTTP status: the
  old form had no `--fail`, and `set -e` misses a curl failure inside a pipeline, so a
  non-200 or a timeout could pass.
- **Prose**: notes the scale-up should be a **hot start** and links the numbers (4.0 s
  vs 85.3 s pod startup, `Qwen/Qwen3-32B`), with the GPU-assignment caveat deferred to
  the FMA guide.


guides/fast-model-actuation-keda/modelserver/kustomization.yaml
```diff
@@ -28,7 +31,7 @@ patches:
         labels:
           llm-d.ai/guide: fast-model-actuation-keda
       spec:
-        replicas: 1
+        replicas: 0
```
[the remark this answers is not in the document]
dumb0002: The cited line is generic KEDA `ignoreNullValues` troubleshooting, not a statement about this metric. `docs/architecture/advanced/autoscaling/hpa-epp.md:95` documents the intended behavior: `"EPP Flow Control enables scale-from-zero by buffering requests while no model server endpoint is ready. A queued request raises the queue-size metric, KEDA activates the target Deployment, and EPP dispatches the request after a model server becomes Ready."` The queue gauge is EPP-side, not backend-side, so it exists with no ready backend — that is what the flowControl gate this guide enables is for, with `activationThreshold: 0` so any `queue depth > 0` activates.

aavarghese: I'm not entirely sure what 0 replicas with no running vllm would mean for the "standup" step in benchmarking. I believe there are some "health check" steps run in standalone that pings the vllm server or something. You'll have to check the benchmarking code to make sure 0 replicas will work as-is?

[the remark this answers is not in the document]
dumb0002: requester pod replica is set to one - this comment has been addressed.

dumb0002: Good catch. Thanks! I update as following:  revert to `replicas: 1`, and rework Installation`step 6` to wait for KEDA's scale-down to 0 before sending the request.

guides/fast-model-actuation-keda/keda/base/kustomization.yaml
```diff
@@ -3,6 +3,9 @@ kind: Kustomization
 resources:
   - ../../../workload-autoscaling/keda-epp-queue/optimized-baseline/base
 
+# pollingInterval 15 is inherited from the upstream base. The two scale-down
```
rubambiza, line 6: This is slightly confusing because it implies something was inherited and then shortened (but we're going from 15 to 30). I am assuming (a) you meant 15s and (b) you are referring to the 15 * 12s retry loop mentioned in the PR summary. Perhaps my understanding is lacking because I have not yet read the rest of the files below, and my answer may be there. This was the first inconsistency I encountered, and I wanted to flag it before I forget.

guides/fast-model-actuation-keda/keda/base/kustomization.yaml
```diff
@@ -24,6 +27,16 @@ patches:
       - op: replace
         path: /spec/maxReplicaCount
         value: 2
+      # Seconds of zero demand before KEDA scales to 0 (that direction only;
+      # scale-up from 0 is the trigger's activation threshold).
```
rubambiza: What is "the" trigger referring to in this context of scale to zero?

guides/fast-model-actuation-keda/modelserver/kustomization.yaml
```diff
@@ -17,6 +17,12 @@ patches:
         modelServerConfig:
           labels:
             llm-d.ai/guide: "fast-model-actuation-keda"
+  # Start at 1 replica, the ScaledObject floor at standup -- NOT 0. The EPP emits
+  # llm_d_epp_flow_control_queue_size only once it has a ready backend endpoint, so
```
rubambiza:
```suggestion
  # llm_d_epp_flow_control_queue_size only once it has a ready backend endpoint. Otherwise,
```

I think it's better to show these in contrast to each other. Moreover, it splits the logical explanation so that the sentence does not try to explain three mechanisms at once (what `llm_d_epp_flow_control_queue_szie` does and what KEDA queries in _both_ cases (and its defaults).

guides/fast-model-actuation-keda/modelserver/kustomization.yaml
```diff
@@ -17,6 +17,12 @@ patches:
...
+  # with 0 replicas the series does not exist, KEDA queries an empty result and
+  # (ignoreNullValues defaults true) silently serves its fallback while reporting
```
rubambiza: What is the logical origin of `(ignoreNullValues defaults true)`?  Is it from KEDA querying an empty result (and thus belongs right before the _and_) or is it set somewhere else as a default (and updated if KEDA does not query an empty result) to influence KEDA's behavior?

guides/fast-model-actuation-keda/guide.yaml
```diff
@@ -106,7 +106,11 @@ deploy:
           -n ${NAMESPACE} --version ${ROUTER_CHART_VERSION}
 
   # Deploys the FMA custom resources (ISC/LauncherConfig/LauncherPopulationPolicy)
-  # AND the server-requesting pods together.
+  # AND the server-requesting pods together, at 1 replica -- the ScaledObject floor.
+  # Starting at 0 would leave the EPP with no ready backend, and the flow-control
+  # queue metric KEDA scales on is not emitted until there is one; KEDA would then
```
rubambiza:
```suggestion
  # queue metric KEDA scales on is not emitted until there is one; Otherwise, KEDA would
```

Same suggestion as [above](https://github.com/llm-d/llm-d/pull/2492/changes#r4014704503) because the previous sentence is describing what happens in the negative case.

guides/fast-model-actuation-keda/guide.yaml
```diff
@@ -121,18 +125,77 @@ deploy:
   # back up on the next queued request.
   keda:
     - run: |
+        export SO="scaledobject/${GUIDE_NAME}-queue"
+
         kubectl apply -n ${NAMESPACE} -k ${REPO_ROOT}/guides/${GUIDE_NAME}/keda/ocp
 
-        kubectl wait --for=condition=Ready --timeout=120s \
-          scaledobject/fast-model-actuation-keda-queue -n ${NAMESPACE}
+        kubectl wait --for=condition=Ready --timeout=120s ${SO} -n ${NAMESPACE}
```
rubambiza:
```suggestion
        kubectl -n ${NAMESPACE} wait --for=condition=Ready --timeout=120s ${SO} 
```

**Nit**:  Puts the global flags before the verb to make it clear we are targeting a particular namespace. Or at least apply consistently with the change on [L138](https://github.com/llm-d/llm-d/pull/2492/changes#diff-b2539132b8f3890b749bd559e12e124ae6e68f28733b870608517cedc3a0bfbeR138), L160, L172, etc. I think it makes it easier to parse the commands during review.

With that being said, I acknowledge that everyone has their own practices and muscle memory.

guides/fast-model-actuation-keda/guide.yaml
```diff
@@ -121,18 +125,77 @@ deploy:
...
+    #
+    # Require the reading to HOLD: Fallback reads False before KEDA's first poll,
+    # so a single sample proves nothing (hence a streak, not `kubectl wait`).
+    - run: |
+        streak=0
+        for _ in $(seq 1 30); do
+          if [ "$(kubectl get ${SO} -n ${NAMESPACE} \
+                -o jsonpath='{.status.conditions[?(@.type=="Fallback")].status}')" = "False" ]; then
+            streak=$((streak + 1))
+            [ "${streak}" -ge 3 ] && break
```
rubambiza: Why 3 in particular? Why not, say, 2? If I understand the loop correctly, we're trying to establish that the behavior holds for at least 20 seconds (since the third sleep is never hit), right?

guides/fast-model-actuation-keda/guide.yaml
```diff
@@ -121,18 +125,77 @@ deploy:
...
+        if [ "${streak}" -lt 3 ]; then
+          echo "KEDA is not reading the EPP queue metric (serving fallback):" >&2
+          kubectl get ${SO} -n ${NAMESPACE} \
+            -o jsonpath='{range .status.conditions[*]}  {.type}={.status} ({.reason}: {.message}){"\n"}{end}' >&2
+          exit 1
+        fi
+        echo "KEDA is scaling on the real metric (Fallback=False held)"
+    # Now wait for KEDA to take the idle requester to 0. Budget covers
+    # cooldownPeriod (30s) plus the HPA's scaleDown stabilization window (30s)
+    # plus polling slack. Reaching 0 is what makes the verification request an
```
rubambiza: What component is doing said polling? For instance, the kubelet polls every 10s or so as a default. A polling slack of 180s (240s - 60s) sounds a bit excessive, no?

guides/fast-model-actuation-keda/README.md
```diff
@@ -187,7 +188,9 @@ helm install ${GUIDE_NAME} \
 
 ### 5. Deploy the Model Server (Dual Pods)
 
-Apply the FMA custom resources — `InferenceServerConfig`, `LauncherConfig`, and `LauncherPopulationPolicy` — **together with** the server-requesting `fma-requester` Deployment in a single `kubectl apply -k modelserver/`:
+Apply the FMA custom resources — `InferenceServerConfig`, `LauncherConfig`, and `LauncherPopulationPolicy` — **together with** the server-requesting `fma-requester` Deployment in a single `kubectl apply -k modelserver/`.
+
+This creates **1 requester pod**, which reserves a GPU and brings up a vLLM instance. KEDA takes over the replica count in [step 6](#6-enable-scale-from-zero-autoscaling-keda) and scales it to zero once idle.
```
rubambiza:
```suggestion
This creates 1 set of matching **requester pod** (which reserves a GPU) and **provider pod** (which brings up a vLLM instance). KEDA takes over the replica count in [step 6](#6-enable-scale-from-zero-autoscaling-keda) and scales it to zero once idle.
```

Worth considering since we're mentioning `Dual Pods` in the header.

guides/fast-model-actuation-keda/README.md
```diff
@@ -200,33 +203,69 @@ kubectl rollout status deployment/fma-requester -n ${NAMESPACE} --timeout=300s
...
+    streak=0
+  fi
+  sleep 10
+done
+
+if [ "${streak}" -lt 3 ]; then
+  echo "KEDA is not reading the EPP queue metric (serving fallback):" >&2
+  kubectl get ${SO} -n ${NAMESPACE} \
+    -o jsonpath='{range .status.conditions[*]}  {.type}={.status} ({.reason}: {.message}){"\n"}{end}' >&2
+  exit 1
```
rubambiza: Is there any action that the user should take with this exit 1? Or at least `echo`-ing some more context from the `kubectl get` above, unless the print statement from the get is implicit with an echo?

guides/fast-model-actuation-keda/README.md
```diff
@@ -264,23 +305,32 @@ kubectl run curl-test --rm -i --restart=Never \
...
+**To watch the scale-up while it runs**, from another terminal:
+
+<!-- llm-d-cicd:skip start -->
+```bash
+kubectl get deployment/fma-requester -n ${NAMESPACE} -w
+```
+<!-- llm-d-cicd:skip end -->
+
+> [!NOTE]
+> **This is [warm start](../fast-model-actuation/README.md#overview), not hot start** — on a fresh deploy there is no sleeping instance to wake, so the launcher creates a new one. Hot start needs a prior scale-down and is the faster path: **4.0 s vs 85.3 s** mean pod startup on `Qwen/Qwen3-32B` ([benchmark results](./benchmark-results/qwen3-32b-h100/README.md#queue-based-autoscaling)).
```
rubambiza:
```suggestion
> **This is [warm start](../fast-model-actuation/README.md#overview), not hot start** — on a fresh deploy there is no sleeping instance to wake, so the launcher creates a new one. Hot start, which needs a prior scale-down, is the faster path: **4.0 s vs 85.3 s** mean pod startup on `Qwen/Qwen3-32B` ([benchmark results](./benchmark-results/qwen3-32b-h100/README.md#queue-based-autoscaling)).
```

**Nit**: I think the faster path is the core point you are trying to make, so the other part should be contextual.

rubambiza (changes requested):
A few suggestions on consistency as well as clarifying questions because it is my first time viewing the KEDA + FMA WLP.

guides/fast-model-actuation-keda/keda/base/kustomization.yaml
dumb0002: `15 sec` s the KEDA **pollingInterval** which we're inheriting. What we're changing is the KEDA's own **cooldownPeriod** from `300 sec. --> 30 sec` - both are different metrics.

guides/fast-model-actuation-keda/keda/base/kustomization.yaml
dumb0002: scale-to-zero uses a **cooldownPeriod** - is how long the metric must stay at/below the activation threshold before scaling back to 0.

guides/fast-model-actuation-keda/modelserver/kustomization.yaml
dumb0002: Fixed, although I used another variation than the proposed in the comment.

guides/fast-model-actuation-keda/guide.yaml
dumb0002: Fixed! Although I used another variation than the proposed in the comment.

guides/fast-model-actuation-keda/guide.yaml
dumb0002: I prefer leave it as-is because it consistent with other guides. For example: https://github.com/dumb0002/llm-d/blob/fix-fma-keda-wlp/guides/optimized-baseline/guide.yaml

guides/fast-model-actuation-keda/guide.yaml
dumb0002: This is not related with vLLM sleep at all but just about registering a prometheus metric series. 

- Request 1 does the actual work — registers the series.
- Requests 2 and 3 are insurance, if the first request fails because the model server is not ready yet.

guides/fast-model-actuation-keda/guide.yaml
dumb0002: KEDA is doing the pulling.  Where did you get this equation:  `180s (240s - 60s) s` ???
We're waiting `240 sec` so it past KEDA cooldownPeriod and HPA scaleDown stablilization window here: https://github.com/dumb0002/llm-d/blob/fix-fma-keda-wlp/guides/fast-model-actuation-keda/guide.yaml#L185.
Basically, we're waiting KEDA to scale down to zero the requester pod before proceeding to the next step.

guides/fast-model-actuation-keda/README.md
dumb0002: Yes, indeed. I added that a launcher pod is also created (`the dual pod concept`).

guides/fast-model-actuation-keda/README.md
dumb0002: Valid point. I added a reference for the user to `guides/workload-autoscaling/keda-epp-queue/README.md#troubleshooting`

guides/fast-model-actuation-keda/README.md
dumb0002: This note is no longer correct after addressing @aavarghese comment's above -  since we're now scaling down to zero and then scaling up. Hence, it should trigger a FMA `hot-start` instead - I updated the note accordingly.

guides/fast-model-actuation-keda/modelserver/kustomization.yaml
dumb0002: `ignoreNullValues` -  it's a scaler config field in `KEDA` trigger configuration with default value `true`, it's not a consequence of the empty query - see https://keda.sh/docs/2.20/scalers/prometheus/#trigger-specification

dumb0002: It's a KEDA Prometheus-scaler field that defaults to true — a precondition on how the scaler behaves, not a consequence of the empty query. We never set it; the trigger inherits the default. Per the scaler docs (https://keda.sh/docs/2.20/scalers/prometheus/#trigger-specification): "Default is true, which means ignoring the empty value list from Prometheus. Set to false the scaler will return error when Prometheus target is lost." So: no ready endpoint → no series → empty result → read as 0 rather than an error. I've reworded the comment to put it in its own clause rather than a parenthetical.

dumb0002:
> A few suggestions on consistency as well as clarifying questions because it is my first time viewing the KEDA + FMA WLP.

@rubambiza, I have addressed all your comments. This PR is ready for another round of review. Thanks!

guides/fast-model-actuation-keda/keda/base/kustomization.yaml
rubambiza: Thanks for clarifying

guides/fast-model-actuation-keda/keda/base/kustomization.yaml
rubambiza: Okay, it sounds like these are different phenomena. May I recommend perhaps rephrasing to "scale-up from 0 is _based on_ the trigger's activation threshold". Otherwise, it sounds like scale-up from 0 is _doing_ the triggering instead of the _being_ triggered.

guides/fast-model-actuation-keda/README.md
```diff
@@ -187,7 +188,9 @@ helm install ${GUIDE_NAME} \
 
 ### 5. Deploy the Model Server (Dual Pods)
 
-Apply the FMA custom resources — `InferenceServerConfig`, `LauncherConfig`, and `LauncherPopulationPolicy` — **together with** the server-requesting `fma-requester` Deployment in a single `kubectl apply -k modelserver/`:
+Apply the FMA custom resources — `InferenceServerConfig`, `LauncherConfig`, and `LauncherPopulationPolicy` — **together with** the server-requesting `fma-requester` Deployment in a single `kubectl apply -k modelserver/`.
+
+This creates 1 [server-requesting pod](../fast-model-actuation/README.md#overview), which reserves a GPU, and the FMA controllers bind it to a [launcher pod](../fast-model-actuation/README.md#overview) — the launcher is what actually runs the vLLM instance. KEDA takes over the replica count in [step 6](#6-enable-scale-from-zero-autoscaling-keda).
```
rubambiza, line 193: I like this description much better!

rubambiza (approved):
Approving with one non-blocking suggestion.

dumb0002:
@rubambiza, thanks! I addressed the now blocking [here](https://github.com/llm-d/llm-d/pull/2492#discussion_r4028960433) comment see below: 

Agreed, that read as if the scale-up were doing the triggering. Went with "governed by the trigger's activation threshold", which also matches keda-epp-queue/README.md:475 ("has crossed its activation threshold").

maugustosilva (approved):

guides/fast-model-actuation-keda/guide.yaml
rubambiza: I added the `coolDownPeriod (30s)` + `scaleDown` stabilization window (30s), and then subtracted from the 240s of the wait being invoked right below.

PR #2493, opened 2026-09-14
 guides/wide-ep-lws: add AMD/ROCm MoRI wide-EP 2P2D recipe for DeepSeek-V3

shikamd123:
Generic wide-ep guides for running an inference with wide expert parallelism using leader-worker deployment.
 DP=EP=16, TP=1 over two pods per role; MoRI-EP for the expert all2all and
 MoRI-IO WRITE mode for KV.


shikamd123:
Note: The guides run 2p2d with dp=ep=16, tp=1, MORI-IO(write mode), MORI-EP(wide-ep)

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/decode.yaml
```diff
@@ -0,0 +1,265 @@
...
+    llm-d.ai/role: decode
+spec:
+  replicas: 1
+  leaderWorkerTemplate:
+    size: 2
+    workerTemplate:
+      metadata:
+        annotations:
+          # Replace the NAD name/namespace with what your cluster publishes (`kubectl get net-attach-def -A`).
+          k8s.v1.cni.cncf.io/networks: |
```
vcave: move to amd-ci/

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/decode.yaml
```diff
@@ -0,0 +1,265 @@
...
+                valueFrom:
+                  secretKeyRef:
+                    name: llm-d-hf-token
+                    key: HF_TOKEN
+              # Must exceed the sidecar's 90 s idle timeout, or pooled connections are reset mid-request.
+              - name: VLLM_HTTP_TIMEOUT_KEEP_ALIVE
+                value: "120"
+
+              # MoRI-EP: must stay identical to prefill.yaml, which carries the per-variable notes.
+              - name: MORI_RDMA_DEVICES
```
vcave: As far as i recall in the PD moriio setup there are no other nics available so setting this may not do anything. If it is necessary, move it amd-ci provider

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/decode.yaml
```diff
@@ -0,0 +1,265 @@
...
+                value: "ionic_0,ionic_1,ionic_2,ionic_3,ionic_4,ionic_5,ionic_6,ionic_7"
+              - name: MORI_RDMA_SL
+                value: "3"
+              - name: MORI_RDMA_TC
+                value: "104"
+              - name: MORI_ENABLE_HOST_PROXY
+                value: "1"
+              - name: MORI_ENABLE_RAIL_ONLY
```
vcave: Double check which of all of there mori settings are provider-specific or generic. For example tc, rail, dmabuf are likely provider-specific.

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/decode.yaml
```diff
@@ -0,0 +1,265 @@
...
+              - name: TORCHINDUCTOR_USE_STATIC_CUDA_LAUNCHER
+                value: "0"
+              - name: HSA_ENABLE_SDMA
+                value: "1"
+              - name: HSA_FORCE_FINE_GRAIN_PCIE
+                value: "1"
+              - name: NCCL_IB_DISABLE
+                value: "1"
+              - name: NCCL_SOCKET_IFNAME
+                value: eth0
```
vcave, line 231: check if this exists in the container

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/prefill.yaml
```diff
@@ -0,0 +1,233 @@
...
+    llm-d.ai/model: DeepSeek-V3
+    llm-d.ai/role: prefill
+spec:
+  replicas: 1
+  leaderWorkerTemplate:
+    size: 2
+    workerTemplate:
+      metadata:
+        annotations:
+          # Replace the NAD name/namespace with what your cluster publishes (`kubectl get net-attach-def -A`).
```
vcave: Overall same comment as for decode regarding base vs amd-ci for NADs and MORI-related env variables

vcave (commented):
Would be good to clean these up in this PR or subsequent one.

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/decode.yaml
shikamd123: addressed

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/decode.yaml
shikamd123: adressed

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/prefill.yaml
shikamd123: adressed

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/decode.yaml
shikamd123: Verified on hardware: eth0 is up, holds the pod IP, owns the default route; net1–net8 carry no IPv4. Kept, comment corrected

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/decode.yaml
shikamd123: Moved to amd-ci, you are right , it is not necessary. /dev/infiniband exposes only uverbs1..8 so ibv_get_device_list already returns those eight in order. Kept explicit because MORI_DISABLE_TOPO=1 maps rank→NIC by list position

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/prefill.yaml
shikamd123: Applied symmetrically. The `amd-ci` patch targets `kind: LeaderWorkerSet` with no name selector, so prefill and decode get the same overrides from one file.

guides/wide-ep-lws/modelserver/amd/vllm-deepseek-v3/base/decode.yaml
```diff
@@ -0,0 +1,294 @@
+apiVersion: leaderworkerset.x-k8s.io/v1
+kind: LeaderWorkerSet
```
ahg-g, line 2: we are moving away from using LWS directly to using DisaggregatedSet, pls see the other guides under wide-ep

ahg-g:
@shikamd123 sorry for the inconvenience, a couple of things happened:
1) As mentioned above, DisaggSet is the API for all wide-ep and p/d guides
2) the lws substring has been dropped from the top level guide name


happy to look at this as soon as you rebase.

shikamd123:
@ahg-g  could you let me know if there is any peer discovery mechanism with the new refactor and disaggregatedset ?

shikamd123:
@ahg-g : Thanks — rebasing onto guides/wide-ep/ and converting to DisaggregatedSet now. One design question before I push, since MoRI-IO's peer requirement differs from NIXL's.

I deployed a throwaway DisaggregatedSet to check pod naming. Child objects are <ds-name>-<template-hash>-<role>, and the hash is stable across restarts but changes on any pod-template edit (confirmed: c4f06366 → ea2cc728 after a one-line change). So the FQDNs this recipe currently passes to the routing sidecar can't survive the migration.

For NIXL that's a non-issue — the sidecar takes no peer flags and the EPP injects x-prefiller-host-port per request, with disaggregatedset-rollout-screener keeping both legs on one revision. I'd like to adopt that.

The gap is that MoRI-IO also needs an ordered per-pod host list at startup, not just a per-request pick. _background_moriio_handshake iterates for cur_dp_rank in range(remote_dp_size) and indexes pod_hosts[pod_index(rank, dp_local)] to do an all-to-all handshake with every remote DP rank. That list reaches the connector as kv_transfer_params["remote_hosts"], which the sidecar builds today from static flags.

Is there an intended way to source that list under DisaggregatedSet? The options I can see are (a) an EPP plugin surfacing the role's pod set rather than a single pick, (b) the per-role -prv headless service, or (c) having the sidecar derive peers from its own subdomain, since both roles share the hash. Happy to implement whichever fits the direction.

PR #2502, opened 2026-09-15
Make DisaggregatedSet the base for the wide-EP GLM-5.2 guide

BenjaminBraunDev:
## What

Converts the wide-EP GLM-5.2 guide to `DisaggregatedSet`, following #2379.

- One DS with prefill and decode roles replaces the LWS prefill/decode manifests.
- Role `rolloutStrategy` pinned to maxSurge 0, maxUnavailable 1 to match the old LWS defaults.
- The coreweave provider and the components patch env, volumes, and containers by name via a DS schema declared once in the base, with shared role specs written once via YAML anchors.
- The 27 deployments only set replicas and group size, so they use JSON6902 against the fixed [prefill, decode] role order.

Name-based patching also fixes a bug: the old no-mtp patch replaced env index 5, which was KV_OFFLOAD_CPU_BYTES on prefill and ENABLE_PROFILING on decode, and left a duplicate ENABLE_MTP where the last entry ("1") wins. So the baseline and offloading benchmark overlays never actually disabled MTP. It now sets a single ENABLE_MTP=0 and keeps those vars.

All 30 kustomizations render semantically identical to the LWS manifests they replace, except the pinned rolloutStrategy, the `llm-d.ai/inference-serving` role label, and the no-mtp fix in the 9 overlays that use it.


guides/wide-ep-lws/modelserver/gpu/vllm-glm-5.2/providers/coreweave/disaggregatedset-schema.json
```diff
@@ -0,0 +1,100 @@
+{
```
ahg-g: we should find a way to avoid doing this over and over, pls investigate.

BenjaminBraunDev: Yeah, working on consolidating repetition. Will open for review once it's in a presentable state.

BenjaminBraunDev: Done. I cut it down to a single DS schema per base and did the same for #2501

ahg-g (approved):

PR #2508, opened 2026-09-16
docs: add multi-endpoint serving guide

DeanKelly751:
**What type of PR is this?**
/kind documentation

Adds a multi-endpoint serving guide under the existing `diffusion-serving` package, consolidating single-endpoint and multi-endpoint deployment patterns in one place per reviewer feedback.

The multi-endpoint guide deploys a single model that serves **multiple modality endpoints** from one checkpoint (e.g. Qwen3-Omni serving text, audio, and images), using a single Gateway API HTTPRoute with multiple exact-path match rules to direct each endpoint to the model's InferencePool.

**Architecture:**

```
/v1/audio/speech         (Exact) → model-pool → model pods  (if model supports TTS)
/v1/audio/transcriptions (Exact) → model-pool → model pods  (if model supports ASR)
/v1/images/generations   (Exact) → model-pool → model pods  (if model supports image gen)
/v1/chat/completions             → catch-all  → model-pool  (or text pool)
```

The EPP uses the standard config (`queue-scorer` + `max-score-picker`, no custom plugins). Pod homogeneity within the pool eliminates the need for architecture-aware filtering.

**This PR includes:**

- `guides/diffusion-serving/multi-endpoint/README.md` — guide covering multi-endpoint deployment with architecture, model capability table, prerequisites, installation, verification per modality, multi-pool deployment, and cleanup
- `guides/diffusion-serving/multi-endpoint/httproutes.yaml` — single HTTPRoute with multiple exact-path match rules for all supported modality endpoints
- `guides/diffusion-serving/multi-endpoint/router/values.yaml` — Helm values for standard EPP config
- `guides/diffusion-serving/multi-endpoint/modelserver/gpu/vllmomni/base/` — Kustomize overlay and deployment patch for vLLM-Omni model server (default: Qwen3-Omni)
- `guides/diffusion-serving/README.md` — updated with single-endpoint and multi-endpoint sections

**Changes from review feedback:**

- Consolidated under `diffusion-serving/` instead of a separate `omni-serving/` package
- Simplified HTTPRoutes into a single resource with multiple path matches (was 3 separate HTTPRoute objects)
- Simplified values.yaml plugin config (removed explicit plugin names)

**Validation:**

- Validated the separate InferencePool + HTTPRoute routing architecture on OpenShift (wetlab-ai) with separate pools, each with its own EPP running standard config. Both EPPs healthy, caches populated, pools reconciled, HTTPRoutes accepted by Gateway.
- Kustomize build passes for the new overlay.

**Release note:**
```release-note
NONE
```


ahg-g:
I think we should the guides under one pkg, either call it omni-serving or diffusion-serving

DeanKelly751:
> I think we should the guides under one pkg, either call it omni-serving or diffusion-serving

Hi @ahg-g,

The reason they're separate is an intentional distinction we're trying to make clearer in llm-d:

diffusion-serving = dedicated single-purpose models that each serve one generation endpoint (e.g. Qwen3-TTS → /v1/audio/speech, FLUX → /v1/images/generations). One model, one endpoint, one namespace per guide.

omni-serving = a single unified model that generates multiple output types from one checkpoint (text, audio, image all from the same vLLM-omni pods). One model, multiple endpoints, with exact-path HTTPRoutes fanning out to the same pool.

This follows the distinction vLLM itself makes:

Model generates only text → vLLM (even if it accepts images/audio as input — that's "multimodal" in the input sense)
Model generates other types (audio, image, video) → vLLM Omni

The deployment architecture is  different: diffusion guides use a PathPrefix: / catch-all because there's only one endpoint, while omni-serving uses multiple exact-path HTTPRoutes because one model serves several endpoints. 
We are just trying to make this distinction clear 

I am happy to add a note to the omni-serving README clarifying how it differs from the diffusion-serving guides if that would help?

guides/omni-serving/router/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+        plugins:
+        - name: queueScorer
+          type: queue-scorer
+        - name: maxScore
+          type: max-score-picker
+        schedulingProfiles:
+        - name: default
+          plugins:
+          - pluginRef: queueScorer
+          - pluginRef: maxScore
```
ahg-g:
```suggestion
        - type: queue-scorer
        - type: max-score-picker
        schedulingProfiles:
        - name: default
          plugins:
          - pluginRef: queue-scorer
          - pluginRef: max-score-picker
```

ahg-g: no need to provide a name in this case, the name will be defaulted.

guides/omni-serving/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+| Model | Engine | Endpoints | HTTPRoutes to deploy |
+|---|---|---|---|
+| **Qwen3-Omni** (default) | vLLM-Omni | `/v1/chat/completions`, `/v1/audio/speech`, `/v1/audio/transcriptions` | `omni-audio-speech`, `omni-audio-transcriptions` |
+| **Qwen3-TTS** | vLLM-Omni | `/v1/audio/speech` | `omni-audio-speech` |
+| **FLUX / Diffusion** | vLLM-Omni | `/v1/images/generations` | `omni-images-generations` |
+| **Future omni model** | vLLM-Omni | All of the above | All HTTPRoutes |
+
+> [!NOTE]
+> This guide differs from the single-endpoint [diffusion-serving guides](../diffusion-serving/) (text-to-speech, text-to-image, image-to-image), which each deploy a dedicated model for one endpoint. Here, a single model may handle **multiple** endpoints, and exact-path HTTPRoutes direct each modality to its InferencePool.
```
ahg-g: This is not a material difference, it is a mode of deployment only, at the end of the day the guides under diffusion-serving deploy vllmomni, I strongly suggest to consolidate under one umbrella pkg, I am not against naming it omni-serving, but we should have two pkgs.

guides/omni-serving/httproutes/httproutes.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - name: inference-gateway
+  rules:
+  - matches:
+    - path:
+        type: Exact
+        value: /v1/audio/speech
+    backendRefs:
+    - group: inference.networking.k8s.io
+      kind: InferencePool
+      name: omni-serving
```
ahg-g: All the routes send the traffic to the same pool, omni-serving; this is confusing as one can rightly point that setting multiple routes is redundant.

Unless the different routes point to different pools with different capabilities, then a single route with a list of paths will achieve the same outcome, but using simpler config.

ahg-g (commented):
Thanks for the clarification, @DeanKelly751. While I understand the conceptual distinction between dedicated single-endpoint models and unified multi-output models, I still believe keeping them in separate top-level packages creates unnecessary fragmentation for the repository.

A few points to consider:

Deployment Mode vs. Distinct Packages: At the end of the day, both setups deploy vllmomni. Whether a deployed model serves one modality or multiple is primarily an operational deployment mode rather than a reason to duplicate package structures. Maintaining two separate packages increases maintenance overhead when manifests or shared configs evolve.

Consolidating under one umbrella: We should consolidate these under a single package (e.g., omni-serving), with sub-guides or sections explaining both patterns: dedicated single-purpose endpoints vs. multi-endpoint setups.

Route Simplification: For the multi-endpoint deployment, since all endpoints currently target the exact same backend pool (omni-serving), creating multiple separate HTTPRoute objects is redundant. A single HTTPRoute containing multiple path match rules achieves the same routing behavior with a significantly simpler and cleaner configuration.

Let's consolidate the guides under one package and simplify the route configuration rather than keeping them split across separate directories.

DeanKelly751:
Thanks @ahg-g for the feedback, I have addressed your points in the latest push:

- Consolidated under one package — removed the separate omni-serving/ directory. The multi-endpoint guide now lives at diffusion-serving/multi-endpoint/ alongside the existing single-endpoint sub-guides, with the parent README updated to cover both patterns.

- Route simplification — consolidated into a single HTTPRoute with multiple path match rules. Users remove the matches their model doesn't support.

- Values.yaml — applied the suggested simplification, types used directly as refs.

Please let me know what you think and thank you for all your help!

ahg-g:
/approve

thanks

PR #2522, opened 2026-09-17
fix(pd-disaggregation): pin llm-d-rocm to v0.8.1, v0.9.0 ships rixl package under a name vllm doesn't recognize

BenjaminBraunDev:
The v0.9.0 image installed NIXL under a name vLLM doesn't look for, so P/D pods can't start. v0.8.1 is the last image where the name matches.

vLLM on ROCm imports `rixl` ([nixl_utils.py](https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/distributed/nixl_utils.py#L39-L43)), v0.9.0 ships `nixl_rocm` ([Dockerfile.rocm](https://github.com/llm-d/llm-d/blob/v0.9.0/docker/Dockerfile.rocm)). Follow-up: build the wheel as `rixl` and re-bump.


BenjaminBraunDev:
Closing. The root cause here is wrong: the upstream vLLM ROCm v0.26.0 base already installs RIXL, so `llm-d-rocm:v0.9.0` contains `rixl` and the import resolves. The pods do crash-loop on v0.9.0, but no container log was captured, so the cause is unknown. Tracking in #2523.

PR #2525, opened 2026-09-17
docs(guides): rename gke-pod-snapshots guide to pod-snapshot

RyanRosario:
This is a followup to #2489. Replace every instance of `gke-pod-snapshot` with `pod-snapshot` so the well lit path is not presented as GKE only.

- guides/gke-pod-snapshots/ -> guides/pod-snapshot/
- router/gke-pod-snapshots.values.yaml -> router/pod-snapshot.values.yaml
- guide.yaml name, GUIDE_NAME, Helm release, and EPP service name
- NAMESPACE: llm-d-gke-pod-snapshots -> llm-d-pod-snapshot
- the llm-d.ai/guide label in kustomization.yaml, podsnapshotpolicy.yaml, and the router values
- namePrefix, storageConfigName, and the Workload Identity service account

Left unchanged:

SNAPSHOT_PROVIDER=gke_gvisor, which is a backend identifier matched in docker/scripts/snapshot/providers.py rather than guide branding, and the modelserver/gpu/vllm/gke/ overlay directory.

The Workload Identity principal embeds both the namespace and the KSA name (.../subject/ns/<NAMESPACE>/sa/<KSA>), so the IAM binding in the prerequisites moved with them. Verified that the ServiceAccount Kustomize renders matches the KSA in the gcloud add-iam-policy-binding command; a mismatch there surfaces only at checkpoint time as a GCS 403.


RyanRosario:
/assign @ahg-g 

Removing `gke` from filenames etc. and changing `snapshots` to `snapshot`. This PR follows from 2489 where there was a request to make the user guide less vendor specific, and only represent GKE as the first implementation.

nicolexin:
LGTM, thanks!

ahg-g:
@RyanRosario pls rebase

PR #2339, opened 2026-08-20
CI - strip out llm-d cuda images + deepep v2

Gregory-Pereira:
Addresses part of: https://github.com/llm-d/llm-d/issues/2338. This currently just represents stripping out the cuda images. I need to make the guide changes to use Deepep HTv2

PR #2530, opened 2026-09-18
fix(makefile): restore the leading tab on the uninstall-docker completion message

chrikrah:
`Makefile:213` starts at column zero, so make does not read it as part of the `uninstall-docker` recipe.

The line contains a colon, in `Remove alias if set:`, so make reads it as a rule rather than reporting a missing separator. The file parses with no warning.

## Two consequences

The recipe ends at line 212, so the completion message never prints:

```
$ make -n uninstall-docker | tail -1
docker stop llm-d-container && docker rm llm-d-container
```

`install-docker` at `:205`, `uninstall-k8s` at `:243` and `uninstall-openshift` at `:272` all print theirs. This is the only one that does not.

And make defines a target out of the text:

```
$ make -p -n | grep '^@echo'
@echo: unalias llm-d"
```

## After

```
$ make -n uninstall-docker | tail -1
echo "docker uninstallation complete. Remove alias if set: unalias llm-d"

$ make -p -n | grep -c '^@echo'
0
```

## The repository already states the rule

`.editorconfig`, added in #2103:

```
# Makefiles require tabs
[Makefile]
indent_style = tab
```

## Scope

One character. A scan of the whole file for a column-zero line inside a recipe block returns line 213 and nothing else. The only other hits were `:302` to `:311`, the space-indented prerequisite continuation under `check-tools`, which is correct as written.

`make uninstall-docker` was run with `-n` only. Running it for real stops and removes a container.

## Duplicate search

`gh issue list --state all` for `Makefile tab` and for `uninstall-docker`, and `gh pr list --state all` for `Makefile`, return nothing covering this.

Open PR #2339 edits this same line, to swap `$(PROJECT_NAME)` for `$(IMAGE_REGISTRY_REPO)`, and keeps the column-zero start. It is `CONFLICTING` today. If it lands first the tab still needs adding, and if this lands first that PR needs a one-line rebase.


maugustosilva:
@chrikrah Thanks for the PR. Please sign your commits so we can merge.

PR #2313, opened 2026-08-18
tiered-prefix-cache: Add TPU multi-host support and restructure overlays

dannawang0221:
- Reorganize TPU base and overlay manifests into `single-host` and `multi-host` subdirectories for TPU v6 and TPU v7.
    - Add multi-host recipe defaults under `recipes/modelserver/base/multi-host/default`.
    - Restructure router configuration values into `router/single-host` and `router/multi-host`.
    - Update `guides/tiered-prefix-cache/README.md` to introduce the `HOST_TYPE` variable (`single-host` | `multi-host`) for TPU deployment and teardown commands.
    - Add benchmark report for Qwen3-Coder-480B-A35B-Instruct on TPU v7 multi-host CPU offload (`vllm-qwen3-coder-480b-a35b-tpuv7-multi-host.md`).

PR #2533, opened 2026-09-18
pd-disaggregation: fold the TPU guide into the main guide, cache TPU weights on the node, pin TPU7x to vllm-tpu v0.26.0

yangligt2:
## Summary

Folds the separate TPU page of the P/D disaggregation guide into the main guide, and fixes two problems found while validating the TPU7x path end to end.

Three commits, each independently revertable:

1. **Fold the TPU guide into the main guide.** `README.tpu.md` was last synced with the shared content on 2026-06-12 and had drifted (dead benchmark anchor, retired `run_only.sh` flow, an env block that sourced `env.sh` before `REPO_ROOT` was set, an unflagged inheritance of the H200-calibrated `peakPrefillThroughput`). It is replaced by a `#### TPU` section next to `#### GPU`, with the dynamic sub-slicing content reduced to a pointer at `docs/infrastructure/providers/gke/dynamic-slicing/`. `README.tpu.md` and the orphaned `benchmark-templates/tpu.yaml` are deleted. The model-server overlays stay where they are; the nightly TPU workflow selects them by `accelerator_type`, and the other CI jobs discover work by path, so no CI change is needed.
2. **Cache model weights on the node for the TPU overlays.** Adds the node-local HuggingFace cache (`hostPath /var/cache/huggingface`, `HF_HOME`) already used by the GKE GPU overlays to the TPU base patches and the dynamic-slice manifests. Without it every container restart re-downloaded the 406 GB TPU7x model into the writable layer, which triggered ephemeral-storage eviction on nodes with other data on the boot disk. The boot-disk requirement is documented.
3. **Pin the TPU7x overlays to `vllm-tpu:v0.26.0`.** In `v0.27.0` through `v0.29.0` the vLLM scheduler reads `connector._kv_transfer_config` (`requires_kv_delivery`), but the bundled `TPUConnectorHMA` skips the base-class initializer, so EngineCore fails at startup with `AttributeError: 'TPUConnectorHMA' object has no attribute '_kv_transfer_config'`. Fixed upstream by vllm-project/tpu-inference#3566 (2026-09-12); no `vllm-tpu` release contains it yet. The shared `tpu-vllm/release` component stays at `v0.27.0` because tiered-prefix-cache's TPU multi-host offload (#2313) needs it; a new `tpu-vllm/release-v0.26.0` component pins only the two HMA overlays. The v6e overlay uses the non-HMA `TPUConnector` and is unaffected. This commit is reverted once a release includes the fix.

## Validation

* Deployed `modelserver/tpu/v7/vllm` from this branch on GKE TPU7x (`2x2x1`, 1 prefill + 1 decode, standalone router). `/v1/completions` and `/v1/chat/completions` succeed through the EPP; the prefill engine logs the KV send and the decode engine the KV pull for every request; decode reports 100% external prefix-cache hits (574/574 prompt tokens over three requests).
* Every `kustomization.yaml` under `guides/pd-disaggregation` builds; `tpu/v6/vllm` renders `v0.27.0`, both TPU7x overlays render `v0.26.0`.
* `make verify`, markdownlint (no new findings beyond the existing `<h4>`-in-`<summary>` pattern), yamllint, and the router `helm template` pass.
* Internal links and anchors in the changed Markdown resolve.

## Note

The regression in commit 3 was not caught by CI because the nightly TPU lane only runs `tpu/v6`. A TPU7x lane for the HMA path would close that gap after we can get capacity for nightly e2e test.

cc: @achandrasekar @ahg-g


yangligt2:
Closing and reopening to re-trigger the kustomize dry-run workflow; its previous attempt failed in the observability kubeconfig-targeting test, which this PR does not touch and which passed on the first run of the same tree.

achandrasekar:
Change looks good overall. @yangligt2 I see you validated on v7x. Have you validated that this doesn't affect the existing GPU and TPU p/d CI (v6e) jobs?

maugustosilva:
/test-nightly nightly-e2e-pd-disaggregation-gke-acc-gpu-vllm-x

yangligt2:
> Change looks good overall. @yangligt2 I see you validated on v7x. Have you validated that this doesn't affect the existing GPU and TPU p/d CI (v6e) jobs?

GPU lanes: not affected. Nothing under modelserver/gpu/, modelserver/amd/, the router values, or the workflows changes. The shared tpu-vllm/release component is unchanged; the pin is a new release-v0.26.0 component referenced only by the two TPU7x overlays. The dry-run check builds every guide kustomization and passes.

TPU v6e lane: verified statically, not run end to end. The nightly applies modelserver/tpu/v6/vllm by path via accelerator_type: tpu/v6; that path is unchanged by the docs fold. v6e inherits the node-local HF cache from the base patches (hostPath /var/cache/huggingface, DirectoryOrCreate, HF_HOME), the same mount the GKE GPU P/D overlay already uses in its nightly. The image pin does not reach v6e: it still renders vllm-tpu:v0.27.0 and uses the non-HMA TPUConnector, which is unaffected by the bug.

guides/recipes/modelserver/components/images/tpu-vllm/release-v0.26.0/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+# Pinned previous release for overlays that use TPUConnectorHMA
+# (tpu_inference.distributed.tpu_connector_hma). In vllm-tpu v0.27.0 through
+# v0.29.0 the vLLM scheduler reads connector._kv_transfer_config
+# (KVConnectorBase_V1.requires_kv_delivery), but the bundled TPUConnectorHMA
+# does not run the base-class initializer, so EngineCore fails at startup with
+# AttributeError: 'TPUConnectorHMA' object has no attribute '_kv_transfer_config'.
+# Fixed upstream by vllm-project/tpu-inference#3566 (2026-09-12); switch the
+# HMA overlays back to ../release once a vllm-tpu release includes that fix.
```
ahg-g, line 10: can it be backported to 27 through 29? does vllm issue patched releases?

